PR #6, opened 2019-09-28
add pynwb get_metadata

bendichter:
the latest pynwb 1.1.0 is having issues, but this works with `pynwb==1.0.3` and `hdmf==1.2.0`.

### DIFF
--- a/dandi/pynwb_utils.py
+++ b/dandi/pynwb_utils.py
@@ -1,16 +1,24 @@
 from pynwb import NWBHDF5IO
 
+metadata_list = ('related_publications', 'lab', 'experimenter', 'keywords', 'session_start_time',
+                 'institution', 'session_description', 'session_id', 'experiment_description')
+
 
 def get_metadata(filepath):
     out = dict()
     with NWBHDF5IO(filepath, 'r') as io:
         nwb = io.read()
-        out['publications'] = nwb.related_publications
-        out['lab'] = nwb.lab
-        out['experimenters'] = nwb.experimenter
-        out['keywords'] = nwb.keywords
-        out['date'] = nwb.session_start_time
+        for key in metadata_list:
+            out[key] = getattr(nwb, key)
         out['subject_id'] = nwb.subject.subject_id
-        out['institution'] = nwb.institution
+        if hasattr(nwb, 'electrodes') and nwb.electrodes is not None:
+            out['number_of_electrodes'] = len(nwb.electrodes)
+        else:
+            out['number_of_electrodes'] = 0
+
+        if hasattr(nwb, 'units') and nwb.units is not None:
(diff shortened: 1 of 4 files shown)


dandi/pynwb_utils.py
```diff
@@ -0,0 +1,16 @@
...
+def get_metadata(filepath):
+    out = dict()
+    with NWBHDF5IO(filepath, 'r') as io:
+        nwb = io.read()
+        out['publications'] = nwb.related_publications
+        out['lab'] = nwb.lab
+        out['experimenters'] = nwb.experimenter
+        out['keywords'] = nwb.keywords
+        out['date'] = nwb.session_start_time
+        out['subject_id'] = nwb.subject.subject_id
```
yarikoptic: Shouldn't we get some session id as well?

setup.py
```diff
@@ -17,7 +17,7 @@
 # Give setuptools a hint to complain if it's too old a version
 # 30.3.0 allows us to put most metadata in setup.cfg
 # Should match pyproject.toml
-SETUP_REQUIRES = ['setuptools >= 30.3.0']
+SETUP_REQUIRES = ['setuptools >= 30.3.0', 'pynwb==1.0.3', 'hdmf==1.2.0']
```
yarikoptic: Let's not pin versions in setup.py - only minimally required version.
requirements.txt (if desired) or pipenv.lock is the way to go for pinning

yarikoptic: For pynwb we can specify `,!=1.1.0` (or something like that, I believe there are examples in nibabel) to avoid installation of the buggy version

dandi/pynwb_utils.py
bendichter: Yeah that could be useful. It's just as easy as the others

setup.py
bendichter: Ok yeah I agree open requirements are better, I just wanted to avoid the buggy versions right now.

yarikoptic:
pushed e46ba108281bb402e4a2ca2170f29135d4ef2573 with some tune ups and added test (didn't establish CI for it yet, so run manually using pytest, that commit has an example if needed). It shows that round trip in pynwb causes tzinfo added to the datetime record.  Although it is minor, it is something to keep in mind! @bendichter , do you think it is a feature (then I will workaround) or a bug (then I will report) within pynwb?

setup.py
yarikoptic: in 44b59755b9380776a357e8ab6cd503c4f3960d45 pushed all needed ;)

yarikoptic:
I guess I will just merge this one for now, and deal with the date round-trip later/separately one way or another

yarikoptic:
thanks!  I will adjust accordingly (probably just set it while defining for a fixture NWBFile)

PR #7, opened 2019-10-01
RF: use explicit example  datetime with tzinfo set to tzutc()

yarikoptic:
Unfortunately test is still not passing, I am no longer certain with what version of pynwb/hdmf tandem I saw the test pass (besides the tzinfo), but when I use older one I see only `None` values. And with current `dev` of pynwb and hdmf I see that it loads some fields as "tuples":
```
(git)hopa:~/proj/dandi/dandi-cli[enh-ls]git
$> python -m pytest -s -vv dandi/tests/test_pynwb_utils.py
==================================================================== test session starts =====================================================================
platform linux -- Python 3.7.3rc1, pytest-3.10.1, py-1.7.0, pluggy-0.8.0 -- /home/yoh/proj/dandi/dandi-cli/venvs/dev3/bin/python
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/home/yoh/proj/dandi/dandi-cli/.hypothesis/examples')
rootdir: /home/yoh/proj/dandi/dandi-cli, inifile:
plugins: localserver-0.5.0, cov-2.6.0, hypothesis-3.71.11
collected 2 items                                                                                                                                            

dandi/tests/test_pynwb_utils.py::test_get_metadata FAILED
dandi/tests/test_pynwb_utils.py::test_pynwb_io PASSED

========================================================================== FAILURES ==========================================================================
_____________________________________________________________________ test_get_metadata ______________________________________________________________________

simple1_nwb = '/home/yoh/.tmp/pytest-of-yoh/pytest-44/data0/simple1.nwb'
simple1_nwb_metadata = {'experiment_description': 'experiment_description1', 'experimenter': 'experimenter1', 'identifier': 'identifier1', 'institution': 'institution1', ...}

    def test_get_metadata(simple1_nwb, simple1_nwb_metadata):
        target_metadata = simple1_nwb_metadata.copy()
        # we will also get some counts
        target_metadata['number_of_electrodes'] = 0
        target_metadata['number_of_units'] = 0
        # subject_id is also nohow specified in that file yet
        target_metadata['subject_id'] = None
>       assert target_metadata == get_metadata(str(simple1_nwb))
E       AssertionError: assert {'experiment_...tution1', ...} == {'experiment_d...tution1', ...}
E         Common items:
E         {'experiment_description': 'experiment_description1',
E          'identifier': 'identifier1',
E          'institution': 'institution1',
E          'keywords': ['keyword1', 'keyword 2'],
[...]

PR #8, opened 2019-10-02
ENH: rudimentary ls and validate commands

yarikoptic:
TODOs (for the future, not PR):
- Proper tests
- Extend interface
- Add recursive files discovery
- More metadata fields to be listed (and provide option to select which ones)

PS Sits on top of #7


dandi/cli/command.py
```diff
@@ -0,0 +1,223 @@
...
+                pass
+            out(rec)
+        out.wait()
+
+#
+# Validate
+#
+
+@main.command()
+@click.argument('paths', nargs=-1, type=click.Path())
```
danlamanna: Should this be `click.Path(exists=True, dir_okay=False)` ?

dandi/cli/command.py
```diff
@@ -0,0 +1,223 @@
...
+        #     len(errors_unique), len(errors))
+        # )
+        for path, errors in errors.items():
+            print(crayons.normal(path, bold=True))
+            for error in errors:
+                print("  {}".format(crayons.red(error)))
+        raise SystemExit(1)
+    else:
+        print(
+            crayons.green("No validation errors among {} files".format(len(files)),
```
danlamanna: I'm unfamiliar with crayons, does it provide something more than `click.style` http://click.palletsprojects.com/en/7.x/utils/#ansi-colors?

dandi/cli/command.py
```diff
@@ -0,0 +1,223 @@
...
+        for path in files:
+            rec = {'path': path}
+            try:
+                rec['size'] = os.stat(path).st_size
+                rec.update(get_metadata_pyout(path))
+            except FileNotFoundError as exc:
+                #lgr.error("File is not available: %s", exc)
+                pass
+            out(rec)
+        out.wait()
```
kyleam: You're using the context manager.  Its `__exit__` already calls `wait`, so you don't need this here (or if you do, it's a bug).

yarikoptic: got it.  There was indeed some oddity in `dandi ls` quitting right away (either not waiting for the callback at all or just not rendering its output), so I added the `.wait()`, but I think it solved nothing, I will strip it away. thanks

dandi/cli/command.py
yarikoptic: I think so (at least for now, until I add directories treatment) - I will change it now. 
As you can feel I am not a `click` expert (yet). Thanks!

dandi/cli/command.py
yarikoptic: oh, I didn't know about that feature of click.  most likely it would be sufficient - I will check. thanks!
I picked up crayons idea from pipenv which also uses click.

yarikoptic:
ok, I have addressed the comments -- thanks for the feedback!
it is ugly and untested but seems to work.  I will merge it but if you see anything in new code worth immediate action, either leave a comment here or add a new issue/PR.

PR #15, opened 2019-10-08
ENH: ls - list NWB (version), lab, subject_id  where available; ignore warnings upon validate

### LINKED ISSUE #13: should we list found (NWB:N?) specification versions?
reflecting upon https://github.com/NeurodataWithoutBorders/pynwb/issues/1077 -- PyNWB fails to open some (NWB 1.0? or just not NWB:N?) NWB files.  So I wondered if we should collect a list of specification versions (Groups) and list them with `dandi ls` before even trying to feed that file to PyNWB?

@bendichter what do you think?

yarikoptic:
Commit comment in 1fd302325803f442162e3757414b5fc90ac1d502 summarizes situation with nwb_version a bit:

    ENH: obtain also "nwb_version", ls it as NWB field in the table
    
    I went for a shorter name in  ls  command since space is scarse.
    
    Unfortunately I could not immediately figure out how to even populate
    that field e.g. using NWBFile.  And PyNWB does not populate it
    either ATM: https://github.com/NeurodataWithoutBorders/pynwb/issues/1085
    
    The only example I found quickly with nwb_version field populated was from
    ///crcns/ssc-7 dataset -- and in particular
    data/L4E_whole_cell/Exp_2015-09-05_001_0001-0162.nwb
    file there.
    
    I really think we should come up with some good, and not that large collection
    of sample .nwb files to test on.  Unfortunately any "real" .nwb file seems to be
    quite large

Closes #13

### DIFF
--- a/dandi/cli/command.py
+++ b/dandi/cli/command.py
@@ -147,9 +147,10 @@ def ls(paths):
             'path',
             'size',
             #'experiment_description',
-            #'lab',
+            'lab',
             'experimenter',
             'session_id',
+            'subject_id',
             'session_start_time',
             #'identifier',  # note: required arg2 of NWBFile
             #'institution',
--- a/dandi/cli/command.py
+++ b/dandi/cli/command.py
@@ -73,10 +73,10 @@ def get_metadata_pyout(path):
 
     if 'nwb_version' not in rec:
         # Let's at least get that one
-        rec['NWB'] = get_nwb_version(path)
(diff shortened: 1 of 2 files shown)


yarikoptic:
bundled in also ef15d04e6b8780c834dc22f7c645345ff81d18c6 to ignore some warnings upon `validate`.  It then would look much "cleaner":
image

yarikoptic:
ok, no objections/comments, let's proceed

PR #16, opened 2019-10-08
enh: add precommit support

satra:


yarikoptic:
never used one, but sounds useful and less painful to introduce this early in the project ;) thanks! let's do it!

PR #11, opened 2019-10-07
RF: use callback mechanism of pyout to process files in parallel

yarikoptic:
unfortunately it stalls for unknown reason with first file record never populated

    $> dandi ls /home/yoh/proj/dandi/nwb-datasets/najafi-2018-nwb/data/FN_dataSharing/nwb/mouse1_fni16_15081{7,8}*nwb
    PATH                                                          SIZE     EXPERIMENTER    SESSION_ID   SESSION_START_TIME   KEYWORDS
    ...use1_fni16_150817_001_ch2-PnevPanResults-170808-190057.nwb 302.9 MB
    ...use1_fni16_150818_001_ch2-PnevPanResults-170808-180842.nwb 227.7 MB Farzaneh Najafi 170808180842 2015-08-17/20:00:00

that sample dataset could be obtained from

   https://github.com/dandi/najafi-2018-nwb

with nwb data hosted on

   http://datasets.datalad.org/?dir=/labs/churchland/najafi-2018-nwb

and linked to that github repo as autoenabled special remote.

PR #17, opened 2019-10-08
ls: Gather get_metadata() information asynchronously

kyleam:
This PR reworks gh-11 and avoids the stalling issue mentioned there.  There's improvements that should be made on pyout's end, but what's here will work with the current release.

I've tested this out on the dandi/najafi-2018-nwb repo with

    $ dandi ls data/FN_dataSharing/nwb/mouse1_fni16_15081{7,8}*.nwb

### DIFF
--- a/dandi/cli/command.py
+++ b/dandi/cli/command.py
@@ -78,13 +78,10 @@ def get_metadata_pyout(path):
     if "nwb_version" not in rec:
         # Let's at least get that one
         try:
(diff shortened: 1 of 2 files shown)


yarikoptic:
THANKS @kyleam ! I am confirming that it works now, and I will start now using that pre-commit beast myself to bring sanity into PR diffs ;)

PR #19, opened 2019-10-09
ENH: Fixed up **the** test to pass, added basic github workflow

yarikoptic:
Let's see if this beast (github workflows) works!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,36 @@
+name: Tests
+
+on:
+  - push
+  - pull_request
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+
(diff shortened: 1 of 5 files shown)


yarikoptic:
ah, apparently I need to sign up per account/institution, so did for dandi:

    You’re in the queue for the beta—we’ll let you know as soon as you have access.
    You’ve agreed to the pre-release terms with your organization @dandi. No further steps are necessary, ....

so a bit of waiting is due

yarikoptic:
apparently it is not clear on how long to wait (asked [here now](https://github.community/t5/GitHub-Actions/Please-give-some-guestimate-on-a-wait-time/m-p/34384#M1897)) so let's proceed and just stay hopefull

PR #26, opened 2019-10-16
NF: upload

yarikoptic:
Somewhat reimplemented so far what possibly girder client could have done for us already

- [x] adjust setup.cfg for new dependencies (keyring)
- [x] actually upload the file(s)
- [x] validate before upload (and option to skip "invalid" files)
- [x] extract and upload metadata
-  test(s) -- dedicated #27 for that

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,3 +22,9 @@ In the source directory
 ```
 pre-commit install
 ```
+## 3rd party components included
+
+### dandi/support/generatorify.py
+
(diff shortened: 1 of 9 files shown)


yarikoptic:
ATM looks like this
image
so by default
- does in parallel up to 10 files
- does validation and skips files with errors
- would skip problematic (not found, e.g. broken symlinks) files
- reports overall upload speed
- provides additional possible error messages in "MESSAGE" column
- uploads metadata (with temp downcasing all values)

Overall -- should work ;-)

I think unless any major objections/bugs detected -- I will merge and cut another alpha release to upload to pypi

PR #35, opened 2020-01-14
ENH: add macos-latest into the matrix of runs for os

yarikoptic:


yarikoptic:
Mammoth in the room was detected and dealt with.  Tests pass! so I will merge

PR #38, opened 2020-02-23
ENH: upload coverage reports to codecov

yarikoptic:
For that to work, we need to provide a secret token and that one would be available only (according to docs) if pull request comes from this repo (so cannot do from my fork... although may be I could set it there as well???)


yarikoptic:
seems to work, so goes into master

PR #39, opened 2020-02-23
ENH(TST): initial test for sanitize_value

yarikoptic:
primarily to test if coverage would be submitted from a PR from
my fork, where I added CODECOV_TOKEN into secrets

PR #40, opened 2020-02-24
ENH(TST): initial test for sanitize_value

yarikoptic:
apparently with current setup cannot submit PRs from my own fork and have their coverage reported : https://github.com/dandi/dandi-cli/pull/39 :-( that is suboptimal

testing by submitting withing our repo - must work

### DIFF
--- a/dandi/tests/test_organize.py
+++ b/dandi/tests/test_organize.py
@@ -1,0 +1,7 @@
+from ..organize import _sanitize_value
+
+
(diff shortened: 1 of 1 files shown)


yarikoptic:
not quite sure what is happening with actions -- why there are two "pull_request"s where one seems to fail to find CODECOV_TOKEN in the secrets. I will see what would happen in some subsequent PR

PR #45, opened 2020-03-09
NF+ENH+RF: full cycle - download, caching of metadata reads, etc

yarikoptic:
I kept too long without posting it, but now that there is substantial amount of changes and some added testing, time to reveal ;)

TODOs:
- [x] now that caching of metadata is in place, finish up RF of `organize` to operate on real files (not just metadata dumps
- [x] adjust upload/download for the treatment of `dandiset.yaml` - yet to check what we decided (no file on archive and just metadata record?)
- [x] rudimentary `register` command
- provide ~~integration test~~ (would need full instance) -- a demo -- on a sample dataset which would do the full loop - organize, register, upload, download - on a "local" instance of the archive
- Some which might happen here or later (would then file separate issues and/or PRs)
  - RF `download` to provide more consistent UI (via pyout) to `upload` reporting on the files which were (re)downloaded or not (later: see #48 )
  - RF+ENH `validate` to (later)
         - validate "dandi" requirements
         - consistent UI


yarikoptic:
@mgrauer - only for you for the Friday-the-13th present, the beast is again compatible with python 3.6 now ;)

yarikoptic:
woohoo - all green, merging while nobody has mentioned! ;-)

mgrauer:
Thanks for the 3.6 support.

PR #54, opened 2020-03-15
ENH: organize - resolve ambiguity by adding _obj- where necessary

yarikoptic:
The need for this was reminded by @satra and implementation uses proposed by him shorter version based on alphanum encoding of the crc32 checksum value, instead of the one I thought about -- first octet of object_id uuid (such change was done in 2845eda4c2fac4341c085d69debfc25e0a65ad94, which also has a brief show case example).

It could be seen needed on an older version of Kriegstein data (now no longer avail in dandiarchive)

- Diff is a bit big since in the first commit (362d346b069c15ca3d542b4b19b7fd3509bb647f ATM) I just refactored code by moving around to prepare for the follow up changes.
- I dropped "special" to "simulate" mode support for non-unique files handling - no need really.

TODOs:
- [x] check which version of nwb-schema (and/or) pynwb (and matnwb) introduced storing `object_id` and add that info into error messages. [nwb-schema 2.1.0](https://nwb-schema.readthedocs.io/en/latest/format_release_notes.html#march-2-2020) and pynwb 1.1.0 which uses that schema
- [x] basic tests
- [x] provide documentation in `--help` on how file names are generated and what constitutes `_obj-` value

### DIFF
--- a/dandi/organize.py
+++ b/dandi/organize.py
@@ -139,11 +139,19 @@ def _assign_obj_id(metadata, non_unique):
     seen_object_ids = {}  # object_id: path
     for r in metadata:
         if r["dandi_path"] in non_unique:
-            object_id = get_object_id(r["path"])
+            try:
+                object_id = get_object_id(r["path"])
+            except KeyError:
+                raise OrganizeImpossibleError(
+                    msg + ". We tried to use object_id but it is absent in %r. "
+                    "It is either not .nwb file or produced by older *nwb libraries. "
+                    "You must re-save files e.g. using recent pynwb." % (r["path"])
+                )
+
             if not object_id:
                 raise OrganizeImpossibleError(
                     msg + ". We tried to use object_id but it was %r for %s. "
-                    " You must re-save files using recent pynwb"
+                    " You might need to re-save files using recent pynwb"
                     % (object_id, r["path"])
                 )
             # shorter version
--- a/dandi/organize.py
+++ b/dandi/organize.py
@@ -8,9 +8,10 @@ from collections import Counter
 import dateutil.parser
 import os.path as op
 
+from .exceptions import OrganizeImpossibleError
 from . import get_logger
 from .consts import dandiset_metadata_file
-from .pynwb_utils import get_neurodata_types_to_modalities_map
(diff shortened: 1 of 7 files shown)


dandi/organize.py
```diff
@@ -96,8 +104,83 @@ def create_unique_filenames_from_metadata(
...
+                    raise RuntimeError(
+                        f"Wrong assumption by DANDI developers that first "
+                        f"CRC32 checksum of object_id would be sufficient.  Please "
+                        f"report: {obj_id} the same for "
+                        f"{seen_object_ids[seen_object_id]}={seen_object_id} "
+                        f"{r['path']}={object_id} "
+                    )
+            r["obj_id"] = obj_id
+            seen_obj_ids[obj_id] = object_id
+            seen_object_ids[object_id] = r["path"]
```
satra: it seems this could benefit from some code improvement. what is the difference between the two variables. semantically seem similar.

dandi/organize.py
```diff
@@ -96,8 +104,83 @@ def create_unique_filenames_from_metadata(
...
+            )
+        _assign_obj_id(metadata, non_unique)
+        _assign_dandi_names(
+            metadata, mandatory, (mandatory_if_not_empty or []) + ["obj_id"]
+        )
+        non_unique = _get_non_unique_paths(metadata)
+        if non_unique:
+            raise OrganizeImpossibleError(
+                "Even after adding obj_id we ended up with non-unique file names. "
+                "Should have not happened: %s" % str(non_unique)
```
satra:
```suggestion
                "Should not have happened: %s" % str(non_unique)
```

dandi/organize.py
yarikoptic: `obj_id` is the shortened version of `object_id`.  There is a comment by those dictionaries definition saying that first maps from short to original, the 2nd one from original to the path.  Indeed names could be better. E.g. `obj_to_object_ids` and `object_id_paths`

yarikoptic:
ok, if there is no strong objections or recommendations, I will soon merge and release 0.4.1

yarikoptic:
failures are due to some difficulty uploading to codecov - unrelated. and @satra gave blessing, so merging

PR #62, opened 2020-03-17
enh: add etelemetry to dandi cli

satra:
@yarikoptic - let me know if you want this check in some other form

@mgrauer - this is a bit of boilerplate that should be moved into etelemetry itself. we have just been lazy!!

### DIFF
--- a/dandi/__init__.py
+++ b/dandi/__init__.py
@@ -36,3 +36,57 @@ lgr = get_logger()
 set_logger_level(lgr, os.environ.get("DANDI_LOG_LEVEL", logging.INFO))
 FORMAT = "%(asctime)-15s [%(levelname)8s] %(message)s"
 logging.basicConfig(format=FORMAT)
+
+
+def check_latest_version(raise_exception=False):
+    """
+    Check for the latest version of the library.
+    Parameters
+    ----------
+    raise_exception: :obj:`bool`
+        Raise a RuntimeError if a bad version is being used
+    """
+    import etelemetry
+    from pkg_resources import parse_version
+
+    logger = lgr
+
+    INIT_MSG = "Running {packname} version {version} (latest: {latest})".format
+
+    latest = {"version": "Unknown", "bad_versions": []}
+    result = None
+    try:
+        result = etelemetry.get_project("dandi/dandi-cli")
+    except Exception as e:
+        logger.warning("Could not check for version updates: \n%s", e)
+    finally:
+        if result:
+            latest.update(**result)
+            if parse_version(__version__) != parse_version(latest["version"]):
+                logger.info(
+                    INIT_MSG(
+                        packname="dandi", version=__version__, latest=latest["version"]
+                    )
+                )
+            if latest["bad_versions"] and any(
+                [
+                    parse_version(__version__) == parse_version(ver)
+                    for ver in latest["bad_versions"]
+                ]
+            ):
+                message = (
+                    "You are using a version of dandi cli with a critical "
+                    "bug. Please use a different version."
+                )
+                if raise_exception:
+                    raise RuntimeError(message)
+                else:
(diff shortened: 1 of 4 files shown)


dandi/__init__.py
```diff
@@ -36,3 +36,57 @@ def set_logger_level(lgr, level):
...
+                    )
+                )
+            if latest["bad_versions"] and any(
+                [
+                    parse_version(__version__) == parse_version(ver)
+                    for ver in latest["bad_versions"]
+                ]
+            ):
+                message = (
+                    "You are using a version of dandi cli with a critical "
```
mgrauer: How about we provide specific info about how to upgrade, both in the bad versions case and in the !latest case?

e.g. `pip install --upgrade dandi`

yarikoptic: What if there is only downgrade to get away from the bad version? 
What if it is installed from Debian package? From conda? ...

dandi/__init__.py
```diff
@@ -36,3 +36,57 @@ def set_logger_level(lgr, level):
...
+                else:
+                    logger.critical(message)
+    return latest
+
+
+# Run telemetry on import for interactive sessions, such as IPython, Jupyter notebooks, Python REPL
+import __main__
+
+if not hasattr(__main__, "__file__"):
+    check_latest_version()
```
yarikoptic: I am not familiar with this approach to see how we are imported... But in principle, shouldn't we just always check and log?

dandi/__init__.py
```diff
@@ -36,3 +36,57 @@ def set_logger_level(lgr, level):
...
+
+def check_latest_version(raise_exception=False):
+    """
+    Check for the latest version of the library.
+    Parameters
+    ----------
+    raise_exception: :obj:`bool`
+        Raise a RuntimeError if a bad version is being used
+    """
+    import etelemetry
```
yarikoptic: The whole function should get under try:except so we don't blow if eg etelemetry import blows or there is a bug in finally handling below

I could add a decorator later

satra: for this last one, since etelemetry is now a dependency i would not put it under try/except. we don't do it for any other import of package on which something is dependent.

satra:
i have also sent PR to etelemetry, if that gets merged and released, we can reduce this boilerplate significantly.

dandi/__init__.py
yarikoptic: if something like `yaml` fails to import -- major functionality of the package is impaired.  If `etelemetry` fails to import -- who really cares?  package should remain fully functional (unless it is a very very bad version ;) ) that is the difference for me.
BTW -- debian folks even already packaged it but it is waiting for acceptance -- lets see if debian ftp masters would raise a red flag ;) https://ftp-master.debian.org/new/python-etelemetry_0.1.2-1.html

satra: since the intent of etelemetry is also to notify bad versions, it's important that it gets imported. it's not just a version check. so i think people should care.

yarikoptic: "should care" != "total lock down, can't use any functionality". and situation is not much different from a user working offline -- no checks could be done, so should he stop? no and the other exception handler allows for that.  So I do not see how the situation of "I failed to import etelemetry" is different.
IMHO etelemetry should be as non-intrusive as possible.

satra: if a package is in the requirements, then it should not be needed to be caught. if we say it is optional it's a separate thing. this PR started with this being a necessary thing for users. so it's in the requirements and therefore should not have a need to be try-excepted.

yarikoptic: we will chat about this over some beers (hopefully this year ;-))

yarikoptic:
tests pass, we should try "in real life", anything else (reducing boiler plate, making more "robust") could be done later, merging

PR #66, opened 2020-03-18
ENH: raise RuntimeError with informative msg when using python2 on setup.py

### LINKED ISSUE #65: prevent attempts of being installed or upgraded using python2
although `pip install dandi` correctly reports
```
ERROR: Package u'dandi' requires a different Python: 2.7.17 not in '>=3.5.1'
```
@bendichter ran into the situation where `pip install --upgrade dandi` failed with
```
bdichter@smaug:~$ pip install --upgrade dandi
Collecting dandi
  Using cached https://files.pythonhosted.org/packages/b9/4d/f2e8243beeb0427e713f4342017768f79fe6917e9cf2dd107ce10d23456c/dandi-0.2.0.tar.gz
  Installing build dependencies ... done
    Complete output from command python setup.py egg_info:
      File "/tmp/pip-install-HoAKb4/dandi/setup.py", line 20, in <module>
        "version": versioneer.get_version(),
      File "versioneer.py", line 1498, in get_version
        return get_versions()["version"]
      File "versioneer.py", line 1452, in get_versions
        keywords = get_keywords_f(versionfile_abs)
      File "versioneer.py", line 985, in git_get_keywords
        except (FileNotFoundError, KeyError):
    NameError: global name 'FileNotFoundError' is not defined
```
the reason for which (pip of python2) was immediately obvious.  I think we should just add a check to setup.py to kaboom with an informative message before versioneer is approached.

yarikoptic:
Closes #65

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,6 +13,12 @@ import sys
 
 from setuptools import setup
 
+if sys.version_info < (3,):
+    raise RuntimeError(
+        "dandi-cli's setup.py requires python 3 or later. "
+        "You are using %s" % sys.version
+    )
+
 try:
     import versioneer
 


yarikoptic:
no objections, so let's proceed

PR #71, opened 2020-04-03
ENH: dandi download (multiple) files

yarikoptic:
- no support to download selected folders and files (well -- we cannot even do that anywhere among drafts AFAIK since we have only either folders or files ATM in a directory)
- comes with a basic unittest and drags along machinery from reproman for flexible and concise annotation of the tests to be skipped.

### DIFF
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -5,6 +5,7 @@ import urllib.parse as up
 from . import girder, get_logger
 from .consts import dandiset_metadata_file
 from .dandiset import Dandiset
+from .utils import flattened
 
 lgr = get_logger()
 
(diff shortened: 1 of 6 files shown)


dandi/download.py
```diff
@@ -124,19 +145,40 @@ def download(
...
+):
     # asset_rec = client.getResource(asset_type, asset_id)
     # lgr.info("Working with asset %s", str(asset_rec))
-
     # In principle Girder's client already has ability to download any
     # resource (collection/folder/item/file).  But it seems that "mending" it
     # with custom handling (e.g. later adding filtering to skip some files,
     # or add our own behavior on what to do when files exist locally, etc) would
     # not be easy.  So we will reimplement as a two step (kinda) procedure.
-
```
yarikoptic, line 135: I believe I didn't do any of those -- must be "pre-commit" or some other evil automation

yarikoptic:
I guess unless we try it, we do not know ;)

satra:
@yarikoptic - if you are doing parallel download please use filelock to prevent concurrent directory creation by multiple process (os.exists is insufficient) and atomic operations are only guaranteed for a single folder, not a hierarchy.

alternatively you can use pydra for local concurrency and caching for download, so you can also resume.

yarikoptic:
I typically just use `os.makedirs(path, exists_ok=True)`. Didn't explicitly check for concurrency gotchas though

satra:
> I typically just use os.makedirs(path, exists_ok=True).

iirc that is not an atomic operation.

satra:
so if you lockfile on a path (without filename) converted to string, it should allow for safely creating paths with that call.

yarikoptic:
as far as I see it -- as long as it swallows exceptions and just checks that target directory exists - what was needed is achieved and there is no need for any locking.

PR #73, opened 2020-04-07
AFAIK no dateutil is needed for newer pythons.

yarikoptic:
We have an additional entry below conditioned on py < 3.7

Let's see if tests pass


yarikoptic:
apparently some tests seems to be flaky and passed only on 2nd try.  unrelated to dateutil, so merging

PR #75, opened 2020-04-09
ENH: sanitize, validate and tolerate better wrong nwb_version

yarikoptic:
As reported on slack, we have use cases with nwb_version somehow `NWB-2.2.0`.
This PR strips NWB- prefix while obtaining the version, and would complain if that is to be happen for files with nwb_version >= 2.1.0 (in a custom comparison, so we defer to proper semver check later). 

`validate` also would now check and report a problem.  This is hopefully a temporary implementation within dandi -- hopefully nwb_schema would introduce constraints via regular expressions so validity of nwb_version could be checked "natively" by pynwb, see https://github.com/NeurodataWithoutBorders/nwb-schema/issues/426

There is no test for `validate` changes here.  I am (slowly) working on a separate branch to RF validation, so they will (hopefully) come there ;)

### DIFF
--- a/dandi/pynwb_utils.py
+++ b/dandi/pynwb_utils.py
@@ -8,6 +8,7 @@ from collections import Counter
 
 import pynwb
 from pynwb import NWBHDF5IO
+import semantic_version
 
 from . import get_logger
 
@@ -34,7 +35,7 @@ def _sanitize_nwb_version(v, filename=None, log=None):
 
     Would log a warning if something detected to be fishy"""
     msg = f"File {filename}: " if filename else ""
-    msg += "Version is"
+    msg += f"nwb_version {v!r}"
 
     if log is None:
         log = lgr.warning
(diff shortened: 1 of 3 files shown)


dandi/pynwb_utils.py
```diff
@@ -187,20 +229,25 @@ def validate(path):
...
-            if errors != errors_:
-                lgr.debug(
-                    "Filtered out %d validation errors on %s",
-                    len(errors_) - len(errors),
-                    path,
-                )
+        # Explicitly sanitize so we collect warnings.
+        # TODO: later cast into proper ERRORs
+        version = _sanitize_nwb_version(version, log=errors.append)
+        if version and semantic_version.validate(version):
```
yarikoptic, line 240: previous versions had some 2.0b or alike, so we are doomed to support those here as well, will adjut

PR #86, opened 2020-04-14
ENH: a very basic implementation of parallel download

yarikoptic:
if works -- why not? ;)  `-J 1` will still cause use of joblib but should avoid parallel execution (whenever needed)

edit : there is no overall progress or ETA yet... might come later (or not)

### DIFF
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -5,7 +5,7 @@ import urllib.parse as up
 from . import girder, get_logger
 from .consts import dandiset_metadata_file
 from .dandiset import Dandiset
-from .utils import flatten, flattened
+from .utils import flatten, flattened, Parallel, delayed
 
 lgr = get_logger()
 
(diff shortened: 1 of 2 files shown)


yarikoptic:
FWIW,  this is using threads, so might still hit CPU ceiling on a speedy connection (I do see it jumping around 100% on smaug with download speed probably ~100MBps when it gets to large files)

dandi/download.py
```diff
@@ -236,7 +239,6 @@ def _get_asset_files(
                     )
                     continue
             dandiset = Dandiset(dandiset_path, allow_empty=True)
-            if not dandiset.path_obj.exists():
-                dandiset.path_obj.mkdir()
+            dandiset.path_obj.mkdir(exist_ok=True)  # exist_ok in case of parallel race
```
yarikoptic, line 242: hm... so the download tests we have do not have any directories, sad... oh well -- I tested locally on a large dataset, was all good.

PR #88, opened 2020-04-23
convert any NWB file to json

bendichter:
Extract the metadata from any NWB file into json.

test with:

```python
with open('test.json','w') as file:
    dandi.metadata.nwb_to_json('ecephys_example.nwb', file)
```

```
{'attributes': {'namespace': 'core',
  'neurodata_type': 'NWBFile',
  'nwb_version': '2.2.2',
  'object_id': '4da8e842-7941-47bc-81ab-01cf27b6c22b'},
 'datasets': {'file_create_date': {'data': '<HDF5 dataset "file_create_date": shape (1,), type "|O">'},
  'identifier': {'data': 'EXAMPLE_ID'},
  'session_description': {'data': 'my first synthetic recording'},
  'session_start_time': {'data': '2020-04-07T18:08:03.053584-04:00'},
  'timestamps_reference_time': {'data': '2020-04-07T18:08:03.053584-04:00'}},
 'groups': {'acquisition': {'groups': {'test_ephys_data': {'attributes': {'comments': 'This data was randomly generated with numpy, using 1234 as the seed',
      'description': 'Random numbers generated with numpy.random.rand',
      'namespace': 'core',
      'neurodata_type': 'ElectricalSeries',
      'object_id': 'f7440a31-1494-45f7-ac6e-dacce90d9174'},
     'datasets': {'data': {'data': '<HDF5 dataset "data": shape (1000, 2), type "<f8">',
       'attributes': {'conversion': '1.0',
        'resolution': '0.001',
        'unit': 'volts'}},
      'electrodes': {'data': '<HDF5 dataset "electrodes": shape (2,), type "<i8">',
       'attributes': {'description': 'the first and third electrodes',
        'namespace': 'hdmf-common',
        'neurodata_type': 'DynamicTableRegion',
        'object_id': 'daa9cec0-54e4-4671-859d-08c916156644',
        'table': 'root/general/extracellular_ephys/electrodes'}},
      'timestamps': {'data': '<HDF5 dataset "timestamps": shape (1000,), type "<f8">',
       'attributes': {'interval': '1', 'unit': 'seconds'}}}}}},
  'general': {'datasets': {'experiment_description': {'data': 'I went on an adventure with thirteen dwarves to reclaim vast treasures.'},
    'experimenter': {'data': '<HDF5 dataset "experimenter": shape (1,), type "|O">'},
    'institution': {'data': 'University of Middle Earth at the Shire'},
    'lab': {'data': 'Bag End Laboratory'},
    'session_id': {'data': 'LONELYMTN'}},
   'groups': {'devices': {'groups': {'trodes_rig123': {'attributes': {'namespace': 'core',
        'neurodata_type': 'Device',
        'object_id': '877a1540-6619-48f3-878f-d36ef706ab26'}}}},
    'extracellular_ephys': {'groups': {'electrodes': {'attributes': {'colnames': "['x', 'y', 'z', 'imp', 'location', 'filtering', 'group', 'group_name']",
        'description': 'metadata about extracellular electrodes',
        'namespace': 'hdmf-common',
        'neurodata_type': 'DynamicTable',
        'object_id': 'a0bbb49e-a614-4363-be97-eccd0a93e7c8'},
       'datasets': {'filtering': {'data': '<HDF5 dataset "filtering": shape (4,), type "|O">',
         'attributes': {'description': 'description of hardware filtering',
          'namespace': 'hdmf-common',
          'neurodata_type': 'VectorData',
          'object_id': 'c836121b-613c-4342-8594-3699ee8c1778'}},
        'group': {'data': '<hdmf.backends.hdf5.h5_utils.BuilderH5ReferenceDataset object at 0x121e69590>',
         'attributes': {'description': 'a reference to the ElectrodeGroup this electrode is a part of',
          'namespace': 'hdmf-common',
          'neurodata_type': 'VectorData',
          'object_id': '62a24ce1-f52d-493b-8e18-33a2656e2561'}},
        'group_name': {'data': '<HDF5 dataset "group_name": shape (4,), type "|O">',
         'attributes': {'description': 'the name of the ElectrodeGroup this electrode is a part of',
          'namespace': 'hdmf-common',
          'neurodata_type': 'VectorData',
          'object_id': 'aa00620d-4abf-4629-b86f-e7facd13ae16'}},
        'id': {'data': '<HDF5 dataset "id": shape (4,), type "<i8">',
         'attributes': {'namespace': 'hdmf-common',
          'neurodata_type': 'ElementIdentifiers',
          'object_id': '815e6e4d-98a7-4b23-b169-5bbb6225d969'}},
        'imp': {'data': '<HDF5 dataset "imp": shape (4,), type "<f8">',
         'attributes': {'description': 'the impedance of the channel',
          'namespace': 'hdmf-common',
          'neurodata_type': 'VectorData',
          'object_id': '39f16b74-3fb8-4301-a180-3aae8f7bd704'}},
        'location': {'data': '<HDF5 dataset "location": shape (4,), type "|O">',
         'attributes': {'description': 'the location of channel within the subject e.g. brain region',
          'namespace': 'hdmf-common',
          'neurodata_type': 'VectorData',
          'object_id': '30ef5927-4013-4997-82da-74ffa6c5af60'}},
        'x': {'data': '<HDF5 dataset "x": shape (4,), type "<f8">',
         'attributes': {'description': 'the x coordinate of the channel location',
          'namespace': 'hdmf-common',
          'neurodata_type': 'VectorData',
          'object_id': 'f7326a67-a5b9-4bda-9d39-36fe0614e545'}},
        'y': {'data': '<HDF5 dataset "y": shape (4,), type "<f8">',
         'attributes': {'description': 'the y coordinate of the channel location',
          'namespace': 'hdmf-common',
          'neurodata_type': 'VectorData',
          'object_id': '38e8ae72-6e42-4d8b-9ec4-a864968a3870'}},
        'z': {'data': '<HDF5 dataset "z": shape (4,), type "<f8">',
         'attributes': {'description': 'the z coordinate of the channel location',
          'namespace': 'hdmf-common',
          'neurodata_type': 'VectorData',
          'object_id': '5767dcbc-9330-4821-bfc2-511753c0676a'}}}},
      'tetrode1': {'attributes': {'description': 'an example tetrode',
        'location': 'somewhere in the hippocampus',
        'namespace': 'core',
        'neurodata_type': 'ElectrodeGroup',
        'object_id': '832951f0-fbdf-42c6-8178-dcd94ae00d39'}}}}}},
  'stimulus': {'groups': {}},
  'units': {'attributes': {'colnames': "['electrodes']",
    'description': 'Autogenerated by NWBFile',
    'namespace': 'core',
    'neurodata_type': 'Units',
    'object_id': '36fec1b3-82d7-4af9-a0af-d9d46d6a5779'},
   'datasets': {'electrodes': {'data': '<HDF5 dataset "electrodes": shape (2,), type "<i8">',
     'attributes': {'description': 'the electrodes that each spike unit came from',
      'namespace': 'hdmf-common',
      'neurodata_type': 'DynamicTableRegion',
      'object_id': '26483af7-474f-4a6b-bce2-0ade6eedfb4e',
      'table': 'root/general/extracellular_ephys/electrodes'}},
    'electrodes_index': {'data': '<HDF5 dataset "electrodes_index": shape (2,), type "<i8">',
     'attributes': {'namespace': 'hdmf-common',
      'neurodata_type': 'VectorIndex',
      'object_id': 'cd507cbc-d580-43de-9a25-696528f5bf4f',
      'target': 'root/units/electrodes'}},
    'id': {'data': '<HDF5 dataset "id": shape (2,), type "<i8">',
     'attributes': {'namespace': 'hdmf-common',
      'neurodata_type': 'ElementIdentifiers',
      'object_id': '760c8d28-a7b0-4782-9395-14647e9a733b'}}}}}}
```


dandi/metadata.py
```diff
@@ -83,3 +88,62 @@ def get_metadata(path):
     meta["nd_types"] = get_neurodata_types(path)
 
     return meta
+
+
+def nwb_to_json(fpath, out_fpath):
+    io = NWBHDF5IO(fpath, 'r')
```
yarikoptic, line 94: - needs a docstring
- could use `@metadata_cache.memoize_path`
- imho we need a function like this which doesn't dump it into a file -- should return the record (dict).  (see below about the name of `dataset_builder_to_json`), thus be RFed into `nwb_to_dict`.  Outside code could easily dump it into json or yaml depending on the use case

dandi/metadata.py
```diff
@@ -83,3 +88,62 @@ def get_metadata(path):
...
+    file = File(fpath)
+    io.read()
+    root_builder = io.get_builder(file['general']).parent
+    out_dict = group_builder_to_json(root_builder)
+    return json.dump(out_dict, out_fpath)
+
+
+def print_dataset(dataset):
+    return str(dataset)  # may want to change
```
yarikoptic, line 103: I don't see a point of having this function.  not to say it doesn't print anythning but rather `== str`

dandi/metadata.py
```diff
@@ -83,3 +88,62 @@ def get_metadata(path):
...
+
+
+def dataset_builder_to_json(dataset_builder):
```
yarikoptic, line 106: it doesn't return "JSON" (there is no such "thing") -- returns `dict`.  
Also it is nwb specific -- I feel that this and other (all?) nwb specific functions should go into `dandi/pynwb_utils.py` (it is ok to "interface" hdmf there too ;))

dandi/metadata.py
```diff
@@ -83,3 +88,62 @@ def get_metadata(path):
     meta["nd_types"] = get_neurodata_types(path)
 
     return meta
+
+
+def nwb_to_json(fpath, out_fpath):
+    io = NWBHDF5IO(fpath, 'r')
+    file = File(fpath)
```
yarikoptic, line 95: should use context manager(s), not plain assignments

dandi/metadata.py
bendichter: This is the simplest way I could think of to handle h5py Dataset objects, but I'm thinking we probably want to change this to expose more metadata about the dataset. Something like:
```
{'size': (100,5),
 'dtype': 'int',
 'path': '/acquisiton/ElectricalSeries/data',
 'compression':('gzip', 5)} 
```
so this function is more of a place-holder until we decide on how to do this

bendichter:
@yarikoptic thanks, I suspected you would have feelings about where this should go. I'll make these changes

bendichter:
@yarikoptic 

* moved to pynwb_utils
* added typehinting and docstrings
* refactored so you can have a metadata dictionary

I haven't implemented the tests though. I looked at your fixtures but wasn't sure how to actually create a file.

yarikoptic:
> I haven't implemented the tests though. I looked at your fixtures but wasn't sure how to actually create a file.

the point is that you don't need to create it -- fixture creates it.  See any test using `simple1_nwb` or `simple2_nwb`, such as e.g. the simplest https://github.com/dandi/dandi-cli/blob/master/dandi/tests/test_pynwb_utils.py#L27

dandi/tests/test_pynwb_utils.py
```diff
@@ -63,3 +63,7 @@ def search(v):
         )
         == "2.1.0"
     )
+
+
+def test_nwbfile_to_dict(simple1_nwb):
+    assert nwbfile_to_metadata_dict(simple1_nwb) == {'attributes': {'namespace': 'core', 'neurodata_type': 'NWBFile', 'nwb_version': '2.2.2', 'object_id': 'f852e236-37db-4e2b-81ef-37f22ba20bb4'}, 'datasets': {'file_create_date': {'data': '<HDF5 dataset "file_create_date": shape (1,), type "|O">'}, 'identifier': {'data': 'identifier1'}, 'session_description': {'data': 'session_description1'}, 'session_start_time': {'data': '2017-04-15T12:00:00+00:00'}, 'timestamps_reference_time': {'data': '2017-04-15T12:00:00+00:00'}}, 'groups': {'general': {'datasets': {'experiment_description': {'data': 'experiment_description1'}, 'experimenter': {'data': '<HDF5 dataset "experimenter": shape (1,), type "|O">'}, 'institution': {'data': 'institution1'}, 'keywords': {'data': '<HDF5 dataset "keywords": shape (2,), type "|O">'}, 'lab': {'data': 'lab1'}, 'related_publications': {'data': '<HDF5 dataset "related_publications": shape (1,), type "|O">'}, 'session_id': {'data': 'session_id1'}}}, 'stimulus': {'groups': {}}}}
```
bendichter, line 69: @yarikoptic I've got a noob question for you. What am I doing wrong here? It looks like a string comparison but they should both be dicts? I'm confused.

bendichter: I figured out the problem. Fresh uuids are generated from each object_id, so these aren't matching. Maybe I can add a seed or something.

yarikoptic: sorry, just saw it.  If there was a way to seed it - would indeed be great. Otherwise I guess we would need to provide some helper to compare structs/dicts ignoring some fields (may be even assuring that they change ;)).

bendichter:
to do: 
* test for link by looking at `GroupBuilder.parent`
* include object_id of links
* smarter handling of datasets

satra:
> smarter handling of datasets

@bendichter - when you say this, does this mean something inside the file? or outside?

bendichter:
@yarikoptic I showed this to the NWB dev team, and they are interested in supporting this within hdmf. I agree it would be good to expose this using the dandi cli, and I also agree that we'll have a better idea of the best way to do that when we understand better how this will be used in practice.

yarikoptic:
> @yarikoptic I showed this to the NWB dev team, and they are interested in supporting this within hdmf. 

GREAT! So, what should be our plan?  you are welcome to prep it here and we could start using it and then you would migrate it to hdmf and eventually we remove it here... or you could start straight in hdmf but then we would delay its adoption here:

> I agree it would be good to expose this using the dandi cli, and I also agree that we'll have a better idea of the best way to do that when we understand better how this will be used in practice.

To do this then you better develop it here and then migrate/adopt into hdmf ;)

bendichter:
Sounds like a plan

dandi/pynwb_utils.py
```diff
@@ -316,3 +320,77 @@ def copy_nwb_file(src, dest):
...
+
+    Parameters
+    ----------
+    node: Builder or attribute value
+    previous_node: Builder or attribute value
+        Used to detect links of node.parent != previous_node
+
+    Returns
+    -------
+
```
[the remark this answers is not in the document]
yarikoptic: actually, last line is `return str(node)`... probably shouldn't happen.  See my other comment about using `str()` in general here.  why do we want to serialize to string right here?

dandi/pynwb_utils.py
```diff
@@ -316,3 +320,77 @@ def copy_nwb_file(src, dest):
...
+            return {'type': 'link', 'object_id': node.attributes['object_id']}  # may want to change
+        else:
+            for category in ('attributes', 'datasets', 'groups'):
+                if hasattr(node, category) and getattr(node, category):
+                    out[category] = {key: traverse_node(val, node)
+                                     for key, val in getattr(node, category).items()
+                                     if traverse_node(val, node)}
+
+        if isinstance(node, DatasetBuilder):
+            out['data'] = str(node['data'])
```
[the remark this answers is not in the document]
yarikoptic: sorry if that would make schema more complicated ;-)

bendichter:
@yarikoptic you are right that scalar values and small lists should not be serialized. I'll fix that.

The bigger issue is how do we handle links and how do we handle `h5py.Dataset` objects. In both of these case, we do not want to load them. 

For links, we do not want to load them because that would just duplicate data in the output json file. My solution it so output e.g. 

```json
{"type": "link", "object_id": "[object_id of target]"}
```

For Datasets, they might be very big and the whole point of this is to separate small metadata from large data objects (with the exception of keywords. I regret making that an HDF5 Dataset. I did not realize at the time that it was possible to have an attribute by an array of variable length strings.)

For now, I am outputting the repr of `h5py.Dataset`, but I don't think this is the best solution. There is meta-data within `h5py.Dataset`s that we will want to pull out such as size, data type, chunking, compression, etc. For this, I think we should follow the lead of Zarr for using a json to express the structure of datasets

yarikoptic:
> For now, I am outputting the repr of `h5py.Dataset`, but I don't think this is the best solution. There is meta-data within `h5py.Dataset`s that we will want to pull out such as size, data type, chunking, compression, etc

So far I did not mind `'data': '<HDF5 dataset "data": shape (1000, 2), type "<f8">'` and I expected that if there is some chunking/compression - it would also be made visible (if not -- better be "fixed" at h5py level).  I do not mind looking into how zarr represents it though.

> The bigger issue is how do we handle links and how do we handle `h5py.Dataset` objects. In both of these case, we do not want to load them.
> 
> For links, we do not want to load them because that would just duplicate data in the output json file. My solution it so output e.g.
> 
> ```json
> {"type": "link", "object_id": "[object_id of target]"}
> ```

I think we should decide on the target purpose of such dumps.  Although may be specific treatments could be done on extracted structure later in the process of its target use. So what do we want to achieve with them? Are they primarily to provide

- **1** some lightweight (without heavy data) summary of the structure of .nwb file?
- **2** visualize metadata within nwb?
- **3** provide metadata so it is easy for a user to then programmatically or via our or pynwb cli/api to display specific portion of the .nwb?

We should keep in mind also that there are generic hdf5 viewers/browsers.   So it feels to me, at least for my own greedy purposes ATM **2** is of most interest but I see how it should be useful for **3** as well.  And I am afraid that the added aspect of **2** is in a conflict with **3**. E.g. example: many entities in the schema have `data` to probably depict that it is the data not metadata? but not really since use of `data` in `'identifier': {'data': 'EXAMPLE_ID'},` makes little sense to me. It makes it even into quite a non-easy to digest nesting like
```
'datasets': {'data': {'data': '<HDF5 dataset "data": shape (1000, 2), type "<f8">',
```
For visualization of metadata having this explicit `data` nesting in e.g. example of `identifier` only gives visual clutter, and I wondered if there is a way to consistently strip this explicit level of having `data` key in the visualization, so accenting what is presented on real metadata.

Also, continuing to `keywords` example , may other datasets seems to be wortwhile displaying (would even be shorter serialization probably), e.g. `'x': {'data': '<HDF5 dataset "x": shape (4,), type "<f8">',`, if it is 2 and/or 3 which is our accent.

bendichter:
Good points. I'm working on a rework over the next few days with a few of
the features you mention. The reason we have the extraneous lone 'data'
field is that it is a dataset that can have attributes. I could skip that
in the case that there are no attributes but I felt like it would make
search more difficult. So yes I agree that the exact form would depend on
how we would use it but the good news is it won't be that hard to change.

yarikoptic:
I feel like I could have used this functionality today ;)  So may be we should just merge, I would interface to  `ls`, we could add a note that actual data/format are still subject to change (like everything here ;)) and then things could be improved later? what do you think @bendichter ?

bendichter:
Sounds good to me

dandi/pynwb_utils.py
```diff
@@ -316,3 +320,77 @@ def copy_nwb_file(src, dest):
...
+    Returns
+    -------
+
+    """
+
+    if isinstance(node, Builder):
+        out = dict()
+        if (previous_node is not None) and (node.parent != previous_node):
+            # it's an imposter! (link)
+            return {'type': 'link', 'object_id': node.attributes['object_id']}  # may want to change
```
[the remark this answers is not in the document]
bendichter: I don't know exactly how this will be used, so I don't how how I want to express it yet.

bendichter:
@yarikoptic I have been working with Zarr to get it to read HDF5 files on S3. Here is the metadata it extracts from an HDF5 file in order to do so. It is very very similar to what we have here, though spread out across a bunch of json files (which are all hidden)

[zarrstore.zip](https://github.com/dandi/dandi-cli/files/4733079/zarrstore.zip)

bendichter:
I'm planning on computing this for each NWB file anyway, so maybe it would be better to use their structure

yarikoptic:
@bendichter I will close it for now -- just trying to get "dashboard" clean to not react to "oh two PRs to review here" all the time ;) don't feel discouraged if willing to pursue it forward - we can always reopen etc

PR #94, opened 2020-05-01
BF+RF: download - use regexes to prescribe known urls, support items and netlify

yarikoptic:
Closes gh-92 :

- supports download from netlify instance (going to the main girder)
- supports /v1/.../download girder urls

Also later added

- retry download up to 3 times (Closes gh-87)

Individual commit messages could provide more details

### DIFF
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -55,6 +55,7 @@ def parse_dandi_url(url):
       in case of multiple files
 
     """
+    lgr.debug("Parsing url %s", url)
     if "#" not in url:
         # assume that it was a dandi notation, let's try to follow redirects
         # TODO: make .head work instead of .get on the redirector
@@ -67,7 +68,7 @@ def parse_dandi_url(url):
         elif r.url != url:
             url = r.url
         else:
-            lgr.warning(f"Redirection did not happen for {url}")
+            lgr.warning("Redirection did not happen for %s", url)
 
     # We will just allow exception to escape if something goes wrong.
     # Warnings above could provide a clue in some cases
@@ -115,7 +116,9 @@ def parse_dandi_url(url):
             f" .*/(folder|collection|dandiset-meta)/ID24 or pointing to "
             f"individual selected items: {url}"
         )
-    return server, asset_type, asset_ids
+    res = server, asset_type, asset_ids
+    lgr.debug("Parsed into %s", res)
+    return res
 
 
 def download(
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -1,124 +1,157 @@
 import os.path as op
+import re
 import requests
-import urllib.parse as up
 
 from . import girder, get_logger
-from .consts import dandiset_metadata_file
+from .consts import dandiset_metadata_file, known_instances
 from .dandiset import Dandiset
+from .exceptions import FailedToConnectError, NotFoundError, UnknownURLError
 from .utils import flatten, flattened, Parallel, delayed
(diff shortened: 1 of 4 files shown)


dandi/girder.py
```diff
@@ -375,7 +381,22 @@ def download_file(self, file_id, path, existing="error", attrs=None):
...
         # when RF - do not forget to remove progressReporterCls in __init__
-        self.downloadFile(file_id, path)
+
+        # Will do 3 attempts to avoid some problems due to flaky/overloaded
+        # connections, see https://github.com/dandi/dandi-cli/issues/87
+        for attempt in range(3):
+            try:
+                self.downloadFile(file_id, path)
+                break
+            except gcl.HttpError as exc:
```
yarikoptic, line 391: this piece of code is not even smoke tested so whoever finds a bug, get a cookie!

dandi/download.py
```diff
@@ -1,121 +1,172 @@
...
+    # into module space, and later we might end up with classes for those anyways
+    id_regex = "[a-f0-9]{24}"
+    id_grp = f"(?P<id>{id_regex})"
+    server_grp = "(?P<server>(?P<protocol>https?)://(?P<hostname>[^/]+)/)"
+    known_urls = {
+        # Those we first redirect and then handle the redirected URL
+        # TODO: Later should better conform to our API, so we could allow
+        #       for not only "dandiarchive.org" URLs
+        "https?://dandiarchive.org/.*": {"handle_redirect": True},
+        "https?://[^/]*dandiarchive-org.netlify.app/.*": {"map_instance": "dandi"},
```
mgrauer, line 25: it seems odd to support netlify, as these domains are not part of our production infrastructure. i'm guessing this makes it easier for you to test/interact with the previews?

satra: for testing dev/master previews, should we deploy a version of it on github simultaneously (through github actions)? then at least it could be reduced to `https://*dandiarchive.org/.*`

yarikoptic: In principle, I could may be make it into development option - available only if DANDI_DEVEL set. But since this is not visible to the user anyways, be it wouldn't really help anything IMHO.
The rationale for adding it - being able to test client on a proposed Ina PR deployment. If we somehow could, it would've been nice to even automate it in dandiarchive tests - feed client a few urls archive is "producing" in its pr state.

yarikoptic: As for an alternative deployment - if way to do it easily per or then sure why not, but then why do we use netlify?

satra: netlfiy allows us to preview any PR easily. while that can be done with github actions, it's a bit more complicated because of the workflow. github actions is replacing what a lot of people used netlify/travis/circle for.

yarikoptic: ok, then for now let's proceed as is ;-)

yarikoptic:
ok, since it seems @mgrauer and @satra had reviewed it at least to some extent, let's merge and then fixup if anything needs to be fixed while we are testing it in real life before kicking it out as a release

dandi/download.py
mgrauer: In the future, I'd like to be able to get away from URLs as an API between the front end and the CLI. I'd prefer something more like where the web UI tells you "use this command to download this dandiset: `dandi download 000023`".

yarikoptic: In the future, I'd prefer web ui to use our API urls ;-)

PR #97, opened 2020-05-07
move etelemetry to main

### LINKED ISSUE #95: etelemetry doesn't seem to run consistently
this line https://github.com/dandi/dandi-cli/blob/master/dandi/__init__.py#L44 returns `False` when dandi cli is called in the terminal.

```
{'__name__': '__main__', '__doc__': None, '__package__': None, '__loader__': <_frozen_importlib_external.SourceFileLoader object at 0x107f7b250>, 
'__spec__': None, '__annotations__': {}, '__builtins__': <module 'builtins' (built-in)>, 
'__file__': '/Users/satra/software/miniconda3/envs/dandi/bin/dandi', '__cached__': None,
 're': <module 're' from '/Users/satra/software/miniconda3/envs/dandi/lib/python3.7/re.py'>, 
'sys': <module 'sys' (built-in)>}
```

so possibly remove the if statement and call it anytime?

satra:
closes #95

### DIFF
--- a/dandi/cli/command.py
+++ b/dandi/cli/command.py
@@ -175,6 +175,15 @@ def main(log_level, pdb=False):
         from ..utils import setup_exceptionhook
 
         setup_exceptionhook()
+    try:
+        import etelemetry
+
+        etelemetry.check_available_version("dandi/dandi-cli", __version__, lgr=lgr)
+    except Exception as exc:
+        lgr.warning(
+            "Failed to check for a more recent version available with etelemetry: %s",
+            exc,
+        )
 
 
 #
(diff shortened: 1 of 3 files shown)


satra:
@yarikoptic - the current cli release will not download a dandiset. i think we should put all prior versions into bad versions. i have not done this here, but can add them in.

till we get to stable API, CLI has to be always upgraded. unless we have separate notions of version for each part of the cli.

yarikoptic:
Thanks! Will release later today

yarikoptic:
For the geeks of it, announced all prior versions as bad in aee1872002023ecb5098b7168e8276e33d155ca6 ;-)

dandi/cli/command.py
```diff
@@ -175,6 +175,15 @@ def main(log_level, pdb=False):
         from ..utils import setup_exceptionhook
 
         setup_exceptionhook()
+    try:
+        import etelemetry
```
yarikoptic, line 179: mental note: apparently according to coverage annotation - none of this code is tested, so whatever invocation we use in the tests, it is not using this top level `main`

Issue #99, opened 2020-05-12
allow for redirector URLs (e.g., identifiers.org, bit.ly)


satra:
```
$ dandi download https://identifiers.org/DANDI:000004
Error: We do not know how to map URL https://identifiers.org/DANDI:000004 to girder.
Regular expressions for known setups:
 - https?://dandiarchive.org/.*
 - https?://[^/]*dandiarchive-org.netlify.app/.*
 - (?P<server>(?P<protocol>https?)://(?P<hostname>[^/]+)/)#.*/(?P<asset_type>folder|collection|dandiset)/(?P<id>[a-f0-9]{24})$
 - (?P<server>(?P<protocol>https?)://(?P<hostname>[^/]+)/)#/folder/[a-f0-9]{24}/selected(?P<multiitem>(/item\+(?P<id>[a-f0-9]{24}))+)$
 - (?P<server>(?P<protocol>https?)://(?P<hostname>[^/]+)/)api/v1/(?P<asset_type>item)/(?P<id>[a-f0-9]{24})/download$
```

more generally follow redirects to determine if you can parse a url, should also help in the future with dois.

yarikoptic:
should be just a matter of adding 
```
"https?://identifiers.org/DANDI:.*": {"handle_redirect": True},
```
to https://github.com/dandi/dandi-cli/blob/master/dandi/download.py#L20

satra:
i meant to go more general. first doing redirect and then checking on whether urls map onto our services. this would allow more generalizability to people creating all kinds of shortcuts (bit.ly, etc.,.)

yarikoptic:
oh... then I will keep this one open.  Meanwhile I will submit a dedicated fix PR for identifiers

yarikoptic:
to support generally -- we could just add catch all 
```
"https?://.*": {"handle_redirect": True},
```
at the end of the list. But we would need to check first if our current code doesn't cause immediate download of the content and just follows redirect, and either in general it would be a good/safe idea ;-)

PR #108, opened 2020-05-27
ENH+OPT: Issue HEAD not GET while checking redirects (whenever we can)

yarikoptic:
and add support for arbitrary redirectors.  Closes: #99 

After https://github.com/dandi/redirector/issues/13 resolved ad-hoc kludge could go away


yarikoptic:
a single failure on 18.04 3.6 is unrelated timeout , so overall it works

yarikoptic:
ok, will be tested in the battle field I guess ;)

PR #113, opened 2020-06-02
RF+ENH: support using probe_id to disambiguate (before obj_id)

yarikoptic:
Apparently we had no issue open for support disambiguation with a sample case of https://dandiarchive.org/dandiset/000022 .

It also "fixes" obviously wrong
```python
mandatory_if_not_empty=["modalities,"]
```
which I am not yet 100% sure how played its evil roles since awhile.

TODOs:
- [x] a unittest to verify correct operation

### DIFF
--- a/dandi/consts.py
+++ b/dandi/consts.py
@@ -23,7 +23,7 @@ metadata_nwb_subject_fields = (
     "subject_id",
 )
 
-metadata_nwb_dandi_fields = ("cell_id", "slice_id", "tissue_sample_id")
+metadata_nwb_dandi_fields = ("cell_id", "slice_id", "tissue_sample_id", "probe_ids")
 
 metadata_nwb_computed_fields = (
     "number_of_electrodes",
@@ -53,6 +53,7 @@ metadata_dandiset_fields = (
     "variables_measured",
     "sex",
     "organism",
+    "probe_ids",
     "number_of_subjects",
     "number_of_cells",
     "number_of_tissue_samples",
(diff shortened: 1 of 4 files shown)


yarikoptic:
Ideally: I need datalad datasets for all dandisets so I could quickly and safely go through them to see possible side-effects of new `organize` features... but that is not to happen "today", thus merging to proceed.

satra:
>  I need datalad datasets for all dandisets so I could quickly and safely go through them to see possible side-effects of new organize features

@yarikoptic - is this something i can help with? re-use piece of download code to walk through and get urls?

yarikoptic:
I am afraid it is more of "download a dandiset", run "organize", check the diff if any (should be none).

If they were datalad datasets -- just a matter of `git status` at the end and then `git reset --hard; git clean -dfx` to bring back to original state ;)  That is what I had been using locally on subset of files within dandisets 22 and 23 locally

Ideally, with some coding we could do a "quick" check across all dandisets without download I believe... filed separate https://github.com/dandi/dandi-cli/issues/114 with details.

satra:
i understand the checking requires download, but presumably creating the datalad dataset would not require download.

yarikoptic:
#114 hypothesises that download could be avoided. It is currently required if someone would want to check

yarikoptic:
> re-use piece of download code to walk through and get urls?

I guess I have missed that. Yes - it is in the heart of the #114

yarikoptic:
ok, time will show and otherwise we wouldn't know, thus going forward

Issue #106, opened 2020-05-25
upload: add "digests" sub-dictionary during upload


yarikoptic:
Ideally should be computed while uploading, but for now could be a separate call before upload.

PR #115, opened 2020-06-02
ENH: Upload digests (only 4 checksums for now)

yarikoptic:
3 right away to capture popular checksums which could possibly later help to identify this file in other places (e.g. datalad uses md5 by default, git annex - sha256, and sha1 iirc used by git)

Theoretically we could provide some progress indication while computing digests which would be valuable for large files, but IMHO not a priority

Unfortunately there is still no unittests for upload, so nothing is tested beyond local tries.

@satra - please recommend another name than `uploaded_digests` for the metadata field.  I kept `uploaded_` prefix consistent with other metadata fields I add upon upload.

TODOs
- [x] (now) add validation upon download that file received correctly.  Will be a naive implementation doing it after download of a file completes.
- (later, possibly) extend with digests for s3 digest with our blocking setting, default ipfs

Closes gh-106

### DIFF
--- a/dandi/cli/cmd_upload.py
+++ b/dandi/cli/cmd_upload.py
@@ -22,6 +22,7 @@ from ..consts import (
     dandiset_identifier_regex,
     dandiset_metadata_file,
     known_instances,
+    metadata_digests,
 )
 
 
@@ -417,8 +418,8 @@ def upload(
                 # TODO: in theory we could also cache the result, but since it is
                 # critical to get correct checksums, safer to just do it all the time.
                 # Should typically be faster than upload itself ;-)
-                digester = Digester(["md5", "sha1", "sha256"])
-                file_metadata_["uploaded_digests"] = digester(path)
+                digester = Digester(metadata_digests)
+                file_metadata_.update(digester(path))
             except Exception as exc:
                 yield skip_file("failed to compute digests: %s" % str(exc))
                 return
--- a/dandi/cli/cmd_upload.py
+++ b/dandi/cli/cmd_upload.py
@@ -97,6 +97,7 @@ def upload(
     """
     from pathlib import Path, PurePosixPath
     from ..dandiset import Dandiset
+    from ..support.digests import Digester
 
     dandiset = Dandiset.find(dandiset_path)
     if not dandiset:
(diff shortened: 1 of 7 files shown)


yarikoptic:
above I meant "@satra feel free to recommend..." -- i.e. you don't have to ;)

satra:
for checksums can we follow simple approach just use  `key: value` instead of trying to embed it in some structure.  just make sure the keys correspond to names here:  http://id.loc.gov/vocabulary/preservation/cryptographicHashFunctions.html (in lowercase) for an asset we will take this info directly into `assetuuid.yaml`.

yarikoptic:
just say how it should look instead of  smth like
```
'uploaded_datetime`: '2020-06-02T18:17:16.239367-04:00',
'uploaded_digests':
{ "md5": "5a5e7a369ef2150492fe59d75daf60c9", "sha1": "ab244c2d5b603d0e54a087e81af78bdf253b08ba", "sha256": "c7b815ce14baa20b056da5e77e828ade913d29065493e00e5a8e451a0f44d761" }
'uploaded_mtime': '2020-06-02T18:14:24.338529-04:00',
...
```
if we just add all of them at the top level (e.g just "md5", "sha1", etc), later we will have hard(er) time fishing out what fields to kill/alter whenever e.g. we decide to alter file on the server side.  Although flatness indeed makes it easier to render, it also makes it easier to provide custom rendering for a specific field with a clear semantic for a collection of arbitrary `key: value` pairs.

satra:
here you go:

```
'uploaded_datetime`: '2020-06-02T18:17:16.239367-04:00',
'uploaded_mtime': '2020-06-02T18:14:24.338529-04:00',
"md5": "5a5e7a369ef2150492fe59d75daf60c9", 
"sha1": "ab244c2d5b603d0e54a087e81af78bdf253b08ba", 
"sha256": "c7b815ce14baa20b056da5e77e828ade913d29065493e00e5a8e451a0f44d761" ,
...
```

yarikoptic:
Eh heh, ok, will do. At least it would make web rendering nicer without extra changes ;-)

yarikoptic:
FTR @magland is using md5 and sha1 in https://github.com/flatironinstitute/kachery so we have those covered - good ;-)

yarikoptic:
ok, Windows is happy, @satra should be happy, let's proceed

PR #116, opened 2020-06-03
BF+ENH: allow for upload of any file (but not that easily for now)

yarikoptic:
commit 808e5894af0c4b613d13fbba444a03a1ff97164b (HEAD -> upload-all, origin/upload-all)

    NF: upload - --allow-any-path development option
    
    Coupled with  --validation ignore  it would allow upload of any file (not only .nwb,
    and dandiset.yaml).  Metadata record for any non .nwb file would be pretty much empty
    ATM BUT it will contain the digests (checksums ATM), thus allowing to verify downloads etc.
    
    Note -- it is a "development" option, so will not be visible/available by default.
    Informed users would need to set DANDI_DEVEL=1 environment variable first.

### DIFF
--- a/dandi/cli/cmd_upload.py
+++ b/dandi/cli/cmd_upload.py
@@ -71,7 +71,7 @@ from ..consts import (
 )
 @devel_option(
     "--allow-any-path",
-    help="For development: allow not only DANDI 'supported' file types/paths",
+    help="For development: allow DANDI 'unsupported' file types/paths",
     default=False,
     is_flag=True,
 )
--- a/dandi/cli/cmd_upload.py
+++ b/dandi/cli/cmd_upload.py
@@ -405,9 +405,9 @@ def upload(
                     yield skip_file("failed to extract metadata: %s" % str(exc))
                     if not file_recs:
                         # remove empty item
-                        yield {"status", "deleting empty item"}
+                        yield {"status": "deleting empty item"}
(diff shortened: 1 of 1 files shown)


dandi/cli/cmd_upload.py
```diff
@@ -69,6 +69,12 @@
     default=False,
     is_flag=True,
 )
+@devel_option(
+    "--allow-any-path",
+    help="For development: allow not only DANDI 'supported' file types/paths",
```
satra, line 74:
```suggestion
    help="For development: allow DANDI 'unsupported' file types/paths",
```

dandi/cli/cmd_upload.py
```diff
@@ -180,13 +192,15 @@ def upload(
 
     # Expand and validate all paths -- they should reside within dandiset
     orig_paths = paths
-    paths = list(find_dandi_files(paths))
+    paths = list(find_files(".*", paths) if allow_any_path else find_dandi_files(paths))
```
satra, line 195: should we allow `.*` (i.e. hidden files?) perhaps we restrict this to non-hidden files

yarikoptic: `.*` here is a regex, and hidden files by default are ignored by `find_files` - there is a dedicated option to allow for them.  I guess regex here could become `[^.].*` but currently effect would be the same, so imho ok to leave as is

Issue #111, opened 2020-06-01
tqdm: errors in the tests with "AssertionError: attempt to release recursive lock not owned by thread"


yarikoptic:
Not a good day for tqdm I guess. Now (after #110 on smaug) it errors tests locally 
```
_________________________________________ test_download_multiple_files __________________________________________

tmpdir = local('/home/yoh/.tmp/pytest-of-yoh/pytest-657/test_download_multiple_files0')

    @mark.skipif_no_network
    def test_download_multiple_files(tmpdir):
        url = (
            "https://gui.dandiarchive.org/#/folder/5e70d3173da50caa9adaf334/selected/"
            "item+5e70d3173da50caa9adaf335/item+5e70d3183da50caa9adaf336"
        )
    
>       ret = download(url, tmpdir)

dandi/tests/test_download.py:74: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
dandi/download.py:230: in download
    for file in files
venvs/dev3/lib/python3.7/site-packages/joblib/parallel.py:1017: in __call__
    self.retrieve()
venvs/dev3/lib/python3.7/site-packages/joblib/parallel.py:909: in retrieve
    self._output.extend(job.get(timeout=self.timeout))
/usr/lib/python3.7/multiprocessing/pool.py:657: in get
    raise self._value
/usr/lib/python3.7/multiprocessing/pool.py:121: in worker
    result = (True, func(*args, **kwds))
venvs/dev3/lib/python3.7/site-packages/joblib/_parallel_backends.py:608: in __call__
    return self.func(*args, **kwargs)
venvs/dev3/lib/python3.7/site-packages/joblib/parallel.py:256: in __call__
    for func, args, kwargs in self.items]
venvs/dev3/lib/python3.7/site-packages/joblib/parallel.py:256: in <listcomp>
    for func, args, kwargs in self.items]
dandi/girder.py:389: in download_file
    self.downloadFile(file_id, path)
venvs/dev3/lib/python3.7/site-packages/girder_client/__init__.py:1207: in downloadFile
    length=int(req.headers.get('content-length', 0))) as reporter:
dandi/girder.py:490: in __init__
    self._pbar = tqdm.tqdm(desc=label, total=length, unit="B", unit_scale=True)
[...]

PR #117, opened 2020-06-04
ENH: add uploaded_by field which would inform what version of dandi client uploaded it

yarikoptic:
I should have done it before 0.5.0 ;-)


yarikoptic:
@satra @mgrauer - any objections/recommendations?

satra:
nope

yarikoptic:
failed test runs seems to be due to #111

mgrauer:
seems like it will be helpful for debugging, no objections

PR #122, opened 2020-06-08
[WIP] add publish subcommand to test organization of dandiset

satra:
This follows up on a s3 based layout for a dandiset (see: https://github.com/dandi/dandiarchive/issues/355)

- [X] download dandiset metadata
- [ ] perform conversion of old schema to new schema for dandiset and assets
- [ ] create manifest demonstrating proposed layout.

### DIFF
--- a/dandi/publish.py
+++ b/dandi/publish.py
@@ -1,0 +1,118 @@
+import os.path as op
+import re
+import requests
+
+from . import girder, get_logger
+from .consts import dandiset_metadata_file, known_instances, metadata_digests
+from .dandiset import Dandiset
+from .exceptions import FailedToConnectError, NotFoundError, UnknownURLError
+from .utils import flatten, flattened, Parallel, delayed
+from .download import parse_dandi_url, _get_asset_files
+
+lgr = get_logger()
+
+
+def publish(
+    dandiset_url,
+    base_dir,
+    *,
+    existing="error",
+    jobs=6,
+    get_files=True,
+    authenticate=False,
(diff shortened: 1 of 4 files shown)


dandi/publish.py
```diff
@@ -0,0 +1,118 @@
...
+            r = requests.head(girder_url)
+            if r.status_code == "404":
+                raise NotFoundError(f"{girder_url} not found")
+            aws_url = r.headers["Location"]
+            aws_url = aws_url.split("?")[0]
+            file["dandiURL"] = f"https://api.dandiarchive.org/v1/asset/{file['id']}"
+            file["awsURL"] = aws_url
+            manifest.append(file)
+    dandiset_meta["assets"] = manifest
+    return dandiset_meta
```
yarikoptic, line 118: @satra - may be I misunderstood, on the call you said that you had extracted asset level metadata according to new schema, but I do not see that here...

satra: not according to new schema. this just cycles through an extracts the metadata currently on the server. for dandiset, the convert.py script updates it to the new schema. the same would have to be done for asset (not done yet).

i thought you were asking for something that recursively downloads all the metadata, and this would do that.

yarikoptic: Since dandi-cli should do its own extraction/upload of metadata (I guess we better shift that code to dandi-core as well?), where is that `convert.py` exactly?  https://github.com/search?q=org%3Adandi+convert.py&type=Repositories didn't gimme anything

PR #124, opened 2020-06-24
make mktemp compatible with MacOS and linux

nicolocin:
There's still an error with `sed` in the `basic_workflow1.sh` script that I don't know how to fix.  If I remove the indent step in sneak the script runs.


yarikoptic:
thanks!

Issue #27, opened 2019-10-19
upload: establish testing


yarikoptic:
#26 introduced `upload` command.  We should enable some kind of unittesting for it

It seems that github actions support docker: https://help.github.com/en/articles/creating-a-docker-container-action and even I found an issue in a now removed "docker" action repository issues: https://webcache.googleusercontent.com/search?q=cache:mPPb1xbgUgEJ:https://github.com/actions/docker/issues/11+&cd=2&hl=en&ct=clnk&gl=us which suggests that docker compose is also available.

But we would need firsts to establish some user creds and obtain the key programmatically first.  @mgrauer - can you help with that?

mgrauer:
We are working on a [docker compose setup](https://github.com/dandi/dandiarchive/pull/53) for the dandiarchive, which might be a solution here. That might also end up being too heavyweight and perhaps we should just mock a few endpoints instead. After we finish that PR I can think about this issue more.

yarikoptic:
There is a working docker compose setup, so it is time to establish within tests (on CI or not) a fixture which would use docker-compose to start an instance of the archive.  Possibly of help and code to adopt: within reproman we have "fixture generators" https://github.com/ReproNim/reproman/blob/master/reproman/tests/fixtures.py#L23  to create fixtures needing specific docker or singularity container etc.  May be could come useful here.

jwodder:
@yarikoptic The images for the docker-compose setup are built by copying in the dandiarchive source, so either dandi-cli's GitHub actions need to check out a copy of dandiarchive (not the cleanest idea) or else the dandiarchive images need to be pushed to Docker Hub or a similar registry.

yarikoptic:
Ok, we already have `dandiarchive` organization on hub.docker  linked to github via `dandibot` user... I have initiated a build for dandiarchive/dandiarchive-client - let's see if that one works out.  I will update whenever I know more.
@jwodder - do you have hub.docker.com account?  I could add you to dandiarchive organization

jwodder:
@yarikoptic Yes; username: jwodder

yarikoptic:
ok, added you to the owners! (as well as developers for that first repo dandiarchive-client).  So I have prepared one for dandiarchive-client, seems to work fine! now up to you @jwodder to link to other components (I have made it to rebuild for master and for any tagged release which are yet to come in dandiarchive).

jwodder:
[...]

PR #130, opened 2020-07-08
ENH: upload - add locking of the dandiset

### LINKED ISSUE #125: Lock dandisets during modification
`dandiarchive` now supports locking dandisets in girder so that other users cannot modify or publish them. The publish workflow is still unreleased (as of 6/24), but the functionality should be implemented in the CLI before release as a safeguard.

There are 3 new endpoints related to locking:
- [/dandi/{identifier}/lock](https://girder.dandiarchive.org/api/v1#!/dandi/dandi_wrapper_0_1) locks a dandiset with the currently authenticated user. If the dandiset is already locked, even by the currently authenticated user, the call will fail. If locking fails, the operation should abort.
- [/dandi/{identifier}/lock/owner](https://girder.dandiarchive.org/api/v1#!/dandi/dandi_wrapper_0_1_2) returns some info on the current owner of the lock on the dandiset. If the dandiset is unlocked, the response body will be `null`. If the dandiset is currently being published, the response will look like this:
```
{
  "email": "publish@dandiarchive.org",
  "firstName": "publish",
  "lastName": "service"
}
```
- [/dandi/{identifier}/unlock](https://girder.dandiarchive.org/api/v1#!/dandi/dandi_wrapper_0_1_2_3_4_5_6) unlocks the dandiset. If the dandiset is not locked by the currently authenticated user, the call will fail.

Before doing any modification to a dandiset in girder (i.e. upload), the CLI should call the lock endpoint. Afterwards, it should call the unlock endpoint.
I recommend a pattern like what was used in [`dandi-publish`](https://github.com/dandi/dandi-publish/blob/master/publish/girder.py#L78):
```
@contextlib.contextmanager
def dandiset_lock(client, dandiset_identifier: str):
    resp = client.post(f'dandi/{dandiset_identifier}/lock')
    if resp.status_code != 200:
        raise GirderError(f'Failed to lock dandiset {dandiset_identifier}')
    try:
        yield
    finally:
        resp = client.post(f'dandi/{dandiset_identifier}/unlock')
        if resp.status_code != 200:
            raise GirderError(f'Failed to unlock dandiset {dandiset_identifier}')
```
This will create a method that can be used in a python `with` statement (like [this](https://github.com/dandi/dandi-publish/blob/master/publish/tasks.py#L17)):
```
with dandiset_lock(client, dandiset.identifier):
    ... do things ...
```

yarikoptic:
It locks the entire dandiset for the entire operation (loop across all
files), even if no actual upload would happen (e.g. if all files
already uploaded and recent).

It has a few TODO items added in the code which ideally should be
resolved before we proceed with merging this PR.

Closes #125 since ATM `upload` is the only one which modifies  an existing dandiset

### DIFF
--- a/dandi/girder.py
+++ b/dandi/girder.py
@@ -1,7 +1,8 @@
-import os
-import os.path as op
+import contextlib
 import json
 import keyring
+import os
+import os.path as op
 import random
 import sys
 import time
@@ -12,6 +13,7 @@ from pathlib import Path
 import girder_client as gcl
 
 from . import get_logger
+from .exceptions import LockingError
 from .utils import ensure_datetime, is_same_time
 from .consts import known_instances_rev, metadata_digests
 from .support.digests import Digester
@@ -31,6 +33,8 @@ class GirderNotFound(Exception):
     pass
 
 
+# TODO: remove if we start to expose this as a Python library which could
+# be longer lived than just a CLI
 @lru_cache(1024)
 def lookup(client, name, asset_type="collection", path=None):
     """A helper for common logic while looking up things on girder by name"""
@@ -447,6 +451,29 @@ class GirderCli(gcl.GirderClient):
         # If that one was not provided, the best we know is the "ctime"
         # for the file, use that one
         return ensure_datetime(attrs.get("mtime", attrs.get("ctime", None)))
+
+    @contextlib.contextmanager
+    def lock_dandiset(self, dandiset_identifier: str):
+        presumably_locked = False
+        try:
+            lgr.debug("Trying to acquire lock for %s", dandiset_identifier)
+            resp = self.post(f"dandi/{dandiset_identifier}/lock")
+            presumably_locked = True
+            # TODO: remove "if resp" 2 instances below after
+            #  https://github.com/dandi/dandiarchive/issues/421 is fixed up
+            if resp and resp.status_code != 200:
+                presumably_locked = False
+                raise LockingError(f"Failed to lock dandiset {dandiset_identifier}")
+
+            yield
+        finally:
+            if presumably_locked:
+                lgr.debug("Trying to release the lock for %s", dandiset_identifier)
+                resp = self.post(f"dandi/{dandiset_identifier}/unlock")
+                if resp and resp.status_code != 200:
+                    raise LockingError(
+                        f"Failed to unlock dandiset {dandiset_identifier}"
+                    )
 
 
 # TODO: our adapter on top of the Girder's client to simplify further
--- a/dandi/girder.py
+++ b/dandi/girder.py
@@ -457,20 +457,20 @@ class GirderCli(gcl.GirderClient):
         presumably_locked = False
         try:
             lgr.debug("Trying to acquire lock for %s", dandiset_identifier)
-            resp = self.post(f"dandi/{dandiset_identifier}/lock")
-            presumably_locked = True
-            # TODO: remove "if resp" 2 instances below after
-            #  https://github.com/dandi/dandiarchive/issues/421 is fixed up
-            if resp and resp.status_code != 200:
-                presumably_locked = False
+            try:
+                self.post(f"dandi/{dandiset_identifier}/lock")
+            except gcl.HttpError:
                 raise LockingError(f"Failed to lock dandiset {dandiset_identifier}")
+            else:
+                presumably_locked = True
 
             yield
         finally:
             if presumably_locked:
                 lgr.debug("Trying to release the lock for %s", dandiset_identifier)
-                resp = self.post(f"dandi/{dandiset_identifier}/unlock")
-                if resp and resp.status_code != 200:
+                try:
+                    self.post(f"dandi/{dandiset_identifier}/unlock")
+                except gcl.HttpError:
                     raise LockingError(
                         f"Failed to unlock dandiset {dandiset_identifier}"
                     )
--- a/dandi/tests/test_girder.py
+++ b/dandi/tests/test_girder.py
@@ -1,0 +1,37 @@
+import pytest
+from ..exceptions import LockingError
+from .. import girder
+
+
+def test_lock_dandiset(local_docker_compose_env):
+    DANDISET_ID = "000001"
+    client = girder.get_client(local_docker_compose_env["instance"].girder)
+    resp = client.get(f"dandi/{DANDISET_ID}/lock/owner")
+    assert resp is None
+    with client.lock_dandiset(DANDISET_ID):
+        resp = client.get(f"dandi/{DANDISET_ID}/lock/owner")
+        assert resp is not None
+    resp = client.get(f"dandi/{DANDISET_ID}/lock/owner")
+    assert resp is None
+
+
+def test_lock_dandiset_error_on_nest(local_docker_compose_env):
+    DANDISET_ID = "000001"
+    client = girder.get_client(local_docker_compose_env["instance"].girder)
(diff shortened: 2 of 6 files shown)


yarikoptic:
it was locked by https://github.com/dandi/dandiarchive/issues/422 which is addressed now.  https://github.com/dandi/dandiarchive/issues/421 is still open but we have a workaround for it in the code here... depending on how/if it would be resolved, we should act accordingly.

yarikoptic:
I will assign @jwodder to proceed with this PR further (after some local testing...)

As for the unit-testing, there is a good elderly https://github.com/dandi/dandi-cli/issues/27 which could now be addressed as well.

jwodder:
@yarikoptic After rebasing this branch on top of PR #158, trying to upload fails at the locking API call with:

```
girder_client.HttpError: HTTP error 401: POST http://localhost:8080/api/v1/dandi/000006/lock
Response text: {"message": "You must be logged in.", "type": "access"}
```

I have `DANDI_API_KEY` set, and I can successfully upload with the key using the branch in PR #158.

mgrauer:
@jwodder You will need a new API key with additional permissions. This gets provided via the GUI in https://github.com/dandi/dandiarchive/pull/439. 

What you can do for now is to go into the girder you are working with, go to your user page, under your API keys, and set the dandicli key as a full user permissions key.

jwodder:
@mgrauer Is there a way to fetch an appropriate API key via the API, or is this GUI-only?  This matters for #27 and #159.

jwodder:
So after changing the API key's scope and throwing away the bad commit from #158, `dandi upload -i local-docker` fails at the API call to `dandi/{dandiset_identifier}/lock` because the server returns a 500 response.  The `/root/.girder/logs/error.log` file from the Girder container contains the following:

```
[2020-08-04 16:16:20,724] ERROR: 500 Error
  File "/girder/girder/api/rest.py", line 648, in endpointDecorator
    val = fun(self, path, params) 
  File "/girder/girder/api/rest.py", line 1223, in POST
    return self.handleRoute(method, path, params)
  File "/girder/girder/api/rest.py", line 965, in handleRoute
    val = handler(**kwargs)
  File "/girder/girder/api/access.py", line 57, in wrapped
    return fun(*args, **kwargs)
  File "/girder/girder/api/describe.py", line 676, in wrapped
    return fun(*args, **kwargs)
  File "/home/girder-dandi-archive/girder_dandi_archive/util.py", line 57, in wrapper
    return func(*args, **kwargs)
  File "/home/girder-dandi-archive/girder_dandi_archive/rest.py", line 333, in lock_dandiset
    locking.lock(identifier, self.getCurrentUser())
  File "/home/girder-dandi-archive/girder_dandi_archive/locking.py", line 91, in lock
    _set_lock(identifier, user)
  File "/home/girder-dandi-archive/girder_dandi_archive/locking.py", line 33, in _set_lock
    dandiset_folder["locked"] = user["_id"]
TypeError: 'NoneType' object does not support item assignment
Additional info:
  Request URL: POST http://localhost:8080/api/v1/dandi/000007/lock
  Query string:
  Remote IP: 172.22.0.1
  Request UID: 1eef0e1a-712d-49d2-a9a1-ef46a41e3a00
```

jwodder:
After editing the code to detect nonexistent Dandisets and then running `dandi register` followed by `dandi upload`, I can confirm (via Girder's REST API web interface) that a lock exists on the Dandiset while uploading.  Other than automated testing — which depends on #27 and #159 — and the TODOs in the code — neither of which apply right now — is there anything else that need to be done for this PR?

yarikoptic:
Given the @mgrauer [opinion](https://github.com/dandi/dandiarchive/issues/421#issuecomment-668268587) about this locking be a temporary measure (while we do not have non-girder API), and the fact that we have workarounds in the code, let's proceed.  If you have a moment @jwodder , please resolve conflicts and push (or force push if rebased).  If not -- I will try to do that later today.

jwodder:
@yarikoptic Conflicts resolved.

mgrauer:
@jwodder 

>Is there a way to fetch an appropriate API key via the API

yes this is all available via the API, look here for the specific route: https://girder.dandiarchive.org/api/v1#!/api95key/api_key_createKey

If you make the post request with an empty `scope` it should create an API key with full user permissions, which is what we need.

yarikoptic:
#159 is almost done, so let's just wait for it to land in master, and then furnish a test here for lock operation.  Due to all the workarounds I would really prefer to have it unit-tested before it lands in master.

yarikoptic:
I think this PR is now ready to be continued with and gain some basic testing of locking to work out correctly.

jwodder:
@yarikoptic How should the tests test for locking?  Should multiprocessing or threading be used to query the lock status while the upload is in progress?  Should the responses library be used to mock the locking API call?  Should pytest-mock's `spy` object be used to assert that some part of the locking code is called?

yarikoptic:
I would have first just tested separately [lock_dandiset](https://github.com/dandi/dandi-cli/pull/130/files#diff-b42fa95e63418eb6dc476c50c5b00299R456) -- that it doesn't allow for additional lock to be issued if there is one already etc; as for upload - indeed, might be best to just mock it that it would use `lock_dandiset` "once" for the upload (I do not remember if I ever mocked context managers and if it would be as easy as a regular function).

jwodder:
@yarikoptic Done.

dandi/girder.py
```diff
@@ -31,6 +33,8 @@ class GirderNotFound(Exception):
     pass
 
 
+# TODO: remove if we start to expose this as a Python library which could
+# be longer lived than just a CLI
```
yarikoptic, line 37: +1

dandi/upload.py
```diff
@@ -108,6 +101,14 @@ def upload(
 
     lgr.debug("Working with collection %s", collection_rec)
 
+    try:
+        girder.lookup(client, girder_collection, path=girder_top_folder)
+    except girder.GirderNotFound:
+        raise ValueError(
```
yarikoptic, line 107: please add a quick test for this codepath (upload to some not registered dataset)

dandi/tests/test_girder.py
```diff
@@ -0,0 +1,37 @@
...
+    assert resp is None
+
+
+def test_lock_dandiset_error_on_nest(local_docker_compose_env):
+    DANDISET_ID = "000001"
+    client = girder.get_client(local_docker_compose_env["instance"].girder)
+    with client.lock_dandiset(DANDISET_ID):
+        with pytest.raises(LockingError) as excinfo:
+            with client.lock_dandiset(DANDISET_ID):
+                pass
```
yarikoptic, line 24: since not intended to be covered, 
```suggestion
                pass  # pragma: no cover
```

dandi/upload.py
jwodder: Done.

dandi/tests/test_girder.py
jwodder: Done.

yarikoptic:
I have just merged another PR #186  and conflicts came up... I took liberty at resolving them via github's "resolve" web UI...

jwodder:
@yarikoptic Was the conflict in `setup.cfg`?  #186 moved `requests` from the `test` extra to `install_requires`; it looks like you kept `requests` in the `test` extra, so now the dependency is duplicated.

yarikoptic:
oh, sorry about that... please fix up to your liking and feel welcome to merge whenever you and CIs are satisfied!

PR #134, opened 2020-07-22
RF/ENH: download - reimplementation to support original and upcoming UI/API

### LINKED ISSUE #111: tqdm: errors in the tests with "AssertionError: attempt to release recursive lock not owned by thread"

yarikoptic:
RFed downloaders.  I thought to make it pretty but it ended up quite messy code.  Default UI though looks familiar since uses pyout as for `ls` and `upload`.

Should be **well** tested before releasing!

- removes use of `tqdm`, Closes #111 

TODOs
- [x] finish support logic for actually downloading via new API
- [x] on the fly (during download) digesting (checksum) of the data
- [x] implement frontend (pyout) analogous to upload
- [x] verify that `upload` or anything else got broken (we have some `upload` unittests going, so let's hope it is all good)
- May be later (for now "static" rendering from pyout if not interactive should suffice): provide an alternative, simpler "format" for really basic reporting so could be used in CI etc.  Probably will just report as json stream or alike.

TODOs after
- (not here) possibly add proper "sync" etc functionality
- while at it simplify all the mess here.  May be it is not worth bothering trying to start downloading while still traversing remote dataset (actually I am not longer sure it actually works that way since I think I have not seen counts in the summary having `>` - I must have broke it).
- ETA would be wrong ATM for downloads where some files are already present.  Probably while reworking for "sync" functionality - a fix would come naturally.


yarikoptic:
@satra @mgrauer -- "Perfect is the enemy of good".  I a bit "overengineered", but it should be a workable solution.  Needs testing. 
Give it a try while its "hot" and I can figure out what I have done (hopefully ;)).

For downloads of "released" dandisets https://github.com/dandi/dandi-publish/issues/112 should be addressed first though.

yarikoptic:
here is some kind of a demo: [asciicast](https://asciinema.org/a/IMIUvKcc2ZGPdtj5IC4lrImBi)

mgrauer:
Thanks for the asciicast, that is some fun viewing 🍿 

The UI looks nice, especially the way it updates multiple lines in parallel.

I have a couple points of discussion I'd like to raise for the CLI commands, they don't necessarily need to be addressed on this PR, but at least to start thinking about them.

1) having the output folder can make it harder to give the user a specific command from the GUI that they can then give to the CLI, because the output folder is specific to the user's setup. i.e. it is easy to tell the user, from the GUI `dandi download DANDI:000007` but not `dandi download -o testdandi DANDI:000007`. Maybe we could put the folder flag at the end or even put a template value in, when we display the command to run from the GUI, e.g. `dandi download DANDI:000007 -output <OUTPUT_DIR>` ?

2) there is still some inconsistency in how people download drafts vs released dandisets currently:

`dandi download -o testdandi DANDI:000007` for latest release
`dandi download -o testdandi 'http://GUI_url_to_draft'` for a draft

Could we change this to something like the following (I'm omitting output folder to simplify): 

`dandi download DANDI:000007` downloads latest published version
`dandi download DANDI:000007:0.200714.1807` downloads this specific published version
`dandi download DANDI:000007:draft` downloads the draft dandiset in its current form

At first I was thinking that maybe `dandi download DANDI:000007` should download a draft if there are no published versions, but upon thinking about it more, the draft is a different enough beast from a release (it is both volatile and mutable) that I'd want the user to have to be explicit about dealing with drafts.

satra:
> `DANDI:000007:draft`

@mgrauer - this is not an acceptable value for the dandi prefix presently in `identifiers.org`, so we would either have to add that as an allowed regex or force people not to use identifiers.org for drafts. also it would be `/` not `:` for version information 

`dandi download DANDI:000007/0.200714.1807` downloads this specific published version
`dandi download DANDI:000007/draft` downloads the draft dandiset in its current form

having a persistent identifier for a draft seems a little weird. so i'm hesitant to add that there, but open to thoughts. instead we could provide the following option through the CLI
```
$ dandi download DANDI:000007
Please select from available versions:
0) Current draft dataset
1) 0.200714.1807
2) ...
```
and the draft url would always be: `https://dandiarchive.org/dandiset/000007/draft`. this would also force people to publish a version to use an identifier.

satra:
also, we would never mint a doi for a dandiset which does not have a published version. i can also add a note to the redirector to specifically point to the draft URL for an unpublished dandiset.

yarikoptic:
@jwodder I invite you to try this one as well.

yarikoptic:
> 1. e.g. `dandi download DANDI:000007 -output <OUTPUT_DIR>` ?

that is already supported out of the box since that is how argparse (which I hope `click` uses underneath) does parsing - it does allow --options to be used at the end of the line.  It is very useful for options which could take arbitrary number of values :

```shell
$> dandi download DANDI:000007 --output-dir myother
2020-08-05 16:30:24,676 [    INFO] Traversing remote dandisets (000007) recursively
PATH                 SIZE DONE    DONE% CHECKSUM STATUS MESSAGE   
dandiset.yaml                                    done   updated   
Summary:                  0 Bytes                1 done 1 updated 
                          <0.00%                                  
^C
Aborted!

$> ls -l myother
total 4
drwx------ 2 yoh yoh 4096 Aug  5 16:30 000007/
```

yarikoptic:
> @mgrauer - this is not an acceptable value for the dandi prefix presently in `identifiers.org`, so we would either have to add that as an allowed regex or force people not to use identifiers.org for drafts. also it would be `/` not `:` for version information
> 
> `dandi download DANDI:000007/0.200714.1807` downloads this specific published version
> `dandi download DANDI:000007/draft` downloads the draft dandiset in its current form

I would also prefer here to use what is supported by/through `identifiers.org`.

Notes:
- I would also to be able to specify a path (to a subject or a specific file)
- There is a "convention" to use `@version` in a url to signal version.
  -  Used to be used by stackbrew to build official docker images, but I think they migrated away from that but some remain: e.g. https://github.com/docker-library/official-images/blob/4967c82e52fe02c24d5c585989fb24ac9eb25ba6/library/couchbase
  - DataLad uses it within URLs for RIA remotes, BUT carefully within a `#fragment` since IIRC `@` is not 100% kosher with allowed symbols in url ;)   But since here it is not a URL, I think it could be Ok

So I wondered if it could be `DANDI:<ID>[@<VERSION>][/PATH]` or `@` isn't good for identifiers?

Re "interactive" `> $ dandi download DANDI:000007` from @satra -- "not sure".  I expressed similar desires for more "interactive" within datalad, but somewhat agree (with Kyle) that for commands which by default are expected to be non-interactive, behavior should be explicitly described (e.g. download latest released version if no version specified), and only with explicit `--interactive|-i` they should offer choices.  Similar to what `git` does.  We aren't git here, but we do expect `dandi` to be used in CIs etc.

yarikoptic:
> also, we would never mint a doi for a dandiset which does not have a published version. i can also add a note to the redirector to specifically point to the draft URL for an unpublished dandiset.

I would prefer to avoid "if".  That would avoid surprises for people who decide to use dandi datasets in their scripts in CI etc, so they could always know what they are downloading.

a. If we decide that it should be most recent published release, then without explicit version specification of "draft" - it should blow.   

b. If we decide that it should not blow -- it should always redirect to a `draft` version.   BUT I do not like this option since then we would need some additional `latest` to point to the latest release.  Thus I would go with `a.`.

yarikoptic:
to resolve conflicts -- rebased against master (next time should just merge - was a bit evolved and I am not sure of benefits from rebase in this branch). download test should still be failing

yarikoptic:
Re windows
```
>           path = "/" + path.split("/", 1)[1]
E           IndexError: list index out of range
```
girder paths are os specific and my sloppy assumption was wrong

yarikoptic:
it seems that windows is happy now . All failures are due to codecov upload failures!  I will submit a PR to not fail workflow if codecov upload fails - it annoyed me enough by now :-/

yarikoptic:
as no concerns were expressed upon requested "manual" testing, and we just keep breeding conflicts, I will resolve conflicts and if all good would merge.

dandi/cli/cmd_download.py
```diff
@@ -48,12 +55,16 @@
 )
 @click.argument("url", nargs=-1)
 @map_to_click_exceptions
-def download(url, output_dir, existing, jobs=6, develop_debug=False):
+def download(url, output_dir, existing, jobs=6, format="pytout"):
```
jwodder, line 58: I don't think there's a situation where click would leave `format` unfilled, and if there is, shouldn't the default be "pyout", not "pytout"?

dandi/consts.py
```diff
@@ -100,6 +112,14 @@
 
 file_operation_modes = ["dry", "simulate", "copy", "move", "hardlink", "symlink"]
 
+#
+# Download (upload?) specific constants
+#
+# Chunk size when iterating a download (and upload) body. Taken from girder-cli
+# TODO: should we make them smaller for download than for upload?
+# ATM used only in download
+MAX_CHUNK_SIZE = os.environ.get("DANDI_MAX_CHUNK_SIZE", 1024 * 1024 * 8)  # 64
```
jwodder, line 121: You need to cast this to an `int`, as the value from `os.environ` is a string.

dandi/dandiapi.py
```diff
@@ -0,0 +1,348 @@
...
+import requests
+
+from . import get_logger
+from .utils import ensure_datetime
+from .consts import MAX_CHUNK_SIZE
+
+lgr = get_logger()
+
+
+class HTTPError(requests.HTTPError):
```
jwodder, line 12: Why not just import HTTPError instead of making a do-nothing subclass?

dandi/dandiapi.py
```diff
@@ -0,0 +1,348 @@
...
+                    client.downloadItem(item, fh)
+
+        In the above example, each request will be executed with the User-Agent header
+        while reusing the same TCP connection.
+
+        :param session: An existing :class:`requests.Session` object, or None.
+        """
+        self._session = session if session else requests.Session()
+
+        yield self._session
```
jwodder, line 59: This should be:

```
try:
    yield self._session
finally:
    self._session.close()
    self._session = None
```

so that the session gets closed on error.

dandi/dandiapi.py
```diff
@@ -0,0 +1,348 @@
...
+            params=parameters,
+            data=data,
+            files=files,
+            json=json,
+            headers=_headers,
+            **kwargs,
+        )
+
+        # If success, return the json object. Otherwise throw an exception.
+        if result.status_code not in (200, 201):
```
jwodder, line 144: I suggest using "`if not result.ok`" instead, as that only regards 4xx and 5xx status codes as errors.  There are other 2xx codes that aren't errors, plus the 3xx codes.

dandi/download.py
```diff
@@ -177,17 +263,75 @@ def follow_redirect(url):
...
+    # TODO: unduplicate with upload. For now stole from that one
+    # We will again use pyout to provide a neat table summarizing our progress
+    # with upload etc
+    import pyout
+    from .support import pyout as pyouts
+
+    # dandi.cli.formatters are used in cmd_ls to provide switchable
+    pyout_style = pyouts.get_style(hide_if_missing=False)
+
+    rec_fields = ("path", "size", "done", "done%", "checksum", "status", "message")
```
jwodder, line 276: If we're going to be passing tuples of `columns` to pyout, we should add `!= 0.6.0` to the dependency in `setup.cfg` in order to avoid the version that doesn't support them.

dandi/download.py
```diff
@@ -197,112 +341,441 @@ def download(
...
+            # I wonder if we could make writing async with downloader
+            with open(path, "wb") as writer:
+                for block in downloader():
+                    if digester:
+                        downloaded_digest.update(block)
+                    downloaded += len(block)
+                    # TODO: yield progress etc
+                    msg = {"done": downloaded}
+                    if size:
+                        if downloaded > size and not warned:
```
jwodder, line 724: `warned` never gets set to `True`.  Should that be added here?

dandi/cli/cmd_download.py
yarikoptic: oh... so we can avoid duplicating default values within `kwargs`?! let met just strip then them away from `download` and we could do to the rest in a separate PR.  THANK YOU! done in 4798e49fe49be2ea54599bd02d0009b1716e5044

dandi/consts.py
yarikoptic: d'oh, right,THANKS! done in 09b219660e77f88b581959447a5eee8af0f7facf

dandi/dandiapi.py
yarikoptic: I wanted to aggregate non-standard exceptions our code might raise within this module.  I guess I could have just gone with `from requests import HTTPError` into this namespace indeed but then linters would not be happy (not used etc), so I somewhat like this explicit subclassing.  Any disadvantage?

dandi/dandiapi.py
yarikoptic: indeed needs a fix, thanks!  but it should close only if session was not provided (should not close outside session). Fixed in f367313da0b87f86c1138666b50bf27b34cebab9

dandi/dandiapi.py
yarikoptic: THANKS! done in a45b550951fefa89e847052a1eb919fbb480758e

yarikoptic:
Thank you @jwodder for the review and suggestions! I have now pushed all the changes I have referenced.

dandi/support/iterators.py
```diff
@@ -0,0 +1,116 @@
...
+            # race condition HERE between checking for self.finished and
+            if self.finished and queue.empty():
+                break
+            # in general queue should not be empty, but if it is, e.g. due to race
+            # condition with above check
+            try:
+                yield queue.get(timeout=0.001)
+            except Empty:
+                continue
+            if self.reraise_immediately and self._exc is not None:
```
jwodder, line 112: Wouldn't it be better to move this check to the top of the `while` loop in case `get` times out while an exception is raised?

dandi/dandiapi.py
jwodder: There's no real disadvantage other than a tiny bit of overhead.  I'd personally prefer to import HTTPError and add a `noqa` comment to tell flake8 to be quiet.

dandi/dandiapi.py
```diff
@@ -0,0 +1,348 @@
...
+        -------
+
+        """
+        return self.get(f"/dandisets/{dandiset_id}/versions/{version}/assets/{uuid}/")
+
+    def get_dandiset(self, dandiset_id, version):
+        return self.get(f"/dandisets/{dandiset_id}/versions/{version}/")
+
+    def get_dandiset_assets(
+        self, dandiset_id, version, location=None, page_size=None, include_metadata=True
```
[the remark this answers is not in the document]
yarikoptic: I thought it would become "parametrized" for in the query for https://github.com/dandi/dandi-publish/issues/78 resolution, but now it always includes it, so indeed there is no need for it.  Removed in 515538eaccdba53d7f3f289c4e5e32f75b29e913

dandi/download.py
yarikoptic: good idea! you did fix for 0.6.0 for upload in cc4d37cda83584c69778e6fcecebc9124dd71ed8  but there is still `ls`, and I just feel it would be just easier to ask to avoid 0.6.0. done in 096c8dd232b665eb6a15aaeb7cf9b9b3aee95d87

dandi/download.py
yarikoptic: yeah, set it to True in 5abd195384356f48e37729ed364f2b89e37cd210 ... hopefully we never run into this part of code.

dandi/support/iterators.py
yarikoptic: indeed! done in e52d728d9294006eedc6a9dd95e0e582f04c45bd

yarikoptic:
> There's no real disadvantage other than a tiny bit of overhead. I'd personally prefer to import HTTPError and add a noqa comment to tell flake8 to be quiet.

I have argued with having our exception in `dandi.exceptions` but even wasn't there... I removed this subclass for now completely in 20718e250a661e4a998641d652704dc3a7ffd407

yarikoptic:
woohoo -- still green, merging before it ripens.

PR #135, opened 2020-07-27
ENH(TEMP): disallow use of a known to be outdated version of dandi client

yarikoptic:
Only if DANDI_ALLOW_OUTDATED env variable is set, it will be just a good old
warning.  I wanted to provision some way to proceed in cases that some other
issue (connectivity, OS, whatnot) prevents people from using dandi cli version
checking

This is a band-aid toward upcoming "dandi-publish" RF until there is a way for a client to interrogate API about supported functionality.

### DIFF
--- a/dandi/cli/command.py
+++ b/dandi/cli/command.py
@@ -186,6 +186,11 @@ def main(log_level, pdb=False):
 
         etelemetry.check_available_version("dandi/dandi-cli", __version__, lgr=lgr)
     except Exception as exc:
+        # This magical undocumented env var would allow to overcome some internal
+        # or external issues by merely disclosing its availability to effected users,
+        # while mandating users to update to the most recent version of the client
+        if not bool(os.environ.get("DANDI_ALLOW_OUTDATED", None)):
+            raise
         lgr.warning(
             "Failed to check for a more recent version available with etelemetry: %s",
             exc,


yarikoptic:
to test this, @satra, is there a way (env var, like `ET_FORCE_CHECK`) to force etelemetry a check (even if it did recently)?

satra:
don't have one at the moment (ET updates every 6 hours or so). i can manually delete the cache if needed. did not want to hand over cache refresh options to the world :)

dandi/cli/command.py
```diff
@@ -186,6 +186,11 @@ def main(log_level, pdb=False):
 
         etelemetry.check_available_version("dandi/dandi-cli", __version__, lgr=lgr)
     except Exception as exc:
+        # This magical undocumented env var would allow to overcome some internal
+        # or external issues by merely disclosing its availability to effected users,
+        # while mandating users to update to the most recent version of the client
+        if not bool(os.environ.get("DANDI_ALLOW_OUTDATED", None)):
+            raise
         lgr.warning(
             "Failed to check for a more recent version available with etelemetry: %s",
```
satra, line 195: this message seems more generic than perhaps we should use? you can use the flag `raise_exception=True` in the call to `check_available_version` and will return a `RuntimeError` if a bad version is running.

yarikoptic: `RuntimeError`?  submitted https://github.com/sensein/etelemetry-client/pull/23 

as for `raise_exception=True` - I should add it indeed but then, depending on etelemetry version I would not be reliably tell apart either it is outdated or some other exception has happened (there is sufficient amount of code outside of try/except in that function).  I don't think we would like to stop on any exception etelemetry is throwing, but on that `BadVersionException` - so I will wait on destiny of that PR

yarikoptic:
I think we better just stick to the "deployment/server specific" handling of versions (than overall one as would be possible with ET), so I will close this PR in favor of #186

PR #137, opened 2020-07-29
ENH: DANDI_LOG_GIRDER env var to instruct to log interactions with girder

yarikoptic:


yarikoptic:
failed one is a codecov upload. Known issue: https://github.com/codecov/codecov-action/issues/81  . I wonder if we should just ignore errors in codecov upload with 
```
fail_ci_if_error: false
```
but for now let's proceed.

PR #139, opened 2020-07-30
ENH: be able to set smaller chunk size, retry up to 3 times on some errors

yarikoptic:
See #136 for more details


yarikoptic:
I worry about this PR approach is that we are resubmitting when possibly girder already accomplished the mission (whatever it was) and would not "correctly" handle retransmission.  And I think we never had a succesful upload by @mhhennig on some file where there were retransmissions. 
Because of any confirmation in real case scenarios that it should work, and since I do not think we can really test such cases reliably, I am hesitant to merge it. I will convert it to Draft, merge and push current master, but will keep it open for visibility and possibly further progress at some point since ideally we should do something like, we just need to figure out how to simulate/test at least some of the possible cases.

yarikoptic:
in particular in the light of the upcoming API work (so no direct correspondence with girder after we are done), I will just close this one

PR #141, opened 2020-08-02
Convert regexes with invalid escapes to raw strings

jwodder:
`'\.'` is not a valid escape sequence in Python strings, and so it generates a SyntaxError under Python 3.8.3.


yarikoptic:
Oh, indeed. Thanks!

PR #142, opened 2020-08-02
Make it possible to build package via PEP 517

jwodder:
This is the simplest possible change to make it possible to build the project with PEP 517 (e.g., using pip or tox).  Without this change, `versioneer.py` is unimportable during the build.  See warner/python-versioneer#193, warner/python-versioneer#192, and pypa/pip#6163 for more information.

This PR also adds `versioneer.py` to `MANIFEST.in` so that it will be present in sdists, which is necessary for wheels built from an sdist to have the correct version set in their metadata.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,3 @@
+# This line includes versioneer.py in sdists, which is necessary for wheels
+# built from sdists to have the version set in their metadata.
+include versioneer.py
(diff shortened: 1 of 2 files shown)


yarikoptic:
Although modifying sys.path worries me, let's hope side effects aren't to happen and proceed

PR #145, opened 2020-08-03
Add a period to the end of the first line in register's docstring

jwodder:
Without this patch, `dandi --help` lists the following description for `register`:

    Register a new dandiset in the DANDI archive This command...


yarikoptic:
Test failure is tqdm, might be gone soon anyways...

PR #144, opened 2020-08-03
Set up testing with tox

jwodder:
`setup.py test` and `tests_require` are deprecated [[1]](https://setuptools.readthedocs.io/en/latest/setuptools.html#test-build-package-and-run-a-unittest-suite) [[2]](https://github.com/pypa/setuptools/issues/1684), with `tox` or similar recommended instead.

PR #148, opened 2020-08-03
Configure CI to run flake8

jwodder:
PR #144 is a prerequisite for this.  (In the future, when PR x depends on PR y, would you prefer it if x's branch extended y's or not?)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,26 @@
+name: Linters
+
+on:
+  - push
+  - pull_request
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Set up environment
+      uses: actions/checkout@v1
+      with:  # no need for the history
+        fetch-depth: 1
+    - name: Set up Python
+      uses: actions/setup-python@v1
+      with:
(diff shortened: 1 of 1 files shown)


yarikoptic:
> In the future, when PR x depends on PR y, would you prefer it if x's branch extended y's or not?

yes, probably to "extend y" so later on, when we need to `git bisect` etc there is lesser chance to end up in a false positive "failing state"

yarikoptic:
I have restarted the CI job since #144 was merged. Came out red.
I guess we need to provide this specific linter with the settings we use for pre-commit/black. We already specify some (max width) in setup.cfg for flake8:
```
[flake8]
max-line-length = 100
```

yarikoptic:
ignore above -- I only now saw #147 ;)

yarikoptic:
could you please rebase on top of master and force push?  Would be nice to get it green to know that we could safely merge

jwodder:
Done, though I don't believe it'll be green until after #147 is merged.

yarikoptic:
#147 is merged now ;) Thanks!

yarikoptic:
all green on rerunning the pull request run -- so good to go AFAIK

PR #149, opened 2020-08-03
Fix warning caused by lack of tzinfo

jwodder:
Without this change, pynwb generates the warning "UserWarning: Date is missing timezone information. Updating to local timezone."


yarikoptic:
Great, thanks!  CI is happy - I am happy ;)

PR #150, opened 2020-08-03
Fix warning caused by `load_namespaces=True`

jwodder:
Without this change, pywnb generates the warning "UserWarning: No cached namespaces found in /private/var/folders/l7/wrkq93d133d8zpn36fmqrq0r0000gn/T/pytest-of-jwodder/pytest-103/data1/simple1.nwb"

Note: I'm not actually sure what the `load_namespaces=True` is trying to accomplish, so it's possible this PR is the wrong way to handle this.


jwodder:
Found a better fix; withdrawn.

PR #151, opened 2020-08-03
Configure pytest to fail on warnings

jwodder:
Warnings generated by pynwb due to `load_namespaces=True` are ignored.

This PR includes the changes from #144.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -20,6 +20,10 @@ commands =
 
 [pytest]
 markers = integration
+filterwarnings =
+    error
+    ignore:No cached namespaces found .*:UserWarning
+    ignore:ignoring namespace '.*' because it already exists:UserWarning
 
 [coverage:run]
 parallel = True


yarikoptic:
Unfortunately code would need to be fixed up I guess so there is no warnings. may be `git history` and `blame` could provide more information on why we are passing those options etc and either they could be removed now or not.

jwodder:
The only warnings I found were the one in #149 and the `load_namespace` warnings, which this PR configures pytest to ignore.  I didn't actually run the combined code, but I believe testing with both #149 and this PR will succeed.

yarikoptic:
> , but I believe testing with both #149 and this PR will succeed.

I have restarted tests, but may be rebase would be needed

jwodder:
@yarikoptic Rebased.

yarikoptic:
failed one is due to #111, workaround finally submitted in #165

PR #155, opened 2020-08-03
Make --log-level case-insensitive and also accept integers

jwodder:


yarikoptic:
WOOHOO! Thanks @jwodder !  I always wanted to figure out how to do it in click properly but never got to it.  I have not tested, but will do later straight in master

PR #156, opened 2020-08-03
BF: remove bug introduced in the rush of 8d1f527e50dec25b599c3e5e4867…

yarikoptic:
Obviously I rushed through ... heh


yarikoptic:
grr, failures are seems to be the tqdm issue https://github.com/dandi/dandi-cli/issues/111 . no upstream fix yet

Issue #118, opened 2020-06-04
introduce -d to upload or just autodetermine top of the dandiset from a given path?


yarikoptic:
I am debating with myself about the need for such an option.  `organize` has it to point to the top of the dandiset. In case of `upload` we can determine the top by presence of `dandiset.yaml` for any path if pointed to while being outside of the dandiset.  It is very similar situation/behavior of what we used to have in datalad for `get`, `install` etc, but it was deprecated in favor of an explicit `-d` when performing operation outside of a dataset.  Complicating factor there was also having both cli and Python API with classes, so it makes sense for any `Dataset("path").command("relpath")` treat `"relpath"` relative to the "path" regardless of the current directory.  Here we have a much simpler usecase.

PR #158, opened 2020-08-04
Fix upload command

jwodder:
Uploading to local-docker using the latest code on master wasn't working for me, so I fixed the errors:

* Use `paths` in place of the commented-out `dandiset_path` argument
* Change a tuple to a list in order to address [a bug](https://github.com/pyout/pyout/issues/110) in pyout's release yesterday

### DIFF
--- a/dandi/cli/cmd_upload.py
+++ b/dandi/cli/cmd_upload.py
@@ -108,14 +108,15 @@ def upload(
     from ..dandiset import Dandiset
     from ..support.digests import Digester
 
-    dandiset = Dandiset.find(dandiset_path)
-    if not dandiset:
-        raise RuntimeError(
-            f"Found no {dandiset_metadata_file} anywhere.  Use 'dandi register', 'download', or 'organize' first"
-        )
-
-    # Should no longer be needed
-    # dandiset_path = Path(dandiset_path).resolve()
+    dandiset_paths = []
+    for p in paths:
+        dandiset = Dandiset.find(Path(p).resolve())
+        if not dandiset:
+            raise RuntimeError(
+                f"Found no {dandiset_metadata_file} anywhere in {p}.  Use 'dandi register', 'download', or 'organize' first"
+            )
+        dandiset_paths.append(dandiset.path)
+    paths = dandiset_paths
 
     # Girder side details:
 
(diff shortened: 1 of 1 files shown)


yarikoptic:
> Change a tuple to a list in order to address a bug in pyout's release yesterday

that one (https://github.com/dandi/dandi-cli/pull/158/commits/722c83a63b9f410904f66894a0fb2e215dafb64e) is good! Thanks.  
I have cherry picked it and pushed into master.  

I have concerns about the 2nd one though:

> Uploading to local-docker using the latest code on master wasn't working for me, so I fixed the errors:

Beyond pyout 0.6.0 issue -- was 2nd commit addressing some issue?
 
> * Use `paths` in place of the commented-out `dandiset_path` argument

It seems it introduces ability to upload multiple dandisets at once, but code below I believe relies on interacting with a single dandiset... 

also it seems to be doing really is removing ability to upload individual files, since `path` could point to specific file(s) and they should not be resolved into dandisets.

jwodder:
I may have misread the code and expected some things from it that weren't intended.  Earlier, I tried running `dandi upload -i local-docker 000006` on master, where `000006` is a directory containing a `dandiset.yaml` plus several folders of `.nwb` files, and it failed with:

```
RuntimeError: Found no dandiset.yaml anywhere.  Use 'dandi register', 'download', or 'organize' first
```

I then edited the code under the assumption that the `upload` command was supposed to accept dandiset.yaml-containing folders as arguments.  Looking further into the code now, it seems that is not the case (though it would be reasonable to support that usage in the future).  I suppose the rest of this PR can be discarded.

yarikoptic:
gotcha! Thanks for the explanation -- makes sense now!  I think it is the https://github.com/dandi/dandi-cli/issues/118 . 
 May be after next release (0.6.0) with support of "publish API" to be introduced in #134 we will re-review current of all cmdline tools and at Python level for 0.7.0 and see how to harmonize etc. (meanwhile I have initiated milestone 0.7.0 but have not yet triaged anything assigned into 0.6.0 either it should be delayed until after it)

yarikoptic:
forgot to add to make it clear: for `upload` ATM you need to be in a dandiset and could point PATHs to files/subfolders.

PR #159, opened 2020-08-04
[GH-27] Define a `local_docker_compose` fixture

jwodder:
This requires `dandiarchive/dandiarchive-{girder,provision}` images to be uploaded to Docker Hub before it can work.

I opted to make the fixture session-scoped so that the tests wouldn't be massively slowed down by repeated creation & destruction of the containers or by the provision container's `sleep 10`.  This does mean that data in the instance persists across tests, which might not be desirable.

Note that, when run under GitHub Actions, any tests using this fixture will run on Ubuntu and Windows but will be skipped on macOS as, according to [this page](https://docs.github.com/en/actions/reference/software-installed-on-github-hosted-runners), the macOS environment lacks Docker.


jwodder:
Status: The most recent commit contains an unfinished smoke test for the `register` command that uses the `local_docker` fixture.  The fixture works fine, and the `register` command exits successfully, but Click isn't capturing any output from `register` and it's making the test fail.

@yarikoptic Any idea why this is?

yarikoptic:
BTW, test also should set `DANDI_DEVEL=1` explicitly while executing the command needing it.  There should be no need to (remember to) set it in outside environment (thus should be no need for tox.ini modifications etc) while running tests.

dandi/tests/fixtures.py
```diff
@@ -107,6 +107,8 @@ def fixture():
 
 @pytest.fixture(scope="session")
 def local_docker():
+    if os.name != "posix":
+        pytest.skip("Docker images require Unix host")
```
yarikoptic, line 111: it doesn't work on OSX? (Darwin) I would have even assumed it might have worked on Windows.  Both have support for Docker.  What specific aspect is missing?

jwodder: `os.name` is `'posix'` on Darwin, so it will (and does) run fine on Macs.  (However, it'll be skipped on GitHub Actions' Mac environments, as they don't have Docker.)  As for Windows, `docker-compose` was failing on GitHub Actions' Windows environments with the message "image operating system "linux" cannot be used on this platform".  I believe we would have to build & upload separate images for Windows in order for the tests to run there.

dandi/tests/data/dandiarchive-docker/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+      - mongo
+    ports:
+      - "8080:8080"
+
+  client:
+    image: dandiarchive/dandiarchive-client
+    depends_on:
+      - girder
+    ports:
+      - "8085:80"
```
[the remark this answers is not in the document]
jwodder: The client container isn't needed for the tests, is it?  Because changing the port numbers makes it unable to load.  (I tried configuring the container to connect to girder on port 8081 by setting the `VUE_APP_API_ROOT` environment variable on it, but that didn't work.)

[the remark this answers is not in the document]
yarikoptic: Good point! I do not think we indeed need anything from web UI client.  Let's indeed get rid of it here - should also speed up testing startup/teardown time I guess.

Note: made a new https://github.com/dandi/dandi-cli/issues/164 to tackle after this one.

dandi/tests/fixtures.py
yarikoptic: Thanks for the explanation!  Could you please also distill it as a comment in the code so anyone who runs into this skip would know why.

dandi/cli/tests/test_register.py
```diff
@@ -0,0 +1,50 @@
...
+    if r.exception is not None:
+        return "".join(format_exception(*r.exc_info))
+    else:
+        return r.output
+
+
+def test_smoke(local_docker):
+    runner = CliRunner(mix_stderr=False)
+    with runner.isolated_filesystem():
+        Path(dandiset_metadata_file).write_text("{}\n")
```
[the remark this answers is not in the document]
jwodder: If there's no `dandiset.yaml` in the current directory, `register` will print the new metadata to stdout instead of saving it in a file.  Is that the intended behavior?  If so, does it matter which route the test takes?  I'll think I'll test both routes for now.

jwodder:
> BTW, test also should set `DANDI_DEVEL=1` explicitly while executing the command needing it. There should be no need to (remember to) set it in outside environment (thus should be no need for tox.ini modifications etc) while running tests.

That won't work.  `DANDI_DEVEL`'s effect of adding options takes effect when the code is imported and the command definitions are processed, not when the commands are run, and so the environment variable needs to be set outside the Python process, e.g., in `tox.ini`.

dandi/cli/tests/test_register.py
```diff
@@ -0,0 +1,78 @@
...
+from click.testing import CliRunner
+import pytest
+import yaml
+
+from ..command import register
+from ...consts import dandiset_identifier_regex, dandiset_metadata_file
+
+
+def yaml_load(s):
+    obj = yaml.load(s, Loader=yaml.BaseLoader)
```
yarikoptic, line 15: note: in some other code places we use `yaml.safe_load` which does not demand `Loader`, but I guess it would be suboptimal here (and thus avoiding need for `yaml_load`) since we would not test an entire document... something to keep in mind I guess (unless may be BaseLoader also loads just first document? I didn't check)

dandi/consts.py
```diff
@@ -83,6 +83,12 @@
     "local-docker": dandi_instance(
         "http://localhost:8080", "http://localhost:8085", None
     ),
+    "local-docker-compose": dandi_instance(
```
yarikoptic, line 86: well, `local-docker` is also via compose.  This one is specifically for testing, so may be name it `"local-docker-tests"` or alike?

setup.cfg
```diff
@@ -71,6 +71,7 @@ test =
     mock
     pytest
     pytest-cov
+    requests ~= 2.20
```
yarikoptic, line 74: why "pinning" of the release is necessary?  Whenever possible I really prefer to avoid pinning (exact or to a series) so we could react to changes in the dependencies etc, and also to not complicate packaging for distributions (like Debian) where only a single (different) version could be available ATM

dandi/consts.py
jwodder: Changed.

setup.cfg
jwodder: A lower version bound of 2.20 is recommended for `requests` because previous versions had security issues.  I personally prefer also using an upper version bound for any dependency that uses or approximates semantic versioning, as that helps to prevent one's project from being installed alongside a newer version of a dependency with which it is incompatible due to API changes in a major version.

jwodder:
@yarikoptic I've changed the code to test the `register()` function directly instead of testing the `register` command, thereby eliminating the external `DANDI_DEVEL` setting.

dandi/cli/tests/test_register.py
```diff
@@ -0,0 +1,78 @@
...
+    return obj
+
+
+def show_result(r):
+    if r.exception is not None:
+        return "".join(format_exception(*r.exc_info))
+    else:
+        return r.output
```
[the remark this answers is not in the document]
jwodder: Removed.

setup.cfg
yarikoptic: lower bound -- agreed 100%.  Upper bound - I can live with in this particular case, although due to our really basic use of `requests` I do not expect requests 3 to break our code. Let's proceed.

PR #160, opened 2020-08-05
Fix some printf-formatting syntax

jwodder:


dandi/pynwb_utils.py
```diff
@@ -125,7 +125,7 @@ def get_neurodata_types_to_modalities_map():
 
                 v_split = v_.__module__.split(".")
                 if len(v_split) != 2:
-                    print("Skipping %s coming from %s" % v_, v_.__module__)
+                    print("Skipping %s coming from %s" % (v_, v_.__module__))
```
yarikoptic, line 128: FWIW, I am not sure even why there is a straight `print` here and not some `lgr.info`... but as the fix goes -- THANKS!

PR #163, opened 2020-08-05
Restrict flake8 to dandi directory

jwodder:
This makes flake8 ignore anything outside the `dandi` directory (such as `venv/`) without having to constantly add to flake8's exclusions list.


yarikoptic:
should we also flake `setup.py` since it is "ours"?

jwodder:
I doubt it'll need linting that often, but sure, added.

PR #165, opened 2020-08-05
BF(workaround): skip test if download blows due to tqdm locking issue

yarikoptic:
ref #111, which will be closed by #134

### DIFF
--- a/dandi/tests/test_download.py
+++ b/dandi/tests/test_download.py
@@ -71,7 +71,15 @@ def test_download_multiple_files(tmpdir):
         "item+5e70d3173da50caa9adaf335/item+5e70d3183da50caa9adaf336"
     )
 
-    ret = download(url, tmpdir)
+    # TEMP: to workaround https://github.com/tqdm/tqdm/issues/982 upstream
+    # at least during testing. Our issue: https://github.com/dandi/dandi-cli/issues/111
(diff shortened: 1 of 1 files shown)


dandi/tests/test_download.py
```diff
@@ -71,7 +71,15 @@ def test_download_multiple_files(tmpdir):
         "item+5e70d3173da50caa9adaf335/item+5e70d3183da50caa9adaf336"
     )
 
-    ret = download(url, tmpdir)
+    # TEMP: to workaround https://github.com/tqdm/tqdm/issues/982 upstream
+    # at least during testing. Our issue: https://github.com/dandi/dandi-cli/issues/111
+    # The plan - to stop using tqdm (by default at least) as part of #134 RF
+    try:
+        ret = download(url, tmpdir)
+    except AssertionError as exc:
```
yarikoptic, line 79: rright - this code path is not tested and I guess was not triggered by this run, but looks "legit" to me... I will nevertheless do a rerun of tests to see if it gets triggered

yarikoptic: AHA -- it did trigger on rerun!
bad thing  - now our coverage (relative to master) would jump back and forth for these 3 lines :-/ ideally there should be a dedicated test to trigger it

PR #167, opened 2020-08-05
BF: import known_instances directly

yarikoptic:
Before it was used via imported binding within girder module, which
we recently flake8-ed away.  Since no "upload" tests, breakage went unnoticed

Thanks @alejoe91 for living on the bleeding edge and reporting back upon the first cut ;-)


yarikoptic:
as promised, coverage became flaky due to #165 .

PR #171, opened 2020-08-06
Test "upload" command

### LINKED ISSUE #27: upload: establish testing

jwodder:
This PR splits the body of the "upload" command off into a function and adds two simple tests.

Note that this removes the `-s` option from the `pytest` invocation in `tox.ini`, as the pyout integration was failing when that option was present.  The tests seem to be fine with that option when running under GitHub Actions, though.

Closes #27.


yarikoptic:
Thank you @jwodder , let's proceed

PR #174, opened 2020-08-07
Add Dockerized redirector to Docker Compose

### LINKED ISSUE #170: bring in dandi/redirector into "test deployment" setup
That should allow for testing redirection to perform correctly, and also "redirect" to the local test deployment so we could use non-girder based URLs in the tests.

Attn @satra as we might need to make redirector to become parametrized so no hardcoded urls (which could be used defaults) are then used by it.

jwodder:
Closes #170.

Before this PR can be accepted, dandi/redirector#16 needs to be merged, after which we can set up the redirector image on Docker Hub.

Also note that this PR uncomments the client container so that the redirector will have something to redirect to, even though the client container currently returns only blank pages due to the Girder container using a non-default port.  [I've asked dandi/dandiarchive to address this point, though.](https://github.com/dandi/dandiarchive/issues/468)

### DIFF
--- a/dandi/consts.py
+++ b/dandi/consts.py
@@ -84,10 +84,7 @@ known_instances = {
         "http://localhost:8080", "http://localhost:8085", None
     ),
     "local-docker-tests": dandi_instance(
-        # "http://localhost:8081", "http://localhost:8086", None
-        "http://localhost:8081",
-        None,
-        None,
+        "http://localhost:8081", "http://localhost:8086", "http://localhost:8079"
     ),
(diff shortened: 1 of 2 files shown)


jwodder:
dandi/redirector#16 has been merged, and I've set up [the Docker image on Docker Hub](https://hub.docker.com/repository/docker/dandiarchive/dandiarchive-redirector).  This PR should be ready to merge now.

yarikoptic:
tests on push failing due to merged #169, let's proceed, thank you @jwodder !

PR #175, opened 2020-08-07
Add "instance" and "instance_id" keys to local_docker_compose return value

### LINKED ISSUE #172: encapsulate relevant information within local_docker_compose and avoid hardcoded or duplicate fixture specific info
ATM `local_docker_compose` returns a dict with only `api_key`, then code hardcodes instance id (`local-docker-tests`). Ideally all that information should come from a fixture.

I also wonder if we could may be make a "test level" fixture which would use the "session level local_docker_compose" fixture so it could also make any test specific changes, such as  `monkeypatch.setenv("DANDI_API_KEY", local_docker_compose["api_key"])`?

All that to possibly make fixture replaceable or used (later) in parametrization of the test, and to avoid duplicate code within tests using the fixture.

jwodder:
Also define a `local_docker_compose_env` fixture that sets `DANDI_API_KEY`.

Closes #172.

### DIFF
--- a/dandi/tests/test_register.py
+++ b/dandi/tests/test_register.py
@@ -13,7 +13,7 @@ def yaml_load(s):
 
 def test_smoke_metadata_present(local_docker_compose, monkeypatch, tmp_path):
     monkeypatch.setenv("DANDI_API_KEY", local_docker_compose["api_key"])
-    dandi_instance = known_instances["local-docker-tests"]
+    dandi_instance = known_instances[local_docker_compose["instance_id"]]
     (tmp_path / dandiset_metadata_file).write_text("{}\n")
     assert (
         register(
@@ -36,7 +36,7 @@ def test_smoke_metadata_present(local_docker_compose, monkeypatch, tmp_path):
 
 def test_smoke_metadata_not_present(local_docker_compose, monkeypatch, tmp_path):
     monkeypatch.setenv("DANDI_API_KEY", local_docker_compose["api_key"])
-    dandi_instance = known_instances["local-docker-tests"]
+    dandi_instance = known_instances[local_docker_compose["instance_id"]]
     assert (
         register(
             dandi_instance,
--- a/dandi/tests/test_register.py
+++ b/dandi/tests/test_register.py
@@ -11,9 +11,8 @@ def yaml_load(s):
     return obj
 
 
-def test_smoke_metadata_present(local_docker_compose, monkeypatch, tmp_path):
-    monkeypatch.setenv("DANDI_API_KEY", local_docker_compose["api_key"])
-    dandi_instance = known_instances[local_docker_compose["instance_id"]]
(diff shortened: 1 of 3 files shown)


dandi/tests/test_register.py
```diff
@@ -34,9 +33,8 @@ def test_smoke_metadata_present(local_docker_compose, monkeypatch, tmp_path):
     # with the given identifier
 
 
-def test_smoke_metadata_not_present(local_docker_compose, monkeypatch, tmp_path):
-    monkeypatch.setenv("DANDI_API_KEY", local_docker_compose["api_key"])
-    dandi_instance = known_instances["local-docker-tests"]
+def test_smoke_metadata_not_present(local_docker_compose_env, monkeypatch, tmp_path):
+    dandi_instance = known_instances[local_docker_compose_env["instance_id"]]
```
yarikoptic, line 37: here and above
```suggestion
    dandi_instance = local_docker_compose_env["instance"]
```
?

dandi/tests/test_upload.py
```diff
@@ -12,14 +12,13 @@
...
-def test_upload(local_docker_compose, monkeypatch, tmp_path):
+def test_upload(local_docker_compose_env, monkeypatch, tmp_path):
     DIRNAME1 = "sub-anm369963"
     FILENAME1 = "sub-anm369963_ses-20170228.nwb"
     DIRNAME2 = "sub-anm372793"
     FILENAME2 = "sub-anm372793_ses-20170508.nwb"
 
-    monkeypatch.setenv("DANDI_API_KEY", local_docker_compose["api_key"])
-    dandi_instance = "local-docker-tests"
+    dandi_instance = local_docker_compose_env["instance_id"]
```
yarikoptic, line 21: here and in another test below (more code would need tune up for the rename):

```suggestion
    dandi_instance_id = local_docker_compose_env["instance_id"]
```
for consistency?

dandi/tests/test_register.py
jwodder: Done.

dandi/tests/test_upload.py
jwodder: Done.

dandi/tests/test_upload.py
```diff
@@ -35,43 +34,42 @@ def test_upload(local_docker_compose, monkeypatch, tmp_path):
         metadata = yaml.safe_load(fp)
     dandi_id = metadata["identifier"]
 
-    client = girder.get_client(known_instances[dandi_instance].girder)
+    client = girder.get_client(known_instances[dandi_instance_id].girder)
```
yarikoptic, line 37: here and similarly in the other tests:

```suggestion
    client = girder.get_client(local_docker_compose_env["instance"].girder)
```

so we could avoid using `known_instances` as much as possible and take advantage of encapsulation of all relevant information within the fixture.

jwodder: Done.

yarikoptic:
AWESOME! Thank you!

PR #176, opened 2020-08-07
register(): Adjust dandi_instance argument

### LINKED ISSUE #173: register (Python API): move dandi_instance to the tail of options and default to "dandi"
similar to how was done for `upload`.

`instance` is more of a development option ATM, so definetely should not be the first and positional argument

jwodder:
This PR includes the changes in #175 in order to make use of the new "identifier_id" key.

Closes #173.


yarikoptic:
a single failure is due to codecov, #177 merged in master already will mitigate.

Thank you!

PR #177, opened 2020-08-07
BF(workaround): do not fail test workflow if codecov fails

yarikoptic:
Failures to upload to codecov come and go.  AFAIK there is no way to
retry and thus we will just allow to skip if codecov "errors"


yarikoptic:
woohoo -- all green, even the flaky test tested the exception ;)  I will annotate it (separate PR) it as no cover to avoid flakiness...

PR #178, opened 2020-08-07
BF+ENH: workaround tqdm exception, test girder retries loop, ensure sleep of at least 1 sec

yarikoptic:
This should also stabilize coverage reporting since tested code paths shouldn't vary

While at it
- added smoke testing of logging when `_DANDI_LOG_GIRDER` is set to a single run in the matrix
- added a test that our tqdm progress reporter does not blow when tqdm blows with assertion error

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
           - ubuntu-18.04
           - macos-latest
         python:
-          - 3.6
+          - 3.6  # will also be used to test with _DANDI_LOG_GIRDER set
           - 3.7
           - 3.8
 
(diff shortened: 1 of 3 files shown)


yarikoptic:
arghm, Windows runs are not happy with my shell conditional
```
Run if [[ 3.6 == 3.6 ]]; then
ParserError: D:\a\_temp\8f725733-8c3d-4357-b9d9-0483b8aeb4c1.ps1:2
Line |
   2 |  if [[ 3.6 == 3.6 ]]; then
     |    ~
     | Missing '(' after 'if' in if statement.
```
smells like it is a powershell or some other beast so it would not be that easy to set that env variable for a single specific matrix run... @jwodder  - have any ideas on how I could accomplish the mission here of having env var set only for some single matrix run?

yarikoptic:
Thank you @jwodder -- that worked with minor tuneups (needed to use `'` not `"` for a string; then was too slow to realize that `_` prefix is not desired).  Now it is all green.  I know that it might be not code parts you are familiar with, but I would appreciate the review if I have not missed something obvious.

yarikoptic:
Great, thank you for the review, let's proceed.

PR #180, opened 2020-08-10
Eliminate pyyaml dependency

jwodder:
There's no reason to use two YAML libraries, so this PR gets rid of the one that doesn't support round-tripping comments.

**Note**: dandi-cli currently uses the API introduced in ruamel.yaml 0.15.0 (featuring `ruamel.yaml.YAML()` and its `load()` and `dump()` methods, as opposed to using the `load()` and `dump()` functions directly in `ruamel.yaml`), and this PR does that as well.  However, [the ruamel.yaml docs](https://yaml.readthedocs.io/en/latest/basicuse.html) state that the new API is "still in the process of being fleshed out" and recommend that production software use pre-0.15 versions of the library.  Should we follow that advice?

### DIFF
--- a/dandi/tests/test_register.py
+++ b/dandi/tests/test_register.py
@@ -1,13 +1,8 @@
 import re
 
-import ruamel.yaml
-
 from ..consts import dandiset_identifier_regex, dandiset_metadata_file
 from ..register import register
-
-
-def yaml_load(s):
-    return ruamel.yaml.YAML(typ="base").load(s)
+from ..utils import yaml_load
 
 
 def test_smoke_metadata_present(local_docker_compose_env, monkeypatch, tmp_path):
@@ -22,7 +17,7 @@ def test_smoke_metadata_present(local_docker_compose_env, monkeypatch, tmp_path)
         is None
     )
     with (tmp_path / dandiset_metadata_file).open() as fp:
-        metadata = yaml_load(fp.read())
+        metadata = yaml_load(fp, typ="base")
     assert metadata
     assert metadata["name"] == "Dandiset Name"
     assert metadata["description"] == "Dandiset Description"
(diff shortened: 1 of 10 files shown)


dandi/tests/test_register.py
```diff
@@ -1,14 +1,13 @@
 import re
 
-import yaml
+import ruamel.yaml
 
 from ..consts import dandiset_identifier_regex, dandiset_metadata_file
 from ..register import register
 
 
 def yaml_load(s):
```
yarikoptic, line 9: This seems to be a common pattern.  May be we could just move this helper into `dandi.utils` to reside alongside with `yaml_dump`, and just equip with `typ=None` kwarg so where needed we could switch to `safe` or any other?

jwodder: Done.

dandi/utils.py
```diff
@@ -347,9 +350,17 @@ def yaml_dump(rec):
...
+    yaml.default_flow_style = False
+    out = io.StringIO()
+    yaml.dump(rec, out)
+    return out.getvalue()
+
 
-    return yaml.safe_dump(rec, default_flow_style=False)
+def yaml_load(f, typ=None):
+    # `f` can be either a string or a file-like object.
+    # `typ` defaults to "rt" (round-trip)
```
yarikoptic, line 362: just not to delay - now that it is in `utils`, it better come with a proper docstring.  We (loosely, might need fixes) following numpy docstring style through the codebase.

dandi/validate.py
```diff
@@ -42,7 +41,7 @@ def validate_file(filepath):
 def validate_dandiset_yaml(filepath):
     """Validate dandiset.yaml"""
     with open(filepath) as f:
-        meta = yaml.safe_load(f)
+        meta = yaml_load(f, typ="safe")
```
yarikoptic, line 44: just a mental note unrelated to this PR -- apparently we nohow/nowhere validate `dandiset.yaml`? will be handled properly anyways soon with RFing of metadata and validation

dandi/utils.py
jwodder: Done.

yarikoptic:
Great, thanks!

PR #186, opened 2020-08-11
Add a get_instance() function

### LINKED ISSUE #184: use redirector's `/server-info` (or just "OPTIONS" call)
A satellite issue to redirector's https://github.com/dandi/redirector/issues/20 to be merged in after redirector gets its portion merged:

- provide a helper (probably with caching) e.g. `get_instance` to avoid direct use of `known_instances` lookup table.  Instead it should given an instance name (or even just a URL later on?), take a `.redirector` url and request that information from it.  If it is a URL provided -- should just talk to it directly (that would in the future allow to specify "instance" via URL not hardcoded name)
- if there is a response from the redirector providing the record:
  - if current client version does not satisfy `cli-version` in the record, raise a dedicated exception which would explain that the instance demands minimal version X and there are some bad versions which must not be used (if any prescribed in that record). 
  - compose instance record using `.url` fields from the record (instead of hardcoded ones in `consts.py`)
- if there is no response from the redirector, issue a warning and continue using hard-coded values for now.  later we will RF it away but probably worth keeping it as a safety blanket for now.

jwodder:
Closes #184

### DIFF
--- a/dandi/tests/test_utils.py
+++ b/dandi/tests/test_utils.py
@@ -266,9 +266,7 @@ def test_get_instance_unknown_url_bad_response():
 
 
 def test_get_instance_id_no_redirector():
-    with pytest.raises(ValueError) as excinfo:
-        get_instance("local-girder-only")
-    assert str(excinfo.value) == "DANDI instance has no known redirector URL"
+    assert get_instance("local-girder-only") is known_instances["local-girder-only"]
 
 
 @responses.activate
--- a/dandi/tests/test_utils.py
+++ b/dandi/tests/test_utils.py
@@ -269,3 +269,29 @@ def test_get_instance_id_no_redirector():
     with pytest.raises(ValueError) as excinfo:
         get_instance("local-girder-only")
     assert str(excinfo.value) == "DANDI instance has no known redirector URL"
+
+
+@responses.activate
+def test_get_instance_bad_version_from_server():
+    responses.add(
+        responses.GET,
+        "https://example.dandi/server-info",
+        json={
+            "version": "1.0.0",
+            "cli-minimal-version": "foobar",
+            "cli-bad-versions": [],
+            "services": {
+                "girder": {"url": "https://girder.dandi"},
+                "webui": {"url": "https://gui.dandi"},
+                "api": {"url": "https://publish.dandi/api"},
+                "jupyterhub": {"url": "https://hub.dandi"},
+            },
+        },
+    )
+    with pytest.raises(ValueError) as excinfo:
+        get_instance("https://example.dandi/")
+    assert str(excinfo.value).startswith(
+        "https://example.dandi/ returned an incorrectly formatted version;"
+        " please contact that server's administrators: "
+    )
+    assert "foobar" in str(excinfo.value)
--- a/dandi/tests/test_utils.py
+++ b/dandi/tests/test_utils.py
@@ -3,12 +3,17 @@ import os.path as op
 import time
 import datetime
 import pytest
+import responses
+from .. import __version__
+from ..consts import dandi_instance, known_instances
+from ..exceptions import BadCliVersionError, CliVersionTooOldError
 from ..utils import (
     ensure_datetime,
     ensure_strtime,
     find_files,
     flatten,
     flattened,
+    get_instance,
     get_utcnow_datetime,
     is_same_time,
     on_windows,
@@ -122,3 +127,145 @@ def test_flatten():
         0,
         1,
     ]
+
+
+@responses.activate
+def test_get_instance_dandi():
+    responses.add(
+        responses.GET,
+        "https://dandiarchive.org/server-info",
+        json={
+            "version": "1.0.0",
+            "cli-minimal-version": "0.5.0",
+            "cli-bad-versions": [],
+            "services": {
+                "girder": {"url": "https://girder.dandi"},
+                "webui": {"url": "https://gui.dandi"},
+                "api": {"url": "https://publish.dandi/api"},
+                "jupyterhub": {"url": "https://hub.dandi"},
+            },
+        },
+    )
+    assert get_instance("dandi") == dandi_instance(
+        girder="https://girder.dandi",
+        gui="https://gui.dandi",
+        redirector="https://dandiarchive.org",
+    )
+
+
+@responses.activate
+def test_get_instance_url():
+    responses.add(
+        responses.GET,
+        "https://example.dandi/server-info",
+        json={
+            "version": "1.0.0",
+            "cli-minimal-version": "0.5.0",
+            "cli-bad-versions": [],
+            "services": {
+                "girder": {"url": "https://girder.dandi"},
+                "webui": {"url": "https://gui.dandi"},
+                "api": {"url": "https://publish.dandi/api"},
+                "jupyterhub": {"url": "https://hub.dandi"},
+            },
+        },
+    )
+    assert get_instance("https://example.dandi/") == dandi_instance(
+        girder="https://girder.dandi",
+        gui="https://gui.dandi",
+        redirector="https://example.dandi/",
+    )
+
+
+@responses.activate
+def test_get_instance_cli_version_too_old():
+    responses.add(
+        responses.GET,
+        "https://example.dandi/server-info",
+        json={
+            "version": "1.0.0",
+            "cli-minimal-version": "99.99.99",
+            "cli-bad-versions": [],
+            "services": {
+                "girder": {"url": "https://girder.dandi"},
+                "webui": {"url": "https://gui.dandi"},
+                "api": {"url": "https://publish.dandi/api"},
+                "jupyterhub": {"url": "https://hub.dandi"},
+            },
+        },
+    )
+    with pytest.raises(CliVersionTooOldError) as excinfo:
+        get_instance("https://example.dandi/")
+    assert str(excinfo.value) == (
+        f"Client version {__version__} is too old!"
+        "  Server requires at least version 99.99.99"
+    )
+
+
+@responses.activate
+def test_get_instance_bad_cli_version():
+    responses.add(
+        responses.GET,
+        "https://example.dandi/server-info",
+        json={
+            "version": "1.0.0",
+            "cli-minimal-version": "0.5.0",
+            "cli-bad-versions": [__version__],
+            "services": {
+                "girder": {"url": "https://girder.dandi"},
+                "webui": {"url": "https://gui.dandi"},
+                "api": {"url": "https://publish.dandi/api"},
+                "jupyterhub": {"url": "https://hub.dandi"},
+            },
+        },
+    )
+    with pytest.raises(BadCliVersionError) as excinfo:
+        get_instance("https://example.dandi/")
+    assert str(excinfo.value) == (
+        f"Client version {__version__} is rejected by server!"
+        f"  Server requires at least version 0.5.0 (but not {__version__})"
+    )
+
+
+@responses.activate
+def test_get_instance_id_bad_response():
+    responses.add(
+        responses.GET,
+        "https://dandiarchive.org/server-info",
+        body="404 -- not found",
+        status=404,
+    )
+    assert get_instance("dandi") is known_instances["dandi"]
+
+
+@responses.activate
+def test_get_instance_known_url_bad_response():
+    responses.add(
+        responses.GET,
+        "https://dandiarchive.org/server-info",
+        body="404 -- not found",
+        status=404,
+    )
+    assert get_instance("https://dandiarchive.org") is known_instances["dandi"]
+
+
+@responses.activate
+def test_get_instance_unknown_url_bad_response():
+    responses.add(
+        responses.GET,
+        "https://dandi.nil/server-info",
+        body="404 -- not found",
+        status=404,
+    )
+    with pytest.raises(RuntimeError) as excinfo:
+        get_instance("https://dandi.nil")
+    assert str(excinfo.value) == (
+        "Could not retrieve server info from https://dandi.nil,"
+        " and client does not recognize URL"
+    )
+
+
+def test_get_instance_id_no_redirector():
+    with pytest.raises(ValueError) as excinfo:
+        get_instance("local-girder-only")
+    assert str(excinfo.value) == "DANDI instance has no known redirector URL"
--- a/dandi/utils.py
+++ b/dandi/utils.py
@@ -11,9 +11,9 @@ import sys
 
 from pathlib import Path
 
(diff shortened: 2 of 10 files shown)


dandi/utils.py
```diff
@@ -504,3 +509,73 @@ def delayed(*args, **kwargs):
...
+        if instance is not None:
+            lgr.warning("Using hard-coded URLs")
+            return instance
+        else:
+            raise RuntimeError(
+                f"Could not retrieve server info from {redirector_url},"
+                " and client does not recognize URL"
+            )
+    server_info = r.json()
+    minversion = Version(server_info["cli-minimal-version"])
```
yarikoptic, line 541: Since we went with simpler `cli-minimal-version` + `cli-bad-versions` (instead of `cli-version` packaging like specification) let's not bring packaging as a dependency.  Our version should be a semver when released, even for release candidates and it should work

```
In [4]: from semantic_version import Version; print(Version('0.6.0-rc1') < Version('0.6.0'))                                                           
True
```
I thought that the tricky part in either case would be the development state version 
```
$> dandi --version
0.5.0+101.ga115c19
```
but it seems to be perfectly fine!
```
In [1]: from semantic_version import Version; print(Version('0.5.0+101.ga115c19') < Version('0.6.0'))                                                  
True

In [2]: from semantic_version import Version; print(Version('0.5.0+101.ga115c19') < Version('0.5.0'))                                                  
False

In [3]: from semantic_version import Version; print(Version('0.5.0+101.ga115c19') < Version('0.5.1'))                                                  
True
```
since provisioned by https://semver.org/spec/v2.0.0.html#spec-item-10

dandi/utils.py
```diff
@@ -504,3 +509,73 @@ def delayed(*args, **kwargs):
...
+    if our_version in bad_versions:
+        raise BadCliVersionError(our_version, minversion, bad_versions)
+    return dandi_instance(
+        girder=server_info["services"]["girder"],
+        gui=server_info["services"]["webui"],
+        redirector=redirector_url,
+    )
+
+
+class CliVersionError(Exception):
```
yarikoptic, line 555: Let's centralize exception definitions in `dandi.exceptions`.  I would probably subclass from `RuntimeError` instead of a generic `Exception`

dandi/utils.py
```diff
@@ -504,3 +509,73 @@ def delayed(*args, **kwargs):
...
+        self.bad_versions = bad_versions
+
+    def server_requirements(self):
+        s = f"Server requires at least version {self.minversion}"
+        if self.bad_versions:
+            s += f" (but not {', '.join(map(str, self.bad_versions))})"
+        return s
+
+
+class CliVersionTooLowError(CliVersionError):
```
yarikoptic, line 568:
```suggestion
class CliVersionTooOldError(CliVersionError):
```
?  even though "Old" might be incorrect in terms of age, I rarely hear "Too Low Version" but do hear "Too Old Version".

dandi/utils.py
yarikoptic: probably worth adding  a docstring (or make `__str__` explicitly an `@abc.abstractmethod`) to note that this is just a base exception, without `__str__` and thus derived classes should be used

jwodder: Done.

dandi/utils.py
jwodder: Changed.

jwodder:
@yarikoptic Some of the tests I wrote for `get_instance()` use `dandi.__version__`, but since the GitHub Actions workflow only checks out the most recent commit, we end up with a version string like `0+untagged.1.g08f7fb1`, which semantic_version errors on.  Thoughts on how to address this?

yarikoptic:
> Thoughts on how to address this?

do a full checkout?

jwodder:
That works, though I was hoping for something involving a smaller checkout (not that this repo is all that large in the first place, though).

New problem: The redirector uses the `GIRDER_URL` environment variable to connect to Girder, yet it also reports the same URL in the /server-info response.  When the redirector is running inside a Docker container, this causes problems[1], as Girder's internal-to-the-Docker-network URL is not the same as its external URL (and the external URL is on localhost, so it can't be used internally by the redirector container).  Should we add another environment variable to the redirector script for specifying the internal/external Girder URL?

[1] Except when I ran the tests on my computer, for some reason.

setup.cfg
```diff
@@ -70,7 +71,7 @@ test =
     mock
     pytest
     pytest-cov
-    requests ~= 2.20
+    responses
```
yarikoptic, line 74: this is a new beast to me!  I have installed system wide python3-responses (0.9.0-2), but then
```
___________________________________________________________________________________________ ERROR collecting dandi/tests/test_utils.py ____________________________________________________________________________________________
dandi/tests/test_utils.py:6: in <module>
    import responses
/usr/lib/python3/dist-packages/responses.py:11: in <module>
    from collections import namedtuple, Sequence, Sized
<frozen importlib._bootstrap>:1039: in _handle_fromlist
    ???
/usr/lib/python3.8/collections/__init__.py:49: in __getattr__
    warnings.warn("Using or importing the ABCs from 'collections' instead "
E   DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
```
so we might need to allow for that warning?  With `pip install --upgrade responses` I got 0.9.0 and then it worked. .. I would till prefer to allow for earlier one to be allowed and suppress that DeprecationWarning -- we know that they fixed it up

jwodder: I've configured pytest to ignore DeprecationWarnings from responses now.

yarikoptic:
TL;DR: most likely indeed it would be best to just go with additional environment variable/configuration (e.g. `GIRDER_URL_INTERNAL` defaulting to the value of `GIRDER_URL`) for redirector, even though I do not quite like that, but at least it would be the most lightweight solution ATM.

Some details: rright... just some additional digesting for myself: redirector itself needs to be able to correspond to girder instance.  For that reason currently `GIRDER_URL: http://girder:8080` is set in `dandi/tests/data/dandiarchive-docker/docker-compose.yml`, but for outside client we would need to provide something like `http://localhost:8079` .  So I thought may be redirector could just contact `http://localhost:8079` but it would then need to be not a `localhost` within docker but `localhost` on the host or the external IP.

On https://stackoverflow.com/a/43541681 found out about magical `host.docker.internal` but it is not there for linux https://github.com/docker/for-linux/issues/264 .  But even if it was supported - it would have probably interferred with possible firewall etc.

Also the same SO page talks about linux only "Use --network="host" in your docker run command, then 127.0.0.1 in your docker container will point to your docker host." but tells that it was removed and did not come back.  

The same SO page references https://github.com/qoomon/docker-host as a solution to establish NAT container as part of the setup... IMHO too heavy weight and seems might still need to require docker tune up to get it working.

yarikoptic:
re `[1]` -- from digesting SO it seems that there is a good amount of differences between how docker behaves across OSes at that level of network routing etc, and IIRC you are using OS X now.

jwodder:
Yes, I am using macOS, though if the tests can resolve the `girder` hostname while running on my computer, I'd expect `host`, `ping`, `nc`, or `curl` to be able to resolve the hostname as well while the containers are up, yet they can't.  Odd.  (Maybe it's a process session thing?)

jwodder:
I created a PR to add the environment variable to the redirector: https://github.com/dandi/redirector/pull/25

yarikoptic:
hm, if I add `    network_mode: host` to the redirector's part in the compose
```
$> git diff
diff --git a/dandi/tests/data/dandiarchive-docker/docker-compose.yml b/dandi/tests/data/dandiarchive-docker/docker-compose.yml
index ef0a109..0b9ad2e 100644
--- a/dandi/tests/data/dandiarchive-docker/docker-compose.yml
+++ b/dandi/tests/data/dandiarchive-docker/docker-compose.yml
@@ -41,6 +41,7 @@ services:
       - girder
     ports:
       - "8079:8080"
+    network_mode: host
     environment:
       GIRDER_URL: http://girder:8080
       GUI_URL: http://localhost:8086
```
with 
```
$> docker --version
Docker version 19.03.7, build 7141c19
$> docker-compose --version
docker-compose version 1.25.0, build unknown
```

then it seems to be able to access `curl http://localhost:8081` from within the container, thus reaching the girder... and looking at `ip addr` withing redirector container -- indeed the full set of host interfaces is there.   So we just need to be careful to not choose some  if it also works on OSX -- might be a solution without tuning the redirector for that, but might need a tune up to be able to run it on a different than 8080 port?

jwodder:
Setting `network_mode: host` and updating `GIRDER_URL` to `http://localhost:8081` appears to work on my computer, though now I can't figure out why the tests can apparently access the redirector yet at the same time in another window `curl -fsSL localhost:8079/server-info` (and also with port 8080) gives me "connection refused".

yarikoptic:
if I parsed the sentence correctly, `8079:` mapping is no longer in effect with `"host"`, so I cannot access it either from the host:
```
$> curl -fsSL localhost:8079/server-info
curl: (7) Failed to connect to localhost port 8079: Connection refused

$> curl -fsSL localhost:8080/server-info
{"version":"1.0.0","cli-minimal-version":"0.5.0","cli-bad-versions":[],"services":{"girder":{"url":"http:\/\/girder:8080"},"webui":{"url":"http:\/\/localhost:8086"},"api":{"url":"https:\/\/publish.dandiarchive.org\/api"},"jupyterhub":{"url":"https:\/\/hub.dandiarchive.org"}}}% 
```

jwodder:
Neither port 8079 nor port 8080 works for me now, and I've just realized that the tests might be failing to connect to the redirector as well, in which case they're falling back to the hard-coded URLs and emitting a (captured & suppressed) warning.

satra:
just some notes from local testing on macos.

i made the following changes to the docker-compose file, which allows me to access the relevant services via localhost, since they are exposed from the docker network. 

```
    environment:
      GIRDER_URL: http://localhost:8081
      GUI_URL: http://localhost:8086
      ABOUT_URL: http://www.dandiarchive.org
```

the GUI_URL doesn't seem to work for me (i.e. it doesn't show up), but girder is accessible and the redirector is accessible.

yarikoptic:
> ` GIRDER_URL: http://localhost:8081`

This would work for dandi-cli testing if we only talk to `/server-info` endpoint and not a path in redirector which requires its communication with girder.  So, to be able to test (in the future) any other `/path` of redirector we would need smth like https://github.com/dandi/redirector/pull/25 .  But for the purpose of this PR, we don't need any other path, so could indeed use this value, right @jwodder ?

> the GUI_URL doesn't seem to work for me

yeap, as we discovered in troubleshooting session with @satra yesterday, reason is: https://github.com/dandi/dandiarchive/issues/474 requiring a fix in the container first.  dandi-cli anyways is not concerned with web ui client ATM anyhow anyways.

setup.cfg
yarikoptic: great, thanks!

dandi/utils.py
yarikoptic: was implemented

dandi/utils.py
```diff
@@ -504,3 +510,44 @@ def delayed(*args, **kwargs):
...
+            lgr.warning("Using hard-coded URLs")
+            return instance
+        else:
+            raise RuntimeError(
+                f"Could not retrieve server info from {redirector_url},"
+                " and client does not recognize URL"
+            )
+    server_info = r.json()
+    minversion = Version(server_info["cli-minimal-version"])
+    bad_versions = [Version(v) for v in server_info["cli-bad-versions"]]
```
yarikoptic, line 543: although unlikely, it is possible that server provides some incorrectly formatted semantic version.  That would cause a generic `ValueError` into user's face like
```
$> python3 -c 'from semantic_version import Version; print(Version("0.6.0a1"))'   
  File "/usr/lib/python3/dist-packages/semantic_version/base.py", line 105, in __init__
    major, minor, patch, prerelease, build = self.parse(version_string, partial)
  File "/usr/lib/python3/dist-packages/semantic_version/base.py", line 303, in parse
    raise ValueError('Invalid version string: %r' % version_string)
ValueError: Invalid version string: '0.6.0a1'
```

I think we better guard for `ValueError` in these two lines, and raise some dedicated or just a ValueError with improved wording, such as 
```python
except ValueError as exc:
    raise ValueError(f"{redirector_url} returned incorrectly formatted version, please contact that server's administrators: {exc}")
```

jwodder:
@yarikoptic Yes, I believe that would work for now.

dandi/tests/test_utils.py
```diff
@@ -122,3 +127,145 @@ def test_flatten():
...
+        get_instance("https://dandi.nil")
+    assert str(excinfo.value) == (
+        "Could not retrieve server info from https://dandi.nil,"
+        " and client does not recognize URL"
+    )
+
+
+def test_get_instance_id_no_redirector():
+    with pytest.raises(ValueError) as excinfo:
+        get_instance("local-girder-only")
```
yarikoptic, line 270: let's for now still allow such instances and then use fields from the hardcoded record, as I have described in the last bullet item of https://github.com/dandi/dandi-cli/issues/184 . E.g. so we could still test against `docker-compose` run of https://github.com/dandi/dandiarchive/tree/master/docker which does not spin up redirector yet.

jwodder:
The tests pass with `GIRDER_URL` set to http://localhost:8081.

dandi/utils.py
jwodder: Done.

dandi/tests/test_utils.py
jwodder: Done.

dandi/utils.py
```diff
@@ -504,3 +510,50 @@ def delayed(*args, **kwargs):
...
+        else:
+            raise RuntimeError(
+                f"Could not retrieve server info from {redirector_url},"
+                " and client does not recognize URL"
+            )
+    server_info = r.json()
+    try:
+        minversion = Version(server_info["cli-minimal-version"])
+        bad_versions = [Version(v) for v in server_info["cli-bad-versions"]]
+    except ValueError as e:
```
yarikoptic, line 545: thinking out loud: sorry for being slow, it got me thinking (since we do not test for version yet etc), what if any of those two is missing? and it kinda makes sense if server does not demand any specific version of the client somehow... I think later some time we might want to allow them to be "optional" but for now we would demand them, and if not present -- blow informatively... or let's just add that later ;)

satra:
> dandi-cli anyways is not concerned with web ui client ATM anyhow anyways.

don't you have a download test for subject folders based on the web ui link? but it's more about processing the link than for the ui to actually work :)

PR #189, opened 2020-08-12
Add "resume" functionality to download

### LINKED ISSUE #182: download: add "resume" functionality
ATM (as of WiP #134) we retry download up to 3 times.  We do it from the beginning of the file, which might be just a waste of bandwidth and "not productive".  Ideally we should just 
- roll back a little to position e.g. `Y=downloaded - 2*chunk_size`
- instruct downloader to get a stream from `Y` (not sure if girder allows, so might need to become only for direct downloads from S3)
- download first chunk (again -- the same up to 3 times rule) and compare to the content in previously downloaded -- it matches -- proceed with writing to the file starting at position `downloaded - chunk_size`, if not -- bail and try downloading from beginning.

edit1: Should probably be implemented within RFed `download` as around this point https://github.com/dandi/dandi-cli/pull/134/files#diff-f84d8150d7873723a79e1057aced4240R717 of instantiating a new `downloader` for the first/next attempt:  so that `downloader` returned by `get_download_file_iter` would need to gain a new `, startat=0` (or alike) option to start downloading from a specific offset. Unfortunately girder's `downloadFileAsIterator` does not provide such ability AFAIK, but eventually we would be going not through girder anyways, but a direct download from the URL.  As for how to do it with `requests`, here is an example: https://stackoverflow.com/a/22894873

jwodder:
~~Note that this PR merges into the #134 branch, not into master.~~ (no longer needed)

Closes #182

### DIFF
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -754,9 +754,11 @@ def _download_file(
                             blob1 = writer.read(len(block))
                             if blob1 == block:
                                 downloaded += len(block)
+                                lgr.debug("Resuming earlier download")
                                 continue
                             else:
                                 downloaded = 0
+                                lgr.debug("Could not resume earlier download")
                                 break
                         first = False
                     if digester:
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -737,12 +737,11 @@ def _download_file(
     # TODO: how do we discover the total size????
     # TODO: do not do it in-place, but rather into some "hidden" file
     downloaded = 0
+    prev_digest = None
     for attempt in range(3):
         try:
             if downloaded:
                 downloaded = max(downloaded - 2 * MAX_CHUNK_SIZE, 0)
-            if digester:
-                downloaded_digest = digester()  # start empty
             warned = False
             # I wonder if we could make writing async with downloader
             with open(path, "a+b") as writer:
@@ -755,13 +754,23 @@ def _download_file(
                             if blob1 == block:
                                 downloaded += len(block)
                                 lgr.debug("Resuming earlier download")
+                                if prev_digest:
+                                    # Set digest state one block back from last
+                                    # run
+                                    downloaded_digest = prev_digest
+                                else:
+                                    assert digester is None
                                 continue
                             else:
                                 downloaded = 0
+                                downloaded_digest = digester()
                                 lgr.debug("Could not resume earlier download")
                                 break
+                        elif digester:
+                            downloaded_digest = digester()  # start empty
                         first = False
                     if digester:
+                        prev_digest = downloaded_digest.copy()
                         downloaded_digest.update(block)
                     downloaded += len(block)
                     # TODO: yield progress etc
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -5,6 +5,7 @@ import os.path as op
 import random
 import re
 import requests
+import socket
 import sys
 import time
 
@@ -191,7 +192,7 @@ class _dandi_url_parser:
         url_server = groups["server"]
         server = cls.map_to[server_type].get(url_server.rstrip("/"), url_server)
 
-        if not server.endswith("/"):
+        if server and not server.endswith("/"):
             server += "/"  # we expected '/' to be there so let it be
 
         if server_type == "girder":
@@ -610,7 +611,11 @@ def _map_to_girder(url):
     # This is a duplicate call from above but it is cheap, so decided to just redo
     # it here instead of passing all the variables + url
     server_type, server_url, asset_type, asset_id = parse_dandi_url(url)
-    server_url = known_instances[known_instances_rev[server_url.rstrip("/")]].girder
+    server_url = known_instances[
+        known_instances_rev[server_url.rstrip("/")]
+        if server_url
+        else "local-docker-tests"
+    ].girder
     server_type = "girder"
     client = girder.get_client(server_url, authenticate=False, progressbars=True)
     # TODO: RF if https://github.com/dandi/dandiarchive/issues/316 gets addressed
@@ -797,13 +802,17 @@ def _download_file(
                 else:
                     break
             # both girder and we use HttpError
-        except requests.exceptions.HTTPError as exc:
+        except (requests.exceptions.RequestException, socket.timeout) as exc:
             # TODO: actually we should probably retry only on selected codes, and also
             # respect Retry-After
-            if attempt >= 2 or exc.response.status_code not in (
-                400,  # Bad Request, but happened with gider:
-                # https://github.com/dandi/dandi-cli/issues/87
-                503,  # Service Unavailable
+            if attempt >= 2 or (
+                isinstance(exc, requests.exceptions.HTTPError)
+                and exc.response.status_code
+                not in (
+                    400,  # Bad Request, but happened with gider:
+                    # https://github.com/dandi/dandi-cli/issues/87
+                    503,  # Service Unavailable
+                )
             ):
                 lgr.debug("Download failed: %s", exc)
                 yield {"status": "error", "message": str(exc)}
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -13,6 +13,7 @@ from urllib.parse import unquote as urlunquote
 from .dandiapi import DandiAPIClient
 from . import girder, get_logger
 from .consts import (
+    MAX_CHUNK_SIZE,
     dandiset_metadata_file,
     known_instances,
     known_instances_rev,
@@ -735,15 +736,29 @@ def _download_file(
 
     # TODO: how do we discover the total size????
     # TODO: do not do it in-place, but rather into some "hidden" file
+    downloaded = 0
     for attempt in range(3):
         try:
-            downloaded = 0
+            if downloaded:
+                downloaded = max(downloaded - 2 * MAX_CHUNK_SIZE, 0)
             if digester:
                 downloaded_digest = digester()  # start empty
             warned = False
             # I wonder if we could make writing async with downloader
-            with open(path, "wb") as writer:
-                for block in downloader():
+            with open(path, "a+b") as writer:
+                first = True
+                for block in downloader(start_at=downloaded):
+                    if first:
+                        writer.seek(downloaded)
+                        if downloaded:
+                            blob1 = writer.read(len(block))
+                            if blob1 == block:
+                                downloaded += len(block)
+                                continue
+                            else:
+                                downloaded = 0
+                                break
+                        first = False
                     if digester:
                         downloaded_digest.update(block)
                     downloaded += len(block)
@@ -762,7 +777,8 @@ def _download_file(
                         # TODO: ETA etc
                     yield msg
                     writer.write(block)
-            break
+                else:
+                    break
             # both girder and we use HttpError
         except requests.exceptions.HTTPError as exc:
             # TODO: actually we should probably retry only on selected codes, and also
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -753,7 +753,10 @@ def _download_file(
                             blob1 = writer.read(len(block))
                             if blob1 == block:
                                 downloaded += len(block)
-                                lgr.debug("Resuming earlier download from position %d", downloaded + len(block))
+                                lgr.debug(
+                                    "Resuming earlier download from position %d",
+                                    downloaded + len(block),
+                                )
                                 if prev_digest:
                                     # Set digest state one block back from last
                                     # run
@@ -764,7 +767,10 @@ def _download_file(
                             else:
                                 downloaded = 0
                                 downloaded_digest = digester()
-                                lgr.debug("Could not resume earlier download. Will possibly retry from the beginning")
+                                lgr.debug(
+                                    "Could not resume earlier download."
+                                    " Will possibly retry from the beginning"
+                                )
                                 break
                         elif digester:
                             downloaded_digest = digester()  # start empty
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -753,7 +753,7 @@ def _download_file(
                             blob1 = writer.read(len(block))
                             if blob1 == block:
                                 downloaded += len(block)
-                                lgr.debug("Resuming earlier download")
+                                lgr.debug("Resuming earlier download from position %d", downloaded + len(block))
                                 if prev_digest:
                                     # Set digest state one block back from last
                                     # run
@@ -764,7 +764,7 @@ def _download_file(
                             else:
                                 downloaded = 0
                                 downloaded_digest = digester()
-                                lgr.debug("Could not resume earlier download")
+                                lgr.debug("Could not resume earlier download. Will possibly retry from the beginning")
                                 break
                         elif digester:
                             downloaded_digest = digester()  # start empty
--- a/dandi/girder.py
+++ b/dandi/girder.py
@@ -104,14 +104,25 @@ class GirderCli(gcl.GirderClient):
             **kw,
         )
 
-    if _DANDI_LOG_GIRDER:
+    def sendRestRequest(self, *args, **kwargs):
         # Overload this core method to be able to log interactions with girder
         # server from the client side
-        def sendRestRequest(self, *args, **kwargs):
+        if _DANDI_LOG_GIRDER:
             lgr.debug("REST>: args=%s kwargs=%s", args, kwargs)
+        try:
             res = super().sendRestRequest(*args, **kwargs)
+        except gcl.HttpError as e:
+            # girder_client raises errors on 2xx values other than 200 and 201,
+            # so we need to salvage those:
+            if e.response.ok:
+                res = e.response
+                if kwargs.get("jsonResp", True):
+                    res = res.json()
+            else:
+                raise
+        if _DANDI_LOG_GIRDER:
             lgr.debug("REST<: %s", str(res))
-            return res
+        return res
 
     def register_dandiset(self, name, description):
         """Register a dandiset and return created metadata record
--- a/dandi/girder.py
+++ b/dandi/girder.py
@@ -375,11 +375,11 @@ class GirderCli(gcl.GirderClient):
             for attempt in range(3):
                 try:
                     path = f"file/{file_id}/download"
-                    headers = None
+                    params = None
                     if start_at > 0:
-                        headers = {"Range": f"bytes={start_at}-"}
+                        params = {"offset": str(start_at)}
                     resp = self.sendRestRequest(
-                        "get", path, stream=True, jsonResp=False, headers=headers
+                        "get", path, stream=True, jsonResp=False, parameters=params
                     )
                     return resp.iter_content(chunk_size=chunk_size)
                 except gcl.HttpError as exc:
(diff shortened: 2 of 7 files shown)


dandi/download.py
```diff
@@ -706,15 +707,29 @@ def _download_file(
...
+                    if first:
+                        writer.seek(downloaded)
+                        if downloaded:
+                            blob1 = writer.read(len(block))
+                            if blob1 == block:
+                                downloaded += len(block)
+                                continue
+                            else:
+                                downloaded = 0
+                                break
```
yarikoptic, line 731: please add `.debug` level lgr messages in both cases

dandi/download.py
```diff
@@ -706,15 +707,29 @@ def _download_file(
...
-                for block in downloader():
+            with open(path, "a+b") as writer:
+                first = True
+                for block in downloader(start_at=downloaded):
+                    if first:
+                        writer.seek(downloaded)
+                        if downloaded:
+                            blob1 = writer.read(len(block))
+                            if blob1 == block:
+                                downloaded += len(block)
```
yarikoptic, line 727: since succesful, and we went back two chunks, either here we would need to refeed a new digester with previously downloaded data, or may be just avoid updating digester below for redownloaded block?  without it -- it would end up with an incorrect digest I think, or am I wrong?

dandi/girder.py
```diff
@@ -353,14 +353,20 @@ def get_download_file_iter(self, file_id, chunk_size=MAX_CHUNK_SIZE):
...
             # Will do 3 attempts to avoid some problems due to flaky/overloaded
             # connections, see https://github.com/dandi/dandi-cli/issues/87
             for attempt in range(3):
                 try:
-                    return self.downloadFileAsIterator(file_id, chunkSize=chunk_size)
-                    break
+                    path = f"file/{file_id}/download"
+                    headers = None
+                    if start_at > 0:
+                        headers = {"Range": f"bytes={start_at}-"}
```
yarikoptic, line 365: oh man -- that would work?! ;)  how could we test I wonder...

dandi/tests/test_download.py
```diff
@@ -140,29 +140,32 @@ def test_download_multiple_files(monkeypatch, tmpdir):
...
+                len(args) > 1
+                and args[1].startswith("file/")
+                and args[1].endswith("/download")
+            ):
+                Mocks.ntries += 1
+                if Mocks.ntries < 3:
+                    raise gcl.HttpError(
+                        text="Failing to download", url=url, method="GET", status=500
+                    )
+            return orig_sendRestRequest(self, *args, **kwargs)
```
yarikoptic, line 161: if any of our test files big enough so we could trigger redownload when somewhere past `2*chunk` sizes?

dandi/download.py
jwodder: Done.

dandi/download.py
jwodder: Commit ba8767f should address this.

dandi/girder.py
jwodder: According to [the SO question you linked](https://stackoverflow.com/questions/22894211/how-to-resume-file-download-in-python/22894873#22894873), that works.  However, not all servers support ranges to begin with.  It's possible to test for support by checking the response for an "Accept-Ranges" header (cf. [this MDN article](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests), which doesn't mention the "start-" syntax).

I just noticed something in that MDN article: The response to a "Range" request has status 206, but `girder_client` regards anything other than 200 and 201 as an error ([see code here](https://girder.readthedocs.io/en/latest/_modules/girder_client.html#GirderClient.sendRestRequest)).

dandi/tests/test_download.py
jwodder: No.  Both of the files in `dandi/tests/data/dandifiles` are under a megabyte.

dandi/girder.py
yarikoptic: Thanks for the detailed analysis!  I guess for `girder` lets
- submit a PR  (that could be helpful to seek feedback from its developers -- may be it couldn't work in principle with it?)
- since it would take time I bet to get PR merged and new girder-client released, we could shim `sendRestRequest` by overloading it, and if raised `HttpError` status is 206, redo handling of the `.response` which seems to be stored in the exception in its entirety.  It is ugly somewhat but I do not see why it should not work. Then we could make shim conditional on the version of girder_client.

dandi/tests/test_download.py
yarikoptic: Let's make a fixture producing sizeable file for local upload/download testing as I have outlined in the review summary?

dandi/girder.py
jwodder: PR submitted: https://github.com/girder/girder/pull/3301

jwodder: `sendRestRequest()` overloaded in commit 89f84d7.

jwodder:
@yarikoptic Is there any particular attribute/structure of an NWB file that should be "filled up"?  Is there a way to just insert random bytes?

yarikoptic:
At first I thought to ask nwb experts on what would be the best .nwb file for testing we could produce... but since it is a file we are to generate dynamically anyways and we just need "size", let's just follow one of the tutorials at https://pynwb.readthedocs.io/en/latest/tutorials/index.html and indeed just fill up with some random junk (or 0s, which would be ok but might be suboptimal for testing download retries ;)).

So, for now, just follow e.g. this tutorial on extracellular recordings (which would be useful also to get a better understanding of structures within .nwb which would be useful for later work with metadata etc): https://pynwb.readthedocs.io/en/latest/tutorials/domain/ecephys.html#sphx-glr-tutorials-domain-ecephys-py  which already has code to populate that "demo" NWB with random data.

dandi/download.py
yarikoptic: thanks ... Going through the code I think that should work indeed!   But logic is a bit tricky (at least to me).  So I think we just need a good number of tests on some file longer than 3 chunks, and have 3 runs with download interruption on 1st chunk, then on 2nd, then on 3rd.  And then some test with fails on multiple chunks (so may be if file was 5 chunks, fail on 3 rd and 5 th).  Just to see that all aspects of restart work out nicely.

dandi/download.py
```diff
@@ -706,16 +707,41 @@ def _download_file(
...
+                                lgr.debug("Resuming earlier download")
```
yarikoptic, line 727:
```suggestion
                                lgr.debug("Resuming earlier download from position %d", downloaded + len(block))
```

dandi/download.py
```diff
@@ -706,16 +707,41 @@ def _download_file(
...
+                                    # Set digest state one block back from last
+                                    # run
+                                    downloaded_digest = prev_digest
+                                else:
+                                    assert digester is None
+                                continue
+                            else:
+                                downloaded = 0
+                                downloaded_digest = digester()
+                                lgr.debug("Could not resume earlier download")
```
yarikoptic, line 738:
```suggestion
                                lgr.debug("Could not resume earlier download. Will possibly retry from the beginning")
```
so we are not left wondering... could be improved with factual "will" or not restart.  Then please place `3` into a variable (`nattempts`?) so there is no hardcoded values in the code if you decide to compare if that is the last one...

dandi/download.py
jwodder: Do you have any suggestions on how to simulate download interruptions?  The best thing I've been able to find for this is https://github.com/Shopify/toxiproxy, but it doesn't handle TLS certificates intelligently, and either way I've been unable to actually connect to its proxy in a minimal setup.

yarikoptic: Isn't our docker instance run over a straight http without any encryption? I would have just tried to mock shim the requests streaming "thing" to abort with some code reaching the desired chunk

jwodder: I'm not clear on exactly what error should be raised when simulating a broken connection.  `HTTPError` (the only thing the download code currently catches) is wrong, as that's only raised in response to 4xx and 5xx status codes, which occur at the beginning of a response.  [This SO post](https://stackoverflow.com/a/25862433/744178) says that a lost connection while iterating through `iter_content` results in a `socket.timeout` error, but I was unable to reproduce this in a minimal example in my own environment; when I tried, the script just hung for 15 minutes and then exited as though `iter_content` had ended naturally.

yarikoptic: I was thinking about 408 Request Timeout . Anyways - this is primarily to test out code/logic and ability to restart from a given point. I wouldn't necessarily aim to 100% match the real life scenario at this point

jwodder: What URL should be passed to `download()` to make it download a given Dandiset from the Docker Compose instance?  If I use `http://localhost:8079/dandiset/{dandi_id}/draft`, dandi-cli makes a HEAD request to the URL which messes with the mocked `iter_content()`.  If I use `http://localhost:8086/#/dandiset/{dandi_id}/draft`, I get `'NoneType' object has no attribute 'endswith'`.

yarikoptic: probably no need for entire dandiset download -- a single URL directly to the file should be enough.

"drafts" would still go through the girder, and only released ones (with a version) through new API.So they should be the "easiest" to try.  But we have not probably interfaced yet the API for "publish". you could find `/publish` end point at http://publish.dandiarchive.org/swagger/ (there is also #122 but I think @satra did not get to trigger it yet).  So might need need to be done first to test "our" (not girder client) direct downloader via requests.

You can mock exclusively that `/dandisets/{version__dandiset__pk}/versions/{version__version}/assets/{uuid}/download/` end point (eg. probably just pass through anything which `not re.search(query, '/versions/.*/assets/.*/download/)`.  You would just need to discover UUID explicitly (via a call which downloader user to get a list of assets) or indeed, as I guess you have tried, download an entire dandiset.

You could similarly pass through for whatever doesn't look like girder's request for a file (do not have a URL handy but also you could see how it looks like if you just print those which are sent out to requests by girder client).

jwodder: OK, I'm getting a bit lost.  Just how do you envision this test being structured?  Note that `responses` does not support raising an error in the middle of a streaming response, if that's what you were getting at.

yarikoptic: Hm... What then happens if eg connection is lost? If some other exception - let's raise that one?

jwodder: [As I said above](https://github.com/dandi/dandi-cli/pull/189#discussion_r473018715), when I tried inducing a lost connection with a test script (by turning off my WiFi), there was no error.  However, an SO post from six years ago says that such an event should raise a `socket.timeout`.

Even if we do determine the proper error to raise, there's still the matter of what to mock and how so that the error gets raised at the right point.

yarikoptic: hm... I have also failed so far (by disabling/enabling WiFi) to trigger requests to give up and raise some exception (but I didn't wait for too long I guess).  I have not realized that requests is really trying hard to maintain that stream.  But I doubt it would try going back and resume if connection does get interrupted somehow (e.g. jumping from one network to another?).

As for what to mock, I would have mocked `get_download_file_iter` which we have in both girder and dandiapi.  Mock shim them would obtain original generator function, but instead of returning that one, would return one which would be a generator which would yield from the original one until it hits the chunk it wants to raise on (implementation of `get_download_file_iter` in dandicli is easier to grasp - there is an explicit loop there).  I would for now raise `socket.timeout` SO talks about.

Related to all this is #198 where also "resume" would be used.

jwodder: I still can't figure out what URL to pass to `download()`.  The mock no longer interferes with `http://localhost:8079/dandiset/{dandi_id}/draft`, but now the request to that URL is returning 500, presumably because of the `GIRDER_URL` issue discovered in #186 that we didn't completely address because editing `serve.py` wasn't necessary at the time.

jwodder:
@yarikoptic Using the new redirector image after merging dandi/redirector#25, the URL `http://localhost:8079/dandiset/{dandi_id}/draft` successfully redirects to `http://localhost:8086/#/dandiset/000004/draft`, but then that causes the tests to fail on [this line](https://github.com/dandi/dandi-cli/blob/00d94d56182186160bb0d1c7546f98a587114210/dandi/download.py#L194) with `'NoneType' object has no attribute 'endswith'`.  This appears to be due to the `local-docker-tests` instance in `known_instances` having a `None` value for its `api` attribute.

yarikoptic:
> but then that causes the tests to fail on [this line](https://github.com/dandi/dandi-cli/blob/00d94d56182186160bb0d1c7546f98a587114210/dandi/download.py#L194) with `'NoneType' object has no attribute 'endswith'`

I think it should be safe to adjust it to `if server and not server.endswith("/"):`

jwodder:
@yarikoptic Now the tests fail [on this line](https://github.com/dandi/dandi-cli/blob/00d94d56182186160bb0d1c7546f98a587114210/dandi/download.py#L584) with `'NoneType' object has no attribute 'rstrip'`.

yarikoptic:
I do not see any "proper" handling there, but would do smth like
```python
    # TODO: workaround - if no server_url, assume "local-docker-tests"
    server_url = known_instances[known_instances_rev[server_url.rstrip("/")] if server_url else "local-docker-tests"].girder
```

jwodder:
Current status: `test_resume_download` is hanging for upwards of ten minutes.  Inserting logging & print statements shows that the hang occurs when the original `GirderCli.get_download_file_iter(file_id, chunk_size)` is called for the second time.  I am unable to reproduce this in a minimal test script.

yarikoptic:
Great!  I will try to dig into it as well .

yarikoptic:
BTW, note that although we found that we could pass Range within the header to girder, it is not necessarily it can support those internally ... in my case the testing eventually (indeed was stuck somewhere, didn't dig yet) fails with
```
dandi/tests/test_download.py:269: in test_resume_download
    download(large_dandiset, tmp_path)
dandi/download.py:308: in download
    for rec in gen_:
dandi/download.py:452: in download_generator
    for resp in _download_generator:
dandi/download.py:725: in _download_file
    for block in downloader(start_at=downloaded):
dandi/tests/test_download.py:258: in new_downloader
    assert start_at == expected_start
E   assert 0 == 25165824
E     -0
E     +25165824
```
may be we are doomed to implement this feature only for "dandiapi" portion (not worth spending time ATM on girder IMHO) and thus wait for being able to talk to django?

jwodder:
@yarikoptic Actually, when I was trying to create an MVCE for the hanging request, I found that Girder didn't seem to honor the "Range" header (despite Girder's response headers indicating that it should), but it did honor an "offset" query parameter mentioned in the interactive documentation, so I changed the GirderCli method to use that in eb7e0d5.

yarikoptic:
I have just merged the #232  so I guess we could progress on this one with a "large file upload/release" fixture and implementing integration testing.

jwodder:
@yarikoptic How exactly do I do a "release" or otherwise take advantage of the "publish" container?  There's no command for that in dandi-cli at the moment.

yarikoptic:
Just go to `/swagger` interface of the running instance, there should be `/publish` end point - something like `/dandisets/{dandiset_id}/versions/{version}/publish` (or the one with `{version}` in the options to the call, don't remember).  Then add interface for it to `dandiapi.DandiAPIClient` class.

jwodder:
Assuming that by `/swagger` you mean the `/api/v1` page of Girder, I see no such endpoint, not even when logged in as `admin` to an instance connected to dandi-publish using the latest versions of the Docker images.

jwodder:
Questions about the endpoint:

* Is `dandiset__pk` supposed to be the dandiset identifier, a Girder ID, or something else?
* What exactly is the "name" parameter in the request body for?
* What exactly is the "dandiset" parameter in the request body for, seeing as the dandiset is already specified in the request path?
* Should the "version" parameter in the request body be set when making the request?  Should the Python method require a "version" argument?

yarikoptic:
> EDIT: Found it on the dandi-publish container.

just in case: http://publish.dandiarchive.org/swagger/

> Is dandiset__pk supposed to be the dandiset identifier, a Girder ID, or something else?

our dandiset identifier, like anything in the API (which should have no IDs from girder), like `000001`

> What exactly is the "name" parameter in the request body for?

worth clearing it up with @dandi/dandiarchive  developers (@brianhelba ?).  May be an annotation for the release (name of the release)

> What exactly is the "dandiset" parameter in the request body for, seeing as the dandiset is already specified in the request path?

yeah, that is an odd one, not sure why it is mandatory if asked for at all!?  May be it tollerates an empty `{}`?

> Should the "version" parameter in the request body be set when making the request? Should the Python method require a "version" argument?

It seems to have a default value assigned, so should be a keyword arg.  In tests might want to set explicitly but should match the pattern.

jwodder:
@yarikoptic It seems that the best way to perform a "publish" action at the moment is via the `POST /dandi/{dandiset_id}` endpoint in Girder, which makes the appropriate calls to dandi-publish.  However, when I call this endpoint in the test fixture, it fails with:

```
girder_client.HttpError: HTTP error 400: POST http://localhost:8081/api/v1/dandi/000004
Response text: {"message": "Failed to sync 000004", "type": "rest"}
```

The logs for the Girder container don't contain anything informative, and the Django container logs contain:

```
[19:58:26] ERROR    Invalid HTTP_HOST header: 'django:8000'. You exception.py:75
                    may need to add 'django' to ALLOWED_HOSTS.                  
           WARNING  Bad Request: /api/dandisets/sync/                 log.py:222
[19:58:26] WARNING  "POST /api/dandisets/sync/?folder-id=5f57e25 basehttp.py:157
                    b71c98353d53b4783 HTTP/1.1" 400 75749   
```

I have filed an issue with dandi-publish to support setting `ALLOWED_HOSTS` via an environment variable: https://github.com/dandi/dandi-publish/issues/148

yarikoptic:
>  It seems that the best way to perform a "publish" action at the moment is via the `POST /dandi/{dandiset_id}` endpoint in Girder, which makes the appropriate calls to dandi-publish. 

IMHO we should avoid de-tour via girder and (may be picking idea from the girder's call) provide adequate record to DANDI API's `/publish/` call .

jwodder:
@yarikoptic Why should Girder be avoided?  Based on conversations in Slack, reimplementing the Girder endpoint would involve two calls to dandi-publish, one or more calls to Girder to get a folder ID, and duplication of the Django credentials given to Girder.

yarikoptic:
ok, to get us going - since DANDI `/publish/` API seems to be not in a usable independently (without interactions with girder) yet (although it could/should have been), lets invoke girder endpoint for now.
But my thinking is that in the longer run we should exercise DANDI API endpoints and avoid girder end points since client should aim to talk to DANDI API not Girder.  To that end filed https://github.com/dandi/dandi-publish/issues/149 and we will just RF our code to talk to `/publish/` directly after that issue is resolved.

yarikoptic:
probably do not spend time messing with girder ATM.  @satra is yet to announce that next week we will push strongly to just get unified API going so all efforts will be spent on that and all "girder" will be hidden behind API so we would be able to just talk to API to accomplish what we need to accomplish.

yarikoptic:
#247 AFAIK provided alternative implementation

PR #192, opened 2020-08-13
map_to_click_exceptions: Convert gcl.HttpErrors to their "message" fields

### LINKED ISSUE #129: UX: provide custom responses for Girder originating exceptions
ATM we just use `str(e)` within `map_to_click_exceptions`.

but instead we could use smth like `get_HttpError_response(e).get('message')` to provide the message itself without much of decoration which could scare the user.

jwodder:
Closes #129

### DIFF
--- a/dandi/cli/command.py
+++ b/dandi/cli/command.py
@@ -9,13 +9,11 @@ from os import path as op
 
 import click
 from click_didyoumean import DYMGroup
-import girder_client as gcl
 
 from .. import get_logger, set_logger_level
 
 from .. import __version__
 from ..consts import dandiset_metadata_file, known_instances
-from ..girder import get_HttpError_response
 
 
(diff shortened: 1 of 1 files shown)


dandi/cli/command.py
```diff
@@ -9,11 +9,13 @@
 
 import click
 from click_didyoumean import DYMGroup
+import girder_client as gcl
```
yarikoptic, line 12: I think so far we did manage to avoid importing it in the path of CLI invocation so e.g. during `dandi --help`... please move it to the point of use -- seems to save 50-100ms from run time.  Minor -- but it all adds up

jwodder: Done.

yarikoptic:
Great.  We will add a test for it (a new click interface to smth raising that exception), may be, later

PR #196, opened 2020-08-13
ENH: pytest.ini

yarikoptic:
individual commits have more reasoning described.


jwodder:
There's already some pytest config in `tox.ini`.  The two sections should be merged together in a single file.  (I would prefer keeping the config in `tox.ini` just because I don't like proliferation of config files.)

Issue #199, opened 2020-08-14
Upload tests are failing


jwodder:
Currently, when I run the test suite on my computer, two tests fail that didn't before: `test_upload` and `test_upload_existing_error`, both from `dandi/tests/test_upload.py`.  They appear to both be failing because the just-uploaded file can't be found on Girder.  This happens when testing commits 82e1820, f31e469, and bee7d50 (the 0.6.0 tag), the last two of which succeeded when initially run on GitHub Actions.  All tests were run using tox in a clean repository with the most recent versions of the Docker images.

My current hypothesis is that a dependency released a new version that broke something.

jwodder:
It appears that the new release yesterday of pynwb is to blame.  Now, `dandi.pynwb_utils.validate()` fails to validate the NWB test data files, with output such as:

```
[VectorIndex/description (units/electrodes_index): argument missing, VectorIndex (units/electrodes_index): incorrect type - expected 'uint8', got 'int32', VectorIndex/description (units/electrodes_index): argument missing, VectorIndex (units/electrodes_index): incorrect type - expected 'uint8', got 'int32', VectorIndex/description (units/spike_times_index): argument missing, VectorIndex (units/spike_times_index): incorrect type - expected 'uint8', got 'int32']
```

satra:
woah - the index types definitely look weird - surely indices can exceed 256? i'm not saying we need int32, having indices be unsigned is good. perhaps something to ping on the pynwb issue. i can't remember where the test files originally came from.

jwodder:
@satra I took the test files used for testing uploads from [Dandiset 000006](https://gui.dandiarchive.org/#/dandiset/000006/draft).

yarikoptic:
It is probably just a matter of pynwb and its schema got "fixed up" to use uint instead of int, those files which we use `sub-anm369963_ses-20170228.nwb` were produced with previous version... and in this case they are so old that produced with even older version which didn't cache schema inside the files I guess:

```
$> dandi ls dandi/**/sub-anm369963_ses-20170228.nwb | grep nwb
- path: dandi/tests/data/dandifiles/sub-anm369963/sub-anm369963_ses-20170228.nwb
  nwb_version: 2.0.2
```
for "permanent fix" we should just switch to using pre-populated dynamically by fixtures files (as WiP for #182) which would (better) be "compatible" with current pynwb  since they would be produced by it.

jwodder:
[...]

PR #200, opened 2020-08-14
Fix upload tests

jwodder:
Generate upload test files on the fly instead of storing (now outdated) files in the repo

Addresses the original problem in #199 but not the new failures


dandi/tests/fixtures.py
```diff
@@ -63,6 +64,35 @@ def simple2_nwb(simple1_nwb_metadata, tmpdir_factory):
     )
 
 
+@pytest.fixture()
+def organized_nwb_dir(simple2_nwb, tmp_path, clirunner):
+    shutil.copy(str(simple2_nwb), str(tmp_path))
+    (tmp_path / dandiset_metadata_file).write_text("{}\n")
+    r = clirunner.invoke(organize, ["-f", "move", "--dandiset-path", str(tmp_path)])
```
yarikoptic, line 71: Coughs use `organize -f copy` instead of explicit shutil.copy above

dandi/tests/fixtures.py
```diff
@@ -63,6 +64,35 @@ def simple2_nwb(simple1_nwb_metadata, tmpdir_factory):
...
+        subject=pynwb.file.Subject(
+            subject_id="lizard001",
+            date_of_birth=datetime(2019, 12, 1, tzinfo=tzutc()),
+            sex="F",
+            species="Gekko gecko",
+        ),
+        **simple1_nwb_metadata,
+    )
+    (tmp_path / dandiset_metadata_file).write_text("{}\n")
+    r = clirunner.invoke(organize, ["-f", "move", "--dandiset-path", str(tmp_path)])
```
yarikoptic, line 90: Same here

dandi/tests/fixtures.py
jwodder: The goal of the fixture is to simply to create a temporary directory containing organized NWB files, a `dandiset.yaml`, and nothing else.  Not mixing in the `simple*.nwb` files at the root of the dandiset makes it cleaner.

yarikoptic: `-f copy` would just do that shutil.copy from their original location into properly named organized files. You just need to point organize command to original paths. Or am I missing the point?

jwodder: Where would it put the organized files?  Next to the originals (not what I want) or in the current directory?

yarikoptic: It should place into the specified dandiset path, otherwise should assumr that curdir is the dandiset path. Dude might need time up if that is not said in --help

yarikoptic:
woohoo -- having merged #201 and rerunning the tests, now all "pull_request" ones are green!

I have (without merging master in) took liberty in 7e6868ef3878abcec8ff21a7d92594e0b6ac7102 to address one of recommended changes to use `organize -f copy` for the first fixture only, since in the 2nd an additional file is created straight in that fixture/path so `-f copy` would indeed not be appropriate there since it would leave that original file also in that dandiset, so I decided to just leave it as is.

I have also added a tiny tune up in 43f1f39c204d985c30aab81b04b320a8d156ae6a to one of the asserts to just take advantage that AFAIK sum can happily "sum" booleans which just would cast to int.

Also in 5ab7fff85b6238c5a476e7abe905b7a71f7169f4 to "quickly" test that change I added a "dummy" `test_fixtures` which would just "smoke test" the fixtures which could be faster than just to wait for full blown `test_upload.py` run.  But timing were somewhat surprising to me -- since I was assuming that the "organized_" fixtures were session level:

```
$> python -m pytest -v dandi/tests/test_upload.py dandi/tests/test_fixtures.py
============================================== test session starts ==============================================
platform linux -- Python 3.8.5, pytest-4.6.9, py-1.8.1, pluggy-0.13.0 -- /home/yoh/proj/dandi/dandi-cli/venvs/dev3/bin/python
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/home/yoh/proj/dandi/dandi-cli/.hypothesis/examples')
rootdir: /home/yoh/proj/dandi/dandi-cli, inifile: tox.ini
plugins: mock-3.2.0, cov-2.10.0, xdist-1.32.0, timeout-1.3.3, forked-1.1.3, hypothesis-4.36.2
collected 6 items                                                                                               

dandi/tests/test_upload.py::test_upload PASSED                                                            [ 16%]
dandi/tests/test_upload.py::test_upload_existing_error PASSED                                             [ 33%]
dandi/tests/test_upload.py::test_upload_locks PASSED                                                      [ 50%]
dandi/tests/test_upload.py::test_upload_unregistered PASSED                                               [ 66%]
dandi/tests/test_fixtures.py::test_organized_nwb_dir PASSED                                               [ 83%]
dandi/tests/test_fixtures.py::test_organized_nwb_dir2 PASSED                                              [100%]

=========================================== slowest 10 test durations ===========================================
54.17s setup    dandi/tests/test_upload.py::test_upload
12.59s teardown dandi/tests/test_fixtures.py::test_organized_nwb_dir2
2.05s setup    dandi/tests/test_fixtures.py::test_organized_nwb_dir2
1.99s setup    dandi/tests/test_upload.py::test_upload_existing_error
1.97s setup    dandi/tests/test_upload.py::test_upload_unregistered
1.92s setup    dandi/tests/test_fixtures.py::test_organized_nwb_dir
1.90s setup    dandi/tests/test_upload.py::test_upload_locks
0.61s call     dandi/tests/test_upload.py::test_upload
0.32s call     dandi/tests/test_upload.py::test_upload_existing_error
0.28s call     dandi/tests/test_upload.py::test_upload_locks
```
then I checked the code to realized that they weren't.  So I added  one more commit which made clirunner and those into session level fixtures: 099c682e93f9bf8ce16f7ed23bfd50e206404f69 .

I hope those changes are ok with you @jwodder - I just wanted to rush this PR up, so we could get master green and kick out a quick bugfix release.  Feel welcome to submit any desired changes to my changes as a PR to follow up.

dandi/tests/fixtures.py
yarikoptic: actually I was wrong here  - since one more file created straight in that path... so I just left it alone for now

dandi/tests/test_upload.py
```diff
@@ -100,16 +91,14 @@ def test_upload_locks(local_docker_compose_env, mocker, monkeypatch, tmp_path):
...
-    (tmp_path / dandiset_metadata_file).write_text("identifier: '999999'\n")
-    (tmp_path / DIRNAME).mkdir(exist_ok=True, parents=True)
-    copyfile(DANDIFILES_DIR / DIRNAME / FILENAME, tmp_path / DIRNAME / FILENAME)
-    monkeypatch.chdir(tmp_path)
+    (organized_nwb_dir / dandiset_metadata_file).write_text("identifier: '999999'\n")
+    monkeypatch.chdir(organized_nwb_dir)
     with pytest.raises(ValueError) as excinfo:
-        upload(paths=[DIRNAME], dandi_instance=dandi_instance_id, devel_debug=True)
+        upload(paths=[dirname], dandi_instance=dandi_instance_id, devel_debug=True)
     assert str(excinfo.value) == (
```
yarikoptic, line 102: for some reason coverage marks all code here as "red" as it is not run on CI -- we will need to double check.

yarikoptic:
all green "pull_request" still, I will proceed.

PR #201, opened 2020-08-16
BF(TST): redirector now directs to new web ui addresses for drafts of dandisets

yarikoptic:
no girder based urls any more.  

Hopefully that resolves our now failing tests.


yarikoptic:
on ubuntus it is the upload tests are failing, I guess #200 is addressing.  For the purpose of this PR fix, seems to be in effect, so merging

PR #209, opened 2020-08-25
Give download an option for only downloading dandiset.yaml/assets

### LINKED ISSUE #132: download: add an option to not download the entire dandiset
the UX problem is that we do allow to download an arbitrary directory, files, etc.

We need an option for a download at dandiset level to not download individual assets, and just get the `dandiset.yaml`.

@satra thinks about `--no-assets` option.

Could be `--download all,dandiset.yaml,assets` so we could make it more flexible in the future if we get other types of entities.

jwodder:
Closes #132.

### DIFF
--- a/dandi/cli/cmd_download.py
+++ b/dandi/cli/cmd_download.py
@@ -88,11 +88,11 @@ class ChoiceList(click.ParamType):
 @map_to_click_exceptions
 def download(url, output_dir, existing, jobs, format, download_types):
     """Download a file or entire folder from DANDI"""
-    # First boring attempt at click commands being merely an interface to
-    # Python function
-    from ..download import download
+    # We need to import the download module rather than the download function
+    # so that the tests can properly patch the function with a mock.
+    from .. import download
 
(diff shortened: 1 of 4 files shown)


dandi/cli/cmd_download.py
```diff
@@ -4,6 +4,29 @@
...
 
 
+class ChoiceList(click.ParamType):
+    name = "choice-list"
+
+    def __init__(self, values):
+        self.values = set(values)
+
+    def convert(self, value, param, ctx):
+        if value is None or isinstance(value, set):
```
yarikoptic, line 14: I think it would be wortwhile to reincarnate unit testing of `cli` interface, while mocking out actual functionality (`dandi.download.download` in this case)

jwodder: I added unit tests for the `--download` option.  Is that what you had in mind?

yarikoptic: YES!  Thank you!

PR #210, opened 2020-08-25
BF/RF: upload - Path for file paths, and PurePosixPath for anything dealing with girder

yarikoptic:
TODO
-  confirm failing on windows -- impossible in tests ATM
- [x] fix up -- tentative (untested) fix is pushed in c1daa2322ef2ade67b5ff277d90fab70d6a44fc0

### DIFF
--- a/dandi/upload.py
+++ b/dandi/upload.py
@@ -1,6 +1,4 @@
 from datetime import datetime
-import os
-import os.path as op
 import re
 import sys
 import time
@@ -28,7 +26,7 @@ def upload(
     allow_any_path=False,
     devel_debug=False,
 ):
-
+    # PurePosixPath to be cast to for paths on girder
     from pathlib import Path, PurePosixPath
     from .dandiset import Dandiset
     from .support.digests import Digester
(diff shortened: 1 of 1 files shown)


yarikoptic:
d'oh -- that effort was somewhat useless since `local_docker_compose_env` is not available on Windows so I guess those tests do not even run.

yarikoptic:
I will try to test it on windows very soon, but if I fail to do so I would like to have it merged anyways if no other issues are found. Giving away to @jwodder for the review.

dandi/tests/test_upload.py
```diff
@@ -30,14 +32,16 @@ def test_upload(local_docker_compose_env, monkeypatch, organized_nwb_dir2):
     client = girder.get_client(local_docker_compose_env["instance"].girder)
     for dname in [dirname1, dirname2]:
         with pytest.raises(girder.GirderNotFound):
-            girder.lookup(client, collection_drafts, path=f"{dandi_id}/{dname}")
+            girder.lookup(client, collection_drafts, path=op.join(dandi_id, dname))
```
jwodder, line 35: `girder.lookup()` passes `path` to `resourceLookup()`, [the documentation for which](https://girder.readthedocs.io/en/latest/python-client.html#girder_client.GirderClient.resourceLookup) states that "The path must be an absolute Unix path".  All of the changes to `girder.lookup()` in this PR need to be undone.

dandi/upload.py
```diff
@@ -148,10 +145,27 @@ def skip_file(msg):
...
     # provided to pyout to cause it to halt
     def process_path(path, relpath):
+        """
+
+        Parameters
+        ----------
+        path: Path
+          Non Pure (OS specific) Path
+        relpath:
+          For location on Girder.  Will be casted to PurePosixPath
```
jwodder, line 155: "Cast", not "casted".

dandi/tests/test_upload.py
yarikoptic: d'oh -- right, indeed all paths to girder should be the POSIX ones.  dropping that 24820e0fc316128fbf85c4dc384c09b31d6733ee commit

dandi/upload.py
yarikoptic: thanks! will fix for that

yarikoptic:
heh, finally established some ad-hoc "Windows VM contacts host docker compose instance" setup for testing (with https://github.com/dandi/dandi-cli/pull/223) -- our test_upload.py tests fail neither with master nor after reverting ce4fc58e37a2b73bf0b05f5e49ebd5169ffc6fdd (the "fix") :-(

PR #215, opened 2020-08-25
ENH(TST): add a very basic test for all commands providing some --help

yarikoptic:
Primarily a smoke test for options rendering etc.


yarikoptic:
Great, thank you @jwodder for the review!

PR #216, opened 2020-08-26
Make filepaths OS-agnostic

### LINKED ISSUE #211: remove POSIX filenames assumptions in the tests
This is a follow up to https://github.com/dandi/dandi-cli/pull/210
where I did that in `test_upload`.  So the goal is to check other tests and provide the same (or similar?) changes in other test files.

jwodder:
Closes #211.

### DIFF
--- a/dandi/tests/test_upload.py
+++ b/dandi/tests/test_upload.py
@@ -1,3 +1,4 @@
+import os
 import pytest
 
 from .. import girder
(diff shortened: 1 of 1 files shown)


yarikoptic:
and there are no other tests (in other `test_` files) making similar assumptions? (just want to make sure before merging)

jwodder:
@yarikoptic I searched for every occurrence of a forward slash or the strings `"."` and `".."`.  These are the only occurrences that aren't part of URLs or pathlib operators.

yarikoptic:
great, thanks!

PR #217, opened 2020-08-26
Fix some find_files() regexen

jwodder:
Without this change, find_dandi_files() would match files whose names ended with "dandiset.yaml" (e.g., "foodandiset.yaml").  I don't think that's intended.

Using `$` instead of `\Z` causes a regex to also accept strings that end with a newline.  That's not desirable here.


yarikoptic:
Great, thank you @jwodder !

PR #220, opened 2020-08-26
Handle symlinks in tests on Windows

### LINKED ISSUE #213: "tox" run failing tests on Windows
I am a proud user of a VM with Windows 7 and conda + pip installed dandi development environment...  Running `tox` resulted in:

```
====================================================================================== short test summary info =======================================================================================
FAILED dandi/support/tests/test_cache.py::test_memoize_path - OSError: [WinError 1314] A required privilege is not held by the client: 'file.dat' -> 'C:\\Users\\DataLad\\AppData\\Local\\Temp\\pyte...

FAILED dandi/support/tests/test_iterators.py::test_IteratorWithAggregation - assert False
FAILED dandi/tests/test_organize.py::test_organize_nwb_test_data[simulate] - assert 1 == 0
FAILED dandi/tests/test_organize.py::test_organize_nwb_test_data[symlink] - assert 1 == 0
FAILED dandi/tests/test_organize.py::test_organize_nwb_test_data[auto] - assert False
FAILED dandi/tests/test_organize.py::test_detect_link_type[True-True-symlink] - AssertionError: assert 'hardlink' == 'symlink'
FAILED dandi/tests/test_organize.py::test_detect_link_type[True-False-symlink] - AssertionError: assert 'copy' == 'symlink'
============================================================================== 7 failed, 56 passed, 9 skipped in 39.81s ==============================================================================
Exception ignored in: <ssl.SSLSocket fd=460, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('10.0.2.15', 51095), raddr=('3.19.164.171', 443)>
ResourceWarning: unclosed <ssl.SSLSocket fd=460, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('10.0.2.15', 51095), raddr=('3.19.164.171', 443)>
ERROR: InvocationError for command 'C:\Users\DataLad\PycharmProjects\dandi-cli\.tox\py3\Scripts\coverage.EXE' run -m pytest -v dandi (exited with code 1)
______________________________________________________________________________________________ summary _______________________________________________________________________________________________
  lint: commands succeeded
ERROR:   py3: commands failed

```

not sure why does not fail for us on Windows github CI runs.

jwodder:
I don't have a Windows machine to test this on, but I expect it'll work.

This PR does NOT contain a fix for the bug I pointed out in the comments of #213.

I don't know why `test_IteratorWithAggregation` was failing, so this PR does not address it.

Fixes #213.

### DIFF
--- a/dandi/tests/test_organize.py
+++ b/dandi/tests/test_organize.py
@@ -1,6 +1,7 @@
 import os
 from glob import glob
 import os.path as op
+from pathlib import Path
 import ruamel.yaml
 
 from ..consts import file_operation_modes
@@ -115,6 +116,19 @@ def test_organize_nwb_test_data(nwb_test_data, tmpdir, clirunner, mode):
         nwb_test_data = op.relpath(nwb_test_data, cwd)
         outdir = op.relpath(outdir, cwd)
 
+    src = Path(tmpdir, "src")
+    src.touch()
+    dest = Path(tmpdir, "dest")
+    try:
+        dest.symlink_to(src)
+    except OSError:
+        symlinks_work = False
+    else:
+        symlinks_work = True
+
+    if mode in ("simulate", "symlink") and not symlinks_work:
+        pytest.skip("Symlinks not supported")
+
     input_files = op.join(nwb_test_data, "v2.0.1")
 
     cmd = ["-d", outdir, "--files-mode", mode, input_files]
@@ -146,7 +160,7 @@ def test_organize_nwb_test_data(nwb_test_data, tmpdir, clirunner, mode):
 
     if mode == "simulate":
         assert all((op.isabs(p) != relative) for p in produced_paths)
-    elif mode == "symlink" or mode == "auto":
+    elif mode == "symlink" or (mode == "auto" and symlinks_work):
         assert all(op.islink(p) for p in produced_nwb_paths)
     else:
         assert not any(op.islink(p) for p in produced_paths)
@@ -216,11 +230,12 @@ def test_ambiguos_probe1():
     ],
 )
 def test_detect_link_type(monkeypatch, tmp_path, sym_success, hard_success, result):
+    def succeed_link(src, dest):
+        pass
+
     def error_link(src, dest):
         raise OSError("Operation failed")
 
-    if not sym_success:
-        monkeypatch.setattr(os, "symlink", error_link)
-    if not hard_success:
-        monkeypatch.setattr(os, "link", error_link)
+    monkeypatch.setattr(os, "symlink", succeed_link if sym_success else error_link)
+    monkeypatch.setattr(os, "link", succeed_link if hard_success else error_link)
     assert detect_link_type(tmp_path) == result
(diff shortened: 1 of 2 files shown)


yarikoptic:
I have submitted a tune up to that script: https://github.com/datalad/datalad/pull/4833 . Running 
```
~datalad/datalad-master/tools/eval_under_testloopfs python -m pytest -s -v dandi/tests/test_organize.py dandi/support/tests/test_cache.py
```
lead to
```
dandi/tests/test_organize.py:144: in test_organize_nwb_test_data
    assert r.exit_code == 0
E   assert 1 == 0
E     -1
E     +0
```
and
```
dandi/support/tests/test_cache.py:120: in test_memoize_path
    check_new_memoread(0, "content", True)
dandi/support/tests/test_cache.py:100: in check_new_memoread
    assert len(calls) == ncalls + 1 + int(expect_new)
E   assert 3 == 4
E     -3
E     +4
```
fails.

yarikoptic:
on windows VM -- those seems to not fail :-/  only `test_IteratorWithAggregation` -- I will look into fixing that one in a separate PR

jwodder:
The `test_organize_nwb_test_data` failure was because vfat apparently doesn't support hardlinks.  I've pushed a commit to make the test skip the "hardlink" mode when they're not supported.

I don't know why the cache test is failing, but the only references to symlinks are after the failure point, and I see no references to hard links.

dandi/tests/test_organize.py
```diff
@@ -115,6 +116,31 @@ def test_organize_nwb_test_data(nwb_test_data, tmpdir, clirunner, mode):
...
+    try:
+        dest.unlink()
+    except FileNotFoundError:
+        pass
+    try:
+        os.link(src, dest)
+    except OSError:
+        hardlinks_work = False
+    else:
+        hardlinks_work = True
```
yarikoptic, line 137: I am ok to merge as is to progress.  

But I feel that we already have logic in multiple places, so it would be worthwhile to introduce some helpers. E.g. (design is just suggestive):
- a helper `get_filesystem_capabilities(dirpath, capabilities=('symlinks', 'hardlinks'))` which would may be return a set of supported capabilities.
- And some shim for `tmpdir` fixture (`tmpdir_caps`) fixture, which could 
  - enhance `tmpdir` with args `skip_missing=None` where we would list capabilities which if missing would cause test to be skipped
  - `tmpdir` provided structure should be enhanced with `.caps` or `.capabilities` with the set from above helper or add a set of `.supports_X` for X being each capability.
  then in each test we could avoid doing all such checks, but just `if tmpdir.supports_symlinks`

What do you think @jwodder ?

jwodder: Augmenting `tmpdir` seems a little like overkill at the moment.  Perhaps just using `supports_{sym,hard}links(dir)` helpers that wrap `get_filesystem_capabilities()` would be enough.  Other than that, sounds like a plan.

yarikoptic: ok, let's then proceed with merge of this PR for now as is. Helper could followup in its own PR

PR #221, opened 2020-08-26
BF(TST): detect slow machine (e.g. Win7 VM) and increase sleeps for such case to 0.1

yarikoptic:
Partially addresses #213


yarikoptic:
eh -- last test "fixup" was not entirely good - mac run failed
```
_________________________ test_IteratorWithAggregation _________________________
dandi/support/tests/test_iterators.py:54: in test_IteratorWithAggregation
    assert got == [0]
E   assert [0, 1] == [0]
E     Left contains one more item: 1
E     Full diff:
E     - [0]
E     + [0, 1]
```

yarikoptic:
I think it is good now, but I will restart sweep of tests once more and then if still green it could be merged.

PR #222, opened 2020-08-26
Fix bugs in test_memoize_path

jwodder:
This PR applies the fixes I suggested in [this comment](https://github.com/dandi/dandi-cli/issues/213#issuecomment-680996204).

### DIFF
--- a/dandi/support/tests/test_cache.py
+++ b/dandi/support/tests/test_cache.py
@@ -5,6 +5,7 @@ import sys
 import time
 
 from dandi.support.cache import PersistentCache
+from dandi.utils import on_windows
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


yarikoptic:
covered now but windows ain't happy ;)

jwodder:
I believe the Windows failure may be due to a change in Python in 3.8 (Note that the Windows tests fail for 3.6 and 3.7 but pass for 3.8).  Specifically, in Python 3.8, [`os.path.realpath()`](https://docs.python.org/3/library/os.path.html#os.path.realpath) started resolving symlinks on Windows, whereas it didn't before, thereby changing the behavior of [this line](https://github.com/dandi/dandi-cli/blob/be0bd80589dfd76f63cbfe14533992d536361cd8/dandi/support/cache.py#L82).

[This SO thread](https://stackoverflow.com/q/43333640/744178) offers two workarounds:

* Use `os.readlink()` instead, which will result in different behavior when a directory component is a symlink
* Use yet another third-party library

yarikoptic:
Let's look forward -- @jwodder please just skip that test on windows with `python < 3.8` -- one extra call should still be safe/correct operation (just less of caching) I guess.

jwodder:
@yarikoptic Done.

yarikoptic:
Great -- thanks!

PR #223, opened 2020-08-27
90% of way to enable testing from windows VM over to an instance running on the host

yarikoptic:
On the host ATM I did

```
(git)lena:~/proj/dandi/dandi-cli[master]git
$> git diff
diff --git a/dandi/tests/data/dandiarchive-docker/docker-compose.yml b/dandi/tests/data/dandiarchive-docker/docker-compose.yml
index 5b825e2..2ce67d3 100644
--- a/dandi/tests/data/dandiarchive-docker/docker-compose.yml
+++ b/dandi/tests/data/dandiarchive-docker/docker-compose.yml
@@ -42,8 +42,8 @@ services:
     ports:
       - "8079:8080"
     environment:
-      GIRDER_URL: http://localhost:8081
-      GUI_URL: http://localhost:8086
+      GIRDER_URL: http://192.168.1.34:8081
+      GUI_URL: http://192.168.1.34:8086
       ABOUT_URL: http://www.dandiarchive.org
 
 volumes:
diff --git a/dandi/tests/test_upload.py b/dandi/tests/test_upload.py
index e7a3163..1100695 100644
--- a/dandi/tests/test_upload.py
+++ b/dandi/tests/test_upload.py
@@ -9,6 +9,7 @@ from ..utils import yaml_load
 
 
 def test_upload(local_docker_compose_env, monkeypatch, organized_nwb_dir2):
+    import pdb; pdb.set_trace()
     nwb_files = list(organized_nwb_dir2.glob(f"*{os.sep}*.nwb"))
     assert len(nwb_files) == 2
     dirname1 = nwb_files[0].parent.name
```

thus hardcoding my external IP (docker serves on it as well as localhost) to be pointed to by the redirector for the services.

May be you see @jwodder a better way to accomplish the mission?

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -55,6 +55,9 @@ development command line options.
   If handling of metadata has changed while developing, set this env var to
   `clear` to have cache `clear()`ed before use.
 
+- `DANDI_INSTANCEHOST` -- defaults to `localhost`. Point to host/IP which hosts
+  a local instance of dandiarchive.
+
 ## Sourcegraph
 
 The [Sourcegraph](https://sourcegraph.com) browser extension can be used to
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -56,11 +56,14 @@ development command line options.
   `clear` to have cache `clear()`ed before use.
 
 - `DANDI_INSTANCEHOST` -- defaults to `localhost`. Point to host/IP which hosts
-  a local instance of dandiarchive.
+  a local instance of dandiarchive. Typically,
+  `DANDI_REUSE_LOCAL_DOCKER_TESTS_API_KEY`
+  should also be set.
 
-- `DANDI_REUSE_LOCAL_DOCKER_TESTS_API_KEY` -- use to make `local_docker_compose*`
-  fixtures work by contacting an instance (possibly at `DANDI_INSTANCEHOST`)
-  instead of starting a new one.
+- `DANDI_REUSE_LOCAL_DOCKER_TESTS_API_KEY` -- make the
+ `local_docker_compose*` fixtures use the given API key for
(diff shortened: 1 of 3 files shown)


DEVELOPMENT.md
```diff
@@ -55,6 +55,13 @@ development command line options.
   If handling of metadata has changed while developing, set this env var to
   `clear` to have cache `clear()`ed before use.
 
+- `DANDI_INSTANCEHOST` -- defaults to `localhost`. Point to host/IP which hosts
```
jwodder, line 58: Mention that, when this is set, `DANDI_REUSE_LOCAL_DOCKER_TESTS_API_KEY` should be set as well.

DEVELOPMENT.md
```diff
@@ -55,6 +55,13 @@ development command line options.
...
+  a local instance of dandiarchive.
+
+- `DANDI_REUSE_LOCAL_DOCKER_TESTS_API_KEY` -- use to make `local_docker_compose*`
```
jwodder, line 61: Suggested description: "Make the `local_docker_compose*` fixtures use the given API key for `DANDI_INSTANCEHOST` instead of spinning up a new environment with a new key."

DEVELOPMENT.md
yarikoptic: although in my case and probably thus "typically" it is the case -- it not generally so since we could have some instance running e.g. on public IP, so we could still `docker-compose up` it and interact with it.

yarikoptic: but I will add a comment that the API_KEY should typically be set as well.

DEVELOPMENT.md
yarikoptic: done, thanks! pushed

PR #224, opened 2020-08-27
BF+ENH: do not puke while flattening and allow to work on broken links

yarikoptic:
See individual commit messages for more info

### DIFF
--- a/dandi/cli/cmd_ls.py
+++ b/dandi/cli/cmd_ls.py
@@ -4,6 +4,8 @@ import time
 
 import click
 
+from collections import defaultdict
+
 from .base import lgr, map_to_click_exceptions
 
 from ..utils import safe_call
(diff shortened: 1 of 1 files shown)


dandi/cli/cmd_ls.py
```diff
@@ -4,6 +4,8 @@
 
 import click
 
+from _collections import defaultdict
```
jwodder, line 7: `_collections` is an implementation detail.  You should import from `collections` instead.

yarikoptic: I could have sworn to type `collections`!  not sure how that `_` came about.  Force pushed the fix. Thanks!!

yarikoptic:
ok, recommendation was resolved, let's proceed

PR #206, opened 2020-08-23
ref: add schema models to dandi library

satra:
This is a start of a PR to bring models over to this library and using it to formalize both dandiset and asset metadata. I will ping folks when this is ready for review. 

I've also created a branch in the main repo, so that multiple people can push/merge to this PR.

PR #226, opened 2020-08-27
Function for converting NWB file to AssetMeta instance

jwodder:
This PR requires PR #206.

@satra Is this what you had in mind?


satra:
@jwodder - kind of. unless we change get_metadata to already do the mapping, you will likely need a transform or function that computes the asset field given the metadata. so perhaps something like: `setattr(asset, field, get_asset_field(metadata, field))`

```
def get_asset_field(metadata, field):
    if field == "wasDerivedFrom": 
        biosample = Biosample.unvalidated()
        for biosample_field in biosample.__fields__:
            setattr(biosample, biosample_field, get_asset_field(metadata, biosample_field))
        biosample = Biosample(**biosample)
    if field == "age":
        age = metadata["age"]
        # convert age to ISO 8601
        # create property value
        return property_value
    ...
```

and then after it is done you can check validation by: `AssetMeta(**asset.dict())`

yarikoptic:
FWIW: I merged #206 into master.

yarikoptic:
So, overall flow I think should be 

```
NWB -1-> metadata dict (unharmonized) -2-> AssetMeta
```

`-1->` is pretty much our `get_metadata` but we might need to extend it with getting more fields from nwb, so that `dict` would be sufficient for `-2->`.  

I think having an explicit `-1->` (instead of just feeding `-2->` with the nwb file) would be useful for initial harmonization if later we start also dealing with other file types (neuroimaging etc).

Somewhat of a "tricky" part is that e.g. Digests we estimate during upload (not on the fly yet, like we already do in download) and enhance metadata record with them.  So we should indeed allow for `-2->` without full validation right away -- or better with validation but allowing some fields to not be defined (such as Digests) before we estimate them.

As for `-2->` -- could be indeed the function as @satra suggested, but as we briefly touched upon yesterday, it might be worth right away to set them up within our model classes? e.g. instead of 

```python
contentSize: str = Field(nskey="schema")
path: str = Field(None, nskey="dandi")
```
have smth like
```python
contentSize: str = Field(nskey="schema", adapter="size")
path: str = Field(None, nskey="dandi", adapter="path")
```

for basic 1-to-1 mappings, but also allow for callables like 

```python
def adapter_age(metadata):
        age = metadata["age"]
        # convert age to ISO 8601
        # create property value
        return property_value
```
and then have
```python
age: Optional[PropertyValue] = Field(...allwhatithasnow, adapter=adapter_age)
```

Then for `DandiBaseModel` to define class method `def from_metadata(cls, metadata, allow_missing=None)` which would just iterate over `cls` attributes, use specified adapters to compose, validate (allowing for missing) and return the structure.  While iterating, it would not fail right away if any `KeyError` is raised, but collect a set of them to raise at the end with a single informative exception telling which keys (and may be for which attributes) are still required.

NB. `from_metadata`  could have avoided taking care about validation altogether, and we just provide additional helper to validate allowing for missing attributes, but then we would not be able to raise that informative message about which metadata fields are needed, which IMHO would be quite useful especially now while we are just establishing the mappings etc.

Re `BioSample`: When  looping it would discover that it is of (or a list of) type which is subclass of DandiBaseModel (like `BioSample` is) it would just delegate to that `cls.from_metadata(metadata, allow_missing=None)`.  

- may be "allow_missing" will need to allow for hierarchical specification to be able to pass into such nested calls, but I do not think it is needed ATM - I know only Digests

@satra @jwodder - does above make sense? ;)

satra:
makes sense to me. the only reason i would suggest the standalone function for now instead of an adapter is that we may want to push nwb to align as closely as possible in the future. 

and if ever we move to attrs from pydantic, it already has a notion of converter. the only reason we are using pydantic is to stay close to jsonschema and fastapi for now.

jwodder:
I have enhanced the code according to the above suggestions.  However, I would appreciate it if someone could provide a comprehensive example of NWB metadata (as extracted with `get_metadata()`) and the `AssetMeta` that it should be converted to so that I know that I'm handling & converting everything correctly.

satra:
@jwodder - a starting point would be to extract the nwb metadata for a few files from each of the datasets we have access to rather than something completely comprehensive. to save time, i think @yarikoptic  was going to do this on his backup system.

jwodder:
@satra I'm interested in the `AssetMeta`, not the base metadata.  For example, given a `dict` with all of the following fields:

https://github.com/dandi/dandi-cli/blob/00414f59d087e2eca9c63271642d7c5a02950150/dandi/consts.py#L9-L45

... where should everything end up in the resulting `AssetMeta`?  Which fields need special parsing?  Etc.

satra:
@jwodder - you kind of have to go field by field and ask is this information somewhere in the base metadata. and remember some fields are themselves structures - like biosample. 

easiest place to start is likely: https://github.com/dandi/dandi-cli/blob/00414f59d087e2eca9c63271642d7c5a02950150/dandi/consts.py#L22

most of which go into biosample.

we may also need to augment the model to allow for other things like number of electrodes, etc.,.  there is no specific place for this in the model. 

other aspects such as experimenter could be mapped to contributor with role Researcher. 

I think most of the fields have some description, but feel free to ask if something is unclear.

jwodder:
@satra Specific questions, then:

* How are `TypeModel` classes like `SexType` supposed to be used?  What should be `identifier`, and what should be `name`?
* What sort of input values are expected for `assayType` and `anatomy`, and how do they map to lists of values?
* How should a plain age like "5 years" be converted to a `PropertyValue`?  Would it just be `PropertyValue(value="P5Y")`?

satra:
> How are TypeModel classes like SexType supposed to be used? What should be identifier, and what should be name?

identifier should be the corresponding url from an ontology, name would be the label. more generally we can translate a few of those to enum types as we have done here: 

https://github.com/dandi/dandi-cli/blob/master/dandi/models.py#L87

> What sort of input values are expected for assayType and anatomy, and how do they map to lists of values?

assayType should come from OBI (which is a large ontology for biomedical investigation). the question here  is how to map nwb "datatypes"  to this.

> How should a plain age like "5 years" be converted to a PropertyValue? Would it just be PropertyValue(value="P5Y")?

correct, with unitText set to "Years from brith" for now or for gestational weeks "Weeks from conception".

> What should be done for the various required/defaultless fields like BioSample.assayType and AssetMeta.dataType that have no corresponding field in get_metadata()?

let's ask a few folks for hints.

@tgbugs - we are back to where do we find enumerations for all kinds of things and @bendichter  we are looking at how we get people to put the right info into nwb or get out of existing nwb (which does not have any ontology attached yet). the specific considerations here are: 
- modality 
- measurementTechnique
(the above two could come from: https://docs.google.com/presentation/d/1OzLDeTvebHXvn07_oagUcSH0xOuxzrP2GSKTdt5jJlg/edit#slide=id.g8daabd40c5_0_0)
- assayType (kind of processing that was done to the sample)
- dataType  (kind of processing that was done to the file)

all types are all listed here: 

https://github.com/dandi/dandi-cli/blob/master/dandi/models.py#L87

and here:

https://github.com/dandi/dandi-cli/blob/master/dandi/models.py#L159

bendichter:
@satra 

I think modality and measurementTechnique would mostly come from which neurodata_types are being used in the NWB file. For instance, 

If `ElectricalSeries` is used, it's "electrophysiology." If the file contains an `IntracellularElectrode`, I guess that's also electrophysiology? Does the ontology differentiate between intracellular and extracellular electrophysiology? NWB does.

If `Position` is used, it's a "behavioral approach." 

If `TwoPhotonSeries` is used, it's a "cell population imaging." 

Many experiments are combinations of these. Does this schema allow multiple modalities/measurementTechniques?

bendichter:
Would you be able to provide an example of assayType or dataType? I'm not sure if NWB stores this information.

yarikoptic:
> Does this schema allow multiple modalities/measurementTechniques?

yes, https://github.com/dandi/dandi-cli/blob/master/dandi/models.py#L574 -- all those are `List`:
```python
    modality: List[ModalityType] = Field(readonly=True, nskey="dandi")
    measurementTechnique: List[MeasurementTechniqueType] = Field(
        readonly=True, nskey="schema"
    )
    variableMeasured: List[PropertyValue] = Field(readonly=True, nskey="schema")
```
I think we have decided to postpone dealing with assets which have no useful data recorded (thus not optional - requires at least a single entry in the list), and I guess it might help to facilitate us to provide heuristics for all which we care about ATM ;) (since cannot just leave empty if we find no relevant one)

tgbugs:
All the [techniques](https://scicrunch.org/scicrunch/interlex/search?q=technique&l=technique) and [modalities](https://scicrunch.org/scicrunch/interlex/search?q=ilxtr_modality&l=ilxtr_modality) are in InterLex now. @tmsincomb have you had a chance to ingest the subClassOf for the modalities (the modelling of the techniques isn't amenable to ingesting those right at the moment)?

tmsincomb:
@tgbugs 
I start up the superclass update for modality. I was going to fix the update first, but I'll do that now to avoid issues.

jwodder:
@satra I've added a test case for conversion of a metadata `dict` to an `AssetMeta` to `dandi/tests/test_metadata.py`.  Please take a look at it and tell me if it does what you want.

tmsincomb:
@tgbugs The modalities superclasses are in now.

tgbugs:
@tmsincomb thanks! The subclassOf closure for modalities (experimental approaches) is visible [here](https://scicrunch.org/scicrunch/interlex/view/ilx_0770148#children).

tgbugs:
@bendichter at the technique level the ontology distinguishes between [intracellular](http://uri.interlex.org/base/ilx_0739521) and [extracellular](http://uri.interlex.org/base/ilx_0739578).

yarikoptic:
FWIW: I have submitted information to get an RRID for dandi-cli.  I will report back whenever it gets registered

yarikoptic:
I think it might be good time to also introduce here RF to `validate.py` (and possibly `cmd_validate.py`) to use this new schema/function.  I guess a good start is [`validate_dandi_nwb`](https://github.com/dandi/dandi-cli/blob/master/dandi/validate.py#L48).  Then new functionality could easily be tested on a sample of real files from dandiarchive dandisets.

yarikoptic:
dandi-cli, RRID is `SCR_019009`  .  It is not yet in the system (takes up to 8 business days) but will get there

jwodder:
I have rewritten `validate_dandi_nwb()` to use the new models & metadata extraction.

@yarikoptic Should the RRID be added to the code at this time?  If so, how, exactly?

yarikoptic:
> I have rewritten validate_dandi_nwb() to use the new models & metadata extraction.

awesome -- I will review and try it out later today.

> @yarikoptic Should the RRID be added to the code at this time? If so, how, exactly?

ideally -- yes.  I believe we thought about smth like
```python
metadata["wasGeneratedBy"] = Activity(
  identifier="SCR_019009", 
  name="RRID",
  description="dandi-cli", 
  endDate=<current date>,
  # ideally later we should add contributor somehow! e.g. could come from 
  # git config or some explicit dandi config whenever we get there
  # contributor=...
) 
```
(or `name` and `identifier` swapped, @satra ?)

satra:
```
metadata["wasGeneratedBy"] = Activity(
  "identifier": <uuid4()>,
  "name": "NWB conversion"
  "description": "Data converted using DANDI cli"
  "wasAssociatedWith": Software(identifier={"propertyID": "RRID", "value": "SCR_019009"},
                                name="DANDI Command Line Interface",  description="dandi-cli 0.6.4", 
                                version="0.6.4",
                                url="https://github.com/
                                ),
  "startedAt": <timestamp>
  "endedAt": <timestamp>
  ...
) 
```

dandi/metadata.py
```diff
@@ -83,3 +88,161 @@ def get_metadata(path):
...
+        metadata["digest"] = digest
+        metadata["digest_type"] = digest_type
+    metadata["contentSize"] = op.getsize(nwb_path)
+    metadata["encodingFormat"] = "application/x-nwb"
+    asset = metadata2asset(metadata)
+    end_time = datetime.now().astimezone()
+    asset.wasGeneratedBy = models.Activity(
+        identifier=str(uuid4()),
+        name="NWB conversion",
+        description="Data converted using DANDI cli",
```
satra, line 233: just a note that `startedAt` and `endedAt` should reflect the creation of the asset. which may be a little more complicated than this. this `startedAt` and `endedAt` seems to be more closely associated with metadata creation.

```suggestion
        description="Metadata generated by DANDI cli",
```

since the description is about the asset and it may have multiple activities that generate the asset (conversion, organization) or its metadata (this function).

for now this is ok, but we may need to get more detailed with `isPartOf` and `hasPart` keywords.

yarikoptic: I think we might need to make generated by into a list. Current activity about metadata, so time stamps are for time when converted by dandi cli. But we might want to add activity on nwb generation/convertion - that one we might later deduce how it was generated etc

dandi/metadata.py
```diff
@@ -230,7 +230,7 @@ def nwb2asset(nwb_path, digest=None, digest_type=None):
     asset.wasGeneratedBy = models.Activity(
         identifier=str(uuid4()),
         name="NWB conversion",
-        description="Data converted using DANDI cli",
```
yarikoptic: This name again should be adjusted, see my other comment on making it into a list

yarikoptic:
keywords -- IMHO should probably be optional, but even with this patch

```
$> git diff
diff --git a/dandi/models.py b/dandi/models.py
index 2907d24..9a37fab 100644
--- a/dandi/models.py
+++ b/dandi/models.py
@@ -475,7 +475,8 @@ class CommonModel(DandiBaseModel):
     protocol: Optional[List[str]] = Field(None, nskey="dandi")
     ethicsApproval: Optional[List[EthicsApproval]] = Field(None, nskey="dandi")
     license: List[License] = Field(nskey="schema")
-    keywords: List[str] = Field(
+    keywords: Optional[List[str]] = Field(
+        None,
         title="Keywords",
         description="Keywords or tags used to describe "
         "this content. Multiple entries in a "

```

validation fails (ignore the semver one since it is indeed old/incorrect) for me on our "test file" with

```
$> dandi validate ../nwb-datasets/000027/sub-RAT123/sub-RAT123.nwb        
2020-09-08 11:56:50,754 [ WARNING] A newer version (0.6.4) of dandi/dandi-cli is available. You are using 0.6.2+75.gc20d788.dirty
../nwb-datasets/000027/sub-RAT123/sub-RAT123.nwb: 2 error(s)
  warning: nwb_version '2.0b' is not a proper semantic version. See http://semver.org
  1 validation error for AssetMeta
keywords
  value is not a valid list (type=type_error.list)
Summary: Validation errors in 1 out of 1 files

```
and even if I remove `min_items=1,` ... why is that so?

satra:
> All the techniques and modalities are in InterLex now.

@tgbugs - is there a place i can quickly pull labels + uris

satra:
@yarikoptic - without knowing what you put into keywords hard to know what the issue is.

yarikoptic:
I think they are empty - that was my point.. I will double check later in actual value - may be it is something. That is a file from nwb tests dataset

satra:
i think it has to be None or a list. i don't think it can be an empty list.

satra:
but i could be wrong will have to check with pydantic.

jwodder:
@yarikoptic I cannot reproduce the "keywords" validation failure.  If I use commit c20d788, or if I make the changes in your patch and remove `min_items=1`, either way, running `dandi validate` on the file from 000027 gives me:

```
/Users/jwodder/Downloads/sub-RAT123.nwb: 1 error(s)
  warning: nwb_version '2.0b' is not a proper semantic version. See http://semver.org
Summary: Validation errors in 1 out of 1 files
```

yarikoptic:
grr... I might have mixed up virtualenvironments or smth.  With my patch above (without even removing min_items) it is performing correctly.  I have pushed 6811bc97b693dbe6929c8745018ec0d1af1eb3f4 here.  I guess time to get serious and run on some dandisets in the archive.  Try on some sample by manually downloading using `dandi download` and meanwhile I will look into establishing  a full dump of them on drogon.

jwodder:
I ran `dandi validate` from this branch on the Dandisets from the dandiarchive backup; it took about ten hours.  In brief, the results were:

- The files for 000003 failed validation with the following errors each:

    ```
    VectorIndex/description (units/spike_times_index): argument missing
  VectorIndex (units/spike_times_index): incorrect type - expected 'uint8', got 'int64'
  warning: nwb_version '2.0b' is not a proper semantic version. See http://semver.org
    ```

- The files for 000004 - 000011, 000019 - 000023, 000028, 000034, and 000035 failed validation with errors of the form:

    ```
    VectorIndex/description (units/electrodes_index): argument missing
  VectorIndex (units/electrodes_index): incorrect type - expected 'uint8', got 'int32'
  VectorIndex/description (units/electrodes_index): argument missing
  VectorIndex (units/electrodes_index): incorrect type - expected 'uint8', got 'int32'
  VectorIndex/description (units/spike_times_index): argument missing
  VectorIndex (units/spike_times_index): incorrect type - expected 'uint8', got 'int32'
    ```

- 000012 is similar to the previous but also includes some warnings of the form:

    ```
    /home/jwodder/dandi-cli/.tox/py3/lib/python3.8/site-packages/pynwb/icephys.py:309: UserWarning: Unit 'V' for VoltageClampStimulusSeries 'index_8' is ignored and will be set to 'volts' as per NWB 2.1.0.
    /home/jwodder/dandi-cli/.tox/py3/lib/python3.8/site-packages/pynwb/icephys.py:309: UserWarning: Unit 'A' for VoltageClampSeries 'index_0' is ignored and will be set to 'amperes' as per NWB 2.1.0.
    ```

- 000017 is similar to the above but also includes a warning of the form:

    ```
    /home/jwodder/dandi-cli/.tox/py3/lib/python3.8/site-packages/hdmf/spec/namespace.py:60: UserWarning: Loaded namespace 'AIBS_ecephys' is missing the required key 'version'. Version will be set to 'None'. Please notify the extension author.
    ```

- The NWB file for 000025 failed validation with:

    ```
    Failed to validate /mnt/backup/dandi/dandiarchive-replica/000025/001_140709EXP_A1_ndx_v3.nwb: data type 'ICEphysFile' not found in namespace core
  Failed to read metadata: 'NWBGroupSpec' object has no attribute 'shape'
    ```

- The NWB files for 000027 and 000029 failed validation with:

    ```
    warning: nwb_version '2.0b' is not a proper semantic version. See http://semver.org
    ```

- All other Dandisets validated without error.

dandi/metadata.py
```diff
@@ -83,3 +88,169 @@ def get_metadata(path):
...
+            sec += delta.microseconds / 1000000
+        s += f"T{sec}S"
+    if s == "P":
+        s += "0D"
+    return s
+
+
+def extract_sex(metadata):
+    if metadata.get("sex") is not None:
+        return models.SexType(identifier="sex", name=metadata["sex"])
```
satra, line 158: for each of these extractors we need to create a search tool. one that maps the existing metadata value to some known entity. i'll try to create a quick first pass at this.  but i think we are close now. we need to do a similar metadata map for the existing dandiset metadata schema as well.

dandi/metadata.py
```diff
@@ -83,3 +88,213 @@ def get_metadata(path):
...
+            value_id = None
+            value = "Other"
+        elif value.startswith("http"):
+            value_id = value
+            value = None
+        else:
+            raise ValueError(f"Cannot interpret sex field: {value}")
+        return models.SexType(identifier=value_id, name=value)
+    else:
+        return ...
```
satra, line 179: @jwodder and @yarikoptic - i added these mappers for sex (above) and species (below).  for the other mappers it would be good to know what's already in the nwb files across datasets.

satra:
@jwodder @yarikoptic - do you think we should add the dandiset metadata mapper from old version to new in this PR or a different PR?

dandi/metadata.py
```diff
@@ -83,3 +88,213 @@ def get_metadata(path):
...
+
+
+def extract_species(metadata):
+    value = metadata.get("species", None)
+    if value is not None:
+        value = value.lower()
+        if "mouse" in value or "mus" in value:
```
yarikoptic, line 186:
```suggestion
        if "mouse" in value or value.startswith("mus"):
```

dandi/metadata.py
```diff
@@ -83,3 +88,213 @@ def get_metadata(path):
...
+            value_id = "http://purl.obolibrary.org/obo/NCBITaxon_10090"
+            value = "House mouse"
+        elif "human" in value or "homo" in value:
```
yarikoptic, line 189:
```suggestion
        elif "human" in value or value.startswith("homo"):
```

dandi/metadata.py
yarikoptic: just to minimize possible false hits a bit.  Ideally `if/elif` should be replaced with "map to all hits" and "if multiple hits -- blow with error" so we at least do not breed some hits from ambigous entries (have none in mind besides "musrat" ;)).

yarikoptic:
This PR already tunes `validate_dandi_nwb`, thus if released we would start enforcing schema upon users trying to validate/upload.  I am not sure  if we are ready for that yet... we could add temporary treatment of `DANDI_SCHEMA` env variable and if defined -- use new logic, if not -- old.  This would allow then to merge and even let curious folks to use it, and eventually we would just remove that treatment when "we are ready". WDYT?

yarikoptic (changes requested):
Let's right away make mapping a bit more "robust" by at least catching ambiguous  items.

Alternative would be:
- establish some `schema_mappings.yaml` which would be hosted within this repo and contain exact values we encounter in data to be mapped to the corresponding term.  We will update it whenever we encounter new loose values
- make `dandi-cli` fetch updated version once in a while (or forcefully if requested) and cache  it locally on user's drive to be used instead of the copy shipped within.  This way we could let people update mapping without updating dandi-cli (although since its releases are quite automated now, we could even not bother about this feature for now)

edit: let's forget about `.yaml` file idea, probably would not be scalable etc, so  IMHO we should keep it a .py, but may be RF later on so we could indeed fetch the freshier "mapping" file to be used instead of the shipped one.

yarikoptic:
@jwodder please
- make linters happy
- unless there are objections/better ideas, please add an env var as described in https://github.com/dandi/dandi-cli/pull/226#issuecomment-736920997 so we could retain current behavior in master while allowing for this new mapping etc.

jwodder:
@yarikoptic I've satisfied the linters and added an `if os.environ.get("DANDI_SCHEMA"):` block to `validate_dandi_nwb()` to control whether to use the behavior from before or after this PR.

yarikoptic:
Great, thank you @jwodder -- let's proceed

PR #227, opened 2020-08-28
DOC: docstring for sleeping_range used in tests

yarikoptic:


yarikoptic:
not sure why coverage jumped up (besides may be we had some run failing in master before) but I like it ;)

PR #229, opened 2020-09-02
BF: address for changes in returned by API records

yarikoptic:
- datetime stamp now has zone `Z` suffix which breaks "stock" datetime.fromisoformat parsing -- use `datetutil` implementation
- metadata for dandiset now returned not within `dandiset` subkey of metadata record.

"Good" news -- no real released dandisets are out there yet, so no user should be impacted AFAIK.

### DIFF
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -401,10 +401,18 @@ def download_generator(
 
         # Handle our so special dandiset.yaml
         if dandiset and get_metadata:
+            dandiset_metadata = dandiset.get("metadata", {})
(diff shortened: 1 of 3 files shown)


yarikoptic:
Wow! Locking also seems to be impaired - so the situation might be worse than I thought. I will later fetch newer images and reproduce - tests passed for need locally.

jwodder (approved):

yarikoptic:
woohoo -- restarted the tests and pull requests are now green, no locking issues I had observed ... Let's proceed then

PR #231, opened 2020-09-03
Run GitHub Actions tests daily

### LINKED ISSUE #230: Add execution of ci tests on cron
So we get alerted if things break while we don't actively push changes on the client code

jwodder:
This configures the "test" workflow to be run once a day at 6 AM UTC, presumably when no one is working on the code.

Closes #230.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,8 +1,10 @@
 name: Tests
 
 on:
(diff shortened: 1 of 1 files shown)


yarikoptic:
thanks , failures are unrelated and should be gone by now, merging

PR #232, opened 2020-09-04
Add dandi-publish to Docker Compose setup

### LINKED ISSUE #164: Add "publish" (AKA API) service to testing docker compose setup
There was an attempt to provide a unified docker compose setup but it was not finished: https://github.com/dandi/dandi-publish/pull/71 .

To test new download of #134 without contacting the mothership deployment, and any other future interaction with dandiarchive through API we will need to improve upon docker compose recipe of https://github.com/dandi/dandi-cli/pull/159  to include also the publish/API service.

jwodder:
Closes #164.


yarikoptic:
Great! now we should be able to test `/publish` and thus downloads of versioned datasets etc. Thank you @jwodder !

PR #233, opened 2020-09-04
RF: move+RF code for navigating dandi archive url from download into (new) dandiarchive module

yarikoptic:
So that it could be used also for ls or any other code which needs to
go through the thing (dandiset, folder, assets()) in the archive.
ATM we need to flip between Girder and DANDI API, so hat helper abstracts
that away a bit more on top of having GirderCli and DandiAPIClient.

At this point it is hoped that RFing was done well enough to not break code but there is no real benefit.
`git diff --color-moved=zebra master...` could be of benefit to see what code moved vs was deleted/added.

Following TODO(s) would take advantage of RF

- [x] Add ability to `ls` to list remote dandiset/folder/file on the archive 

I think I am done with this one for now before it grew too big etc.

Now it should `ls` dandisets and their assets from the archive (drafts or published versions).  Minimal harmonization is done for girder records to look more like DANDI API returned records.  

Some interesting invocation could be 
```
cmd="DANDI_CACHE=ignore DANDI_DEVEL=1 dandi ls -f yaml -r https://dandiarchive.org/dandiset/000027" ; diff -Naur <(eval $cmd/draft) <(eval $cmd/0.200721.2222)
```
to show differences in records we have for draft (from girder) vs published (from API) version.  With `-f yaml` there is an issue that keys are not sorted (TODO: https://github.com/dandi/dandi-cli/issues/235) .

Notes:

- "auto" mode would now switch to "yaml" if there is at least one URL specified in cmdline
- "pyout" mode would not work nicely ATM since we would need to "flatten" nested records as we are getting them from the archive.  TODO but later. I think the primary use case ATM is to be able to quickly `ls` local or remote dandisets/files in yaml or json forms.


dandi/cli/cmd_ls.py
```diff
@@ -99,39 +119,47 @@ def ls(paths, fields=None, format="auto", recursive=False):
...
+                # TODO: harmonization for pyout
+                rec = asset
+            else:
+                raise TypeError(asset)
+
             if not rec:
-                errors["Empty record"].append(path)
-                lgr.debug("Skipping a record for %s since emtpy", path)
+                errors["Empty record"].append(asset)
+                lgr.debug("Skipping a record for %s since emtpy", asset)
```
jwodder, line 162: "Empty" is misspelled.

dandi/cli/cmd_ls.py
```diff
@@ -251,3 +279,11 @@ def fn():
...
 
     return fn
+
+
+def is_url(s):
+    """Very primitive url detection for now
+
+    TODO: redo
+    """
+    return s.startswith("http") and "://" in s
```
jwodder, line 289: Suggestion:

```
return s.lower().startswith(("http://", "https://"))
```

jwodder (approved):

dandi/cli/cmd_ls.py
yarikoptic: Thank you! Away from laptop ATM, would you be so kind to push fix and merge if all tests are still good. I tuned up your other suggestion to become a proper suggestion for GitHub (see how) and merged - thanks!

Issue #238, opened 2020-09-13
lock Error when calling dandi upload


kaeldai:
I'm trying to upload data from the Allen Institute but when I call `dandi upload` I get the following error:

`2020-09-12 10:08:49,180 [    INFO] Found 101 files to consider`
`Error: Failed to lock dandiset 000039`

The repo has already been created: https://gui.dandiarchive.org/#/dandiset/000039/draft. `dandi validate` was successful and my 000036/ directory contains 33 sub-*/ folders with valid sym links.

Thanks,
Kael

yarikoptic:
Was there some interrupted upload?
 We definitely should provide more information though on the nature of the existing lock. I will look into fixing it tomorrow, thanks for the report.

kaeldai:
I tried a few times throughout the day and still got the same error message. I was doing this through a VPN so the internet was working.

yarikoptic:
A little update - instead of just pruning the lock so you could progress forward, @mgrauer will check first on girder server side on what has lead to that stale lock to come to exist, and then we will remove it so you could proceed with the upload.

mgrauer:
@yarikoptic We looked in Girder and don't see any lock on that dandiset.

If you go to the [swagger page API endpoint](https://girder.dandiarchive.org/api/v1#!/dandi/dandi_wrapper_0_1_2_3_4_5_6) and run the unlock endpoint, it will say that the dandiset isn't currently locked.

yarikoptic:
Thank you Mike! Might then be some other issue (permissions etc) and client being stupid to report it properly, I will later provide further guidance to get closure.

yarikoptic:
Dear @kaeldai , please install the version from #239 via e.g. `pip3 install --upgrade git+https://github.com/dandi/dandi-cli@enh-lock2` and rerun the upload. Error message should provide more information then which would help to identify actual problem.

kaeldai:
Hi @yarikoptic - I'm now getting a slightly different error message:

`2020-09-14 18:21:30,475 [    INFO] Found 101 files to consider`
`Error: Failed to lock dandiset 000039 due to: You must be logged in.`

My dandi account is linked to my github account. I checked and git's user.name and user.email is correct on the machine I'm trying to run a dandi upload from.

yarikoptic:
I have looked through the code and not sure yet how this could happen since we do authenticate into girder first.  Please 
[...]

PR #239, opened 2020-09-15
ENH: provide original exception message while preparing LockingError exception

yarikoptic:
Ref #238

### DIFF
--- a/dandi/girder.py
+++ b/dandi/girder.py
@@ -139,6 +139,7 @@ class GirderCli(gcl.GirderClient):
         api_key = os.environ.get("DANDI_API_KEY", None)
         if api_key:
             self.authenticate(apiKey=api_key)
+            lgr.debug("Successfully authenticated using the key from the envvar")
             return
 
         if self._server_url in known_instances_rev:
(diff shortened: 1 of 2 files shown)


dandi/girder.py
```diff
@@ -150,8 +158,11 @@ def dandi_authenticate(self):
...
                 )
                 keyring.set_password(app_id, "key", api_key)
+                lgr.debug("Stored key in keyring")
+
             try:
                 self.authenticate(apiKey=api_key)
+                lgr.debug("Successfully authenticated using the key")
                 break
             except Exception as exc:
                 sys.stderr.write("Failed to authenticate: {}".format(exc))
```
yarikoptic, line 168: hm, I have not realized that we are not testing this since we always just rely on the env var... oh well -- girder will be gone soon, so we have better things to worry about and let's pretend that this code could not be the reason for #238 ;-)

yarikoptic:
Since master ATM in a "funny" state, I am for now purposefully no merging this PR which I rebased straight on the latest release.  We might even want to cut release from this point.

yarikoptic:
#238 was resolved (wrong scoped key) and IMHO there is nothing else we could/should do about it ATM, and I think it would be ok to merge it into master now

PR #241, opened 2020-09-15
Failing download + upload test

### LINKED ISSUE #240: "integration" test for `upload` to use main archive dandiset metadata record
I am afraid `dandi upload` is broken ATM since dandiset metadata structure has changed (now metadata is under "dandiset" key), and there were no coordination with having client account for that + release.

But I guess this `dandi` cli `organize` and `upload` for dandiset still uses old schema.  So our tests pass (since testing in isolation) -- but real upload fails since on the main deployment metadata records are different... heh

I think we need a test for that: download 000027 from the main archive, and upload to the local fixture instance.

Meanwhile I will look into just fixing up the code so `upload` works again.

jwodder:
I left the `register()` call commented out in case it's needed later.

Closes #240.

### DIFF
--- a/dandi/tests/test_upload.py
+++ b/dandi/tests/test_upload.py
@@ -4,6 +4,7 @@ import pytest
 
 from .. import girder
 from ..consts import collection_drafts, dandiset_metadata_file
+from ..download import download
 from ..register import register
 from ..upload import upload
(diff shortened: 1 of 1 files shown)


yarikoptic:
it is included in #242 now. Thank you!

PR #242, opened 2020-09-15
BF: dandiset metadata record structure/assumptions

### LINKED ISSUE #240: "integration" test for `upload` to use main archive dandiset metadata record

yarikoptic:
In #233 (post 0.6.4 release, current master) I have tried to unify metadata records, but did that incorrectly for dandisets. I forgot that those would also be used during `upload` which caused fiasco in uploading sample 000029 (web ui is also not tolerant to bad metadata)

As part of this PR I also
- adjuster `register` to always return record given by the server, and adjusted (and refactored a bit) corresponding test of register
- included test by @jwodder from #241 (Closes #240) and adjusted it to use `register`
- `dandiset.yaml` would no longer be uploaded at all to the archive.  we already add a comment that all changes are to be lost, so we should actually do that (i.e. not upload so all changes lost)

### DIFF
--- a/dandi/register.py
+++ b/dandi/register.py
@@ -43,6 +43,12 @@ def register(name, description, dandiset_path=None, dandi_instance="dandi"):
                     dandiset_metadata_file,
                     dandiset.path,
                 )
+        else:
+            lgr.info(
+                "No dandiset path was provided and no dandiset detected in the path."
+                " No %s will be modified",
+                dandiset_metadata_file,
+            )
 
     client = girder.get_client(dandi_instance.girder)
     dandiset = client.register_dandiset(name, description)
--- a/dandi/register.py
+++ b/dandi/register.py
@@ -11,6 +11,20 @@ lgr = get_logger()
 
 
 def register(name, description, dandiset_path=None, dandi_instance="dandi"):
+    """Register a dandiset
+
+    Parameters
+    ----------
+    name: str
+    description: str
+    dandiset_path: str, optional
+    dandi_instance: str, optional
+
+    Returns
+    -------
+    dict
+      Metadata record of the registered dandiset
+    """
     dandi_instance = get_instance(dandi_instance)
     if not dandiset_path and op.exists(dandiset_metadata_file):
         dandiset = Dandiset.find(os.getcwd())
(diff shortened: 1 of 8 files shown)


dandi/register.py
```diff
@@ -35,15 +49,13 @@ def register(name, description, dandiset_path=None, dandi_instance="dandi"):
...
+    lgr.info(
+        f"Registered dandiset {dandiset['identifier']} at {url}. Please visit and adjust metadata."
+    )
+
     if dandiset_path:
+        lgr.info(f"Adjusting {dandiset_path} with obtained metadata")
         ds = Dandiset(dandiset_path, allow_empty=True)
         ds.update_metadata(dandiset)
-        return None
-    else:
```
jwodder, line 43: I feel like the function should at least warn if no dandiset path is provided or detected.

jwodder (approved):

dandi/register.py
yarikoptic: Fair enough.  I have returned `.info` message in 7aa6877653c083ab53f9ecb865ba868734d14414 but in a location in the code above.  Since here we could get within an existing dandiset but which already has an `identifier` which we refuse above to modify (and thus do not set `dandiset_path`), so it would not be a matter of "not provided or detected", so this location would be suboptimal.

jwodder:
The help text for the `register` command's `--dandiset-path` option says "If not specified, content for the file will be printed to the screen," but this PR seems to eliminate that behavior.  Is that intentional?

yarikoptic:
oh, not intentional.  Thanks for catching that!  could you please adjust code to just print registered dandiset id and adjust help string accordingly?

jwodder:
Do you mean print the ID always or only when `--dandiset-path` is not specified?

yarikoptic:
I think it is ok to print it always (unless fails to register)

jwodder:
@yarikoptic Commit added.

yarikoptic:
Thanks, let's proceed

PR #251, opened 2020-10-07
Script for creating Datalad datasets from Dandiarchive backup

jwodder:


tools/backups2datalad.py
```diff
@@ -0,0 +1,301 @@
...
+
+"""
+IMPORTANT NOTE ABOUT GIT CREDENTIALS
+
+This script uses `datalad create-sibling-github` to create GitHub repositories
+that are then pushed to GitHub.  The first step requires either GitHub user
+credentials stored in the system's credentials store or else a GitHub OAuth
+token stored in the global Git config under `hub.oauthtoken`.  In addition,
+pushing to the GitHub remotes happens over SSH, so an SSH key that has been
+registered with a GitHub account is needed for the second step.
```
yarikoptic, line 11: this https://github.com/datalad/datalad/wiki/FAQ#how-to-configure-insecure-local-credentials-store could be of help in cases of remote/server setups

yarikoptic (commented):
just a single comment.  upon a quick review -- code looks really nice.  yet to try myself

tools/backups2datalad.py
```diff
@@ -0,0 +1,301 @@
...
+                        except Exception:
+                            if self.ignore_errors:
+                                log.warning("cp command failed; ignoring")
+                                continue
+                            else:
+                                raise
+                        log.info("Adding asset to dataset")
+                        ds.repo.add([deststr])
+                        log.info("Adding URL %s to asset", bucket_url)
+                        ds.repo.add_url_to_file(deststr, bucket_url)
```
yarikoptic, line 175: here -- use `batch=True` mode for `add_url_to_file`?

tools/backups2datalad.py
```diff
@@ -0,0 +1,301 @@
...
+        with requests.Session() as self.session:
+            for did in self.get_dandiset_ids():
+                log.info("Syncing Dandiset %s", did)
+                ds = Dataset(str(self.target_path / did))
+                if not ds.is_installed():
+                    log.info("Creating Datalad dataset")
+                    ds.create(cfg_proc="text2git")
+                    ds.config.set("annex.backends", "SHA256E", where="local")
+                gitattrs = ds.pathobj / ".gitattributes"
+                gitattrs.write_text(gitattrs.read_text().replace("MD5E", "SHA256E"))
```
yarikoptic, line 80: A better way was "formalized" later on in https://github.com/datalad/datalad/issues/4978 avoiding this need for explicit replace:
```
import datalad
import datalad.api as dl
datalad.cfg.set('datalad.repo.backend', 'SHA256E', where='override')
ds = dl.Dataset('/tmp/dssss').create()
```

tools/backups2datalad.py
```diff
@@ -0,0 +1,301 @@
...
+        self.target_path = target_path
+        self.ignore_errors = ignore_errors
+        self.gh_org = gh_org
+        self.session = None
+        self._s3client = None
+
+    def run(self):
+        self.target_path.mkdir(parents=True, exist_ok=True)
+        with requests.Session() as self.session:
+            for did in self.get_dandiset_ids():
```
yarikoptic, line 72: Please RF so that it could be ran for a specific dandiset id.  The this loop would just reuse that function.  Log for the dandiset sync could be dumped somewhere under `.dandi/logs` with datetime based filename and prefix e.g. `sync-20201008<TIME>-PID.log` and committed to git directly (later we might reconsider location and/or "commit to git")

tools/backups2datalad.py
jwodder: Is just adding `batch=True` sufficient, or do I need to add something else at the end of the batch?

tools/backups2datalad.py
jwodder: Does this work in the most recent release of datalad?

yarikoptic: it should IIRC, try

tools/backups2datalad.py
yarikoptic: yeap.  Upon `.save` (or `.repo.commit`) all batched annexes will be stopped/finished before proceeding (there is a lingering issue which was worked-around in a rush but IIRc only in master, more info https://github.com/datalad/datalad/issues/4998)

tools/backups2datalad.py
jwodder: Done.

tools/backups2datalad.py
jwodder: Updated.

tools/backups2datalad.py
jwodder: Done.

yarikoptic:
pushed minor ENHs... please add
- crash if assetstore path provided doesn't have asset store (no `girder-assetstore`)
- option with a regex for files to consider and skip the rest -- so we could get to problematic one right away

jwodder:
@yarikoptic Done.

yarikoptic:
ok, pushed the fix for that particular issue (a297195eef37a3a31d19183b35123fa4be06170a)

next crash is due to the fact that file is under git, so there is no key to get checksum from.  Just use digester and compute checksum or rely mtime (since not under git) for those (files under git, not annex)

jwodder:
@yarikoptic Done.

yarikoptic:
Great, thanks!  Just one more thing:  please store the list of assets as e.g. `.dandi/assets.json` and `datalad save` it (should go to git).
  - while saving let's cater to git a bit (humans can pipe through jq) -- use `datalad.support.json_py.dump` without enabling compression.

oh, and how long a full rerun takes (when no updates)?

jwodder:
@yarikoptic Should `assets.json` just be a list of the pathnames of the assets, or should there be something more in there?

yarikoptic:
> @yarikoptic Should `assets.json` just be a list of the pathnames of the assets, or should there be something more in there?

Just a full list of full records you receive from server. We later can use them to cheaply extract metadata for datalad (ATM they aren't harmonized yet)

jwodder:
With all assets already in place, the script takes 59 minutes and 30 seconds to run.

jwodder:
> > @yarikoptic Should `assets.json` just be a list of the pathnames of the assets, or should there be something more in there?
> 
> Just a full list of full records you receive from server. We later can use them to cheaply extract metadata for datalad (ATM they aren't harmonized yet)

`datalad.support.json_py.dump` can't be used for that, as the asset metadata contains `datetime` objects, which it doesn't know how to serialize.

yarikoptic:
oh, then serialize any such object before dumping -- the dump should be just a regular json any software/human could "read"

PR #255, opened 2020-10-21
ENH: establish backup (via rclone special remote) and make it suitable for a cron job

yarikoptic:
TODOs:
- [ ] make `backup_remote` into an option, which would be None by default and otherwise - name of the rclone remote to push to - so could be tested without any backup linked, and also against some temporary test backup placement
- [ ] make it so that if there were no actual changes to the content (could use `ds.status` or `ds.diff` to interrogate), we do not just save logs.  Always saving logs causes a commit and push to github.  I would like to run this script on cron, and quite regularly (e.g. hourly).
- [ ] make jobs (passed to `push` ATM) into a cmdline option


jwodder:
@yarikoptic Regarding the second point: If an exception occurs while processing a given repository, should a logfile be created?  It won't be committed at that time, but it will be committed the next time the script is successfully run.

yarikoptic:
That is a very good idea -- we need logs for troubleshooting.
What I was actually thinking for how to implement -- we could always store the logs but the just move them away somewhere (e.g. `.git/dandi/logs`) if there is no relevant changes to save.  It should not be done if run fails.  Anyways most likely before next rerun we would desire to `git clean -dfx; git reset --hard` to get to clean state.  Actually please also check for that into the code -- `if ds.repo.dirty: raise RuntimeError("dirty repo, clean or save before running")` -- before we proceed with fetching etc.)

jwodder:
@yarikoptic Done.

yarikoptic:
oh, I have managed to start this branch on older time point, please resolve conflicts

jwodder:
@yarikoptic Merge conflicts resolved.

tools/backups2datalad.py
```diff
@@ -100,25 +117,48 @@ def run(self, dandisets=()):
...
+                            self.backup_remote,
+                            [],
+                            type="rclone",
+                            external=True,
+                            config={
+                                "chunk": "1GB",
+                                "target": self.backup_remote,  # I made them matching
+                                "prefix": "dandi-dandisets/annexstore",
+                                "embedcreds": "no",
+                                "uuid": "727f466f-60c3-4778-90b2-b2332856c2f8"
```
yarikoptic, line 131: Just a mental note to myself that I forgot about this uuid. Wouldn't be appropriate for some other of beyond testing... Will need to become an option

tools/backups2datalad.py
```diff
@@ -100,25 +117,48 @@ def run(self, dandisets=()):
...
+                    log.info("Creating GitHub sibling for %s", ds.pathobj.name)
+                    ds.create_sibling_github(
+                        reponame=ds.pathobj.name,
+                        existing="skip",
+                        name="github",
+                        access_protocol="ssh",
+                        github_organization=self.gh_org,
+                        publish_depends=self.backup_remote,
+                    )
+                    ds.config.set("branch.master.remote", "github", where="local")
```
yarikoptic, line 154: Note: This one isn't needed on each run but mustn't hurt

yarikoptic:
Thank you! I will test in real life ;-)

jwodder:
@yarikoptic Note that the script requires [this datalad PR](https://github.com/datalad/datalad/pull/5065), which I was expecting to get merged first.

Issue #252, opened 2020-10-08
Automate releases/changelog


yarikoptic:
even if I've heard about it, didn't "register", but it seems to be quite powerful and flexible: https://github.com/intuit/auto
(was mentioned in https://github.com/physiopy/phys2bids/issues/311 ;)

Unfortunately I have not found mentioning of a plugin for automagic publishing to pypi but it is trivial to do so could be just baked into the corresponding workflow I guess. I did ~~not dare to  suggest yet to datalad folks since some agreement was reached at some point to "not bother"~~ (apparently I did click "Save" instead of "Cancel" ;) https://github.com/datalad/datalad/issues/4999), so would be nice to start here, gain some experience and then "transfer" to other projects if seem beneficial.

We should look into automating
- changelog composition
- release to pypi:
   - conda-forge will pick up by itself and send a PR, alternatively/ideally in the long run -- we could of cause mint a PR, wait for it to finish and report back if all green and when we merge here, merge also that PR against conda-forge -- full automation!!!
- later: 
  - release for debian as: build source package (so I could download, re-review, sign locally and upload to debian proper) + backports for neurodebian (probably can just automate)
  - anything else?

jwodder:
@yarikoptic I've continued working on the proof-of-concept repo I mentioned [here](https://github.com/datalad/datalad/issues/4999#issuecomment-707278319), and it now automatically publishes to PyPI whenever a new release is created.  (You're a repo collaborator; feel free to try it out.)  Should I create a PR to apply the config to dandi-cli now?

yarikoptic:
GREAT, Thank you @jwodder !  Give us (may be @satra would find some time to check it out as well) some time (hours/few days) to try/see that proof-of-concept first before applying it here.

satra:
go for it.

yarikoptic:
done

PR #257, opened 2020-11-02
Set up workflow with auto for releasing & PyPI uploads

jwodder:
Addresses part of #252.

**Note:** I would recommend not merging this PR until after the next release of dandi-cli so that we can start fresh with labelling all of our PRs.

This PR also needs to be accompanied by the following changes:
- [x] Create the necessary labels for `auto` by running `GH_TOKEN=... auto create-labels` in a copy of this repository containing `.autorc`
- [x] A GitHub release must be created for the most recent version of dandi-cli
- [x] An auth token for uploading assets for the `dandi` project to PyPI must be saved as a secret named "`PYPI_TOKEN`"
- [x] Modify `CHANGELOG.md` so its sectioning matches the format used by `auto`; specifically, eliminate the part above the topmost version section and remove a `#` from all remaining headers

### DIFF
--- a/.autorc
+++ b/.autorc
@@ -1,0 +1,7 @@
+{
+    "onlyPublishWithReleaseLabel": true,
+    "baseBranch": "master",
+    "author": "auto <auto@nil>",
+    "noVersionPrefix": true,
+    "plugins": ["git-tag"]
+}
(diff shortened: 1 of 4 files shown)


yarikoptic:
thanks you @jwodder  for taking care about the release! I guess we should be ready to proceed with this one to see how this whole machinery would work.  I should remember to use labels

yarikoptic:
Do you want me to mint a PYPI token or would you? Then we could test right away with a fix for #264

jwodder:
@yarikoptic You do it.

yarikoptic:
done -- added PYPI_TOKEN scoped to dandi on pypi

PR #267, opened 2020-11-06
Make keyring lookup more flexible

### LINKED ISSUE #263: Establish some local keyring store if there is no default one available
We use keyring (as well as we do in datalad).  That one by default considers the desktop's (be it a Mac, Gnome, whatever) credential store, but falls short if session is pure terminal session.  See https://github.com/dandi/dandi-cli/issues/262#issuecomment-721988907 .

What I think we should do (instead of just `get_password`)
- check with keyring if credential store(s) available.
- if not: ask user either we should establish one local in his/her account.  https://github.com/datalad/datalad/wiki/FAQ#how-to-configure-insecure-local-credentials-store provides howto for a simple insecure one but we could give an option to make it secure (protected by passphrase) -- I believe keyring allows for that.
- if yes and multiple: if we have that one explicitly created, it should be used instead probably or may be we could check if password available in any (unless keyring does it automagically)?

jwodder:
Closes #263.

### DIFF
--- a/dandi/girder.py
+++ b/dandi/girder.py
@@ -885,8 +885,7 @@ def keyring_lookup(service_name, username):
 
     from keyring.core import get_keyring, load_config, load_env
     from keyring.backend import get_all_keyring
-    from keyring.errors import InitError
-    from keyring.util.platform_ import config_root
+    from keyring.errors import KeyringError
     from keyrings.alt.file import EncryptedKeyring
 
     kb = load_env() or load_config()
@@ -895,7 +894,7 @@ def keyring_lookup(service_name, username):
     kb = get_keyring()
     try:
         password = kb.get_password(service_name, username)
-    except InitError as e:
+    except KeyringError as e:
         lgr.info("Default keyring errors on query: %s", e)
         if isinstance(kb, EncryptedKeyring):
             lgr.info(
@@ -915,7 +914,7 @@ def keyring_lookup(service_name, username):
             if askyesno(
                 "Would you like to establish an encrypted keyring?", default=True
             ):
-                keyring_cfg = Path(config_root()) / "keyringrc.cfg"
+                keyring_cfg = Path(keyringrc_file())
                 if keyring_cfg.exists():
                     lgr.info("%s exists; refusing to overwrite", keyring_cfg)
                 else:
@@ -932,3 +931,9 @@ def keyring_lookup(service_name, username):
             raise
     else:
         return (kb, password)
+
+
+def keyringrc_file():
+    from keyring.util.platform_ import config_root
+
+    return op.join(config_root(), "keyringrc.cfg")
--- a/dandi/girder.py
+++ b/dandi/girder.py
@@ -9,6 +9,12 @@ import time
 from functools import lru_cache
 from pathlib import Path
 
+from keyring.core import get_keyring, load_config, load_env
+from keyring.backend import get_all_keyring
+from keyring.errors import KeyringError
+from keyring.util.platform_ import config_root
+from keyrings.alt.file import EncryptedKeyring
+
 import girder_client as gcl
 
 from . import get_logger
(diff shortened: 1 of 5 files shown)


dandi/girder.py
```diff
@@ -845,3 +844,99 @@ def __exit__(self, exc_type, exc_value, tb):
...
+
+    Returns a keyring backend and the password it holds (if any) for the given
+    service and username.
+    """
+
+    from keyring.core import get_keyring, load_config, load_env
+    from keyring.backend import get_all_keyring
+    from keyring.errors import InitError
+    from keyring.util.platform_ import config_root
+    from keyrings.alt.file import EncryptedKeyring
```
yarikoptic, line 889: why not to import them all on top?

setup.cfg
```diff
@@ -43,6 +43,7 @@ install_requires =
     fasteners
     h5py ~= 2.9
     joblib
+    pycryptodomex
```
yarikoptic, line 46: why this one is needed? I do not see it (or `Crypto`) imported -- could comment be added right here?

dandi/girder.py
yarikoptic: or may be keep them here but we do need some unit (via mockups) and/or integration (do specify `PYTHON_KEYRING_BACKEND` pointing to some pre-crafted file) test to verify this logic.  It would also be sad if we miss the moment when keyring introduces some regression etc and we do not notice since no coverage here

dandi/girder.py
```diff
@@ -845,3 +844,99 @@ def __exit__(self, exc_type, exc_value, tb):
...
+                        "[backend]\n"
+                        "default-keyring = keyrings.alt.file.EncryptedKeyring\n"
+                    )
+                return (kb, None)
+            raise
+    else:
+        return (kb, password)
+
+
+def askyesno(question):
```
yarikoptic, line 934: please initiate some helper module, e.g. `dandi/support/ui.py` or alike with this function so it could be used elsewhere when/if needed

yarikoptic: add support for default?  IMHO in this case `yes` is a reasonable default.  Should be visualized in the prompt as well

dandi/girder.py
```diff
@@ -157,7 +156,7 @@ def dandi_authenticate(self):
                     "Account/API keys "
                     "in Girder) for {}: ".format(client_name)
                 )
-                keyring.set_password(app_id, "key", api_key)
+                keyring_backend.set_password(app_id, "key", api_key)
```
yarikoptic, line 159: could you please mockup a unittest for this logic? I know that it is old code, but I forgot that we did not test it (due to env var making it return above I guess).

yarikoptic (changes requested):
Primary concern: need of test(s).

setup.cfg
jwodder: It's for the EncryptedKeyring backend.

dandi/girder.py
jwodder: Done.

setup.cfg
jwodder: Comment added.

dandi/girder.py
jwodder: Done.

jwodder:
@yarikoptic Tests added.

setup.cfg
```diff
@@ -73,6 +74,7 @@ style =
     pre-commit
 test =
     coverage
+    pyfakefs ~= 4.0
```
yarikoptic, line 77: note to myself -- provides `fs` fixture.

yarikoptic:
Seems to work nicely.  One gotcha to keep in mind (add to our handbook) is that it would require to set the password for the encrypted keyring:
```
2020-11-09 11:26:40,026 [    INFO] Configuring /home/yoh/.local/share/python_keyring/keyringrc.cfg to use EncryptedKeyring as default backend
Please provide API Key (created/found in My Account/API keys in Girder) for dandi: fu1oj0NC2QCbSkhCe6V2cTDalwpJOfaPiiVGGs2P
Please set a password for your new keyring: 
Please confirm the password: 
Error: blank passwords aren't allowed.
Please set a password for your new keyring: 
Please confirm the password: 
```
initiated https://github.com/dandi/handbook/issues/6

PR #269, opened 2020-11-09
Copy files with `cp --reflink=auto` where supported

### LINKED ISSUE #268: Make copy_file (used at least by `organize`) use `cp --reflink=auto` where available
We have a TODO comment in https://github.com/dandi/dandi-cli/blob/419e8de4db1e7f36c0e6675bbe70707b9769a42d/dandi/utils.py#L358 and do use `cp --reflink=always` in https://github.com/dandi/dandisets/blob/master/tools/backups2datalad.py#L394 .

I think in the client, we should 
- sense if `cp` has `--reflink=auto` support and if so
- use `cp --reflink` instead of `shutil.copy*` in such cases.  Unless there is some pythonic way to perform CoW

jwodder:
Closes #268.

### DIFF
--- a/dandi/utils.py
+++ b/dandi/utils.py
@@ -7,6 +7,7 @@ import os.path as op
 import platform
 import re
 import shutil
+import subprocess
 import sys
 
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


yarikoptic:
eh, looks good, thought that I would restart only failing ones but restarted all.  I would have liked some debug level messages, but I guess we could add those as need arises, let's proceed

PR #271, opened 2020-11-10
Include item path in "Multiple files found for item" message

### LINKED ISSUE #270: make "Multiple files for one item found; using oldest one" more informative
i.e. for which item (path) there are multiple files? makes it difficult to "debug"

jwodder:
Closes #270.


yarikoptic:
thanks!

PR #272, opened 2020-11-10
Add healthchecks for the Postgres and minio Docker containers

jwodder:


dandi/tests/data/dandiarchive-docker/docker-compose.yml
```diff
@@ -4,7 +4,7 @@
 # <https://github.com/dandi/dandi-publish/blob/master/docker-compose.override.yml>,
 # but using images uploaded to Docker Hub instead of building them locally.
 
-version: '3'
+version: '2.1'
```
yarikoptic, line 7: do you remember why 2.1 and not 3?  would be nice to know FTR (could have even been a comment or at least in a commit message)

yarikoptic:
Left a comment/question for my own education sake... but otherwise -- it is green and hopefully would stay more green than before! Thank you @jwodder !

dandi/tests/data/dandiarchive-docker/docker-compose.yml
jwodder: Because version 3 of the Compose file format doesn't support healthcheck-based dependencies.  Apparently, Docker wants people to write applications that are robust against dependency failure to begin with instead of relying on Docker's healthcheck feature.

yarikoptic: Great, thank you!

PR #275, opened 2020-11-19
Support h5py 3.0

### LINKED ISSUE #259: Update for h5py v3.0
Version 3.0 of h5py (a dependency of pynwb) was released on October 30, and it brought with it [changes to its string handling](https://docs.h5py.org/en/stable/strings.html) that make several tests (e.g., `dandi/tests/test_pynwb_utils.py::test_get_metadata`) fail.  For now, I have added `h5py ~= 2.9` to dandi-cli's dependencies so that the new version won't be installed alongside it, but at some point we need to update for the new library version.

### LINKED ISSUE #274: Remove make compatible with h5py 3.x
I already forgot the details why it was added and what needs to be fixed where. To facilitate ros3 driver use, were need to become compatible, and in general avoid introducing upper version limits

jwodder:
Closes #259 and closes #274.

### DIFF
--- a/dandi/pynwb_utils.py
+++ b/dandi/pynwb_utils.py
@@ -180,6 +180,10 @@ def _get_pynwb_metadata(path):
                 # serialize into a basic container (list), since otherwise
                 # it would be a closed Dataset upon return
                 value = list(value)
+            if isinstance(value, (list, tuple)) and all(
+                isinstance(v, bytes) for v in value
+            ):
+                value = type(value)(v.decode("utf-8") for v in value)
             out[key] = value
 
         # .subject can be None as the test shows
--- a/setup.cfg
+++ b/setup.cfg
@@ -41,7 +41,6 @@ install_requires =
     email-validator
     etelemetry >= 0.2.0
     fasteners
-    h5py ~= 2.9
     joblib
     pycryptodomex  # for EncryptedKeyring backend in keyrings.alt
     pydantic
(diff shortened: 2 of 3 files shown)


yarikoptic:
tests aren't happy (didn't look in detail)
and I guess as soon as passes, we should release, so adding the label

jwodder:
@yarikoptic The tests are failing on Python 3.6 and 3.7 (but not 3.8) because loading h5py produces "RuntimeWarning: numpy.ufunc size changed, may indicate binary incompatibility. Expected 192 from C header, got 216 from PyObject".  I am currently unable to reproduce this in an environment with only h5py installed, though I am considering filing a bug report with h5py.

jwodder:
Bug report filed: https://github.com/h5py/h5py/issues/1765

yarikoptic:
great, thanks!  could we allow for that warning temporarily and see if any actual tests  fail?

jwodder:
@yarikoptic After configuring the warning to be ignored, all tests pass.  Should we keep it like that and merge now?

yarikoptic:
I would say yes since no breakages were detected and thus nothing more to be done on our side.  But is it possible to add a comment to `tox.ini` above that line with a URL to your bug report against h5py so it is easy later to figure out why we have it?

jwodder:
@yarikoptic Yes; comment added.  Feel free to hit the merge button to release when you're ready.

Issue #153, opened 2020-08-03
download: while running within dandiset should take that dandiset as default to download from


yarikoptic:
So that instead of 
```
lena:/tmp/000030
$> dandi download
Error: No URLs were provided.  Cannot download anything
```
the result operation should be analogous to
```
cd ..
$> dandi download http://localhost:8092/#/dandiset/5f282afffb242e375f756695
2020-08-03 11:41:09,963 [    INFO] Traversing remote dandisets (000030) recursively
PATH                                         SIZE      DONE    DONE% CHECKSUM STATUS    MESSAGE             
dandiset.yaml                                                                 skipped   already exists      
sub-anm369962/sub-anm369962_ses-20170309.nwb                                  skipped   same time and size  
Summary:                                     0 Bytes   0 Bytes                2 skipped 1 already exists    
                                             +796.9 kB 0.00%                            1 same time and size
```

the problem is that `download` deduces which instance (global or some local "docker") to talk to from the URL provided.  So I guess the only way to quickly achieve desired ATM would be to add option already existing in `upload`:

```shell
$> DANDI_DEVEL=1 dandi upload --help                                                     
Usage: dandi upload [OPTIONS] [PATHS]...
...
  -i, --dandi-instance [dandi|local-docker|local-girder-only]
                                  For development: DANDI instance to use
```
so it could be then `DANDI_DEVEL=1 dandi download -i local-docker [URLs]` , and if instance specification conflicts with the one from URLs - raise an exception

Issue #140, opened 2020-07-30
download: RF for most logical handling of `--output-dir` and (specified) dandiset


yarikoptic:
ATM  we download into `--output-dir/DANDISET_ID/` if URL points to the dandiset or just `--output-dir/PATH` if PATH points to the folder or file.  Kinda simple/sensible -- `output-dir` points to a directory to download specified entity under.
But as we would introduce download of versioned dandisets, such behavior might be suboptimal. What if I want to download a specific version of the dandiset under a specifically named directory (e.g. `000001-v1`) and draft under plain `000001`.  `download` of cause could just (always?) append version of the dandiset into the dandiset top level directory name, or it could become an option and then we just keep current behavior.
Minor, but something to decided on

PR #283, opened 2020-11-24
Update for new API at https://api.dandiarchive.org/api

### LINKED ISSUE #277: RF DandiAPIClient et al to use new WiP API
Haven't looked yet either prior server is anyhow accessible at all, if not -- we can just migrate `master` to use what is coming to be provided https://github.com/dandi/dandi-api/ .
- [ ] RF DandiAPIClient
- [ ] RF tests fixture from installing old server to install this new one
- [ ] Implement `upload` using API (not girder)
     - If we are to do it straight in master -- RF `upload` so depending on the target server, it would either upload to girder or to API 
- [ ] ...

Eventually (after we test and migrate over to the new "API" server from girder)
- [ ] RM fixture/functionality interfacing girder altogether

jwodder:
Closes #277.

### DIFF
--- a/dandi/consts.py
+++ b/dandi/consts.py
@@ -79,7 +79,7 @@ dandi_instance = namedtuple("dandi_instance", ("girder", "gui", "redirector", "a
 
 # So it could be easily mapped to external IP (e.g. from within VM)
 # to test against instance running outside of current environment
-instancehost = os.environ.get('DANDI_INSTANCEHOST', 'localhost')
+instancehost = os.environ.get("DANDI_INSTANCEHOST", "localhost")
 
 known_instances = {
     "local-girder-only": dandi_instance(
@@ -97,7 +97,7 @@ known_instances = {
         f"http://{instancehost}:8081",
         f"http://{instancehost}:8086",
         f"http://{instancehost}:8079",
-        None,  # TODO: https://github.com/dandi/dandi-cli/issues/164
+        f"http://{instancehost}:8000/api",
     ),
     "dandi": dandi_instance(
         "https://girder.dandiarchive.org",
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -307,12 +307,14 @@ class DandiAPIClient(RESTFullAPIClient):
     def upload(self, dandiset_id, version_id, asset_path, asset_metadata, filepath):
         filehash = Digester(["sha256"])(filepath)["sha256"]
         lgr.debug("Calculated sha256 digest of %s for %s", filehash, filepath)
-        self.post("/uploads/validate/", json={"sha256": filehash})
         try:
-            self.get(f"/uploads/validations/{filehash}/")
-        except requests.HTTPError:
-            lgr.debug("Blob does not already exist on server")
-            blob_exists = False
+            self.post("/uploads/validate/", json={"sha256": filehash})
+        except requests.HTTPError as e:
+            if e.response.status_code == 400:
+                lgr.debug("Blob does not already exist on server")
+                blob_exists = False
+            else:
+                raise
         else:
             lgr.debug("Blob is already uploaded to server")
             blob_exists = True
@@ -355,6 +357,7 @@ class DandiAPIClient(RESTFullAPIClient):
                 },
             )
             self.post(resp["complete_url"])
+            self.post("/uploads/validate/", json={"sha256": filehash})
         while True:
             lgr.debug("Waiting for server-side validation to complete")
             resp = self.get(f"/uploads/validations/{filehash}/")
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -322,7 +322,10 @@ class DandiAPIClient(RESTFullAPIClient):
             lgr.debug("Beginning upload")
             resp = self.post(
                 "/uploads/initialize/",
-                json={"file_name": asset_path, "file_size": os.path.getsize(filepath)},
+                json={
+                    "file_name": f"{dandiset_id}/{version_id}/{asset_path}",
+                    "file_size": os.path.getsize(filepath),
+                },
             )
             object_key = resp["object_key"]
             upload_id = resp["upload_id"]
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -19,6 +19,7 @@ class RESTFullAPIClient(object):
     def __init__(self, api_url):
         self.api_url = api_url
         self._session = None
+        self._headers = {}
 
     @contextmanager
     def session(self, session=None):
@@ -46,6 +47,7 @@ class RESTFullAPIClient(object):
         :param session: An existing :class:`requests.Session` object, or None.
         """
         self._session = session if session else requests.Session()
+        self._session.headers.update(self._headers)
 
         try:
             yield self._session
@@ -124,6 +126,7 @@ class RESTFullAPIClient(object):
         if json_resp and "accept" not in _headers:
             _headers["accept"] = "application/json"
 
+        lgr.debug("%s %s", method.upper(), url)
         result = f(
             url,
             params=parameters,
@@ -133,6 +136,7 @@ class RESTFullAPIClient(object):
             headers=_headers,
             **kwargs,
         )
+        lgr.debug("Response: %d", result.status_code)
 
         # If success, return the json object. Otherwise throw an exception.
         if not result.ok:
@@ -208,6 +212,11 @@ class RESTFullAPIClient(object):
 
 
 class DandiAPIClient(RESTFullAPIClient):
+    def __init__(self, api_url, token=None):
+        super().__init__(api_url)
+        if token is not None:
+            self._headers["Authorization"] = f"token {token}"
+
     def get_asset(self, dandiset_id, version, uuid):
         """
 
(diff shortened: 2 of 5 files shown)


yarikoptic:
Let's add `download` support so we could have a full of upload/download via `dandi-cli`. 
I am still wondering on either we should aim to have this new support in `master` (and have just a new "instance" of dandiarchive to switch to use in DANDI_DEVEL mode) or have a dedicated branch.  What do you think @jwodder -- a branch or straight in master?

jwodder:
@yarikoptic I think it'd be easier to work on if it were merged into master sooner rather than later.

jwodder:
@yarikoptic Download of what, exactly?  Just one asset per method call?  Any specific API?

yarikoptic:
> @yarikoptic Download of what, exactly? Just one asset per method call? Any specific API?

download of an asset, e.g. the one you would upload in the test. Yes, just one asset at a time but support multiple: 

given a dandiset id, and path (and possibly a version, default to `draft`) use
- `​/dandisets​/{version__dandiset__pk}​/versions​/{version__version}​/assets​/paths​/` with `path` parameter - get a list of assets (UUIDs)
- and then loop `/dandisets/{version__dandiset__pk}/versions/{version__version}/assets/{uuid}/download/` 

after we have that interface, it should be "integrated" into our current code/logic for `upload` and `download`.  I will outline that in a dedicated issue

> @yarikoptic I think it'd be easier to work on if it were merged into master sooner rather than later.

Ok, let's aim then for `master`/released version of dandi-cli to be able to support both "old" dandiarchive.org and new api.dandiarchive.org

jwodder:
@dchiquit 

* After uploading an asset with the path `testing/simple1.nwb`, it's listed in `/dandisets/{dandiset_id}/versions/{version}/assets/` but not `/dandisets/{dandiset_id}/versions/{version}/assets/paths/`, regardless of whether the `path` query parameter is set to `/` or `testing/`; the endpoint is just returning an empty list.  I don't think that's right.

* Is it intentional for `/dandisets/{dandiset_id}/versions/{version}/assets/paths/` to return a list as the top-most structure?  I would have expected a dict with "results" and "next" keys, like `/dandisets/{dandiset_id}/versions/{version}/assets/`.

yarikoptic:
may be you could figure out a fix for  https://github.com/dandi/dandi-api @jwodder for the `/paths` issue before @dchiquit can get back to us?  familiarizing with that codebase would be important in the long run.

dchiquit:
That endpoint was patched together to provide data for the Dandiset browser in the web UI, so it wasn't designed very thoughtfully and has some weird inconsistencies.
For one thing, the correct argument to use is `path_prefix`, not path. I assume it's not paginated because whoever implemented the "file browser" widget didn't want to bother with pages. The swagger page presumes pagination for some reason, I will see if I can disable that.
`assets/paths/` assumes that `path_prefix` is a directory, and will automatically append a trailing slash, even if you pass it an empty string. I set up some assets on the draft of dandiset 000001 to demonstrate:
* https://api.dandiarchive.org/api/dandisets/000001/versions/draft/assets/paths/?path_prefix=no-root-slash should show two files. They both have a path that starts with `no-root-slash/`.
* https://api.dandiarchive.org/api/dandisets/000001/versions/draft/assets/paths/?path_prefix=/root-slash should show a file and a subdirectory. They both have a path that starts with `/root-slash`.

We never made a design decision as to whether or not paths should start with a root slash or not. We can adjust this if one implementation is better for the CLI.

What exactly are you using the `assets/paths/` endpoint for? The `assets/` endpoint should work fine for fetching all assets. `assets/paths/` only returns strings because that's all the UI needs to render.

jwodder:
@yarikoptic 

> given a dandiset id, and path (and possibly a version, default to draft)

By "path", do you mean that the download function should take an exact path to a single asset file at a time, or should it also accept paths to folders?

yarikoptic:
> @yarikoptic
> 
> > given a dandiset id, and path (and possibly a version, default to draft)
> 
> By "path", do you mean that the download function should take an exact path to a single asset file at a time, or should it also accept paths to folders?

folders as well.  logic I guess should be added to deduce if it was to an asset or a folder: if a single asset returned by API and basename matches the basename of the path in url -- it was a single asset.  If multiple assets or not matching -- take basename of the url's path and that would be the folder, unless it was an empty path - so it is the top of the dandiset, use dandiset ID as the folder name

jwodder:
@dchiquit What endpoint would you recommend for looking up the IDs and paths of all assets whose paths have a given prefix?  Do I just have to filter `/dandisets/{dandiset_id}/versions/{version}/assets/` myself?

dchiquit:
@jwodder Yes, that's probably safer than using `/assets/paths/`. I will see what I can today to add filtering directly to `/assets/`.

dchiquit:
@jwodder Looks like we did add some rudimentary filtering, that is the `path` parameter you were using earlier. It is also present on `/assets/`, it will filter the results to only include assets whose `path` begins with the parameter. The filter uses a simple string prefix match, so it does not maintain the illusion of a directory structure based on the path: Filtering on `/foo` would return both `/foo/bar.txt` and `foo/fiz/buz.txt`.

https://api.dandiarchive.org/api/dandisets/000001/versions/draft/assets/?path=/ should return two assets, `/root-slash/file.txt` and `/root-slash/foo/bar/test.txt`.

jwodder:
> > @yarikoptic
> > > given a dandiset id, and path (and possibly a version, default to draft)
> > 
> > 
> > By "path", do you mean that the download function should take an exact path to a single asset file at a time, or should it also accept paths to folders?
> 
> folders as well. logic I guess should be added to deduce if it was to an asset or a folder: if a single asset returned by API and basename matches the basename of the path in url -- it was a single asset. If multiple assets or not matching -- take basename of the url's path and that would be the folder, unless it was an empty path - so it is the top of the dandiset, use dandiset ID as the folder name

Forcing the user to use folder names they didn't ask for seems too high-level, like something that should be implemented on top of the basic download methods.  Counterproposal: There are two download methods:

* `download_asset(asset_filename, filepath, ...)` downloads a single asset and saves it at `filepath`.  If `asset_filepath` is not the name of a single asset, it errors.
* `download_asset_directory(asset_dirname, dirpath, ...)` downloads a directory of assets and saves them in `dirpath`.  Each asset is saved to a subpath of `dirpath` calculated by stripping off `asset_dirname`.

satra:
since asset paths are always in the namespace of the dandiset, would the following be useful:

`<root>/dandi-<id>/<version>/<path>`

and the function would construct something like this

```python
basedir = basedir or os.getcwd()
namespace = "dandi"
Path(basedir) / f"{namespace}-{dandiset} / f"{version}" / assetpath
```

`assetpath` should always be a relative path so no leading path separator

and when we transition to personal dandi infrastructure, the namespace could change. we could make `version` optional, but that could result in conflicts on a file system.

jwodder:
@satra Are you suggesting this for the path at which to download an asset?  I believe that the basic download method should be low-level, allowing the user to specify whatever complete target filepath they want.

satra:
> Are you suggesting this for the path at which to download an asset?
yes

my suggestion is primarily about a reasonable default. we could be returning a list of assets as the result of a search. in that scenario this could be a default local layout. for a single asset, indeed a user should be able to override the location to store the asset, but the default location could be something like the one i listed.

yarikoptic:
> Forcing the user to use folder names they didn't ask for seems too high-level, like something that should be implemented on top of the basic download methods. Counterproposal: There are two download methods:
> 
> * `download_asset(asset_filename, filepath, ...)` downloads a single asset and saves it at `filepath`.  If `asset_filepath` is not the name of a single asset, it errors.

I think we would want `download_asset(asset_uuid, filepath, ...)` as the basic download mechanism, and then `download_asset_bypath(asset_path, filepath, ...)` (`asset_path` not just `asset_filename`, but path as in the archive)

> * `download_asset_directory(asset_dirname, dirpath, ...)` downloads a directory of assets and saves them in `dirpath`.  Each asset is saved to a subpath of `dirpath` calculated by stripping off `asset_dirname`.

ok, just make it plural, and again "path" not "name" since IMHO "name" is more of a "basename"? i.e. `download_assets_directory(assets_dirpath, dirpath, ...)`

yarikoptic:
> ```python
> basedir = basedir or os.getcwd()
> namespace = "dandi"
> Path(basedir) / f"{namespace}-{dandiset} / f"{version}" / assetpath
> ```
> 
> `assetpath` should always be a relative path so no leading path separator
> 
> and when we transition to personal dandi infrastructure, the namespace could change. we could make `version` optional, but that could result in conflicts on a file system.

I would prefer to avoid custom prefixes, like `dandi-` since we do not have them in the archive, and the same dandiset could be (eventually) present in multiple archives.  I would just keep it an `id`.  Anyways -- ATM we better just aim for parity with current `download` behavior, while just also supporting new API: what directory etc it would create/layout into should be a separate issue(s), such as #153 and #140 .

satra:
> I would prefer to avoid custom prefixes, like dandi- since we do not have them in the archive, and the same dandiset could be (eventually) present in multiple archives.

the archive does not have a root location for a dandiset only an identifier `<namespace_prefix>:<number>` (e.g., `DANDI:000003`)  so this would actually be reflecting that. the only change was to replace `:` with `-` to make it OS neutral.

yarikoptic:
Filed a dedicated issue #321 for that to not derail discussion here (where goal is "feature parity" ATM)

jwodder:
@yarikoptic This PR should be done now.

yarikoptic:
Looks good, and I did not spot any change which could impact operation of the client while talking with current girder-based instance, so should be good to go.  Thank you @jwodder !

PR #286, opened 2020-11-30
Some schema updates

satra:


yarikoptic:
@satra may be just push directly into nwb2asset which is also used for #226 so we could finally merge that one after one more review cycle (to make sure that no side-effects for any current behavior of master)?

satra:
i wanted to fix some tests here before merging this.

yarikoptic:
ok.  I will then push just the merge of master you have made (f110465bf57b0b5cb328791b3f1d4a7190f61c77) into that PR so this diff starts to look sensible and we have conflict resolved in that PR

satra:
i'm going to merge this. i think this is ok.

PR #287, opened 2020-11-30
updated just models

satra:


yarikoptic:
Marked as "internal" since AFAIK those are not yet affecting the client's functionality, merging

dandi/models.py
```diff
@@ -179,6 +181,15 @@ class SpeciesType(TypeModel):
...
 
+    dxdate: Optional[List[date]] = Field(
+        None,
+        title="Dates of diagnosis",
+        description="Dates of diagnosis",
+        readOnly=True,
+        nskey="dandi",
+        rangeIncludes="schema:Date",
+    )
+
```
yarikoptic, line 192: hm, shouldn't this (and possibly other changes) have affected the produced json schema?  according to the workflow no changes being pushed to `dandi/schema` repository by https://github.com/dandi/dandi-cli/runs/1476406805?check_suite_focus=true

satra: it should have.

yarikoptic: filed #289 289

PR #296, opened 2020-12-02
Replace askyesno() with click.confirm()

jwodder:
It turns out that `askyesno()` is already provided by a library we're already using.  The only advantage of the former is that it supports repeated prompting when no default is given and the user enters a empty line, but we weren't using that feature anyway.


yarikoptic:
and there would be no side-effect of some kind if we use `dandi` as python package, and thus not going through click's cmdline parsing/handling?  just want to make sure

yarikoptic:
Hm, not entirely 100% optimal (outside code, e.g. soon to be introduced https://github.com/datalad/datalad/blob/master/datalad/support/parallel.py) might need to catch that exception too then in some very rare cases when underlying dialog happens to be embedded.  But I think it is tolerable and we could later shim it if need arises. Let's proceed

PR #299, opened 2020-12-02
Add title to `Field` calls where necessary

AlmightyYakob:
This is something I've noticed for a while now, and figured I'd just add these titles. Almost all of the titles are just the variable name with capitalization and spaces, so I doubt there's much wrong here, but if there is please lmk.

### DIFF
--- a/dandi/models.py
+++ b/dandi/models.py
@@ -138,7 +138,8 @@ class PropertyValue(DandiBaseModel):
     propertyID: Union[IdentifierType, AnyUrl, str] = Field(
         None,
         title="Property ID",
-        description="A commonly used identifier for the characteristic represented by the property.",
+        description="A commonly used identifier for"
+        "the characteristic represented by the property.",
         nskey="schema",
     )
 
--- a/dandi/models.py
+++ b/dandi/models.py
@@ -125,17 +125,19 @@ class DandiBaseModel(BaseModel):
 
 
 class PropertyValue(DandiBaseModel):
-    maxValue: float = Field(None, nskey="schema")
-    minValue: float = Field(None, nskey="schema")
-    unitCode: Union[str, AnyUrl] = Field(None, nskey="schema")
-    unitText: str = Field(None, nskey="schema")
+    maxValue: float = Field(None, title="Max Value", nskey="schema")
+    minValue: float = Field(None, title="Min Value", nskey="schema")
+    unitCode: Union[str, AnyUrl] = Field(None, title="Unit Code", nskey="schema")
+    unitText: str = Field(None, title="Unit Text", nskey="schema")
     value: Union[str, bool, int, float, List[Union[str, bool, int, float]]] = Field(
         None, nskey="schema"
     )
     valueReference: "PropertyValue" = Field(
-        None, nskey="schema"
+        None, title="Value Reference", nskey="schema"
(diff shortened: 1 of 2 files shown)


yarikoptic:
although "explicit better than implicit" I generally prefer to avoid duplication.
I wonder why just not at the level of UI to apply the rule of having `camelCased` to become `Camel Cased` and apply rules to downcase `of`, `to` etc if encountered upon splitting?

yarikoptic:
FWIW, might be better to have that "code" to be actually here within schema conversion tools, so produced .json would already have titles.  That would centralize the code and not offload it upon UIs.

satra:
@AlmightyYakob - i sent some changes on top of yours (feel free to merge that). i was mostly focusing on the non-autogenerated ones. the reason license was misbehaving is that pydantic does not generate titles automatically for fields that have a type that's an enum. (i don't know why). so i fixed titles for all of them.

AlmightyYakob:
> FWIW, might be better to have that "code" to be actually here within schema conversion tools, so produced .json would already have titles. That would centralize the code and not offload it upon UIs.

I agree, as otherwise it's more code that splits the schema used in the UI from the source schema. I realize that title formatting is a pretty small change, but JSON Schema is designed to be able to specify this information, without intervention from whatever is displaying it.

yarikoptic:
> > FWIW, might be better to have that "code" to be actually here within schema conversion tools, so produced .json would already have titles. That would centralize the code and not offload it upon UIs.
> 
> I agree, as otherwise it's more code that splits the schema used in the UI from the source schema. I realize that title formatting is a pretty small change, but JSON Schema is designed to be able to specify this information, without intervention from whatever is displaying it.

ok, filed https://github.com/dandi/dandi-cli/issues/301 so could be done after to not hold this PR hostage

jwodder:
I've submitted a PR on the head repository to fix the failing tests: https://github.com/AlmightyYakob/dandi-cli/pull/3

yarikoptic:
well, don't we need PRs to come from branches in this repository to get access to some secrets? then this PR should be redone from a branch here anyways.  @jwodder , please just submit a new PR from a branch here to supersede this PR. @AlmightyYakob is also part of the team , so he should be able to push directly to that branch if would be needed

jwodder:
@yarikoptic Are you sure superseding this PR is necessary?  The only secret the test workflow relies on is for pushing to Codecov.

yarikoptic:
we can wait for all the dance with merging of a PR for a PR to happen and then ignore codecov, or could just streamline with superseded PR with all the commits desired to bring it to the finish line together.  up to you ;)

AlmightyYakob:
> @AlmightyYakob is also part of the team , so he should be able to push directly to that branch if would be needed

Do I have contributor access on this repo? If so, then in the future I'm happy to deal with branches here directly.

yarikoptic:
> > @AlmightyYakob is also part of the team , so he should be able to push directly to that branch if would be needed
> 
> Do I have contributor access on this repo? If so, then in the future I'm happy to deal with branches here directly.

AFAIK yes, you are part of the dandi-cli team https://github.com/orgs/dandi/teams/dandi-cli/members

yarikoptic:
ok, since @satra did not raise any concerns, I guess it is good to go!

PR #303, opened 2020-12-03
Temporary workaround: prevent upload of files larger than 67108864000

yarikoptic:
See https://github.com/dandi/dandiarchive/issues/517 for more information

If looks kosher and doesn't break anything, we should get it into the masses, unless we find a proper solution to that issue.

we might want to announce all prior versions of dandi-cli as "bad" in .et

### DIFF
--- a/dandi/upload.py
+++ b/dandi/upload.py
@@ -166,6 +166,14 @@ def upload(
             try:
                 path_stat = path.stat()
                 yield {"size": path_stat.st_size}
+                size_cut_off = 67108864000
+                if path_stat.st_size > size_cut_off:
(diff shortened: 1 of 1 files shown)


satra:
i think we should fix the issue because we will be getting TB size files from the U01 project.

satra:
we can do a quick fix to limit it, but we should fix it urgently.

yarikoptic:
I know that it should be fixed urgently... but we also should prevent data loss on uploads which might happen meanwhile, and may be fix isn't trivial.  I will wait until the end of the day and then proceed/merge.

PR #304, opened 2020-12-03
Autogenerate schema element titles

### LINKED ISSUE #301: Automatically mint "title" for entities in json schema if not explicitly given
This is a continuation to changes/discussion in https://github.com/dandi/dandi-cli/pull/299#issuecomment-737522905 . That PR adds a good number of explicit `title` entries even when they could be trivially created, e.g. `title="Value Reference"` for `valueReference` at https://github.com/dandi/dandi-cli/pull/299/files#diff-295856c56542fa0dda691f493995baed932f49e94744fbacfbd7412db8c79f17R135

So we should remove hardcoded "duplicate" entries and produce title for all entries where no explicit one is defined, so all UIs could consistently display them

jwodder:
Closes #301.

Note that there doesn't seem to be a way to apply this to enum types.


dandi/models.py
```diff
@@ -123,21 +125,28 @@ def to_dictrepr(__pydantic_cls__: Type[BaseModel]):
             .replace(__pydantic_cls__.__name__, "dict")
         )
 
+    class Config:
+        @staticmethod
+        def schema_extra(schema: Dict[str, Any]) -> None:
+            schema["title"] = name2title(schema["title"])
+            for name, prop in schema.get("properties", {}).items():
+                if prop.get("title") is None or prop["title"] == name.title():
+                    prop["title"] = name2title(name)
```
yarikoptic, line 134: we need a test for this.  Please add a little test which would pretty much do what `dandi/tools/pubschemata.py` does and ensure that we do get `title` set as expected in at least one record

yarikoptic (changes requested):
just asking for a simple unittest.  Otherwise looks great to me, thank you @jwodder !

May be @satra could have a peek as well

dandi/models.py
jwodder: Test added.

satra:
LGTM - can we check that the generated json schema has the titles (especially for the enumerated types like `License`)?

jwodder:
@satra As I said above, there doesn't seem to be a way to autogenerate titles for enum types.

dandi/models.py
```diff
@@ -525,28 +498,20 @@ class CommonModel(DandiBaseModel):
...
-        None,
-        title="Study Target",
-        description="What the study is related to",
-        nskey="dandi",
-    )
-    license: List[LicenseType] = Field(
-        title="License", description="License of item.", nskey="schema"
+        None, description="What the study is related to", nskey="dandi"
     )
+    license: List[LicenseType] = Field(description="License of item.", nskey="schema")
```
satra, line 503: @jwodder - then we should not remove the title field that was added to enum types. There are a few of these. 

```
AccessType = create_enum(AccessTypeDict)
RoleType = create_enum(RoleTypeDict)
RelationType = create_enum(RelationTypeDict)
LicenseType = create_enum(LicenseTypeDict)
IdentifierType = create_enum(IdentifierTypeDict)
DigestType = create_enum(DigestTypeDict)
```

jwodder: Oh, you mean the enum *fields*!  No, those titles are autogenerated fine.  It's the enum entries in the `definitions` section that don't get titles.

yarikoptic (approved):
thank you @jwodder , looks good to me.  @satra -- if `enum` issue is "settled" , feel welcome to click Merge ;)

yarikoptic:
Let's proceed, and if anything -- fixup after! ;)  "release is cheap" these days (and we do not need a release for this anyways) but it would avoid possible conflicts happen we decide to tune the schema

PR #306, opened 2020-12-04
Compare uploaded file size against what download headers report

### LINKED ISSUE #305: add a check that HEAD on the uploaded file reports correct size
in light of https://github.com/dandi/dandiarchive/issues/517 we should verify more that upload succeeded.  ATM we only do 
```python
            # Get uploaded file id
            file_id, current = client.isFileCurrent(
                item_rec["_id"], path.name, path.absolute()
            )
            if not current:
                yield skip_file("File on server was unexpectedly changed")
                return
```
so after that we should verify that it is of a correct size (not according to girder, but from HEAD on the download)

jwodder:
Closes #305.

### DIFF
--- a/dandi/tests/test_upload.py
+++ b/dandi/tests/test_upload.py
@@ -1,6 +1,7 @@
 import os
 import re
 import pytest
+import responses
 
 from .. import girder
 from ..consts import collection_drafts, dandiset_metadata_file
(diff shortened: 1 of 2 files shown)


jwodder:
@yarikoptic I think the 70 GB file you added to Dandiset 000027 is messing with the download tests and making the VMs run out of disk space.

satra:
i think @mgrauer added that. should perhaps be removed now.

yarikoptic:
rright! forgot that to alert @mgrauer -- use https://gui.dandiarchive.org/#/dandiset/000029/draft  for any such test uploads. 00027 is used in testing and should be kept lean.

yarikoptic:
I deleted it now.

mgrauer:
Got it, thanks for info.

jwodder:
Tests are working fine now.

satra:
@jwodder - it may be worthwhile to add a mocktest that returns a different file size after upload.

yarikoptic:
@jwodder , I am echoing @satra -- please add a test but also make it parametric 
- in one run -- mocked return size failure
- in another -- do produce huge file (ATM over 65G) and upload.  It should be executed **only** if `DANDI_DEVEL` envvar is set.  I assume that minio has the same logic of paginating 1,000 of responses, so we should hit it ATM.  More of relevant discussion etc is at https://github.com/girder/girder/issues/3318

jwodder:
@yarikoptic How should that be parametric?  Those seem like two separate tests.  I assume that, for the second one, the returned size should not be mocked and the test should check for failure.

jwodder:
@yarikoptic Exactly what behavior should the second test expect?  Should it test that [this error](https://github.com/dandi/dandi-cli/blob/330accde911c6f2b042967ebb3d1b3914b35a3b7/dandi/upload.py#L169-L176) is raised or something else?

yarikoptic:
could be just two separate tests indeed. 
re which exception -- may be at some point it would become girder client raising some exception.  I would just make sure that there is any exception raised for now

jwodder:
Tests added.

yarikoptic:
Thank you @jwodder , let's proceed.  I will check the huge file test "locally" on some box later on

PR #307, opened 2020-12-05
fix: rat to common lab rat

satra:


dandi/metadata.py
```diff
@@ -189,9 +189,9 @@ def extract_species(metadata):
         elif "human" in value or value.startswith("homo"):
             value_id = "http://purl.obolibrary.org/obo/NCBITaxon_9606"
             value = "Human"
-        elif "rat" in value:
-            value_id = "http://purl.obolibrary.org/obo/NCBITaxon_10117"
-            value = "House rat"
+        elif "norvegicus" in value:
+            value_id = "http://purl.obolibrary.org/obo/NCBITaxon_10116"
+            value = "Brown rat"
```
yarikoptic, line 194: What do we do with other rats? (I see house ready removed, not used?)

satra:
i couldn't find a house rat in the current datasets ;) i think this is where we will need additional UI and update the NWB file properly. if NWB already has URI, we won't touch it.

satra:
i think we should try to improve NWB metadata before upload if we can.

yarikoptic:
Yes, for that we need to add a mode/command to suggest replacements and do them in the files. Unrelated to this pr, so let's proceed

PR #310, opened 2020-12-08
Set chunk size on per-file basis; limit to 1000 chunks; upload files up to 400GB ATM

### LINKED ISSUE #309: enable upload of files larger than current limit of 65G
Underlying issue is multifaceted ATM, it is in 
- the girder S3 backend support: it does not handle pagination of responses from S3 for multi-part uploads (https://github.com/girder/girder/issues/3318), so handles only the first 1,000 of them (out of possible up to 10,000)
- the girder client: hardcodes chunk size to `MAX_CHUNK_SIZE = 1024 * 1024 * 64`

so we are ending up with about 64G (`1000!=1024` ;)) maximal size.  I am not sure how long it would take for both girder server **and** client to get fixed up, or for our API server (re)implementation to come to replace current girder based setup, so I think we should better just workaround to raise the limit of maximal file size users could upload *now* since we already do have such use cases.

S3 allows for upload of chunks up to 5GB (current - 64GB), but we are working through nginx, so there is also a limit on the size of the PUT request (let's call it `NGINX_MAX_CHUNK_SIZE`).  Current limit is set to 100MB but probably could be raised to 400MB (@mgrauer will check if anything forbidding that).

So the plan would be
- have a test in place to verify that we do fail ATM (requested as part of the https://github.com/dandi/dandi-cli/pull/306#issuecomment-740030234)
- resurrect https://github.com/dandi/dandi-cli/pull/139 (or just reimplement more cleanly?) to be able to 
   - manually set the chunk size
   - switch from default 64M (`GirderClient.MAX_CHUNK_SIZE`) size to  `max(64M, filesize/1000)` if we could tune CHUNK_SIZE on per file basis.  If that one exceeds `NGINX_MAX_CHUNK_SIZE` (we should add a constant I guess, although ideally server should provide such information -- added to now transferred https://github.com/dandi/dandi-cli/issues/308).  But if we could not base it on file-size, we could just go for `NGINX_MAX_CHUNK_SIZE` for the value of `GirderClient.MAX_CHUNK_SIZE`.  If with a given chunksize we need more than 1000 chunks -- we raise an exception (and remove our ad-hoc check on filesize ATM).  With 400M chunk we should be able to handle uploads of up to 400G which should cover current use cases.

jwodder:
Closes #309.

### DIFF
--- a/dandi/girder.py
+++ b/dandi/girder.py
@@ -556,7 +556,7 @@ class GirderCli(gcl.GirderClient):
                         f"Failed to unlock dandiset {dandiset_identifier} due to: {msg}"
                     )
 
-    NGINX_MAX_CHUNK_SIZE = 100 * (1 << 20)
+    NGINX_MAX_CHUNK_SIZE = 100 * (1 << 20)  # 100 MiB
 
     def _uploadContents(self, uploadObj, stream, size, progressCallback=None):
         """
@@ -579,7 +579,7 @@ class GirderCli(gcl.GirderClient):
         offset = 0
         uploadId = uploadObj["_id"]
 
-        chunk_size = max(1 << 26, size // 1000)
+        chunk_size = max(self.MAX_CHUNK_SIZE, (size + 999) // 1000)
         if chunk_size > self.NGINX_MAX_CHUNK_SIZE:
             raise Exception("File requires too many chunks to upload")
 
--- a/dandi/girder.py
+++ b/dandi/girder.py
@@ -556,6 +556,71 @@ class GirderCli(gcl.GirderClient):
                         f"Failed to unlock dandiset {dandiset_identifier} due to: {msg}"
                     )
 
+    NGINX_MAX_CHUNK_SIZE = 100 * (1 << 20)
+
+    def _uploadContents(self, uploadObj, stream, size, progressCallback=None):
+        """
+        Uploads contents of a file.  Overridden so that the chunk size can be
+        set on a per-file basis.
+
+        :param uploadObj: The upload object contain the upload id.
+        :type uploadObj: dict
+        :param stream: Readable stream object.
+        :type stream: file-like
+        :param size: The length of the file. This must be exactly equal to the
+            total number of bytes that will be read from ``stream``, otherwise
+            the upload will fail.
+        :type size: str
+        :param progressCallback: If passed, will be called after each chunk
+            with progress information. It passes a single positional argument
+            to the callable which is a dict of information about progress.
+        :type progressCallback: callable
+        """
+        offset = 0
+        uploadId = uploadObj["_id"]
+
+        chunk_size = max(1 << 26, size // 1000)
+        if chunk_size > self.NGINX_MAX_CHUNK_SIZE:
+            raise Exception("File requires too many chunks to upload")
+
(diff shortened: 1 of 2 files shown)


dandi/girder.py
```diff
@@ -556,6 +556,71 @@ def lock_dandiset(self, dandiset_identifier: str):
...
+        :type size: str
+        :param progressCallback: If passed, will be called after each chunk
+            with progress information. It passes a single positional argument
+            to the callable which is a dict of information about progress.
+        :type progressCallback: callable
+        """
+        offset = 0
+        uploadId = uploadObj["_id"]
+
+        chunk_size = max(1 << 26, size // 1000)
```
yarikoptic, line 582: why not to use original super-class's chunk size instead of hardcoded `1 << 26` here?

shouldn't it be `math.ceil(size/1000)` so we do not end up with 1001 chunk?

yarikoptic (changes requested):
also please add in comments sizes (as `#  ??MB`) for the lines where it is hardcoded like `100 * (1 << 20)` to make it easier to grasp/identify.

dandi/girder.py
jwodder: * I misread the original assignment and thought that `1 << 26 != MAX_CHUNK_SIZE`.  Fixed now.

* Done (but without getting floats involved).

jwodder:
@yarikoptic Comment added to `NGINX_MAX_CHUNK_SIZE` line.

dandi/girder.py
```diff
@@ -556,6 +556,71 @@ def lock_dandiset(self, dandiset_identifier: str):
                         f"Failed to unlock dandiset {dandiset_identifier} due to: {msg}"
                     )
 
+    NGINX_MAX_CHUNK_SIZE = 100 * (1 << 20)  # 100 MiB
```
yarikoptic, line 559:
```suggestion
    NGINX_MAX_CHUNK_SIZE = 400 * (1 << 20)  # 100 MiB
```
@mgrauer raised up the limits to 400, so I will suggest and accept right away to proceed

jwodder: You should probably update the comment as well.

dandi/girder.py
```diff
@@ -556,6 +556,71 @@ def lock_dandiset(self, dandiset_identifier: str):
                         f"Failed to unlock dandiset {dandiset_identifier} due to: {msg}"
                     )
 
+    NGINX_MAX_CHUNK_SIZE = 400 * (1 << 20)  # 100 MiB
```
yarikoptic, line 559: Yarik fell into its own trap ! ;)
```suggestion
    NGINX_MAX_CHUNK_SIZE = 400 * (1 << 20)  # 400 MiB
```

yarikoptic: I will merge locally and force push

dandi/girder.py
yarikoptic: yeap yeap -- I did

yarikoptic:
I am yet to try on some real use case... and then we will merge for a release

yarikoptic:
eh, will test "as released" ;)  unittests pass so I am quite certain that it all should be great!  Thank you @jwodder !

Issue #313, opened 2020-12-11
h5py started to error on install with python 3.9


satra:
In #312 i am seeing a h5py installation error.

yarikoptic:
which looks like
https://github.com/dandi/dandi-cli/pull/312/checks?check_run_id=1534598643

```
    creating build\lib.win-amd64-3.9\h5py\tests\test_vds
    copying h5py\tests\test_vds\test_highlevel_vds.py -> build\lib.win-amd64-3.9\h5py\tests\test_vds
    copying h5py\tests\test_vds\test_lowlevel_vds.py -> build\lib.win-amd64-3.9\h5py\tests\test_vds
    copying h5py\tests\test_vds\test_virtual_source.py -> build\lib.win-amd64-3.9\h5py\tests\test_vds
    copying h5py\tests\test_vds\__init__.py -> build\lib.win-amd64-3.9\h5py\tests\test_vds
    running build_ext
    error: Could not find module 'hdf5.dll' (or one of its dependencies). Try using the full path with constructor syntax.
    Loading library to get version: hdf5.dll
ERROR: Command errored out with exit status 1: 'c:\hostedtoolcache\windows\python\3.9.0\x64\python.exe' -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pip-install-9vwfpdn9\\h5py_a7684753b3e14584ba4a2f8a1f29e3d5\\setup.py'"'"'; __file__='"'"'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pip-install-9vwfpdn9\\h5py_a7684753b3e14584ba4a2f8a1f29e3d5\\setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record 'C:\Users\runneradmin\AppData\Local\Temp\pip-record-4jhjrah4\install-record.txt' --single-version-externally-managed --compile --install-headers 'c:\hostedtoolcache\windows\python\3.9.0\x64\Include\h5py' Check the logs for full command output.
Error: Process completed with exit code 1.
```
with such an error -- go figure... but it is not windows but 3.9 specific -- fails on osx and ubuntu as well with 3.9, e.g.
```
    running build_ext
[...]

PR #315, opened 2020-12-14
Install hdf5 for Python 3.9 tests

jwodder:
On December 9, hdmf released a new version that restricted its h5py dependency version range to `<3,>=2.9`; as a result, installing dandi with a recent version of pip will install a pre-3.0 version of h5py.  However, pre-3.0 versions of h5py do not have Python 3.9 wheels available on PyPI, and so the package must be built from source for Python 3.9, which requires an hdf5 library to be installed.  This PR installs such a library for Ubuntu and macOS when testing Python 3.9; testing Python 3.9 on Windows is disabled for now as installing hdf5 on that OS is apparently quite involved.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,6 +25,8 @@ jobs:
           - 3.8
           - 3.9
         exclude:
+          # Temporarily disabled due to h5py/hdf5 dependency issue
+          # See <https://github.com/dandi/dandi-cli/pull/315>
           - os: windows-2019
             python: 3.9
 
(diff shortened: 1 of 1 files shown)


yarikoptic:
Awesome! This would close #313

jwodder:
h5py is installing fine now under Python 3.9; the current test failures are a separate issue, resolved by #314.

yarikoptic:
Thanks! I have merged #314

yarikoptic:
Could you please rebase so we get a clean rerun of workflows with tests fixed?

jwodder:
@yarikoptic Rebased.

PR #326, opened 2020-12-28
Require keyring backends to be initialized before running any tests

jwodder:
On December 22, keyring released version 21.6, which no longer initializes keyring backends on import, instead delaying initialization until any functions are called.  As a result, when dandi-cli’s tests are run, the EncryptedKeyring backend is only initialized in the middle of individual tests, and these tests happen to use pyfakefs.  However, initializing EncryptedKeyring leads to a use of the subprocess module, which is unusable while pyfakefs is in effect, and the tests end up failing.

This PR forces the keyring backends to be initialized when the tests are loaded, before any of them run, thereby fixing the currently-failing tests.

The only thing I don’t understand is why this was only causing a problem on macOS.

### DIFF
--- a/dandi/tests/test_girder.py
+++ b/dandi/tests/test_girder.py
@@ -1,7 +1,7 @@
 import os.path
 
+from keyring.backend import get_all_keyring
 from keyring.backends import fail, null
-from keyring.core import init_backend
 from keyring.errors import KeyringError
 from keyrings.alt import file as keyfile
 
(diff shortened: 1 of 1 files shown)


yarikoptic (changes requested):
I think ideally it be within some session wide pytest fixture instead of test module "global" namespace thus triggered immediately upon import. and also may be condition on version of keyring or ideally (if available) some indicator on either backend was already initialized or not.

yarikoptic:
don't we need `init_backend` call somewhere in client code itself, e.g. `dandi/girder.py`?

jwodder:
@yarikoptic I have moved the call to a module-scoped fixture and changed it to be more idempotent.

> don't we need init_backend call somewhere in client code itself, e.g. dandi/girder.py?

No, the new version of keyring calls `init_backend()` automatically as soon as a backend is requested.  We only need to call it explicitly in the tests because we need the initialization to happen before pyfakefs goes into effect.

yarikoptic:
Great, thank you @jwodder .  Happy New Year! ;)

PR #338, opened 2021-01-22
Add functions for validating metadata against JSON Schema and use in tests

### LINKED ISSUE #332: add testing that a prototypical dandiset metadata validates using converted to json schema
`tools/pubschemata.py` publishes schema as json schema to https://github.com/dandi/schema .  Conversion to json might "mutate" it so a correct according to pydantic's model metadata would not validate using that json schema and `jsonschema` python package (is that what you were using @dchiquit? found no jsonschema import in dandi-api though).

- move `publish_model_schemata` from `tools/pubschemata.py`  into `dandi/metadata.py`
- add a function to validate metadata given provided json schema directory (as produced by `publish_model_schemata`)
- verify that metadata validates using that schema and jsonschema:
   - add tests: in http://github.com/dandi/dandi-cli/blob/HEAD/dandi/tests/test_metadata.py we already have some quite detailed sample cases for dandiset level metadata migration.

jwodder:
Closes #332.

### DIFF
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -249,11 +249,13 @@ def extract_wasDerivedFrom(metadata):
 
 
 def extract_wasAttributedTo(metadata):
-    return [
-        extract_model(
-            models.Participant, metadata, identifier=metadata.get("subject_id")
-        )
-    ]
+    wat = extract_model(
+        models.Participant, metadata, identifier=metadata.get("subject_id")
+    )
+    if all(v is None for v in wat.dict().values()):
+        return []
+    else:
+        return [wat]
 
 
 def extract_digest(metadata):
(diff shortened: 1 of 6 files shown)


yarikoptic (approved):
Looks good to me. If I get it right, I think we might need to get back to `license` to become a list in a conversion code when we try to migrate existing dandiset.yaml s .

@satra -- ok with changes to the model (that TempOptional etc)?

satra:
we are still using unvalidated in the tests. is that still intended? 

just a note that all the tempoptionals are readonly in the sense we expect these to be set on the server side. but perhaps we should insert them in the test data.

yarikoptic:
hm, indeed... but may be one step at a time, i.e. get this merged/used and then "re-introduce" them when server is populating them or does it already?

satra:
i leave it to you.

PR #342, opened 2021-01-26
BF: allow to handle an item with multiple files

yarikoptic:
How that happened, nobody knows.  But since we are making judgement anyways
only on the metadata in the item, we will just allow to remove the entire item
and reupload that file

attn @CodyCBakerPhD


CodyCBakerPhD:
@yarikoptic Yep, this worked! Those files are now uploaded, thanks.

PR #346, opened 2021-01-26
Give AssetMeta and DandiMeta `json_dict()` methods for better dictification

jwodder:
Since it seems we're going to be doing this sort of conversion frequently, this PR adds a method to `CommonModel` for recursively converting to a `dict` while also converting enum values to strings.

### DIFF
--- a/dandi/models.py
+++ b/dandi/models.py
@@ -1,6 +1,7 @@
 from copy import deepcopy
 from datetime import date
 from enum import Enum
+import json
 import sys
 from typing import Any, Dict, List, Optional, Type, Union
 
@@ -659,6 +660,14 @@ class CommonModel(DandiBaseModel):
     relatedResource: Optional[List[Resource]] = Field(None, nskey="dandi")
 
     wasGeneratedBy: Optional[Activity] = Field(None, nskey="prov")
+
+    def asdict(self):
+        """
+        Recursively convert the instance to a `dict` of JSONable values,
+        including converting enum values to strings.  Unset and `None` fields
+        are omitted.
+        """
+        return json.loads(self.json(exclude_unset=True, exclude_none=True))
 
 
 class DandiMeta(CommonModel):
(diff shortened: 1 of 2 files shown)


dandi/models.py
```diff
@@ -563,6 +564,14 @@ class CommonModel(DandiBaseModel):
         None, readOnly=True, nskey="prov"
     )
 
+    def asdict(self):
```
yarikoptic, line 567: I only wonder if it would not bring us confusion as there is `.dict` provided by pydantic as well... may be something like `dict_via_json` or `json_dict`. Although uglier, and not following of `as*` convention - it would be closer to pydantic's `dict`, `json`?

jwodder: Of those suggestions, I prefer `json_dict`, though not as much as I like `asdict`.

yarikoptic: Let's proceed with `json_dict` then.

dandi/tests/test_metadata.py
```diff
@@ -1265,7 +1302,421 @@ def test_dandimeta_migration(schema_dir):
...
+            "single-neurons",
+        ],
+        license=["dandi:CCBY40"],
+        name=(
+            "A NWB-based dataset and processing pipeline of human single-neuron"
+            " activity during a declarative memory task"
+        ),
+        schemaVersion="1.0.0-rc1",
+        repository="https://dandiarchive.org/",
+    )
```
yarikoptic, line 1721: just a note: although I love tests, I think eventually such "sample test" (instead of just concise targetted unittest which would test on some smaller metadata record) might become too much of a burden to maintain

dandi/models.py
jwodder: Renamed.

yarikoptic:
please rebase on/merge master @jwodder - let's get it merged

jwodder:
@yarikoptic Rebased.

PR #347, opened 2021-01-26
Handle uploading already-present files in new API

### LINKED ISSUE #345: dandi-api: upload - (re)introduce handling of existing files
We should reintroduce handling of existing files similar to girder based implementation 
https://github.com/dandi/dandi-cli/blob/master/dandi/upload.py#L342
since dandi-api does verify that there is no asset with a given path (https://github.com/dandi/dandi-api/issues/65). Since
 we have no locking, it would be a "racy" code but ATM I do not see another way around: we just would need to report an error if having uploaded an new file we still fail to mint a new asset for it (if another process uploaded for that path meanwhile as well).

TODOs
- [x] Test for correct behavior upon possible values of `existing`
- [x] Test should be added which would
  - trigger 'publish', i.e. making a release
  - upload a new file (into a draft)
  - verify that we can still get released file from the published version and from a draft version
  - remove file in the draft version
  - verify that we can still get released file from the published version

jwodder:
Closes #345.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -487,15 +487,8 @@ class DandiAPIClient(RESTFullAPIClient):
     def download_asset_bypath(
         self, dandiset_id, version, asset_path, filepath, chunk_size=MAX_CHUNK_SIZE
     ):
-        try:
-            # Weed out any assets that happen to have the given path as a
-            # proper prefix:
-            (asset,) = (
-                a
-                for a in self.get_dandiset_assets(dandiset_id, version, path=asset_path)
-                if a["path"] == asset_path
-            )
-        except ValueError:
+        asset = self.get_asset_bypath(dandiset_id, version, asset_path)
+        if asset is None:
             raise RuntimeError(f"No asset found with path {asset_path!r}")
         self.download_asset(
             dandiset_id, version, asset["uuid"], filepath, chunk_size=chunk_size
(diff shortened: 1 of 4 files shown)


dandi/dandiapi.py
```diff
@@ -518,3 +518,5 @@ def get_asset_bypath(self, dandiset_id, version, asset_path):
             )
         except ValueError:
             return None
+        else:
+            return asset
```
yarikoptic, line 522: Just for my own sake: what is the benefit from having this in else instead of at the end of the try/except block or even just after it?

jwodder: The line should be outside of the `try:` block because the `try:` is only supposed to catch failures in the `(asset,)` assignment (True, a `return` can't raise an exception, so it makes no practical difference here, but placing the line outside of the `try:` is in keeping with the practice of making `try:` enclose only what you want to catch errors from and nothing else.)  As for why to put it under `else` instead of just unindenting it a level, that's largely a stylistic choice, a consequence of my preference for writing things like:

```
if x:
    return y
else:
    return z
```

instead of:

```
if x:
    return y
return z
```

yarikoptic: gotcha, thanks.  Just didn't want to miss some unknown to me property.  I should use `try/else` more often I think

jwodder:
The PR includes the requests tests, but they currently fail (and are marked xfailing) due to https://github.com/dandi/dandi-cli/issues/345#issuecomment-768334263 and https://github.com/dandi/dandi-api/issues/70 not yet being addressed.

yarikoptic:
Thank you @jwodder . I guess we are just doomed to wait for https://github.com/dandi/dandi-api/issues/70 to get implemented.

yarikoptic:
there is `dandisets_versions_assets_delete` endpoint now, so it is possible to proceed

jwodder:
Found another blocking bug: https://github.com/dandi/dandi-api/issues/76

yarikoptic:
https://github.com/dandi/dandi-api/issues/76 was addressed, so should be possible to proceed forward.

jwodder:
A new blocking bug: https://github.com/dandi/dandi-api/issues/81

yarikoptic:
I have just merged https://github.com/dandi/dandi-api/pull/84 which should provide remedy (even if later to be refactored) so we could proceed here.

jwodder:
@yarikoptic Tests finally pass.

yarikoptic:
awesome -- let's proceed!

PR #348, opened 2021-01-27
Update  and simplify models to support automated editor generation

satra:
An attempt to reconcile the meditor and issues in the model.


satra:
here is a pointer to a codepen with the generated schema: https://codepen.io/satra-the-reactor/pen/NWbKbrj?editors=1010

satra:
just a note that i had to remove `exclude_unset` when exporting the json file from `DandiMeta` to make it work, otherwise it would remove the `schemaKey`

yarikoptic:
> just a note that i had to remove `exclude_unset` when exporting the json file from `DandiMeta` to make it work, otherwise it would remove the `schemaKey`

didn't check details, but note that it might flood `dandiset.yaml` with bunch of undefined fields.  But I guess if we are to aim for machine-readability ATM anyways, should not matter indeed ;)

dandi/tests/test_metadata.py
```diff
@@ -118,95 +100,55 @@ def test_metadata2asset(schema_dir):
...
+      "species": {
+        "identifier": "http://purl.obolibrary.org/obo/NCBITaxon_9606",
+        "name": "Human"
+      }
+    }
+  ]
+}"""
+    data_as_dict = json.loads(json_data)
+    assert data == AssetMeta(**data_as_dict)
     # We need to convert `data` to a `dict` this way instead of with `.dict()`
```
jwodder, line 150: This comment no longer applies now that the line below has been deleted.

dandi/tests/test_metadata.py
```diff
@@ -118,95 +100,55 @@ def test_metadata2asset(schema_dir):
...
     # so that enums will be converted to strings.
-    data_as_dict = json.loads(data.json(exclude_unset=True, exclude_none=True))
```
jwodder, line 209: Deleting this line assumes that the input JSON is the same as the output from `json.loads(data.json(...))`.  It would probably be better to add an assertion that the two forms are equal (both here and in the other tests).

jwodder (changes requested):

dandi/models.py
```diff
@@ -600,12 +700,19 @@ def check_data(cls, values):
...
     # From server (requested by users even for drafts)
-    manifestLocation: TempOptional[List[AnyUrl]] = Field(readOnly=True, nskey="dandi")
+    manifestLocation: TempOptional[List[HttpUrl]] = Field(readOnly=True, nskey="dandi")
 
     # On publish
     version: TempOptional[str] = Field(readOnly=True, nskey="schema")
-    doi: Optional[Union[AnyUrl, str]] = Field(
-        None, title="DOI", readOnly=True, nskey="dandi"
+    doi: Optional[str] = Field(
+        None, title="DOI", readOnly=True, pattern="^10\.[A-Za-z0-9.\/-]+", nskey="dandi"
```
jwodder, line 708: `\.` is not a valid escape sequence.  Either the string should be made raw (with an `r` prefix) or another backslash should be added.  Without this, certain versions of Python produce a syntax error.

satra:
@jwodder - how do you think we should fix the remaining linting issues. should we move the json document into a file?

jwodder:
@satra Why is it necessary to store serialized JSON in the code instead of the equivalent Python objects?  The latter would make it possible to wrap long strings.

satra:
i like the readability of those json objects. the intent of using json is that it would be good to ensure that the validation works  on json input. the thing that doesn't work are the long strings.

yarikoptic:
@jwodder we need your help
- I have pushed the tiny fix to allow for identifier to have `DANDI:` prefix when we load it from file 4a7b1849ab9ab9f7c4409d7727e7c7fe232ead30, unfortunately it is not sufficient
- need to fix something about metadata or its validation to account for changes here: metadata validation fails while running `python -m pytest -s -v dandi/tests/test_upload.py::test_new_upload_download` 
```
PATH                          SIZE    ERRORS UPLOAD STATUS MESSAGE
dandiset.yaml                                                     lidation error for Software\nidentifier\n  str type expected (type=type_error.sub-mouse001/sub-mouse001.nwb                                     
Summary:                      0 Bytes    
```
but since we do not raise in `upload()` ATM we do not even know about that but subsequent test for succesful upload fails with 

```
/home/yoh/proj/dandi/dandi-cli-master/dandi/tests/test_upload.py:268: in test_new_upload_download
    nwb_file2, = tmp_path.glob(f"{dandiset_id}{os.sep}*{os.sep}*.nwb")
E   ValueError: not enough values to unpack (expected 1, got 0)
```
- linting is not happy: lots of E501 (too long line) in `dandi/tests/test_metadata.py` but me trying to add 
```
diff --git a/dandi/tests/test_metadata.py b/dandi/tests/test_metadata.py
index 339efe1..ca15604 100644
--- a/dandi/tests/test_metadata.py
+++ b/dandi/tests/test_metadata.py
@@ -213,7 +213,7 @@ def test_metadata2asset_simple1(schema_dir):
     validate_asset_json(data_as_dict, schema_dir)
 
 
-def test_dandimeta_migration(schema_dir):
+def test_dandimeta_migration(schema_dir):  # noqa: E501
     data = migrate2newschema(
         {
             "dandiset": {
```
is of no effect.

jwodder:
@yarikoptic Regarding the long lines, see the brief conversation with @satra right above your comment.

yarikoptic:
Also, since I guess `auto` picks up changelog entries from the PR title, @satra could you please expand the title for this PR so it best reflects envisioned changes?

yarikoptic:
> @yarikoptic Regarding the long lines, see the brief conversation with @satra right above your comment.

but it doesn't provide an immediate remedy.  I was expecting to be able to just disable that particular check in that particular test, but I fail somehow -- I expect that I am doing something not completely right, or  `flake8` does not allow for establishing such skips at the level of the function?

jwodder:
@yarikoptic flake8 does not support function-level `noqa` comments.  The best way to ignore all of the checks would be to add the following to the `[flake8]` section in `setup.cfg`:

```
per-file-ignores =
    dandi/tests/test_metadata.py:E501
```

jwodder:
@yarikoptic @satra The reason the upload test is failing is because `Software.identifier` is supposed to be a string, yet [here](https://github.com/satra/dandi-cli/blob/4a7b1849ab9ab9f7c4409d7727e7c7fe232ead30/dandi/metadata.py#L310) it's being set to a `dict`.  Which entry should be changed, and to what?  (Also, once that's fixed, adjustments will have to be made for the `description` and `schemaKey` fields.)

satra:
let's change that to: `RRID:SCR_019009` 

also it looks like this has diverged from master and will need to be reconciled. @jwodder  - could you take a stab at this? let me know if you have any questions.

yarikoptic:
> @yarikoptic flake8 does not support function-level `noqa` comments. The best way to ignore all of the checks would be to add the following to the `[flake8]` section in `setup.cfg`:
> 
> ```
> per-file-ignores =
>     dandi/tests/test_metadata.py:E501
> ```

Thanks for the information.  I am ok to go this route

jwodder:
Branch rebased on top of the latest master.

satra:
thank you @jwodder - looks great.

yarikoptic:
Let's then proceed ;-)

Issue #352, opened 2021-02-01
FOI: 3.9 with h5py 2.10.0 started to fail to import


yarikoptic:
`ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject`

```
Run python -m pytest -s -v --cov=dandi --cov-report=xml dandi
ImportError while loading conftest '/home/runner/work/dandi-cli/dandi-cli/dandi/conftest.py'.
dandi/conftest.py:1: in <module>
    from .tests.fixtures import *  # noqa: F401, F403
dandi/tests/fixtures.py:12: in <module>
    import pynwb
/opt/hostedtoolcache/Python/3.9.1/x64/lib/python3.9/site-packages/pynwb/__init__.py:7: in <module>
    import h5py
/opt/hostedtoolcache/Python/3.9.1/x64/lib/python3.9/site-packages/h5py/__init__.py:46: in <module>
    from ._conv import register_converters as _register_converters
h5py/_conv.pyx:1: in init h5py._conv
    ???
h5py/h5t.pyx:1: in init h5py.h5t
    ???
E   ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject
Error: Process completed with exit code 4.
```
my wild guess is that due to pynwb limiting h5py to 2.x series, but we did test/impor ok just recently with such restrictions.  Did not look in detail on what is going on and where to complain - I do not think it is something we could address on our end.

jwodder:
I think the failure may have started due to the release of numpy 1.20.0 on the 30th.

jwodder:
On further inspection, this may also be due to a bug or something in pip's new resolver; hdmf requires `numpy <1.19.4, >=1.16`, yet numpy 1.20.0 is getting installed.  I am unable to reproduce this on my local machine running macOS with Python 3.9.

PR #353, opened 2021-02-01
Configure & apply isort via pre-commit

### LINKED ISSUE #350: Sorting import statements
The ordering of `import` statements in the code is all over the place.  Fortunately, there are two major projects that can help out in different ways (using one or the other, not both):

* [`flake8-import-order`](https://github.com/PyCQA/flake8-import-order) — a plugin for flake8 that, if used, would cause the linter to fail if any imports are out of order
    * The plugin supports a variety of import order styles.  I would recommend either the `appnexus` style or [my custom `jwodder` style](https://pypi.org/project/flake8-import-order-jwodder/).
* [`isort`](https://github.com/PyCQA/isort) — a program that can be added to the pre-commit configuration to automatically edit source files to put `import` statements in a reasonable order.
    * I recommend the following configuration for isort:

            profile = "black"
            force_sort_within_sections = true
            known_first_party = ["dandi"]

jwodder:
Closes #350.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,3 +13,8 @@ repos:
     hooks:
     -   id: black
         exclude: ^(dandi/_version\.py|versioneer\.py)$
+-   repo: https://github.com/PyCQA/isort
+    rev: 5.7.0
+    hooks:
+    -   id: isort
+        exclude: ^(dandi/_version\.py|dandi/due\.py|versioneer\.py)$
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ Options:
 
 See [DANDI Handbook](https://www.dandiarchive.org/handbook/10_using_dandi/)
 for examples on how to use this client in various use cases.
- 
+
 ## Development/contributing
 
 Please see [DEVELOPMENT.md](./DEVELOPMENT.md) file.
--- a/dandi/__init__.py
+++ b/dandi/__init__.py
@@ -1,5 +1,6 @@
 import logging
 import os
+
 from . import _version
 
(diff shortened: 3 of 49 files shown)


.pre-commit-config.yaml
```diff
@@ -13,3 +13,8 @@ repos:
     hooks:
     -   id: black
         exclude: ^(dandi/_version\.py|versioneer\.py)$
+-   repo: https://github.com/PyCQA/isort
+    rev: 5.7.0
+    hooks:
+    -   id: isort
+        exclude: ^(dandi/_version\.py|versioneer\.py)$
```
yarikoptic, line 20: also exclude `dandi/due.py` please

yarikoptic (changes requested):
minor request  + should not we also list it in `setup.cfg` for `style`?

yarikoptic:
and I guess we should merge it asap (I assume 3.9 fails are all due to #352 , didn't look)

.pre-commit-config.yaml
jwodder: Do you want that to be excluded from black as well?

jwodder:
> should not we also list it in setup.cfg for style?

pre-commit takes care of installing isort itself, so only pre-commit needs to be listed under `style`.

.pre-commit-config.yaml
yarikoptic: ideally - yes, I thought it was so.  It is a "standard" stab file from duecredit, we might take a new version in some time in the future.  I guess I have missed adding it to exclusions when I have added it.

yarikoptic:
all green (but 3.9) - let's proceed to not accumulate conflicts

PR #354, opened 2021-02-01
Add DANDI_TESTS_PERSIST_DOCKER_COMPOSE envvar for reusing Docker containers across test runs

### LINKED ISSUE #337: wishlist: add option to make docker_compose_setup persist across sessions
starting that fixture takes awhile which complicates rapid local development/testing.  I am not sure if there is some native pytest way to announce scope outside of `session`, it does seems to allow to change scope dynamically (https://docs.pytest.org/en/stable/fixture.html#dynamic-scope) but not sure if would help here.  so may be there could be some env var (e.g. DANDI_TESTS_PERSIST_DOCKER_COMPOSE=1|0) which if defined at all would instruct to first check if may be there is already a running instance and thus avoid `docker-compose up` while also not doing `docker-compose down` upon teardown unless == `0`; or choose other/better values than 1 and 0)

jwodder:
Closes #337.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -67,6 +67,10 @@ development command line options.
  `local_docker_compose*` fixtures use the given API key for
  `DANDI_INSTANCEHOST` instead of spinning up a new environment with a new key.
 
+- `DANDI_TESTS_PERSIST_DOCKER_COMPOSE` -- When set, the tests will reuse the
+  same Docker containers across test runs instead of creating & destroying a
+  new set on each run.  Set this environment variable to `0` to cause the
+  containers to be destroyed at the end of the next run.
 
 ## Sourcegraph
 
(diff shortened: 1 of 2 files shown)


yarikoptic:
AWESOME!  On my limited trials works nicely, thank you @jwodder .  Some fresh conflicts are to be resolved, and then couldbe merged

Note to myself,  there is still
```
Starting dandiarchive-docker_postgres_1 ... done
Starting dandiarchive-docker_minio_1    ... done
Starting dandiarchive-docker_mongo_1    ... done
Starting dandiarchive-docker_rabbitmq_1 ... done
Starting dandiarchive-docker_girder_1   ... done
```
upon each invocation, and some container blinks into existence but goes away when I run the single test...

jwodder:
@yarikoptic Conflicts resolved.

Issue #349, opened 2021-01-28
incomplete info for a file


djarecka:
Can't get enough info about `/sub-mouse-ZZPUG/sub-mouse-ZZPUG_ses-20181207-sample-1_slice-20181207-slice-1_cell-20181207-sample-1_icephys.nwb` from `DANDI:000008` when running `dandi ls`: no `metadata`, no `sha256` etc.

can't find this info also on website: https://gui.dandiarchive.org/#/dandiset/000021/draft/files?_id=5f2de703ee8baa608594ba82&_modelType=folder

yarikoptic:
could you provide an exact command 

<details>
<summary>cannot reproduce on a local copy</summary> 

```shell
(git-annex)lena:~/proj/dandi/dandisets/000008[master]git
$> datalad get sub-mouse-ZZPUG/sub-mouse-ZZPUG_ses-20181207-sample-1_slice-20181207-slice-1_cell-20181207-sample-1_icephys.nwb
get(ok): sub-mouse-ZZPUG/sub-mouse-ZZPUG_ses-20181207-sample-1_slice-20181207-slice-1_cell-20181207-sample-1_icephys.nwb (file) [from web...]   
(dev3) 1 54696 [1].....................................:Mon 01 Feb 2021 10:53:08 AM EST:.
(git-annex)lena:~/proj/dandi/dandisets/000008[master]git
$> dandi ls sub-mouse-ZZPUG/sub-mouse-ZZPUG_ses-20181207-sample-1_slice-20181207-slice-1_cell-20181207-sample-1_icephys.nwb
/usr/lib/python3/dist-packages/hdmf/spec/namespace.py:484: UserWarning: Ignoring cached namespace 'hdmf-common' version 1.1.3 because version 1.3.0 is already loaded.
  warn("Ignoring cached namespace '%s' version %s because version %s is already loaded."
/usr/lib/python3/dist-packages/hdmf/spec/namespace.py:484: UserWarning: Ignoring cached namespace 'core' version 2.1.0 because version 2.2.5 is already loaded.
  warn("Ignoring cached namespace '%s' version %s because version %s is already loaded."
- age: P98D
  cell_id: 20181207_sample_1
  date_of_birth: 2018-08-31 00:00:00-04:00
  experiment_description: Current clamp square
  experimenter:
  - Matteo Bernabucci
  genotype: Pvalb-Cre/wt; Ai9/wt
[...]

PR #355, opened 2021-02-01
ENH: upload: --existing=overwrite-metadata to upload fresh metadata even if file already there

### LINKED ISSUE #105: upload: only metadata
Since we are still working out our schema of metadata to be kept on the remote dandiarchive server, it would be useful to have a mode of operation where it 
- would not upload file if it was not changed
- instead of removing a girder item it would just adjust its metadata with up-to-date information (current metadata schema, possibly new validations, etc)

yarikoptic:
To facilitate cases where we want to ask users to reupload while using newer dandi-cli which would extract metadata differently

Closes #105 

TODOs:
- [x] Development safeguard (`DANDI_UPLOAD_SAME_MUST_EXIST` env var) for paranoids so if we operate under assumption that we have up to date local data files, we do not mistakenly reupload (outdated) data.
- [ ] test on sample cases to verify correct operation (see e.g. #349 as one of the target use cases)

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -72,6 +72,9 @@ development command line options.
   new set on each run.  Set this environment variable to `0` to cause the
   containers to be destroyed at the end of the next run.
 
+- `DANDI_UPLOAD_SAME_MUST_EXIST` -- primarily for internal use with
+ `--existing=overwrite-metadata` so we do not upload some outdated file.
+
 ## Sourcegraph
 
 The [Sourcegraph](https://sourcegraph.com) browser extension can be used to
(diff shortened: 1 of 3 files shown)


jwodder (approved):

yarikoptic:
"tested" in real life, managed to delete file with some version before this one... so I better wait to have done more testing or if I add some unittest to guarantee that  there is no side effects... and we might not need it in the long run with a new API server since if data blob is the same - we should not reupload it.

yarikoptic:
I will close it for now ... can resurrect if the need arises again and would need to do it also for dandi-api interface

PR #356, opened 2021-02-01
Fix numpy dependency issue in tests

### LINKED ISSUE #352: FOI: 3.9 with h5py 2.10.0 started to fail to import

jwodder:
Fixes #352.

Fix inspired by [this comment](https://github.com/pypa/pip/issues/9542#issuecomment-770379808).

Unfortunately, I do not see a way to ensure that users of dandi-cli who do not already have numpy installed do not run into this problem.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,6 +55,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip wheel
+        pip install 'numpy<1.19.4'
         pip install ".[test]"
 
     - name: Set environment variable when using Python 3.6 to cover more code
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,6 +55,13 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip wheel
+        # As of 2021-02-01, the latest version of hdmf (v2.3.0) requires numpy
+        # >=1.16, <1.19.4, so we end up with numpy 1.19.3 installed.  However,
+        # due to the release of numpy 1.20.0 on January 30, h5py gets built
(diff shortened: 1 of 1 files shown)


yarikoptic:
Thank you @jwodder !!!! I expect the issue to be properly resolved "upstream" and workaround would let us keep going, thus merging

jwodder:
@yarikoptic I'm not sure if this will be fixed upstream, as it only affects h5py 2.x, and the project is already on to 3.1.

PR #357, opened 2021-02-02
Add dandiset ID override for upload

dchiquito:
With DANDI_DEVEL on, you can now specify `-d 123456` to upload the
dandiset directly to dandiset `123456`, regardless of the contents of
`dandiset.yaml`. This makes migrating dandisets easier, as the API
server does not have a way to create a dandiset with a given identifier.

I tested this with:
```
dandi download https://dandiarchive.org/dandiset/000027/draft
cd 000027
dandi upload -i dandi-api --validation ignore --allow-any-path --upload-dandiset-metadata -d 001001
```

You can see that https://gui-beta-dandiarchive-org.netlify.app/#/dandiset/001001 now contains 18.35 KB worth of files, exactly the size of dandiset 000027.

Running the `map_ids` script will not work at the moment because dandiset 000027 specifies the identifier as "DANDI:000027" in the metadata which the script does not support, but we're one step closer.

### DIFF
--- a/dandi/cli/cmd_upload.py
+++ b/dandi/cli/cmd_upload.py
@@ -39,6 +39,8 @@ from ..consts import collection_drafts
 #
 # TODO: should always go to dandi for now
 @instance_option()
+# TODO remove this once migration is complete
+@devel_option("-d", "--identifier", help="Override the identifier in dandiset.yaml")
 # TODO: should always go into 'drafts' (consts.collection_drafts)
 @devel_option(
     "-c", "--girder-collection", help="For development: Girder collection to upload to"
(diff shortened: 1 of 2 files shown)


yarikoptic:
CHEATER! ;) We can try this, but do not get surprised if we start getting various side-effects from "a difficult life of a dataset identifier" when either API would freak out when uploaded metadata has a different one, or mapping breaks or who knows what. I still think that any of the ideas I suggested https://github.com/dandi/dandi-api/issues/69 would have been cleaner and thus more robust.

yarikoptic:
with https://github.com/dandi/dandi-api/pull/87 should no longer be needed

PR #366, opened 2021-02-08
Add script for migrating Dandiset metadata

### LINKED ISSUE #360: "migrate" existing dandisets (only dandiset.yaml for now) to new API  server
Blockers (AFAIK should be easy to address, so may be just send a PR @jwodder ):
- [x] ~~https://github.com/dandi/dandi-api/issues/57~~ no longer pertinent

Overall workflow:

- Given a full collection of dandisets (so later we could upload data files as well), which is at `dandi@drogon:/mnt/backup/dandi/dandisets` or for now (since no data) just on a local copy which could be obtained quickly using `datalad install -r -J5 https://github.com/dandi/dandisets`
- make a script which takes path(s) of dandisets to operate on, and for each given dandiset, 
  - uses `POST ​/dandisets​/`  to create a new one while providing metadata with original identifier (now possible, thanks to https://github.com/dandi/dandi-api/pull/87)
  - uploads dandiset.yaml to that new dandiset (I am still a bit reluctant to merge it in, but we could do that if would make things easier)

Script should have option to delete a dataset if already exists so we would just recreate it with that identifier for a "clean start"

jwodder:
Closes #360.

### DIFF
--- a/tools/migrate-dandisets.py
+++ b/tools/migrate-dandisets.py
@@ -1,0 +1,37 @@
+#!/usr/bin/env python3
+import click
+import requests
+
+from dandi.dandiapi import DandiAPIClient
+from dandi.dandiset import APIDandiset
+
+
+@click.command()
+@click.option(
+    "-d", "--delete-extant", is_flag=True, help="Delete Dandisets that already exist"
+)
+@click.argument("api_url")
+@click.argument("token")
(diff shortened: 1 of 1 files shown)


yarikoptic:
I will test it in the battlefield ;) Thank you @jwodder

Issue #369, opened 2021-02-09
Make validation on local assets not require identifier and contentUrl


yarikoptic:
following brief (not-yet-)discussion with @satra on slack, our AssetMeta record already has some fields which are known/set only on the server side: 
- `identifier: UUID4 = Field(readOnly=True, nskey="schema")`
- `contentUrl: Optional[List[HttpUrl]] = Field(None, readOnly=True, nskey="schema")`
- ... may be more ...

Also there is 
- `path: str = Field(None, nskey="dandi")` which we do not provide as part of the metadata record but rather as parameter to `PUT` call.  Since in current state of things a new asset UUID is minted only upon publish, I think we should just upload metadata record with path, hence https://github.com/dandi/dandi-api/issues/91 but I could be shown that it is a bad idea

jwodder:
Is is strictly necessary for these fields to be stored under `"metadata"` instead of at the top-level of the asset object alongside `created` etc.?

satra:
we can add anything at the top level for archive interaction, but the metadata field has to correspond to our schema and should allow us to change the schema without database migrations.

yarikoptic:
but it seems we need to "separate" / subclass, or otherwise annotate schema for a local asset -vs- asset on the server.

satra:
you need to decide which fields to check for locally during validation. it will be a bit more complicated to separate out fields, since many are shared through the commonmodel between dandiset and asset.

yarikoptic:
#373 initiated the approach where could use BareAssetMeta (instead of AssetMeta) for validation of the local assets.

Otherwise validation of any asset would fail ATM, which leads to e.g.
```
dandi@drogon:/mnt/backup/dandi/dandi-api-datasets$ git grep -B1 'field required (type=value_error.missing)' | grep identifier | nl | tail
  6500	000053/VALIDATION.errors-identifier
  6501	000053/VALIDATION.errors-identifier
  6502	000053/VALIDATION.errors-identifier
  6503	000053/VALIDATION.errors-identifier
  6504	000053/VALIDATION.errors-identifier
  6505	000053/VALIDATION.errors-identifier
  6506	000053/VALIDATION.errors-identifier
  6507	000053/VALIDATION.errors-identifier
  6508	000053/VALIDATION.errors-identifier
  6509	000053/VALIDATION.errors-identifier
```
fails.

PR #373, opened 2021-02-09
RF/NF: Identifiable and BareAssetMeta to describe an asset anywhere

yarikoptic:
Identifiable is a mix-in class for anything with an identifier.
It is "extracted" from the CommonModel since we might not have identifiers defined locally.

BareAssetMeta provides a base class to AssetMeta which could then be populated only on the server side.

This PR is more of an experiment as a possible solution to #369 .  In converted schema it should all still look the same (didn't check though), may be only for some reordering of keys within schema.

### DIFF
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -3,7 +3,7 @@ import json
 import os.path as op
 from pathlib import Path
 import re
-from uuid import UUID, uuid4
+from uuid import uuid4
 
 import jsonschema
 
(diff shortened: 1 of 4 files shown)


jwodder:
@yarikoptic Exactly what else needs to be done for this PR?

yarikoptic:
Use it to fix #369

jwodder:
@yarikoptic Is this the sort of thing you had in mind?  I've tested this out with https://github.com/dandi/dandi-api-datasets/pull/8, and all the assets in 000053 passed validation.

yarikoptic:
> @yarikoptic Is this the sort of thing you had in mind? I've tested this out with [dandi/dandi-api-datasets#8](https://github.com/dandi/dandi-api-datasets/pull/8), and all the assets in 000053 passed validation.

THANKS!  I think it is *it* ;)

yarikoptic:
I guess let's proceed

PR #378, opened 2021-02-10
Add `include_metadata=False` parameter to asset-listing DandiAPIClient methods

### LINKED ISSUE #376: DandiAPIClient.get_dandiset_*assets: add option include_metadata=False
Primary use case is `dandi ls` to quickly display assets with their metadata, and be able to align what we display for local assets and remote assets.  Later it would be of use for `dandi diff` (https://github.com/dandi/dandi-cli/issues/72)

So, the idea is that if it is set to True, assets in the returned list would have `metadata` field with their metadata.  Two possible strategies (choice is yours) to have it implemented

- do it on the server side (https://github.com/dandi/dandi-api/issues/100) and just pass that option to the endpoint call.  But according to @dchiquito that might require django trickery to switch serializers depending on the parameter, so might apparently be "a more involved solution" than it seems to blissfully django-ignorant me.
- do it on `dandi-cli` side:  iterate through assets and query their metadata via `dandisets_versions_assets_read` endpoint (I guess could be done via asyncio/threading/whatnot to get some speedup).  We will be able to RF it away later whenever implemented on the server side

In either case, a unittest should ensure that we are getting metadata whenever we request it.

jwodder:
Closes #376.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -285,9 +285,9 @@ class DandiAPIClient(RESTFullAPIClient):
         while True:
             for asset in resp["results"]:
                 if include_metadata:
-                    asset["metadata"] = self.get_asset(
-                        dandiset_id, version, asset["uuid"]
-                    )["metadata"]
+                    asset_rec = self.get_asset(dandiset_id, version, asset["uuid"])
+                    if "metadata" in asset_rec:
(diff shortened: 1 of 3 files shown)


yarikoptic (approved):
pushed two tune ups (one is to address my own comment).  If tests pass -- good to go (please merge) and if I broke something - would need to be fixed ;)

PR #377, opened 2021-02-10
Support downloading folders and latest Dandiset version

jwodder:
Closes #340.

PR #380, opened 2021-02-10
Add sample Dandiset test fixture

jwodder:
This change allows us to cut down on a bunch of duplicate code in the tests.

This PR is built on top of #377.  Only the last commit is specific to this branch.


yarikoptic:
nice -- thank you @jwodder !

PR #381, opened 2021-02-11
List extra auto labels in DEVELOPMENT.md

jwodder:
These labels are not listed in auto's documentation, but they are still recognized & used, as can be seen by dumping a repository's auto config with `auto info -v`.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -118,3 +118,6 @@ organized by label.  `auto` recognizes the following PR labels:
   component; this is the default label for unlabelled PRs
 - `internal` — for changes only affecting the internal API
 - `documentation` — for changes only affecting the documentation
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -118,3 +118,6 @@ organized by label.  `auto` recognizes the following PR labels:
   component; this is the default label for unlabelled PRs
 - `internal` — for changes only affecting the internal API
 - `documentation` — for changes only affecting the documentation
+- `tests` — for changes to tests
```
yarikoptic, line 121: we had `testing` with a single resolved issue #164, so I reassigned it to tests and removed testing label (I guess we could have just renamed testing into tests)

yarikoptic (approved):
Let's try to not forget to use them (in particular `dependencies`)

PR #371, opened 2021-02-09
Set "path" in metadata when uploading to new API

jwodder:
Closes #368.

PR #382, opened 2021-02-12
Set asset path via metadata only

jwodder:
Merge if & only if https://github.com/dandi/dandi-api/pull/105 has been merged.

Replaces #371.


yarikoptic:
https://github.com/dandi/dandi-api/pull/105 was merged, testing is broken ATM, so I will just merge and  try in real life. Thank you @jwodder !

PR #383, opened 2021-02-12
Make `dandi download -i <instance>` run in a Dandiset download that Dandiset

### LINKED ISSUE #153: download: while running within dandiset should take that dandiset as default to download from

jwodder:
Closes #153.

Note that the tests for this change must be run with `DANDI_DEVEL=1`, which also enables `test_enormous_upload_breaks_girder`.  As a result, the new tests are not run under CI.

### DIFF
--- a/dandi/cli/cmd_download.py
+++ b/dandi/cli/cmd_download.py
@@ -2,7 +2,9 @@ import os
 
 import click
 
-from .base import map_to_click_exceptions
+from .base import instance_option, map_to_click_exceptions
+from ..consts import known_instances, known_instances_rev
+from ..dandiarchive import parse_dandi_url
 
 
 class ChoiceList(click.ParamType):
@@ -71,6 +73,7 @@ class ChoiceList(click.ParamType):
     default="all",
     show_default=True,
 )
+@instance_option()
 # Might be a cool feature, not unlike verifying a checksum, we verify that
 # downloaded file passes the validator, and if not -- alert
 # @click.option(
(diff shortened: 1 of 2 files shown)


yarikoptic:
> Closes #153.
> 
> Note that the tests for this change must be run with `DANDI_DEVEL=1`, which also enables `test_enormous_upload_breaks_girder`. As a result, the new tests are not run under CI.

couldn't those specific tests mock patch env just for their own `CliRunner` invocation?

jwodder:
@yarikoptic No, `DANDI_DEVEL` needs to be set when the code is loaded in order for the `--dandi-instance` option to be available.

yarikoptic:
oh, ok. Thanks for the explanation.  Let's proceed then as is

PR #384, opened 2021-02-15
Give known_urls human-readable display strings

jwodder:
This PR associates each entry in `known_urls` with a string describing the pattern that should be more readable than the regexes themselves; the strings are then used in place of the regexes in the exception message when raising `UnknownURLError`.

The PR also reformats the `known_urls` source to be, in my opinion, more readable.

### DIFF
--- a/dandi/dandiarchive.py
+++ b/dandi/dandiarchive.py
@@ -110,68 +110,116 @@ def _map_to_girder(url):
 
 
 class _dandi_url_parser:
-    # Defining as a class with all the attributes to not leak all the variables etc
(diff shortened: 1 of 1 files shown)


yarikoptic:
Looks indeed better!  I do not see a point in compiling regexes right there (`re` does a good job caching compiled ones) but I guess that makes it more explicit, thus good with me.

didn't look through code and why tests are failing

jwodder:
@yarikoptic The tests are failing because of https://github.com/dandi/dandi-api/issues/112.

jwodder:
@yarikoptic The tests are passing now.

PR #386, opened 2021-02-15
Populate wasDerivedFrom

### LINKED ISSUE #385: populate 'wasDerivedFrom'
Originally posted by @satra in https://github.com/dandi/dandi-api-datasets/pull/6#issuecomment-778664746 :

    wasDerivedFrom is intended to be a chain of biosamples from cell id, slice id, tissuesample id, to subject id.

"chain" here is nesting, i.e. that cell biosample should have wasDerivedFrom pointing to slice and so on.  Any particular level could be skipped if not present

jwodder:
Closes #385.

I am marking this as a draft because further edits will be necessary if & when #373 is merged.

### DIFF
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -253,11 +253,17 @@ def extract_model_list(modelcls, id_field, id_source, **kwargs):
 
 def extract_wasDerivedFrom(metadata):
     derived_from = None
-    for field in ("tissue_sample_id", "slice_id", "cell_id"):
+    for field, sample_name in [
+        ("tissue_sample_id", "tissuesample"),
+        ("slice_id", "slice"),
+        ("cell_id", "cell"),
+    ]:
         if metadata.get(field) is not None:
             derived_from = [
                 models.BioSample(
-                    identifier=metadata[field], wasDerivedFrom=derived_from
+                    identifier=metadata[field],
+                    wasDerivedFrom=derived_from,
+                    sampleType=models.SampleType(name=sample_name),
                 )
             ]
     return derived_from
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -231,8 +231,6 @@ def extract_anatomy(metadata):
 def extract_model(modelcls, metadata, **kwargs):
     m = modelcls.unvalidated()
     for field in m.__fields__.keys():
-        if modelcls == models.BioSample and field == "wasDerivedFrom":
-            continue
         value = kwargs.get(field, extract_field(field, metadata))
         if value is not Ellipsis:
             setattr(m, field, value)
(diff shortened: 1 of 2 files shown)


yarikoptic:
FWIW I have just merged #373 to not block this one

jwodder:
@yarikoptic PR updated.

dandi/metadata.py
```diff
@@ -319,7 +308,17 @@ def nwb2asset(nwb_path, digest=None, digest_type=None):
 
 
 def metadata2asset(metadata):
-    return extract_model(models.BareAssetMeta, metadata)
+    m = extract_model(models.BareAssetMeta, metadata)
+    derived_from = None
+    for field in ("subject_id", "tissue_sample_id", "slice_id", "cell_id"):
+        if metadata.get(field) is not None:
+            derived_from = [
+                models.BioSample(
```
yarikoptic, line 316: Here identifier wouldn't provide semantic of what that identifier of. 
@satra - should we need to define classes for each one of those types of biosample at different levels ?  For subject_id we already have `Participant` in the model (but it is not a BioSample, so we would need to either make it a subclass of  BioSample or may be better allow wasDerivedFrom to be
```python
    wasDerivedFrom: Optional[Union(BioSample, Participant)]] = Field(None, nskey="prov")
```
instead of current
```python
    wasDerivedFrom: Optional[List["BioSample"]] = Field(None, nskey="prov")
```
(unless you see the need for list with the hierarchy), and then define subclasses of `BioSample`: `Tissue`, `Slice`, `Cell`.  WDYT?

yarikoptic: I see that we have already 
```
    wasAttributedTo: List[Participant] = Field(
        None, description="Participant(s) to which this file belongs to", nskey="prov"
    )
```
so I guess we should not place subject id in `wasDerivedFrom` and rely on participant be linked only in `wasAttributedTo`, correct @satra?

satra: correct.  for BioSample, we can either create subclasses or set `sampleType` - it's a new field i added this morning.

jwodder:
Converting this back to a draft in anticipation of changes in #392 that will necessitate changes here.

dandi/metadata.py
jwodder: Assuming we want to set the `sampleType` on each `BioSample` in `wasDerivedFrom`, what would the various fields of the `SampleType` be?

yarikoptic: I was wondering about that (`sampleType`) myself but forgot to ask @satra . as far as I see it boils down to use of an ontology, which we do not really do yet, so I would not sure if we can do anything about it ATM.  But it would be nice for all those referring to OBI to at least have a link to it alongside somewhere (I failed to find the ultimate source ATM via google).

satra:
we can create our own types for now (cell, slice, block). let me see if i can find something. for the moment can you add strings: "cell, slice, tissuesample" to match to what we have?

@bendichter - any thoughts on types of tissues or samples? also what does probeid mean? doesn't seem like a biosample thing, would be good for us to know where to put that? probably some other entity.

jwodder:
@satra 

> for the moment can you add strings: "cell, slice, tissuesample" to match to what we have?

Do you mean that those should be used as the `name` fields of the respective `SampleType`s?

satra:
yes `name` for now.  we may eventually change that to an Enum or create derived classes (more likely, since that would allow us to add additional metadata). i'll look into those.

yarikoptic:
looks good to me, despite a bit odd `tissuesample` but since it is only temporary (I guess before we switch to use a vocab/ontology) IMHO it is ready to proceed. WDYT @satra?

satra:
go for it - i'm sure this will change over the next few weeks, but we have a framework in place.

PR #388, opened 2021-02-16
Support passing precomputed file digest to DandiAPIClient upload methods

jwodder:
Currently, when uploading a file to the new API with the `upload()` function / `dandi upload` command, its digest is computed twice: once for use in the metadata and again for use in the validation API.  This PR lets DandiAPIClient's upload methods accept a precomputed digest so that we only need to digest each file once.

PR #391, opened 2021-02-16
Cache file digests and check for change in digest when uploading

### LINKED ISSUE #389: fscache checksum computation on a file
Inspired by the necessity of #388.

I initially had reservations about that but I think it should be ok - we can always disable caching using `DANDI_CACHE` to troubleshoot etc.    I guess we just need a helper function 

```python
@checksums.cache
def get_digest(filepath, digest='sha256'):
    return  Digester([digest])(filepath)[digest]
```
?

that should help significantly rerunning `upload` on files.

If we cached the checksum, we could call it even twice (in context of #388) and actually take advantage of that -- if checksum changes from one point to another during upload - we have an good indication that user is trying to upload a file which is still being modified.  WDYT @jwodder  ?

jwodder:
Closes #389.

This PR reverts #388.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -348,9 +348,7 @@ class DandiAPIClient(RESTFullAPIClient):
             dandiset["metadata"] = dandiset_metadata.pop("dandiset")
         return dandiset
 
-    def upload(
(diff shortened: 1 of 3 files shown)


yarikoptic:
Looks good, thank you @jwodder!

Issue #343, opened 2021-01-26
model: access doesn't survive R/T


yarikoptic:
```
(Pdb) DandiMeta.unvalidated(identifier='0001').dict(exclude_none=True)
{'identifier': '0001', 'schemaVersion': '1.0.0-rc1', 'access': [{'status': <AccessType.Open: 'dandi:Open'>}], 'repository': 'https://dandiarchive.org/'}
(Pdb) DandiMeta.unvalidated(**json.loads(DandiMeta.unvalidated(identifier='0001').json(exclude_none=True))).dict(exclude_none=True)
{'identifier': '0001', 'schemaVersion': '1.0.0-rc1', 'access': [{'status': 'dandi:Open'}], 'repository': 'https://dandiarchive.org/'}
```

so you can see that it putates from `[{'status': <AccessType.Open: 'dandi:Open'>}]` to ` 'access': [{'status': 'dandi:Open'}]`

satra:
we will likely need to hack it in our models for the time being  through custom parsers for embedded types. there is an open issue around this issue (https://github.com/samuelcolvin/pydantic/issues/1409) that may soon see a PR.

yarikoptic:
or should we just remove them for now ? (#344) ;)

satra:
that's not the only place where the roundtrip will run into an issue. so removing them doesn't solve the issue.

yarikoptic:
yes, but it is one of them so provides partial temporal workaround which would help to proceed ATM AFAIK. Sure thing if there is an alternative route which could resolve them in short term - we could go that way too (but waiting on PR in pydantic + release might take awhile)

satra:
we definitely cannot wait for their PR, but we can change our models to support round tripping and exporting cleaner json. this is really about serialization and deserialization.

satra:
btw, i did this check at the level of the model and this seems to work. i have to check how well it works with nested models:

```
In [13]: AccessRequirements(status=AccessType.Open).json()
Out[13]: '{"status": "dandi:Open", "email": null, "contactPoint": null, "description": null, "embargoedUntil": null}'

In [15]: AccessRequirements(**json.loads(AccessRequirements(status=AccessType.Open).json
    ...: ()))
Out[15]: AccessRequirements(status=<AccessType.Open: 'dandi:Open'>, email=None, contactPoint=None, description=None, embargoedUntil=None)
```

satra:
also simple nesting works:

```
[...]

PR #392, opened 2021-02-17
enh: account for samples, sessions and participants

satra:
this should clarify how to store biosample and participant info.


dandi/metadata.py
```diff
@@ -260,6 +260,14 @@ def extract_wasAttributedTo(metadata):
         return [wat]
 
 
+def extract_wasGeneratedBy(metadata):
+    wat = extract_model(models.Session, metadata, identifier=metadata.get("session_id"))
```
satra, line 264: ideally this would be name, not identifier.

dandi/models.py
```diff
@@ -791,6 +849,12 @@ class BareAssetMeta(CommonModel):
     wasAttributedTo: List[Participant] = Field(
         None, description="Participant(s) to which this file belongs to", nskey="prov"
     )
+    wasGeneratedBy: Optional[List[Union[Session, Project, Activity]]] = Field(
```
jwodder, line 852: This overrides the `CommonModel.wasGeneratedBy` attribute that is currently set in `nwb2asset()`.  Perhaps you should pick a different name?

satra: it's the same field and should be converged as necessary.

jwodder: The field in the parent class is a scalar, while this is a list.  That's bad inheritance.

satra: then the parent class should be adjusted to be a list. a given dandiset or asset could have been generated as as function of many activities. especially since we don't parcellate out wasGeneratedBy to specific elements.

yarikoptic:
So, if I get it right, it is the `wasGeneratedBy` and `Session` (and placing aside added cross-participant relationship) which after this PR is merged could be used in addition to https://github.com/dandi/dandi-cli/pull/386 which populates `wasDerivedFrom` and does not concern with the Session.  

But I fail to understand how this PR addresses the question of `tissue`, `slice`, `cell` id placement which I have asked in https://github.com/dandi/dandi-cli/pull/386/files#r577631718 , could you please clarify @satra ?

CI errors are not due to nwb validation (ignore those warnings) but due to
```
dandi/tests/test_validate.py:8: in test_validate_simple1
    assert not errors
E   AssertionError: assert not ['3 validation errors for BareAssetMeta\nwasGeneratedBy\n  value is not a valid list (type=type_error.list)\nwasDerive...e.not_allowed)\nwasAttributedTo -> 0 -> identifier\n  none is not an allowed value (type=type_error.none.not_allowed)']
```
so probably relates to https://github.com/dandi/dandi-cli/pull/392#discussion_r577748245 above

dandi/models.py
satra: i adjusted the parent class.

jwodder:
The tests are currently failing because certain `Session` objects are somehow being validated against the `Activity` model, and the different `schemaKey` values are leading to a validation error.  This is only happening on Python 3.6, yet I cannot reproduce it when running the tests in a `python:3.6` or `python:3.6.12` Docker container.  I am not sure what exactly is causing the problem or how to fix it.

yarikoptic:
I wonder if that is due to the `Union[Session, Project, Activity]` and pydantic somehow failing to match `Session` first (or may be order is not that well defined, although dicts should be ordered as of cpython 3.6) and thus being just yet another manifestation of #343 and may be the fact .  I will try to reproduce locally ...

satra:
`schemaKey` is intended to disambiguate that `Union` so perhaps there is a place where we are not adding `schemaKey` to the JSON output. i'll check as well.

yarikoptic:
> I cannot reproduce it when running the tests in a `python:3.6` or `python:3.6.12` Docker container

does smth like
```
python3 <(curl --silent -L http://get.datalad.org) -E /tmp/dandi-miniconda3.6.sh miniconda --path /tmp/dandi-miniconda3.6 --batch --spec '-y python=3.6' conda-env -n dandi --spec='-c conda-forge -y --only-deps dandi'
source /tmp/dandi-miniconda3.6.sh
pip install -e .[test] 
python -m pytest --pdb dandi/tests/test_metadata.py::test_metadata2asset_simple1
```
work for you @jwodder ?

<details>
<summary>in my case I ran into that numpy/h5py order of installation isssue IIRC</summary> 

```shell
$> python -m pytest --pdb dandi/tests/test_metadata.py::test_metadata2asset_simple1
ImportError while loading conftest '/home/yoh/proj/dandi/dandi-cli-master/dandi/conftest.py'.
dandi/conftest.py:1: in <module>
    from .tests.fixtures import *  # noqa: F401, F403
dandi/tests/fixtures.py:13: in <module>
    import pynwb
/tmp/dandi-miniconda3.6/envs/dandi/lib/python3.8/site-packages/pynwb/__init__.py:7: in <module>
    import h5py
/tmp/dandi-miniconda3.6/envs/dandi/lib/python3.8/site-packages/h5py/__init__.py:46: in <module>
    from ._conv import register_converters as _register_converters
h5py/_conv.pyx:1: in init h5py._conv
    ???
h5py/h5t.pyx:1: in init h5py.h5t
    ???
E   RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 80 from C header, got 88 from PyObject
```
</details>

but as the story goes -- something is indeed odd with resolving of the order in pydantic on 3.6:
```
(Pdb) Union[ml.Session, ml.Activity](**data_as_dict['wasGeneratedBy'][0])
*** pydantic.error_wrappers.ValidationError: 1 validation error for Activity
schemaKey
  unexpected value; permitted: 'Activity' (type=value_error.const; given=Session; permitted=('Activity',))
```
(and the same if I flip Session and Activity order) although

```
*(Pdb) ml.Session(**data_as_dict['wasGeneratedBy'][0])
Session(identifier=None, name='session_id1', description=None, startDate=None, endDate=None, wasAssociatedWith=None, schemaKey='Session')
```
so session itself is ok... I see Satra is pushing smth...

satra:
found one - let's see what it does.

yarikoptic:
I don't think it would be the one or if it would be the one fixing the test it is not the one representative of the issue, since we do have schemaKey (in the examples above) but:
```
(Pdb) p data_as_dict['wasGeneratedBy'][0]
{'name': 'session_id1', 'schemaKey': 'Session'}

(Pdb) Union[ml.Activity, ml.Session](**data_as_dict['wasGeneratedBy'][0])
*** pydantic.error_wrappers.ValidationError: 1 validation error for Activity
schemaKey
  unexpected value; permitted: 'Activity' (type=value_error.const; given=Session; permitted=('Activity',))
```
so something finicky with python3.6/pydantic

satra:
i merged master and now my latest changes are failing on validation. looking into it now.

satra:
i found the current issue:

in this function, validation is run on full asset metadata (https://github.com/dandi/dandi-cli/blob/master/dandi/metadata.py#L549)

however metadata2asset no longer returns identifier, hence the test is failing because the json output does not contain an identifier.

satra:
faking an identifier for testing, but we will need to resolve the issue.

satra:
i think a fix would be to have ` validate_asset_json(data_as_dict, schema_dir)` take a model class as input, generate the schema and then validate it. this would allow using `BareAssetMeta`, which is what nwb2asset is using. however on metadata from the server we could use `AssetMeta`.

satra:
this is back now to the py 3.6 failures + a memoize one.

yarikoptic:
<details>
<summary>FWIW issue replicates in py3.6 with current pydantic master</summary> 

```shell
$> python -m pytest --pdb dandi/tests/test_metadata.py::test_metadata2asset_simple1
====================================================== test session starts ======================================================
platform linux -- Python 3.6.10, pytest-5.4.1, py-1.8.1, pluggy-0.13.1
rootdir: /home/yoh/proj/dandi/dandi-cli-master, inifile: tox.ini
plugins: cov-2.8.1
collected 1 item                                                                                                                

dandi/tests/test_metadata.py F
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
dandi/tests/test_metadata.py:231: in test_metadata2asset_simple1
    assert data == BareAssetMeta(**data_as_dict)
../../misc/pydantic/pydantic/main.py:394: in __init__
    raise validation_error
E   pydantic.error_wrappers.ValidationError: 1 validation error for BareAssetMeta
E   wasGeneratedBy -> 0 -> schemaKey
E     unexpected value; permitted: 'Activity' (type=value_error.const; given=Session; permitted=('Activity',))
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> entering PDB >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB post_mortem (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /home/yoh/proj/misc/pydantic/pydantic/main.py(394)__init__()
-> raise validation_error
(Pdb) import pkg_resources
(Pdb) pkg_resources.get_distribution('pydantic')
pydantic 1.7.3 (/home/yoh/proj/misc/pydantic)

```

although git tag is odd

```
$> git -C /home/yoh/proj/misc/pydantic describe --tags
v1.7.2-76-gb7a8ef2
```
since they had v1.7.3 in a dedicated release branch off 1.7.2
</details>

<details>
<summary>Following minimalistic example</summary> 

```python
from pydantic import BaseModel, Field

from typing import Union
from typing_extensions import Literal

class Activity(BaseModel):
  name: str = Field(
    title="Title",
    description="The name of the logical session associated with the asset.",
    max_length=150,
    nskey="schema",
  )
  schemaKey: Literal["Activity"] = Field("Activity", readOnly=True)

class Session(Activity):
  schemaKey: Literal["Session"] = Field("Session", readOnly=True)

print(Union[Session,Activity](**{'name': 'session_id1', 'schemaKey': 'Session'}))
```
</details>

<details>
<summary>seems to replicate with that python 3.6/master pydantic</summary> 

```shell
(dandi-py3.6) 1 58631 ->1 [2].....................................:Wed 17 Feb 2021 07:04:27 PM EST:.
(git)lena:~/proj/misc/pydantic[master]git
$> python /tmp/pyda.py               
  File "/tmp/pyda.py", line 18, in <module>
    print(Union[Session,Activity](**{'name': 'session_id1', 'schemaKey': 'Session'}))
  File "/home/yoh/proj/misc/pydantic/pydantic/main.py", line 394, in __init__
    raise validation_error
pydantic.error_wrappers.ValidationError: 1 validation error for Activity
schemaKey
  unexpected value; permitted: 'Activity' (type=value_error.const; given=Session; permitted=('Activity',))

```
</details>

<details>
<summary>but fails to run with system wide python 3.8: `TypeError: Cannot instantiate typing.Union`, so I guess I am missing smth needed?</summary> 

```shell
$> python3.8 /tmp/pyda.py    
  File "/tmp/pyda.py", line 18, in <module>
    print(Union[Session,Activity](**{'name': 'session_id1', 'schemaKey': 'Session'}))
  File "/usr/lib/python3.8/typing.py", line 729, in __call__
    result = self.__origin__(*args, **kwargs)
  File "/usr/lib/python3.8/typing.py", line 339, in __call__
    raise TypeError(f"Cannot instantiate {self!r}")
TypeError: Cannot instantiate typing.Union
```
</details>

satra:
i'm not sure you can using `typing` like this

`print(Union[Session,Activity](**{'name': 'session_id1', 'schemaKey': 'Session'}))`

```python

class Test(BaseModel):
     field1: Union[Session,Activity]

Test(field1={'name': 'session_id1', 'schemaKey': 'Session'}) 
```

example
```python
In [14]: Test(field1={'name': 'session_id1', 'schemaKey': 'Activity'}).field1
Out[14]: Activity(name='session_id1', schemaKey='Activity')

In [15]: Test(field1={'name': 'session_id1', 'schemaKey': 'Session'}).field1
Out[15]: Session(name='session_id1', schemaKey='Session')
```

.github/workflows/test.yml
```diff
@@ -20,7 +20,6 @@ jobs:
           - ubuntu-18.04
           - macos-latest
         python:
-          - 3.6  # will also be used to test with _DANDI_LOG_GIRDER set
```
jwodder, line 23: The workflow step for setting `DANDI_LOG_GIRDER` should be updated to trigger on 3.7 instead of 3.6.

yarikoptic: I have pushed update for all the 3.6 --> 3.7

dandi/metadata.py
```diff
@@ -254,6 +254,16 @@ def extract_wasAttributedTo(metadata):
...
+    if all(v is None for k, v in wat.dict().items() if k != "schemaKey"):
+        return []
+    else:
+        return [wat]
+
+
+def extract_wasGeneratedBy(metadata):
+    wat = extract_model(
+        models.Session, metadata, identifier=None, name=metadata.get("session_id")
+    )
```
yarikoptic, line 266: @jwodder would you please RF this code to avoid duplication?  it seems we need some helper like `extract_simple(field, metadata)` which could be used as a `partial` instead of 3 identical copies now

dandi/metadata.py
yarikoptic: suggest the change? ;)

dandi/metadata.py
jwodder: I factored out the conversion to a list, but the different treatment of the fields in `extract_wasGeneratedBy()` prevents factoring down to a `extra_model_list(modelcls, identifier_field, metadata)` function.

yarikoptic: but would work as `extract_model_list(modelcls, identifier_field, metadata, **kw)` where you would pass e.g. `identifier=None` through via `kw` (and as a part of the `partial`) right?

jwodder: Code compressed further.

yarikoptic:
ok, let's proceed and see where it takes us ;) Thank you @satra and @jwodder ... and here we go into the bright future without Python 3.6 ;)

PR #394, opened 2021-02-17
Use PUT endpoint to replace pre-existing assets on upload

### LINKED ISSUE #393: dandi-api: just reuse (modify) an existing asset when uploading a new version
@dchiquito in slack:

I noticed some DELETE calls to assets, presumably so that they could be recreated. You should be able to PUT an asset with a different sha256 to change which blob it points to.

Indeed in current implementation (where new asset UUIDs are minted upon release) it should be ok to adjust an existing asset instead of creating a new one.  It would allow for some kind of "tracking the life" of an asset if we ever get to annotate for that

jwodder:
Closes #393.


yarikoptic:

PR #397, opened 2021-02-18
Use fscacher

### LINKED ISSUE #396: since fscacher is already on pypi: strip original code and use fscacher package
so we do not spread effort in fixinup here and there, let's have fscacher 0.1 released, and use it in dandi code and fix anything needed in fscacher in fscacher

jwodder:
Closes #396.


yarikoptic:
Looks good to me and CI doesn't complain -- so all is good, thank you!

PR #399, opened 2021-02-18
Add CI run with dev version of pynwb

### LINKED ISSUE #398: add a test run which would test against dev of hdmf and pynwb
later we will also add some testing on sample .nwb files with various nwb extensions of direct interest to DANDI (relates to https://github.com/NeurodataWithoutBorders/pynwb/issues/1338).  Meanwhile, although with a fear of having not immediately treatable failed CI runs for PRs, I think should benefit from having at least a single run (python and OS of personal preference) which would ensure that our state/changes are still ok with the direction where hdmf/pynwb is going.

jwodder:
Closes #398.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,7 +28,7 @@ jobs:
           - false
         include:
           - os: ubuntu-18.04
-            python: 3.6
+            python: 3.7
             dev_deps: true
         exclude:
           # Temporarily disabled due to h5py/hdf5 dependency issue
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,6 +24,12 @@ jobs:
           - 3.7
           - 3.8
           - 3.9
+        dev_deps:
+          - false
+        include:
+          - os: ubuntu-18.04
+            python: 3.6
+            dev_deps: true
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -24,6 +24,12 @@ jobs:
           - 3.7
           - 3.8
           - 3.9
+        dev_deps:
+          - false
+        include:
+          - os: ubuntu-18.04
+            python: 3.6
```
yarikoptic, line 31:
```suggestion
            python: 3.7
```
since most likely we would jump to 3.7 as minimal in #392

yarikoptic (approved):
I will accept my recommendation

FTR: submodule with schema does get installed by pip
```
Run pip install git+https://github.com/NeurodataWithoutBorders/pynwb
Collecting git+https://github.com/NeurodataWithoutBorders/pynwb
  Running command git clone -q https://github.com/NeurodataWithoutBorders/pynwb /tmp/pip-req-build-uc921st4
  Cloning https://github.com/NeurodataWithoutBorders/pynwb to /tmp/pip-req-build-uc921st4
  Running command git submodule update --init --recursive -q
```

PR #401, opened 2021-02-18
Disabling logging to stderr while using pyout

### LINKED ISSUE #358: pyout + logging: make them co-exist
NB A rough idea ATM

ATM logging interfers with pyout (used for upload/download/ls) output: log line would be printed to stderr but then overriden by pyout on display thus only screwing up pyout rendering since a new line added, and otherwise not visible to the user.  There is really no ideal solution to make them co-exist peacefully. E.g. even if we use (could not find ATM) pyout's ability for "write through" - those log lines would appear above rendering, and likely off the screen already.
But we already store all logs into a temp file which we announce to the user.  So I think a viable way forward would be
- when using pyout, define our logging handler to be only the file one + "pyout handler"
- when logging during pyout, "pyout handler" should inform pyout display that a log line was issued, and we add a counter (ideally actually separating by different log levels, e.g. ERRORS:1, WARNINGS:2 etc) to LOG column for that item
- tricky part is association between log lines and pyout records, but I think could be worked out

jwodder:
A partial solution to #358

### DIFF
--- a/dandi/cli/formatter.py
+++ b/dandi/cli/formatter.py
@@ -1,8 +1,6 @@
 import datetime
 import sys
 
-import pyout
-
 from .. import get_logger
 from ..support import pyout as pyouts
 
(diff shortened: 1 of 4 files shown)


yarikoptic:
Didn't test but logic-wise looks good, thank you @jwodder  - let's take it to the battle fields for testing ;-)  I would just not mark the original issue closed with this one, I think this is a temporary measure and we might improve upon later on.

PR #409, opened 2021-02-22
Set Content-MD5 header when uploading asset parts

### LINKED ISSUE #407: dandi-api: for upload to S3 add integrity "marker"
@satra reminded about this feature of uploads to S3: https://aws.amazon.com/premiumsupport/knowledge-center/data-integrity-s3/ to provide base64 encoded md5sum.  If possible, we better provide it as well, although it would require now adding computation of md5 in addition to sha256 used by the archive - Digester can do that, so at least we will not re-read the file twice.

jwodder:
Closes #407.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -1,4 +1,6 @@
+from base64 import b64encode
 from contextlib import contextmanager
+from hashlib import md5
 import os.path
 from pathlib import Path
 from time import sleep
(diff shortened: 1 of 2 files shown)


satra:
it may be worthwhile adding a test that fails to upload because the wrong md5 was added but passes when the correct md5 is added

dandi/dandiapi.py
```diff
@@ -437,7 +447,13 @@ def iter_upload(self, dandiset_id, version_id, asset_metadata, filepath):
                             part["part_number"],
                             part["size"],
                         )
-                        r = storage.put(part["upload_url"], data=chunk, json_resp=False)
+                        chunk_md5 = md5(chunk).digest()
```
yarikoptic, line 450: note to myself: so we need it per chunk, thus cannot (ab)use Digester to get both md5 and sha256 at the same time.  Here at least chunk will already be in memory so should not be much of a burden.

yarikoptic:
concur with @satra, so might be worth mocking out `md5` digest computation for some sample upload and see if it fails -- otherwise we might be believing blindly ;)

jwodder:
@yarikoptic @satra Test with wrong md5 header added.

yarikoptic:
passed for where it most matters, merging ;) thank you @jwodder !

PR #413, opened 2021-02-22
Capture all dandi log messages when testing

jwodder:
To make debugging failed tests easier.


yarikoptic:
windows freaked out but I think it is unrelated, let's proceed

Issue #405, opened 2021-02-22
Make validation of the upload query server/update less frequently + improve yields to pyout


yarikoptic:
ATM we are bombarding API server with validation requests too frequently.  E.g. (although this is parallel uploads):
```
2021-02-22T09:53:38-0500 [DEBUG   ] dandi Waiting for server-side validation to complete
2021-02-22T09:53:38-0500 [DEBUG   ] dandi GET https://api.dandiarchive.org/api/uploads/validations/c2c71c26b433c6cd317822d5bf6484f9fa0dce07293a9eecc46f3dfafbb84125/
2021-02-22T09:53:38-0500 [DEBUG   ] dandi Response: 200
2021-02-22T09:53:38-0500 [DEBUG   ] dandi Waiting for server-side validation to complete
2021-02-22T09:53:38-0500 [DEBUG   ] dandi GET https://api.dandiarchive.org/api/uploads/validations/d32f2cc328dedec938c2a28c734a9ec4f3cc2f347b13fccee3bd5d5ede1cf35c/
2021-02-22T09:53:38-0500 [DEBUG   ] dandi Response: 200
2021-02-22T09:53:38-0500 [DEBUG   ] dandi Response: 200
2021-02-22T09:53:38-0500 [DEBUG   ] dandi Waiting for server-side validation to complete
2021-02-22T09:53:38-0500 [DEBUG   ] dandi GET https://api.dandiarchive.org/api/uploads/validations/81a6e764bda0c0cfa4b96d200b103d347e7d7c7c4ee82665b5b9784f8d2084cf/
2021-02-22T09:53:38-0500 [DEBUG   ] dandi Waiting for server-side validation to complete
2021-02-22T09:53:38-0500 [DEBUG   ] dandi GET https://api.dandiarchive.org/api/uploads/validations/f07681c40d0aae99f82836605115290927a25971fcada11b6d502e0e48a73f13/
2021-02-22T09:53:39-0500 [DEBUG   ] dandi Response: 200
2021-02-22T09:53:39-0500 [DEBUG   ] dandi Waiting for server-side validation to complete
2021-02-22T09:53:39-0500 [DEBUG   ] dandi GET https://api.dandiarchive.org/api/uploads/validations/ee9eac72ff68e02f413c1e97445b712d20b9bc692c1cc7c1c4b0433b857e59b1/
2021-02-22T09:53:39-0500 [DEBUG   ] dandi Response: 200
[...]

PR #417, opened 2021-02-22
upload(): Only yield first "validating" status to pyout

jwodder:
Completes #405.

### DIFF
--- a/dandi/upload.py
+++ b/dandi/upload.py
@@ -799,10 +799,18 @@ def _new_upload(
             # Upload file
             #
             yield {"status": "uploading"}
+            validating = False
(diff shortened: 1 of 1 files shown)


yarikoptic:
I thought it is not needed since we yield at this point: https://github.com/dandi/dandi-cli/blob/master/dandi/upload.py#L258

jwodder:
@yarikoptic That's for client-side validation before upload.  This status is yielded while waiting for server-side validation after upload.

yarikoptic:
d'oh -- right.  We should come up with some different term for the things. Ok, let's proceed then

yarikoptic:
Thank you

PR #419, opened 2021-02-23
Revert PR #409 (Content-MD5 header)

jwodder:
See [this comment](https://github.com/dandi/dandi-cli/issues/414#issuecomment-784347724) and dchiquito's response


yarikoptic:
eh, ok -- so it seems we cannot do this additional upload integrity/validation checking without informing the api-server about those per each chunk since it is the one which is to mint those urls.

PR #420, opened 2021-02-24
Distinguish between pre- and post-validation when uploading

jwodder:
This PR changes the names of the "validation" statuses when uploading to "pre-validation" (done client-side before uploading) and "post-validation" (done server-side after uploading) so that one can more accurately tell what state an upload is in.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -490,7 +490,7 @@ class DandiAPIClient(RESTFullAPIClient):
                     )
                 break
             before_time = monotonic()
-            yield {"status": "validating"}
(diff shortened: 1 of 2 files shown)


yarikoptic:
I even wonder if we should make it even more to the point
- metadata check
- upload check

Using check instead of validation since shorter. Wdyt?

jwodder:
@yarikoptic I feel like "upload check" doesn't really convey much meaning.  Also, "metadata check" is the same length as "pre-validation".

yarikoptic:
I meant that "metadata check" would be shorter than "metadata validation".  But ok -- let's proceed with pre/post.

PR #422, opened 2021-02-25
Assorted code cleanup

jwodder:
This is a collection of minor edits to the code that aren't really worth their own PRs.  See each commit for descriptions of the changes.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -21,8 +21,6 @@ description = Command line client for interaction with DANDI archive elements
 long_description = file:README.md
 long_description_content_type = text/markdown; charset=UTF-8
 platforms = OS Independent
-provides =
-    dandi
 project_urls =
     Source Code = https://github.com/dandi/dandi-cli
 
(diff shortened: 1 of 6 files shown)


setup.cfg
```diff
@@ -21,8 +21,6 @@ description = Command line client for interaction with DANDI archive elements
 long_description = file:README.md
 long_description_content_type = text/markdown; charset=UTF-8
 platforms = OS Independent
-provides =
-    dandi
```
yarikoptic, line 25: just curious: where it potentially could be used? I found no description on https://packaging.python.org/guides/distributing-packages-using-setuptools

yarikoptic:
Looks good. Just curious about "provides". Let's proceed

setup.cfg
jwodder: Nothing in the ecosystem currently uses this field.  There are occasional discussions around using "Provides" to implement support for things like virtual packages, but nothing ever comes of it.

PR #423, opened 2021-02-25
Use iter_content() instead of raw.stream()

jwodder:
I believe that `iter_content()` is the preferred way to iterate over a streaming `requests.Response`.  Note that this comes with a slight change in behavior, as `iter_content()` calls `raw.stream()` with `decode_content=True`, which causes Content-Encodings to be decoded, which I believe is the correct thing to do.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -330,7 +330,7 @@ class DandiAPIClient(RESTFullAPIClient):
             # if result.status_code not in (200, 201):
             result.raise_for_status()
 
-            for chunk in result.raw.stream(chunk_size, decode_content=False):
+            for chunk in result.iter_content(chunk_size=chunk_size):
                 if chunk:  # could be some "keep alive"?
                     yield chunk
             lgr.info("Asset %s successfully downloaded", uuid)


yarikoptic:
for a download of files content "as is" I do not see why we should decode anything.

I do remember that some servers allow for "transparent" compression of the content , so contentencoding would come something like `gzip`, but when I last tried it (which was awhile) I just remember that it was not possible to tell apart a file which was encoded only for transfer from what was originally a gzip'ed file... I do not remember if that needs to be somehow pre-negotiated by the client to trigger such on the fly transfer encoding...

What contentencoding's do we receive from the API server?

overall -- I am not yet confident about such a change, in particular because there is no known issue it would be solving.

jwodder:
@yarikoptic Content encoding is something that's applied by HTTP; it's not a property of the file being transferred.  If a file being downloaded has a Content-Encoding of gzip, it means that the server has gzipped the actual file for the purposes of transfer and you need to ungzip it to get the same bytes as on the server's filesystem.  [See Wikipedia for more information.](https://en.wikipedia.org/wiki/HTTP_compression)

> I do not remember if that needs to be somehow pre-negotiated by the client to trigger such on the fly transfer encoding...

By default, `requests` sends "Accept-Encoding: gzip, deflate" in its requests, so any server we connect to could potentially use either of those encodings.

> What contentencoding's do we receive from the API server?

I am not aware of any that are actually used at the moment.  This is more about future-proofing the code by making it do the right thing ahead of time.

yarikoptic:
Sounds good.  Not sure what problems I had that while back, but sounds like indeed this should be the way forward.  Since we do AFAIK (but for a paranoid me: https://github.com/dandi/dandi-cli/issues/427) verify the checksum for downloads (https://github.com/dandi/dandi-cli/blob/master/dandi/download.py#L446) we should at least catch if something goes wrong.

PR #424, opened 2021-02-25
Report dandi version in User-Agent header

jwodder:
This PR sets a User-Agent header when communicating with the new API that contains the version of dandi in use.


yarikoptic:
GREAT IDEA! Thank you @jwodder !

PR #428, opened 2021-02-25
Error if sha256 digest is missing from asset being downloaded

### LINKED ISSUE #427: dandi-api: download - make verification of the checksum "guaranteed"
ATM in the code we verify download digests only if digests are provided to https://github.com/dandi/dandi-cli/blob/master/dandi/download.py#L446 .   I think we need to remove "if" for dandi-api instances - there always must be a digest.

jwodder:
Closes #427.

### DIFF
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -170,8 +170,7 @@ def download_generator(
                     asset["uuid"],
                 )
                 if "sha256" not in asset:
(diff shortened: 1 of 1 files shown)


yarikoptic:
thanks! Let's see if we hit it somehow

PR #429, opened 2021-02-27
Set DJANGO_DANDI_SCHEMA_VERSION in docker-compose.yml

jwodder:
As of https://github.com/dandi/dandi-api/pull/83, the dandi-api Django setup requires the `DJANGO_DANDI_SCHEMA_VERSION` environment variable to be set in order to run; without it, our tests fail.


yarikoptic:
Great, thank you

PR #435, opened 2021-03-01
schema: minor spotted typo fixes

yarikoptic:


yarikoptic:
something is up with the macos-latest... not even a log.  Otherwise I do not think this PR is worth much of deliberation ;)

PR #437, opened 2021-03-01
Retry requests on ConnectionErrors

jwodder:
I've encountered `requests.exceptions.ConnectionError: HTTPSConnectionPool(host='api.dandiarchive.org', port=443): Pool is closed.` a couple times while testing uploads to the new API.  This PR should take care of that.


yarikoptic:
Looks good, thank you!

Issue #402, opened 2021-02-19
have we reached the ultimate limit of abuse of github actions?


yarikoptic:
https://github.com/dandi/dandi-cli/runs/1931672765?check_suite_focus=true

`Test populate_dandiset_yaml.py` on master errorred out with 

```
Run actions/checkout@v2
2021-02-19T00:33:24.2606301Z ##[group]Determining the default branch
2021-02-19T00:33:24.2611932Z Retrieving the default branch name
2021-02-19T00:33:24.6670616Z API rate limit exceeded for installation ID 3038186.
2021-02-19T00:33:24.6671799Z Waiting 10 seconds before trying again
2021-02-19T00:33:34.6690964Z Retrieving the default branch name
2021-02-19T00:33:35.0126023Z API rate limit exceeded for installation ID 3038186.
2021-02-19T00:33:35.0126735Z Waiting 15 seconds before trying again
2021-02-19T00:33:50.0283471Z Retrieving the default branch name
2021-02-19T00:33:50.3645753Z ##[error]API rate limit exceeded for installation ID 3038186.
2021-02-19T00:33:50.3799605Z Post job cleanup.
```

WDYT @jwodder ?

satra:
for the moment closed by #403 - but will have to monitor and see if it pops up again - especially when we are testing multiple PRs and commits.

yarikoptic:
seems like #403 was not enough 

https://github.com/dandi/dandi-cli/pull/415/checks?check_run_id=1955110878

```
  Retrieving the default branch name
  API rate limit exceeded for installation ID 3038186.
  Waiting 14 seconds before trying again
  Retrieving the default branch name
  API rate limit exceeded for installation ID 3038186.
  Waiting 11 seconds before trying again
  Retrieving the default branch name
  Error: API rate limit exceeded for installation ID 3038186.
```
I guess I should mirror them on datasets.datalad.org and we could clone from there -- probably less efficient but oh well.

satra:
can we do a shallow retrieval? also perhaps another potential reason to create a single assets.yaml (this is only going to grow).

yarikoptic:
# of files should not matter much (unless they count per object and do not repack).  I think it is more about clone attempts in general (although I could be wrong).

yarikoptic:
Let's try to switch to a regular/direct `git clone` instead of  `actions/checkout@v2` action  @jwodder.

yarikoptic:
Let's again assume that we have resolved it (in  #443) ;)

PR #443, opened 2021-03-01
Check out dandi/dandi-api-datasets with direct `git clone`

jwodder:
Closes (?) #402.


yarikoptic:
relevant run succeeded, so it is more likely to help than not.  Let's proceed. Thanks @jwodder !

PR #444, opened 2021-03-01
Simplify release workflow

jwodder:
This eliminates the use of `auto version` (a failure in which led to a failed & unsolicited attempt at uploading to PyPI) in favor of triggering the build & upload via auto's `exec` plugin and its `afterRelease` hook.

### DIFF
--- a/.autorc
+++ b/.autorc
@@ -3,5 +3,13 @@
     "baseBranch": "master",
     "author": "auto <auto@nil>",
     "noVersionPrefix": true,
-    "plugins": ["git-tag"]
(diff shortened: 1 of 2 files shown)


yarikoptic:
not verse in `auto` enough to review.  If works (or you think @jwodder it should work) -- good with me

jwodder:
@yarikoptic This setup has already been used for releases of datalad-installer and fscacher, and I've tested it in a test repository.  It works.

yarikoptic:
Thank you @jwodder !

PR #445, opened 2021-03-01
Add --new-schema option to `ls` command for converting output

jwodder:


yarikoptic:
What did you decide not to go with --version as of validate?

I know that we have only one version here ATM, but I wonder if for alignment with validate and for future proofing we might want to allow for explicit version specification

jwodder:
@yarikoptic Because the conversion functions don't take any `schema_version` arguments.  Would you like them to?

yarikoptic:
sure, let's do that for uniformity.  Later we will change default version to the current supported... I guess majority of code for version handling etc is already in `validate_dandiset_yaml`

jwodder:
@yarikoptic PR edited.

yarikoptic:
Having a test (and we have already sales of nwb files with target converted metadata) would have been nice, but let's proceed

Issue #249, opened 2020-09-29
Python API: design


yarikoptic:
Development prior this moment was targeting the command line client, and making it informative and "nice looking" to the users.  So some functionality (e.g. `ls`) still exists only as a "click" interface without underlying Python function.

Some, more recently added or refactored, interfaces did get proper Python functions (e.g. `download`) but they seems
  to provide neither "the best" (if any) return value Python coders would expect to receive, nor visual feedback to
    the users using those functions "interactive" in Python shell or more likely - jupyter shell or notebook.  E.g. in    Jupyter notebook ATM `download` reports only INFO message that it is about to traverse the dandiset, and then no visual output until it is done (#246), and `download` returns no value.
`validate` - is a generator which returns some (cryptic) records. 
`ls` - I think lacks any "pure Python" interface, and I guess others are not much better.

So it is time to harmonize Python interfaces and display produced by dandi commands.  Overall I think we should adopt
 DataLad's approach(es) which I mention below

### Functions vs generators

`download`, `upload`, `validate`, `ls`, `organize`: I think those should become/stay **generators** since they are
 intended to be ran for notable amount of time, so we better "update" outside code as soon as we know.

`register`: could be just a function since it just needs to return a single value.

### Returned values

We should make "returned" (yielded) values sensible for Python coding, in particular by default (i.e. without tuning  any options we might introduce).  In DataLad majority of commands now operate by returning "result records" and there is quite elaborate wrappers and invocation kwarg options which tune up returned values and behavior (datalad's `--on-failure {ignore,continue,stop}` e.g. to stop iterating as soon as first problem happened or not, etc) - it makes it flexible to compose multiple "interfaces" etc (but brings pains to debug at times; but here we will not have deep nesting of them).  In Python API DataLad's interfaces have 

```
result_xfm : {'datasets', 'successdatasets-or-none', 'paths', 'relpaths', 'metadata'} or callable or None, optional
  if given, each to-be-returned result status dictionary is passed to
  this callable, and its return value becomes the result instead. This
  is different from `result_filter`, as it can perform arbitrary
  transformation of the result value. This is mostly useful for top-
  level command invocations that need to provide the results in a
  particular format. Instead of a callable, a label for a pre-crafted
  result transformation can be given. [Default: None]
return_type : {'generator', 'list', 'item-or-list'}, optional
  return value behavior switch. If 'item-or-list' a single value is
  returned instead of a one-item return value list, or a list in case
[...]

PR #447, opened 2021-03-03
Add new API design doc

jwodder:
prior discussion in https://github.com/dandi/dandi-cli/issues/249

### DIFF
--- a/doc/design/python-api-1.md
+++ b/doc/design/python-api-1.md
@@ -184,6 +184,7 @@ Designs for an improved Python API
     * Give `CommonModel` a `validate()` method that just does `type(self)(**self.dict())`
     * Add a `UserInputError` exception class subclassing `ValueError`, use it for all errors caused by bad user input, and make `map_to_click_exceptions` only remap it and nothing else to `click.UsageError`
     * Make the return value of `parse_dandi_url()` a structured class
+        * Give it `get_dandiset(client: DandiAPIClient) -> RemoteDandiset` and `get_assets(client: DandiAPIClient) -> Iterator[RemoteAsset]` methods
     * Rename the `download()` function to `download_url()`?
     * Add a `process_uploads(Iterable[Iterator[UploadProgressDict]], show_progress=True) -> List[RemoteAsset]` function that takes a collection of upload iterators and consumes them in parallel, using pyout if `show_progress` is true
     * Add a `process_downloads(Iterable[Iterator[DownloadProgressDict]], show_progress=True) -> ???` function that takes a collection of download iterators and consumes them in parallel, using pyout if `show_progress` is true
--- a/doc/design/python-api-1.md
+++ b/doc/design/python-api-1.md
@@ -70,10 +70,10 @@ Designs for an improved Python API
         * `get_assets_under_path(path: str) -> Iterator[RemoteAsset]`
         * `get_asset_by_path(path: str) -> Optional[RemoteAsset]`
         * `get_asset_by_uuid(uuid: str) -> Optional[RemoteAsset]` — Returns `None` if the given asset does not exist
-        * `download(target_dir: Union[str, Path], paths: Optional[List[str]] = None, show_progress=True, existing="error", get_metadata=True) -> ???`
+        * `download(target_dir: Union[str, Path], paths: Optional[List[str]] = None, show_progress=True, existing="error", get_metadata: Optional[bool] = None) -> ???`
             * Use this to replace the `download()` function?
             * The elements of `paths` are interpreted the same way as the argument to `get_assets_under_path()`
-            * `get_metadata` controls whether to create a `dandiset.yaml` file; should it default to `False` when `paths` is not `None`?
+            * `get_metadata` controls whether to create a `dandiset.yaml` file.  When `None`, it defaults to `True` if `paths` is `None` or empty, `False` otherwise.
             * TO DO: Add an argument for controlling whether to create a directory under `target_dir` with the same name as the Dandiset ID?
 
     * Methods of the `DraftDandiset` class (a subclass of `RemoteDandiset` used only for mutable draft versions):
--- a/doc/design/python-api-1.md
+++ b/doc/design/python-api-1.md
@@ -1,0 +1,185 @@
+Designs for an improved Python API
+==================================
+
+* New base classes:
+    * `Dandiset`
+        * `identifier: str`
+        * `created: datetime`
+        * `modified: datetime`
+        * `version: Optional[Version]`
+        * `get_all_assets() -> Iterator[Asset]`
+        * `get_assets_under_path(path: str) -> Iterator[Asset]` — Returns all assets whose paths are either under the given folder or equal the given filepath
+        * `get_asset_by_path(path: str) -> Optional[Asset]` — Finds the unique asset with the given exact path; returns `None` if no such asset
+        * `get_metadata() -> DandiMeta`
+   * `Asset`:
+        * `path: str`
+        * `sha256: str` — Should this only be on `RemoteAsset`?
+        * `size: int`
+        * `created: datetime` — Should this only be on `RemoteAsset`?
+        * `modified: datetime` — Should this only be on `RemoteAsset`?
+        * `get_metadata() -> BareAssetMeta`
+
+* `dandi.dandiapi`:
+    * Upon creating a new `RESTFullAPIClient` instance, a new session object should automatically be created and used for all of the client's requests; the `session()` method should thus be eliminated
+        * If we need to support user-supplied session objects, those should be passed in to the constructor
+    * It may or may not be a good idea for the client to be usable as a context manager that closes its session on exit.
+        * Con: Closing the session would affect `RemoteDandiset` etc. instances for the client (see below), making them unusable.
+    * Rename the `send_request()` method to `request()` to parallel `requests.Session.request()`
+    * Rename the `parameters` argument to the HTTP methods to `params` to match `requests`
+    * The `api_url` argument to `DandiAPIClient`'s constructor should default to the URL for the official Dandiarchive instance (which should also be available as a constant)
+    * Give `DandiAPIClient` a classmethod `for_dandi_instance(cls, instance: Union[str, dandi_instance], token=None, authenticate=True)` (Rethink name) that takes either the name of a Dandi instance or an actual `dandi_instance`, fetches a token from the environment, keychain, and/or user if one is not given and `authenticate` is true, and returns a `DandiAPIClient` instance
+    * Most interaction with API resources should be via new `RemoteDandiset`, `Version`, and `RemoteAsset` classes and their methods
+        * These classes (except `Version`?) will all contain a private `_client` attribute referring back to the associated `DandiAPIClient` instance.
+    * Changes to `DandiAPIClient` methods:
+        * Add a `get_dandiset(dandiset_id: Union[str, RemoteDandiset], version_id=None) -> RemoteDandiset` method
+            * **To discuss:** Should the default `version` be "draft" or the most recent version?  Either way, there should be another method that explicitly gets the other option.
+            * **To discuss:** Should this take a `lazy=False` parameter that, if true, constructs a `RemoteDandiset` containing only an identifier, without making any requests at the time of creation?  This would allow calling a `RemoteDandiset`'s methods without having to make a request for the Dandiset's data beforehand.  If & when any non-identifiers attributes of a lazy `RemoteDandiset` are requested, a request is made at that point and the results cached.
+        * Add a `get_all_dandisets() -> Iterator[RemoteDandiset]` method
+            * **To discuss:** What should these objects' `Version`s be?
+        * Retype `create_dandiset` to `create_dandiset(name: str, metadata: DandiMeta) -> RemoteDandiset`
+        * Add a `register_dandiset(dirpath: Union[str, Path], name: str, description: str) -> LocalDandiset` method that, in addition to calling `create_dandiset()`, also creates a `dandiset.yaml` file in `dirpath`?
+        * Add a `get_current_user() -> User` method
+        * Add a `search_users(username: str) -> Iterator[User]` method
+        * Add an `upload(dandiset: LocalDandiset, paths: Optional[List[str]] = None, show_progress=True, existing="refesh", validation="require") -> List[RemoteAsset]` method
+            * Use this to replace the `upload()` function
+            * The elements of `paths` are interpreted the same way as the argument to `get_assets_under_path()`
+        * The remaining methods should be either deleted or made private.
+        * **To bikeshed:** Should the Dandiset and version arguments to these methods be named `dandiset` & `version` or `dandiset_id` & `version_id`?
+    * Where a method can take an ID as either a string or an object, it is permitted for the object to belong to a different client.
+
+    * Attributes & methods of the `RemoteDandiset` class, a subclass of the new `Dandiset` base class, representing a combination of the API's dandiset and version resources:
+        * `identifier: str`
+        * `created: datetime`
+        * `modified: datetime`
+        * `version: Version`
+        * `most_recent_version: Version` ?
+        * `get_version(version_id: Union[str, Version]) -> Version` ?
+        * `get_all_versions() -> Iterator[Version]`
+        * `for_version(version_id: Union[str, Version]) -> RemoteDandiset`
+            * When `version_id == "draft"`, this is the same as `for_draft_version()`
+        * `for_draft_version() -> DraftDandiset`
+        * `delete() -> None`
+        * `get_users() -> Iterator[User]`
+        * `set_users(users: List[User]) -> None`
+            * Should this method also accept strings giving the usernames directly?
+        * `get_metadata() -> DandiMeta`
+        * `publish() -> Version` (Should this be moved to `DraftDandiset`?)
+        * `get_all_assets() -> Iterator[RemoteAsset]`
+        * `get_assets_under_path(path: str) -> Iterator[RemoteAsset]`
+        * `get_asset_by_path(path: str) -> Optional[RemoteAsset]`
+        * `get_asset_by_uuid(uuid: str) -> Optional[RemoteAsset]` — Returns `None` if the given asset does not exist
+        * `download(target_dir: Union[str, Path], paths: Optional[List[str]] = None, show_progress=True, existing="error", get_metadata=True) -> ???`
+            * Use this to replace the `download()` function?
+            * The elements of `paths` are interpreted the same way as the argument to `get_assets_under_path()`
+            * `get_metadata` controls whether to create a `dandiset.yaml` file; should it default to `False` when `paths` is not `None`?
+            * TO DO: Add an argument for controlling whether to create a directory under `target_dir` with the same name as the Dandiset ID?
+
+    * Methods of the `DraftDandiset` class (a subclass of `RemoteDandiset` used only for mutable draft versions):
+        * `set_metadata(metadata: DandiMeta) -> DraftDandiset` — returns the new, modified object
+        * Methods for uploading an individual asset:
+            * `upload_file(filepath: Union[str, Path], metadata: BareAssetMeta, show_progress=True, existing="refesh", validation="require") -> RemoteAsset`
+            * `iter_upload_file(filepath: Union[str, Path], metadata: BareAssetMeta, existing="refesh", validation="require") -> Iterator[UploadProgressDict]`
+            * `upload_asset(asset: LocalAsset, show_progress=True, existing="refesh", validation="require") -> RemoteAsset` ?
+            * `iter_upload_asset(asset: LocalAsset, existing="refesh", validation="require") -> Iterator[UploadProgressDict]` ?
+        * Methods for uploading a directory/collection of assets:
+            * `upload_assets(assets: Iterable[LocalAsset], show_progress=True, existing="refesh", validation="require") -> List[RemoteAsset]`
+
+    * Attributes & methods of the `Version` class:
+        * `identifier: str`
+        * `created: datetime`
+        * `modified: datetime`
+        * `name: str`
+        * `asset_count: int`
+        * `size: int`
+
+    * Attributes & methods of the `RemoteAsset` class (a subclass of `Asset`):
+        * `uuid: str`
+        * `path: str`
+        * `sha256: str`
+        * `size: int`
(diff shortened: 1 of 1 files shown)


doc/design/python-api-1.md
```diff
@@ -0,0 +1,187 @@
...
+    * The `api_url` argument to `DandiAPIClient`'s constructor should default to the URL for the official Dandiarchive instance (which should also be available as a constant)
+    * Give `DandiAPIClient` a classmethod `for_dandi_instance(cls, instance: Union[str, dandi_instance], token=None, authenticate=True)` (Rethink name) that takes either the name of a Dandi instance or an actual `dandi_instance`, fetches a token from the environment, keychain, and/or user if one is not given and `authenticate` is true, and returns a `DandiAPIClient` instance
+    * Most interaction with API resources should be via new `RemoteDandiset`, `Version`, and `RemoteAsset` classes and their methods
+        * These classes (except `Version`?) will all contain a private `_client` attribute referring back to the associated `DandiAPIClient` instance.
+    * Changes to `DandiAPIClient` methods:
+        * Add a `get_dandiset(dandiset_id: Union[str, RemoteDandiset], version_id=None) -> RemoteDandiset` method
+            * **To discuss:** Should the default `version` be "draft" or the most recent version?  Either way, there should be another method that explicitly gets the other option.
```
yarikoptic, line 35: Since `draft` might disappear (e.g. we "archive" a dandiset, preventing any changes and thus removing "draft"), I think we should just stay consistent with what API's `/dandisets/{dandiset__pk}/` returns in `most_recent_version` -- might be a released one or a draft IIRC.  So the default here could stay `None`.  But I wonder if we should complicate with allowing for special handling of `"latest_release"`?

doc/design/python-api-1.md
```diff
@@ -0,0 +1,187 @@
...
+            * **To discuss:** Should this take a `lazy=False` parameter that, if true, constructs a `RemoteDandiset` containing only an identifier, without making any requests at the time of creation?  This would allow calling a `RemoteDandiset`'s methods without having to make a request for the Dandiset's data beforehand.  If & when any non-identifiers attributes of a lazy `RemoteDandiset` are requested, a request is made at that point and the results cached.
```
yarikoptic, line 36: if `RemoteDandiset` constructor does not demand talking to the server, shouldn't it be "lazy" to start with and have properties lazily populated?  (or may be only if no version_id is given)

doc/design/python-api-1.md
```diff
@@ -0,0 +1,187 @@
...
+        * Add a `get_all_dandisets() -> Iterator[RemoteDandiset]` method
+            * **To discuss:** What should these objects' `Version`s be?
```
yarikoptic, line 38: I think boils down to the same - are `RemoteDandisets` lazy at the constructor - do they require communication to the server or not...

doc/design/python-api-1.md
```diff
@@ -0,0 +1,187 @@
...
+        * Retype `create_dandiset` to `create_dandiset(name: str, metadata: DandiMeta) -> RemoteDandiset`
```
yarikoptic, line 39: I think we should remove explicit `name` from the signature.  Should be present in metadata anyways.
It does make it less user-friendly since hard to figure out from the function signature though :-/

A possible choice could be to make it `create_dandiset(*metadata)` and mint the docstring listing the fields based on `DandiMeta` from which we know what is mandatory and what is optional.  I do not know though if it is possible to autogenerate the signature though for extra user/IDE's convenience.
But that would remove (or complicate) possibility to provide extra options which aren't metadata, and not sure how python typing would take it.

So may be just making it `create_dandiset(metadata: DandiMeta)` and minting docstring with a nice description of fields in `metadata` would be the reasonable compromise.

doc/design/python-api-1.md
```diff
@@ -0,0 +1,187 @@
...
+        * Add a `register_dandiset(dirpath: Union[str, Path], name: str, description: str) -> LocalDandiset` method that, in addition to calling `create_dandiset()`, also creates a `dandiset.yaml` file in `dirpath`?
```
yarikoptic, line 40: let's name it may be `create_local_dandiset` since otherwise difference between `create` and `register` is not clear.

doc/design/python-api-1.md
```diff
@@ -0,0 +1,187 @@
...
+==================================
+
+* New base classes:
+    * `Dandiset`
+        * `identifier: str`
+        * `created: datetime`
+        * `modified: datetime`
+        * `version: Optional[Version]`
+        * `get_all_assets() -> Iterator[Asset]`
+        * `get_assets_under_path(path: str) -> Iterator[Asset]` — Returns all assets whose paths are either under the given folder or equal the given filepath
```
yarikoptic, line 11: UX sugarings we should consider add on top
```
@assets -> get_assets()
@metadata -> get_metadata()
__iter__() -> .get_assets()
```
?  they might be tricky with subclassing though since IIRC we cannot just bind the property to get _assets in super class, since then subclasses .assets would us it instead of overloaded...   May be properties should be lazy (cached value) invocations over the methods?  That would give them a better purpose

yarikoptic (commented):
Just some initial comments/notes.  One design decision to make is either `Remote{Dandiset,Asset}` constructor would need to communicate to the server right away or not (and have properties/etc populated only whenever queried for the first time).

doc/design/python-api-1.md
```diff
@@ -0,0 +1,187 @@
+Designs for an improved Python API
+==================================
+
+* New base classes:
+    * `Dandiset`
+        * `identifier: str`
+        * `created: datetime`
+        * `modified: datetime`
+        * `version: Optional[Version]`
+        * `get_all_assets() -> Iterator[Asset]`
```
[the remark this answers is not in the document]
jwodder: I prefer the `get_all_*` nomenclature because otherwise `get_asset()` and `get_assets()` are just a typo apart.

jwodder:
@yarikoptic 

> One design decision to make is either `Remote{Dandiset,Asset}` constructor would need to communicate to the server right away or not (and have properties/etc populated only whenever queried for the first time).

The API resource instances would be constructed from data returned from an API call made by whatever method constructs the resources.  If there are any fields left out of the API response that require another call to fetch, those can be populated lazily/on-demand rather than when the object is constructed.

doc/design/python-api-1.md
jwodder: `name` removed.

doc/design/python-api-1.md
jwodder: Done.

doc/design/python-api-1.md
jwodder: Updated.

doc/design/python-api-1.md
```diff
@@ -0,0 +1,187 @@
...
+        * `get_assets_under_path(path: str) -> Iterator[Asset]` — Returns all assets whose paths are either under the given folder or equal the given filepath
+        * `get_asset_by_path(path: str) -> Optional[Asset]` — Finds the unique asset with the given exact path; returns `None` if no such asset
+        * `get_metadata() -> DandiMeta`
+   * `Asset`:
+        * `path: str`
+        * `sha256: str` — Should this only be on `RemoteAsset`?
```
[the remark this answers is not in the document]
jwodder: Noted.

doc/design/python-api-1.md
jwodder: I'm not clear on what exactly you're proposing here.

doc/design/python-api-1.md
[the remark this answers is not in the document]
yarikoptic: but we would never have `get_asset()` (and if we get -- it would require some argument), so I do not see a typo be a silent problem.

doc/design/python-api-1.md
yarikoptic: smth like

```python
@property
def assets(self):
     if self._assets is None:
            self._assets = self.get_assets()
     return self._assets

def __iter__(self):
     for asset in self.assets:
          yield asset
```

but we can think about all the sugaring later

yarikoptic (commented):
some more comments

doc/design/python-api-1.md
```diff
@@ -0,0 +1,189 @@
...
+        * **To bikeshed:** Should the Dandiset and version arguments to these methods be named `dandiset` & `version` or `dandiset_id` & `version_id`?
+    * Where a method can take an ID as either a string or an object, it is permitted for the object to belong to a different client.
+
+    * Attributes & methods of the `RemoteDandiset` class, a subclass of the new `Dandiset` base class, representing a combination of the API's dandiset and version resources:
+        * `identifier: str`
+        * `created: datetime`
+        * `modified: datetime`
+        * `version: Version`
+        * `most_recent_version: Version` ?
+        * `get_version(version_id: Union[str, Version]) -> Version` ?
```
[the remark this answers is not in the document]
jwodder: The purpose is to be able to query the server for information about just a specific version.

doc/design/python-api-1.md
```diff
@@ -0,0 +1,189 @@
...
+        * `get_versions() -> Iterator[Version]`
+        * `for_version(version_id: Union[str, Version]) -> RemoteDandiset`
+            * When `version_id == "draft"`, this is the same as `for_draft_version()`
+        * `for_draft_version() -> DraftDandiset`
+        * `delete() -> None`
+        * `get_users() -> Iterator[User]`
+        * `set_users(users: List[User]) -> None`
+            * Should this method also accept strings giving the usernames directly?
+        * `get_metadata() -> DandiMeta`
```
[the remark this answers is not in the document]
jwodder: Updated

doc/design/python-api-1.md
```diff
@@ -0,0 +1,189 @@
...
+        * `get_raw_metadata() -> dict` — useful when the metadata is invalid
+        * `publish() -> Version` (Should this be moved to `DraftDandiset`?)
+        * `get_assets() -> Iterator[RemoteAsset]`
+        * `get_assets_under_path(path: str) -> Iterator[RemoteAsset]`
+        * `get_asset_by_path(path: str) -> Optional[RemoteAsset]`
+        * `get_asset_by_uuid(uuid: str) -> Optional[RemoteAsset]` — Returns `None` if the given asset does not exist
+        * `download(target_dir: Union[str, Path], paths: Optional[List[str]] = None, show_progress=True, existing="error", get_metadata=True) -> ???`
+            * Use this to replace the `download()` function?
+            * The elements of `paths` are interpreted the same way as the argument to `get_assets_under_path()`
+            * `get_metadata` controls whether to create a `dandiset.yaml` file; should it default to `False` when `paths` is not `None`?
```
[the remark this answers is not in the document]
jwodder: Updated.

doc/design/python-api-1.md
```diff
@@ -0,0 +1,189 @@
...
+            * TO DO: Add an argument for controlling whether to create a directory under `target_dir` with the same name as the Dandiset ID?
+
+    * Methods of the `DraftDandiset` class (a subclass of `RemoteDandiset` used only for mutable draft versions):
+        * `set_metadata(metadata: DandiMeta) -> DraftDandiset` — returns the new, modified object
```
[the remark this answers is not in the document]
jwodder: Modified, but I thought assets kept their identifiers on metadata change?

doc/design/python-api-1.md
```diff
@@ -0,0 +1,189 @@
...
+        * Methods for uploading an individual asset:
+            * `upload_file(filepath: Union[str, Path], metadata: BareAssetMeta, show_progress=True, existing="refesh", validation="require") -> RemoteAsset`
```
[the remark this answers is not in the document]
jwodder: The idea of this method is to support generic file uploads where the user crafts the metadata themselves.  If we only want to support upload of files that can be represented with `LocalAsset` instances, we can get rid of it.

doc/design/python-api-1.md
```diff
@@ -0,0 +1,189 @@
...
+    * Attributes & methods of the `Version` class:
+        * `identifier: str`
+        * `created: datetime`
+        * `modified: datetime`
+        * `name: str`
+        * `asset_count: int`
+        * `size: int`
+
+    * Attributes & methods of the `RemoteAsset` class (a subclass of `Asset`):
+        * `uuid: str`
```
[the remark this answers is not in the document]
jwodder: Updated.

doc/design/python-api-1.md
```diff
@@ -0,0 +1,189 @@
...
+
+    * Attributes & methods of the `Version` class:
+        * `identifier: str`
+        * `created: datetime`
+        * `modified: datetime`
+        * `name: str`
+        * `asset_count: int`
+        * `size: int`
+
+    * Attributes & methods of the `RemoteAsset` class (a subclass of `Asset`):
```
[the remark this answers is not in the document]
jwodder: It's for exposing the "meta-metadata" associated with an asset that's returned when paginating through asset lists and was previously returned when querying an individual assets.  Are we planning on getting rid of those fields from the API?

doc/design/python-api-1.md
```diff
@@ -0,0 +1,189 @@
...
+        * `uuid: str`
+        * `path: str`
+        * `sha256: str`
```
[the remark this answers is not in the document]
jwodder: I've replaced this with a `get_digest(method)` method on the `Asset` base class.

doc/design/python-api-1.md
```diff
@@ -0,0 +1,189 @@
...
+        * `size: int`
+        * `created: datetime`
+        * `modified: datetime`
+        * `dandiset_id: str`
```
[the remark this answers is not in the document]
jwodder: Until the asset API endpoints are modified to not require the dandiset & version IDs as path components, the `RemoteAsset` class needs this information.

doc/design/python-api-1.md
```diff
@@ -0,0 +1,189 @@
...
+        * `version_id: str`
```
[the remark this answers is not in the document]
jwodder: Yes, Dandiset version.  Same response as above.

jwodder:
@yarikoptic I had an idea about the return values from the blocking upload/download functions.  Since some transfers may not produce assets due to errors or the `existing` setting, I think it may be a good idea for the functions to return structures with the following fields:

* `UploadResult`:
    * `local_asset: LocalAsset` — the asset that was uploaded; will have to be something else if we want to support an `upload_file(filepath, ...)` method
    * `status: TransferStatus` — an enum with `SUCCESS`, `ERROR`, and `SKIPPED` values
    * `message: str` — for `ERROR`, this is the error message; for `SKIPPED`, this is a reason why the asset was skipped; for `SUCCESS`, it doesn't matter
    * `remote_asset: Optional[RemoteAsset]` — The asset created by the upload; only set for `SUCCESS`
* `DownloadResult`:
    * `remote_asset: RemoteAsset` — the asset that was downloaded
    * `status: TransferStatus` — an enum with `SUCCESS`, `ERROR`, and `SKIPPED` values
    * `message: str` — for `ERROR`, this is the error message; for `SKIPPED`, this is a reason why the asset was skipped; for `SUCCESS`, it doesn't matter
    * (Some representation of the resulting downloaded file?  Returning `LocalAsset` would be awkward for the cases where we're not downloading to a local Dandiset.)

What do you think?

yarikoptic:
> @yarikoptic I had an idea about the return values from the blocking upload/download functions.

if they are blocking, so more typical Python user-facing interfaces, why not just to raise an exception? e.g. `IncompleteResultsError` which would have attributes such as `.results` containing the results which were successfully obtained and may be `.failed` for e.g. `Assets` which failed to upload or download.   That would make it easier to write high level Python code, without requiring to analyze all those records/status'es just to e.g. download something and not miss an errored out operation.

jwodder:
@yarikoptic Assuming you're suggesting that the exception only be raised after all threads have ended, that could work, too, but even when there are no errors, I feel there should be some way to indicate which assets were skipped and which weren't.  It doesn't have to be done by returning a list of Result objects; a single Result object with list attributes would work as well.

How about this:

* On success, upload functions return a structure with an `assets: List[RemoteAsset]` attribute for successful uploads plus a `skipped: List[Tuple[LocalAsset, str]]` attribute pairing skipped local attributes with the reasons why they were skipped.
* If an error occurs in an upload function, an exception is raised with the same `assets` and `skipped` attributes as above plus an `errored: List[Tuple[LocalAsset, str]]` attribute pairing failed assets with error messages.
* Download functions return the same, except `LocalAsset` is replaced with `RemoteAsset`, and `RemoteAsset` is replaced with ... something I still haven't decided on.

jwodder:
@yarikoptic Do we want the classes to be "lazy" (not making any API requests until actual data is requested) or not (and thus fetching data upon instance creation)?

yarikoptic:
re lazy: I feel that we might benefit from "lazy" in particular for large dandisets, but for "remote" dandisets we should at least verify that they do exist, so I guess basic metadata record should be fetched upon class instantiation for those.

doc/design/python-api-1.md
```diff
@@ -0,0 +1,194 @@
...
+        * `get_assets() -> Iterator[Asset]`
+        * `get_assets_under_path(path: str) -> Iterator[Asset]` — Returns all assets whose paths are either under the given folder or equal the given filepath
+        * `get_asset_by_path(path: str) -> Optional[Asset]` — Finds the unique asset with the given exact path; returns `None` if no such asset
+        * `get_metadata() -> DandisetMeta`
+   * `Asset`:
+        * `path: str`
+        * `size: int`
+        * `created: datetime`
+        * `modified: datetime`
+            * **Problem:** The API reports `created` and `modified` dates when paginating through a version's assets, but the data is absent when getting an asset's metadata directly from the asset-specific endpoint (Note that, although there are two timestamps for modification times in asset metadata, neither of them match the `modified` time in the pagination results.)
```
yarikoptic, line 19: yeah -- those `created` and `modified` AFAIK are for the specific row in dandiset_versions_assets -- so specific to that dandiset version asset, whenever metadata record modified, should be the one on when metadata record itself was modified (should be before those `created` `modified` above).  If our `Asset` here  is not dandiset/version specific (I think it is not), then I guess we could have only the dateModified from metadata.    For our needs (upload/download `refresh` mode) I think it is the one which is needed. So let's just drop `created` (even if for just now).

doc/design/python-api-1.md
```diff
@@ -0,0 +1,194 @@
...
+        * `get_raw_metadata() -> dict` — useful when the metadata is invalid
+        * `set_metadata(metadata: BareAssetMeta) -> None` — Assuming assets keep their identifier on metadata change, this modifies the instance in-place; otherwise, it should return the new, modified instance
+        * `delete() -> None`
+        * `download(filepath: str, show_progress=True, chunk_size=...) -> ???`
+        * `iter_download(filepath: str, chunk_size=...) -> Iterator[DownloadProgressDict]`
+
+    * Attributes of the `User` class:
+        * `username: str`
+        * `first_name: str`
+        * `last_name: str`
```
yarikoptic, line 116: may be don't bother separating and just keep it in "sync" with API here which just has `name`.

doc/design/python-api-1.md
```diff
@@ -0,0 +1,194 @@
...
+        * `admin: bool`
+        * **Note:** We should ask the dandi-api developers to make the `GET/PUT /dandisets/{dandiset__pk}/users/` endpoints return full user structures, not just ones with only `username` fields, so that instances of this class can be properly populated
```
yarikoptic, line 118: filed https://github.com/dandi/dandi-api/issues/319 .  For now we can proceed without exposing `.admin` and then add it whenever API provides it

doc/design/python-api-1.md
```diff
@@ -0,0 +1,194 @@
...
+        * `current: int` — number of bytes downloaded (formerly "done")
+        * `pct: float` — percentage of bytes downloaded (formerly "done%")
+        * `checksum: str` — `"differs"`, `"ok"`, or `"-"`
+
+    * On success, blocking upload methods for uploading multiple assets return structures with the following fields:
+        * `assets: List[RemoteAsset]` — list of assets successfully uploaded
+        * `skipped: List[Tuple[LocalAsset, str]]` — list of skipped assets, paired with the reasons why they were skipped
+    * For blocking upload methods that upload multiple assets, if an error occurs while uploading, an exception is raised after all threads have completed; this exception has the same `assets` and `skipped` attributes as above plus an `errored: List[Tuple[LocalAsset, str]]` attribute pairing errored assets with error messages
+
+    * **To discuss:** Should methods that accept `DandisetMeta` and `(Bare|Remote)AssetMeta` instances also accept raw `dict`s?  If so, should any validation be done on these `dict`s?
```
yarikoptic, line 154: I think for starters we can just add a dedicated `@classmethod` `from_dict(cls, d: dict, validate:bool=True) -> cls (possible?)` and require such conversion explicitly.  If we keep running into needing that too often -- we would rethink

yarikoptic (changes requested):
Left some comments, primarily on freshly added `**Notes**` and there might be some elderly comments whenever I started the review awhile back but never "Finished"

doc/design/python-api-1.md
jwodder: `created` removed.

doc/design/python-api-1.md
jwodder: The separate fields used to exist in a previous version of the API; updated.

doc/design/python-api-1.md
jwodder: Pydantic already provides `parse_obj()` (validating) and `construct()` (non-validating) methods for that.

doc/design/python-api-1.md
```diff
@@ -0,0 +1,189 @@
+Designs for an improved Python API
```
yarikoptic, line 1: note, in https://github.com/dandi/dandi-cli/pull/712 I renamed `doc/` to `docs/` so if we are to merge this one, might want to rename

jwodder:
@yarikoptic I'd say somewhere between a third and a half of the doc has been implemented.  I would prefer breaking all the tasks in this doc up into multiple issues (possibly organized in a GitHub project).  I don't know why you'd ever merge this PR, as once it's implemented the docstrings should be the canonical location for the features' documentation.  If you really want a nice place to put a write up of the desired API, I think [the repo wiki](https://github.com/dandi/dandi-cli/wiki) is a better place.

yarikoptic:
I don't know myself why we haven't ever merged it , so will do that now -- shouldn't hurt.

PR #449, opened 2021-03-04
fix: change enum to const for single enums

satra:
This should address the change to using `Enum` for `Literal` in pydantic 1.8+. This reverts the schema generation to use `const`. This change can be reverted once https://github.com/koumoul-dev/vuetify-jsonschema-form/issues/212 is addressed.


satra:
merging so that @AlmightyYakob can proceed.

PR #450, opened 2021-03-04
[wip] fixing unitText in parse_age (closes #421)

### LINKED ISSUE #421: age and text don't match on one asset
> curl -X GET "https://api.dandiarchive.org/api/dandisets/000008/versions/draft/assets/812181a1-50c0-40c8-9f09-a3e75d31a40e/" -H  "accept: application/json" 

returns age as `"age":{"value":"P49D","unitText":"Years from birth"}`

the value says 49 days but text says years from birth.

djarecka:
fixing `unitText` in `parse_age`

### DIFF
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -125,7 +125,12 @@ def extract_age(metadata):
             return ...
     else:
         duration = timedelta2duration(start - dob)
-    return models.PropertyValue(value=duration, unitText="Years from birth")
+    if duration.endswith("D"):
+        return models.PropertyValue(value=duration, unitText="Days from birth")
+    elif duration.endswith("S"):
+        return models.PropertyValue(value=duration, unitText="Seconds from birth")
+    else:
+        return models.PropertyValue(value=duration, unitText="Years from birth")
 
 
 def timedelta2duration(delta):
@@ -143,11 +148,11 @@ def timedelta2duration(delta):
     s = "P"
     if delta.days:
         s += f"{delta.days}D"
-    if delta.seconds or delta.microseconds:
+    elif delta.seconds or delta.microseconds:
         sec = delta.seconds
         if delta.microseconds:
             # Don't add when microseconds is 0, so that sec will be an int then
-            sec += delta.microseconds / 1000000
+            sec += delta.microseconds / 1e6
         s += f"T{sec}S"
     if s == "P":
         s += "0D"


dandi/metadata.py
```diff
@@ -143,11 +148,11 @@ def timedelta2duration(delta):
     s = "P"
     if delta.days:
         s += f"{delta.days}D"
-    if delta.seconds or delta.microseconds:
+    elif delta.seconds or delta.microseconds:
```
jwodder, line 151: This should stay as `if`, in case `delta` has both `days` and `seconds`/`microseconds`.

dandi/metadata.py
```diff
@@ -125,7 +125,12 @@ def extract_age(metadata):
             return ...
     else:
         duration = timedelta2duration(start - dob)
-    return models.PropertyValue(value=duration, unitText="Years from birth")
```
jwodder, line 128: As a duration can be a mixture of units, it might be better to just always set the `unitText` to "Time from birth".

satra:
a few notes. we will likely have to transform this information in specific ways from different datasets. however, we did adopt that value of `Age` would follow the ISO-8601  (https://en.wikipedia.org/wiki/ISO_8601) standard. 

so everything would be a duration (https://en.wikipedia.org/wiki/ISO_8601#Durations). thus you don't want to change the value to an integer. it should be an ISO-8601 formatted string. 

the text corresponding to it should use the periodicity (Y,W,D,...) and say "Years, Weeks, Days". to the extent we have information about the reference. we can use "Years from birth", "Weeks (Gestation)" to give additional information about the reference point. 

Age is also a Property Value from schema.org, so it can take a field (https://schema.org/valueReference) and put the reference point in that field. so 

```JSON
{
"value": "P49D",
"unitText ": "Days",
"valueReference":  "BirthReference"
}
```

and we can define a set of enum types for Reference (Birth, Conception, etc.,.)

satra:
reference would have to be omitted if we don't know the answer. but this is the kind of thing we need to have data providers add to the NWB file in some manner (natural language, or more explicitly) (ping @bendichter).

in dandi-cli we should try to interpret the string and turn it as well as possible into the structure.

dandi/metadata.py
djarecka: so it wouldn't really contain any form of "Unit" itself. Any reason why we use `unitText` ?

dandi/metadata.py
djarecka: oh, i see, thanks!

djarecka:
so for value you're using `ISO-8601`- duration, so it's a mixture of values and units. I'm still not completely sure, where I should check for everything else. for example, the idea of having a unit field doesn't work well with `ISO-8601`- duration IMO

satra:
thinking about this a bit more. indeed `unitText` could simply be "ISO-8601 duration". after all that's what the value is, but we do need to add `valueReference` when there is information.

in terms of changes, this is all the heuristic for age extraction. we just have to keep doing this for as many files, till the NWB files themselves become consistent.

bendichter:
A little background on this. The age field in NWB is technically a free-form text string. the way users were entering ages was all over the place, and definitely not machine-readable across datasets, so I started encouraging people to use the 8601 duration standard. We needed a way that would work for all species with a large range of lifespans from hours (e.g. flies) to years (e.g. humans), and to express resolution on those different scales (we wouldn't want to record a human's age in hours if we don't know when exactly they were born).  I'll admit that ISO 8601 duration is not very commonly used but it seems to me to be the best approach given these considerations, and it's an established standard. We currently use this as a "best practice" within NWB, though we have not enforced this since it would be a schema breaking change, meaning that files created before this rule might be invalid after the rule change. That's not the end of the world, but I've been trying to avoid breaking changes. In fact, doing this properly would probably require making a change to the schema language itself, and adding a data type for duration, similar to what we do for datetime at the bottom of [this table](https://schema-language.readthedocs.io/en/latest/specification_language_description.html#dtype).

One thing we could do in NWB is allow users to enter a `timedelta` object and automatically convert that to an 8601 duration string. We haven't done this yet but it's definitely doable and would help encourage properly formatted ages.

For representing the data in DANDI, I think it would be OK to parse the string out into value and units if you want, or to leave it as an 8601 string and make the value "8601 duration" as Satra is suggesting. Either approach makes sense to me.

jwodder:
@djarecka FYI:
* If a PR is a work in progress, it should be a [draft pull request](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests).
* The "Closes #ISSUE_NO" text should be in the pull request summary (the initial comment) in order to autoclose the issue on merge.

djarecka:
@jwodder - ok, will create a draft next time

djarecka:
@satra - is this is what you had in mind? 
` return models.PropertyValue(value=duration, unitText="ISO-8601 duration", valueReference="BirthReference")`

pydantic complains about it `value is not a valid dict (type=type_error.dict)`

satra:
you'll have to modify valueReference in PropertyValue to allow EnumerationTypes in addition to PropertyValue, and then define an additional enumeration type (e.g., AgeReferenceType) similar to (accesstype, roletype).

satra:
these things are defined in models.py and model_types.py

djarecka:
@satra - could you please check the changes for `age`, is it what you had in mind?

dandi/metadata.py
```diff
@@ -147,7 +151,7 @@ def timedelta2duration(delta):
         sec = delta.seconds
         if delta.microseconds:
             # Don't add when microseconds is 0, so that sec will be an int then
-            sec += delta.microseconds / 1000000
+            sec += delta.microseconds / 1e6
```
satra, line 154: do you really want to change this?

dandi/model_types.py
```diff
@@ -573,3 +573,21 @@
...
+            "rdfs:comment": "An enumeration of age reference",
+            "rdfs:label": "Age reference type",
+            "rdfs:subClassOf": {"@id": "schema:Enumeration"},
+        },
+        {
+            "@id": "dandi:BirthReference",
+            "@type": "dandi:AgeReferenceType",
+            "rdfs:comment": "Age since Birth",
+            "rdfs:label": "BirthReference",
+        },
```
satra, line 591: also add a gestational age reference. you will need that for some datasets too: https://en.wikipedia.org/wiki/Gestational_age

satra:
yes this looks good - i left a few comments. you may want to run this through every nwb file to check where it breaks or doesn't do the right thing.

dandi/metadata.py
djarecka: yes... why are you asking?
do you have a reason to keep it? I could revert this change if there is any reason, but can't think of any, so you have to tell me

djarecka: do you prefer 10**6?

satra: what you have is fine :)

dandi/model_types.py
djarecka: should I be adding pointers to wikipedia or other sources, or the short description is enough: https://github.com/dandi/dandi-cli/pull/450/commits/aaa1973d5f71870db058f3684caaa909601b8602

satra: always useful to add references when possible - you can add it to the comment field.

satra:
here is an additional example to use for age:

in this file: 

> curl -o trythis.nwb https://dandiarchive.s3.amazonaws.com/girder-assetstore/a4/36/a43656b9d3f9453aaa1e653c4eeb0226

this is from dandiset 000008

```
 'age': 'P148D',
 'date_of_birth': datetime.datetime(2017, 11, 23, 0, 0, tzinfo=tzoffset(None, -18000)),
```

this should be converted to Age: as "2017-11-23T00:00:00-05:00/P148D" signifying birthdate + age under ISO8601 interval format.

djarecka:
at least for this file the age is calculated based on "date_of_birth" (`datetime.datetime(2017, 11, 23, 0, 0, tzinfo=tzoffset(None, -18000))`) and "session_start_time" (`datetime.datetime(2018, 4, 20, 0, 0, tzinfo=tzoffset(None, -14400))`.  At the end `extract_age` gives `P147DT82800S`

satra:
i don't know how precise we will want this to be overall, but this value sounds good to me as the duration bit: `P147DT82800S` (but i also see an Age value in the metadata of this file, which is this value rounded off to the nearest day. whether rounding is appropriate or not may depend on the species, study, etc.,.). but since age is provided, i would use Age.

this file is a little weird, in the sense that the timezones for the date of birth and start of session are different. possible in the world of science, just weird. i'm just not sure you will have date of birth in every nwb file. so this extension would only be done in cases where it is available. 

also if the age derived mismatches the age provided by more than a unit of the provided age, we should trigger a notification to the user of the mismatch. this bit can be implemented later.

djarecka:
@satra - right now  the code uses `age` if it fails to find `date_of_birth` and `session_start_time`. Do you want me to revert this?

satra:
that is fine for now. i'm saying add date of birth to the age representation as discussed above in alignment with the iso standard.

yarikoptic:
@satra @djarecka - what do we do about this one? should it just be taken out of Draft and merged or more work to be done?

djarecka:
I think the only doubts I had was about `age`  vs  `date_of_birth` and `session_start_time`. But I believe @satra agreed to leave it for now as it is

djarecka:
having issue with merging master - the test I was working on was change to represent age in "23day" instead of the ISO-8601 duration I was working on...

djarecka:
please let me know if you have more suggestions

yarikoptic:
I could only ask for a few more tests dedicated to that conversion function, e.g. returning `P0D` for 0?

djarecka:
@yarikoptic - tests for `parse_age` already exist and cover your example : https://github.com/dandi/dandi-cli/blob/master/dandi/tests/test_metadata.py#L54

I was thinking just about adding a specific tests for `extract_age`, but realized that I'm not sure about the representation we're using. Do we have a reason to use only `S` for time designators? ISO_8601 allows for hours, H, and minutes, M: https://en.wikipedia.org/wiki/ISO_8601

djarecka:
after checking with @satra: keeping `S` only for time designators

yarikoptic:
ok, let's proceed then! Thank you @djarecka !

yarikoptic:
d'oh -- just stroke me that it is likely changing the schema even if so slightly... damn, in the light of the whole discussion within #623 (and "private" discussions with @satra) I guess ideally it should have also boosted at least .patch of the schemaVersion... heh heh.

yarikoptic:
FTR: here is the "inplace" (no version to schema version) inflicted by my rush: https://github.com/dandi/schema/commit/efc9db675f01d3d2d3692deec47def4efbac911e . Well, in principle we could revert change to 0.3.0 and progress .patch and regenerate to stay kosher.    Any objections @satra if I do that or you would prefer me not to?

satra:
this is an example where dandi cli older version would use older schema and may overwrite a rewrite.

yarikoptic:
Well, we would need to revert this PR merge as well and/or boost schema version to get kosher again or schema would just keep getting rewritten in 0.3.0 (add I don't think we should stop ci from doing that)

djarecka:
anything I should have done in this PR? Not completely sure what is wrong

yarikoptic:
oh, "It is not you, it is me" ;)  we dragged on this PR past the point of migration.  Now IMHO we should be more consciousness about changes to metadata schema since we already populated metadata records in the archive upon migration. Hence now all the recent deliberation about version boost in #623 and recent issues like #631 so we might need to establish "proper" releasing of metadata (ATM we immediately render jsonschemas into https://github.com/dandi/schema possibly overriding already existing/used (by web ui) one etc.

djarecka:
@yarikoptic - what should I do add these changes to master? It looks like these changes were reverted and never came back to master. At least some of them, as @satra  pointed in #669.
I can create a new PR, but not sure what is at the end the best way of doing it, since it includes changes to schema

satra:
@djarecka - just send a PR to this branch.

PR #454, opened 2021-03-05
Rename DandiMeta to DandisetMeta

jwodder:
The test failures are due to changes caused by dandi/dandi-api#144.  The fix for them is in #452.


yarikoptic:
Thank you! Let's proceed

PR #455, opened 2021-03-08
Just to trigger CI rerun to see if any changes in dandi-api require changes in the client

yarikoptic:
will also do a local run


yarikoptic:
whatever was broken was fixedup in #452

PR #457, opened 2021-03-10
ENH: more of lgr.debug for multipart upload

yarikoptic:
Primarily was added to resolve the mystery of ETag of the file
from blob store not corresponding to precomputed one.  Which was
(as reported in detail in
https://github.com/dandi/dandi-api/pull/150#issuecomment-795456385
due to us then moving that blob from staging area into blob store
using different chunking logic, with smaller number of chunks


jwodder (approved):

yarikoptic:
Thank you @jwodder for the approval, let's proceed

Issue #431, opened 2021-03-01
add bids validation support


satra:
since we may be getting more bids dandisets, we should start providing bids validation support.

yarikoptic:
I wonder which route should we go -- demand/ship full blown JS bids-validator or may be just go (at least start with) https://github.com/bids-standard/bids-validator#python-library .  I expect it to be improved in near future to make use of now machine readable bids-specification/src/schema (just filed https://github.com/bids-standard/bids-validator/issues/1202 since found no existing issue on that end yet).

satra:
does it check for required keys in json files? for some of these datasets coming from k-space reconstructions without dicoms, we may want to need to enable that.

yarikoptic:
the new `src/schema` AFAIK doesn't yet specify json sidecar fields yet :-/ not sure what JS bids-validator does

PR #459, opened 2021-03-10
Display `dandi validate` errors using logger

### LINKED ISSUE #458: incomplete log file created by validation
the log files created by `dandi validate` do not contain any information about errors.

jwodder:
Closes #458.

### DIFF
--- a/dandi/cli/cmd_validate.py
+++ b/dandi/cli/cmd_validate.py
@@ -2,7 +2,7 @@ import os
 
 import click
 
-from .base import devel_option, map_to_click_exceptions
+from .base import devel_option, lgr, map_to_click_exceptions
 
 
 @click.command()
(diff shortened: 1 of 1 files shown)


yarikoptic:
thank you @jwodder , let it be like this for now.  validation is yet to be RFed later on (#431 etc), so we will tune later

djarecka:
thanks for addressing my issue. Unfortunately, the output is now much less readable IMO

before:
Screen Shot 2021-03-11 at 8 54 54 AM
after:
Screen Shot 2021-03-11 at 8 55 35 AM

Let me know if you want me to open a separate issue.

jwodder:
@yarikoptic What exactly should the output look like?

yarikoptic:
IMHO it is ok (although not colored) -- the only thing is that `%s: ok` I have missed - somewhere path was not provided.
Or @djarecka what readability aspect you have in mind?

jwodder:
@yarikoptic I fixed the "%s: ok" thing in #462.

djarecka:
@yarikoptic - it doesn't provide info about files that are not failing, and yes the color also helps

djarecka:
@jwodder - thanks, your PR helps! The color is gone, but I guess, we can live with this.

yarikoptic:
yes, let's live without colors for now.  Later we will RF for more consistent handling of outputs etc and I hope that is when colors would return ;-)

PR #460, opened 2021-03-11
Change BareAssetMeta.digest to a list

### LINKED ISSUE #451: schema: make digest into a Dict
ATM https://github.com/dandi/dandi-cli/blob/master/dandi/models.py#L835 we have
```python
digest: Digest = Field(nskey="dandi")
```
but as discussing possibly using another checksuming (multihash) for validation of the uploads, we need to tune up to make this into a list or better a dict to populate/use more efficient checksum

jwodder:
Closes #451.

### DIFF
--- a/dandi/consts.py
+++ b/dandi/consts.py
@@ -154,3 +154,6 @@ MAX_CHUNK_SIZE = int(os.environ.get("DANDI_MAX_CHUNK_SIZE", 1024 * 1024 * 8))  #
 # just a structure, better than dict for RFing etc
 class routes(object):
     dandiset_draft = "{dandi_instance.redirector}/dandiset/{dandiset[identifier]}/draft"
+
+
+DANDI_SCHEMA_VERSION = "0.1.0"
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -403,15 +403,14 @@ class DandiAPIClient(RESTFullAPIClient):
         asset_path = asset_metadata["path"]
         filehash = get_digest(filepath)
         lgr.debug("Calculated sha256 digest of %s for %s", filehash, filepath)
-        if (
-            asset_metadata.get("digest") is not None
-            and asset_metadata.get("digest_type") == "SHA256"
-            and asset_metadata["digest"] != filehash
-        ):
-            raise RuntimeError(
-                f"{filepath}: File digest changed; was originally"
-                f" {asset_metadata['digest']} but is now {filehash}"
-            )
+        for digest in asset_metadata.get("digest", []):
+            if digest["cryptoType"] == "SHA256":
+                if digest["value"] != filehash:
+                    raise RuntimeError(
+                        f"{filepath}: File digest changed; was originally"
+                        f" {asset_metadata['digest']} but is now {filehash}"
+                    )
+                break
         try:
             self.post("/uploads/validate/", json={"sha256": filehash})
         except requests.HTTPError as e:
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -279,10 +279,12 @@ extract_wasGeneratedBy = extract_model_list(
 
 def extract_digest(metadata):
     if "digest" in metadata:
-        return models.Digest(
-            value=metadata["digest"],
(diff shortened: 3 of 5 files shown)


yarikoptic:
looks good to me.
- @satra -- all good?
- I wonder if we should boost schema version to rc2 with this to make transition more "conscious" if web UI or api would need to account for the change for one reason or another

yarikoptic:
> also perhaps it will be good for us to start incrementing the `rcN` from now even for practice. it's just that we need some indicator on both the UI side and the API side to check if they are using an old schema in their testing infrastructure.

`dandi-cli`:
```
$> git grep 1.0.0-rc1
dandi/models.py:    schemaVersion: str = Field(default="1.0.0-rc1", readOnly=True, nskey="schema")
dandi/tests/data/dandiarchive-docker/docker-compose.yml:      DJANGO_DANDI_SCHEMA_VERSION: 1.0.0-rc1
dandi/tests/data/dandiarchive-docker/docker-compose.yml:      DJANGO_DANDI_SCHEMA_VERSION: 1.0.0-rc1
dandi/tests/test_metadata.py:  "schemaVersion": "1.0.0-rc1",
dandi/tests/test_metadata.py:  "schemaVersion": "1.0.0-rc1",
dandi/tests/test_metadata.py:  "schemaVersion": "1.0.0-rc1",
```
I think we should move specification into `consts.py` as `DANDI_SCHEMA_VERSION = "1.0.0-rc1"` and make code/tests use it and it would be the one to be used by outside code using `dandi` as a library.  As for metadata records when uploading -- they will include the version. 

An alternative: we make it `1.0.0` right away and treat increments in `1.MINOR` as "breaking changes" while using the last `.PATCH` as non-breaking until we stabilize.  Then dandi-api could right away become tolerant to `.PATCH` version changes and establish some minimal version (such as `1.0.0`).  After we stabilize, starting from `1.MINOR.0` we would say that there would be no breaking changes, and we proceed in fully semver fashion with appropriate treatment by dandi-api.

WDYT @satra @jwodder ?

`dandi-api`: 
- re exposing supported API versions -- yet to be addressed https://github.com/dandi/dandi-api/issues/36 so client could check and announce to the user that possibly `dandi-cli` update is needed.  Related freshly filed: 
- while we are introducing breaking changes (so rc2 is not backward compatible with rc1) there is probably no sensible way to handle such changes on the dandi-api side besides "migrating" dandi-api to use a new version
  - in the long run, it dandi-api, probably should just declare "minimal versions" within any `MAJOR.` version it supports. I.e. could be something like `minimal_metadata_versions = ['1.2.3', '2.0.0']` since within each `MAJOR.` version there should be no breaking changes going forward until the next major version, and within `MAJOR.` we might need to establish minimal supported. WDYT @dchiquito ? (we are not introducing yet `2.0.0` -- just future proofing ideas)

This post also inspired me to https://github.com/dandi/dandi-cli/issues/469 to establish clear(er) provenance for our schema dump and linkage of schema version to dandi-cli version

yarikoptic:
> depending on what @dchiquito extracts during an asset post from metadata, changes in asset post may need changes on the api (aren't there tests on asset creation with the API here? if so that should catch changes, no?)

unlikely, but in other changes, it might also relate to meditor of the web UI which might need to become schema version "aware" as well, thus attn @AlmightyYakob to bring attention to possible "unfreezing" and boosting forward `1.0.0-rc1` version (see/comment on the above)

yarikoptic:
created dedicated https://github.com/dandi/dandi-cli/issues/472 to summarize items toward briefly discussed during squat meeting going back to 0.1.0 for schema version.  @jwodder please add a const and make it 0.1.0 for this "schema release", and we will start incrementing MINOR.PATCH as schema gets further changes

jwodder:
@yarikoptic Constant added.

yarikoptic:
hm, and it didn't trigger any tests failures.  Don't we test against dandi-api instance or it doesn't care/error out ATM upon upload of metadata with unsupported metadata version (I thought that check was added awhile back)?

Anyways -- let's proceed and fixup if/where needed (after all it is green here)

PR #465, opened 2021-03-12
Show digest progress when uploading

### LINKED ISSUE #400: upload: progress indication for "digesting"?
"digesting" could take comparable to upload time, in particular in the case of the parallel `upload`s. Although with caching we at least would not bother redoing it twice if did recently, it would be nice if we have a way to report back to the user on the progress of digesting.  For upload we have a dedicated column `upload` which shows `%`.  But I wonder if we better have a dedicated column "progress" which would be populated (and either cleared or report time on how long it took after it is done) used by whatever 'status' column reports (digesting, or uploading, etc).

jwodder:
Closes #400.

Caveats:

* This means a new cache, so `populate_dandiset_asset_yamls.py` will take 25+ hours the first time it runs after merging this.
* Diskcache is backed by SQLite, which does not do well on NFS filesystems.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -464,7 +464,7 @@ class DandiAPIClient(RESTFullAPIClient):
                         bytes_uploaded += len(chunk)
                         yield {
                             "status": "uploading",
-                            "upload": 100 * bytes_uploaded / total_size,
+                            "pct": 100 * bytes_uploaded / total_size,
                             "current": bytes_uploaded,
                         }
                         parts_out.append(
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -464,7 +464,7 @@ class DandiAPIClient(RESTFullAPIClient):
                         bytes_uploaded += len(chunk)
                         yield {
                             "status": "uploading",
-                            "pct": 100 * bytes_uploaded / total_size,
+                            "upload": 100 * bytes_uploaded / total_size,
                             "current": bytes_uploaded,
                         }
                         parts_out.append(
--- a/dandi/support/digests.py
+++ b/dandi/support/digests.py
@@ -94,8 +94,7 @@ class Digester(object):
 
 class DigestProgress(TypedDict, total=False):
     status: str
-    size: int
-    pct: float
+    #size: int
     digests: Dict[str, str]
 
 
@@ -144,10 +143,10 @@ def get_progressive_digests(
                 for d in digestions:
                     d.update(block)
                 current += len(block)
+                pct = 100 * current / total_size
                 yield {
-                    "status": "digesting",
-                    "size": current,
-                    "pct": 100 * current / total_size,
+                    "status": f"digesting ({pct:.2f}%)",
+                    #"size": current,
                 }
         digested = {n: d.hexdigest() for n, d in zip(digest_tuple, digestions)}
         # Calculate the key again just in case:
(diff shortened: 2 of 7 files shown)


yarikoptic:
> Diskcache is backed by SQLite, which does not do well on NFS filesystems.

this might be a showstopper for such an approach since I would expect a good number of use cases where people would be uploading from some HPC etc where FS is likely to be mounted via NFS.

jwodder:
@yarikoptic Would those users be likely to have their home directories mounted on NFS, or just their NWB files?

satra:
in many HPC centers (including ours) all directories are mounted over nfs or lustre.

yarikoptic:
same here -- I believe underlying is lustre, and then NFS on top... HOME directories included:
```shell
[d31548v@discovery7 ~]$ df . -h
Filesystem                                        Type  Size  Used Avail Use% Mounted on
dartfs-nfs-t3.dartmouth.edu:/ifs/rc/dartfs-hpc/rc nfs4   50G   41G  9.2G  82% /dartfs-hpc/rc
[d31548v@discovery7 ~]$ mount | grep /dartfs-hpc/rc
dartfs-nfs-t3.dartmouth.edu:/ifs/rc/dartfs-hpc/rc on /dartfs-hpc/rc type nfs4 (rw,nosuid,relatime,vers=4.0,rsize=1048576,wsize=1048576,namlen=255,...
```

yarikoptic:
I believe we no longer use/run `populate_dandiset_asset_yamls.py` so we do not need to worry about that.  But the concern about use of sqlite was "real".  @jwodder you can take discovery/ndoli as one of the prototypical systems to try on.
Overall this PR would need a big refresh, but I am also not sure yet if we don't just need to add needed support/interfaces to fscacher  and avoid any extra fanciness.

yarikoptic:
@jwodder can you research into another solution here which would not require on-disk cache? (not clear to me why it needs to be on disk and not some in-memory thread safe data structure)

jwodder:
@yarikoptic We already cache all digests on-disk with fscacher, so unless we're planning on getting rid of that feature, we'll need some way to interact with the on-disk cache in finer detail than the fscacher API currently allows.

yarikoptic:
> @yarikoptic We already cache all digests on-disk with fscacher, so unless we're planning on getting rid of that feature, we'll need some way to interact with the on-disk cache in finer detail than the fscacher API currently allows.

sorry, I am missing why we would necessarily need that.  My thinking:

- `fscacher` caches entire call to `get_digest` and `get_dandietag`
- `fscacher` has no problem with NFS AFAIK (we run benchmarks on NFS)
- there can be some thread-safe `progress_reports: dict` for `filepath: deque` entries which do get populated/appended to within operation of digest compute function.
- `get_digest` is invoked in a separate thread and main thread monitors until thread is done with result (checksum) for that path, and meanwhile if any entry appears in corresponding deque -- takes it and yields up for pyout to update report

what forbids or just "not good" in such or similar implementation ?

jwodder:
@yarikoptic

> there can be some thread-safe `progress_reports: dict` for `filepath: deque` entries which do get populated/appended to within operation of digest compute function.

If the digest-computing function is cached by fscacher, then external variables won't be modified on a cache hit.

yarikoptic:
it is non draft although we never finalized it, so moving to draft, or we might even need to close/redo since lots of changes since then.

yarikoptic:
for now let's just close since would need a big overhaul if we decide to actually have it implemented.

PR #468, opened 2021-03-12
Only call logging.basicConfig() when used as a command

### LINKED ISSUE #466: avoid logging.basicConfig(format=FORMAT) in __init__.py
For being usable as python library we should just set logging backends etc for our `lgr` and not interfere with "global" logging configuration

jwodder:
Closes #466.

### DIFF
--- a/dandi/__init__.py
+++ b/dandi/__init__.py
@@ -41,7 +41,4 @@ def set_logger_level(lgr, level):
 
 
 lgr = get_logger()
-# Basic settings for output, for now just basic
 set_logger_level(lgr, os.environ.get("DANDI_LOG_LEVEL", logging.INFO))
-FORMAT = "%(asctime)-15s [%(levelname)8s] %(message)s"
-logging.basicConfig(format=FORMAT)
--- a/dandi/cli/command.py
+++ b/dandi/cli/command.py
@@ -84,6 +84,7 @@ def main(ctx, log_level, pdb=False):
 
     e.g. dandi upload --help
     """
+    logging.basicConfig(format="%(asctime)-15s [%(levelname)8s] %(message)s")
     set_logger_level(get_logger(), log_level)
 
     # Ensure that certain log messages are only sent to the log file, not the


yarikoptic:
but what will happen if used as a Python module - every script would need to setup logging for our logger and thus we would loose ability to quickly enable logging via `DANDI_LOG_LEVEL` env var?

jwodder:
@yarikoptic `DANDI_LOG_LEVEL` still takes effect via the line in `dandi/__init__.py`, and if a user's script calls `basicConfig()`, then the logging level will be preserved, and dandi's logs will be sent to stderr.  Is that good enough, or do you want the dandi logger to always have a StreamHandler attached so that it always writes to stderr even without the user configuring anything?

yarikoptic:
> Is that good enough, or do you want the dandi logger to always have a StreamHandler attached so that it always writes to stderr even without the user configuring anything?

that is what I thought but I wonder if that is generally a good strategy for a Python module?  I did see double logging when datalad is used as a module and I wonder if that was a reason and should generally be avoided

jwodder:
@yarikoptic [The official Python logging tutorial has this to say about logging in a library](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library):

> It is strongly advised that you *do not add any handlers other than `NullHandler` to your library’s loggers*. This is because the configuration of handlers is the prerogative of the application developer who uses your library. The application developer knows their target audience and what handlers are most appropriate for their application: if you add handlers ‘under the hood’, you might well interfere with their ability to carry out unit tests and deliver logs which suit their requirements.

yarikoptic:
Thank you @jwodder!  One immediate aspect/"application" is logging during testing -- should log handler setup be added into a fixture?

yarikoptic:
may be should be ok to enable it only if `DANDI_LOG_LEVEL` env var is present

jwodder:
@yarikoptic pytest already captures logs, and [there's already a fixture for setting the log level during tests](https://github.com/dandi/dandi-cli/blob/d7d8b2d96df504c2bd736862cb60c6d9513360b4/dandi/tests/fixtures.py#L29-L31).

yarikoptic:
ah ok -- let's proceed then

PR #474, opened 2021-03-17
ENH: prototype for the DANDIEtag "digester"

### LINKED ISSUE #464: NF: dandi.core to contain dandi-s3-etag logic
The main purpose is to provide a collection of core utilities (very "stable") for reuse by both dandi-api and dandi-cli, such as the S3 ETag chunking strategy which ATM resides outside of "dandi" (ref: https://github.com/dandi/dandi-api/issues/146).  We definitely do should not maintain copies of such code, or have it outside of dandi space.
 
I think it would have been better as a dedicated `dandicore` module but if dandi-api would depend on `dandi-cli` anyways (e.g. for metadata validation etc), so the benefit is not so obvious, and I dislike proliferation of modules/packages (requires more effort to maintain distributions etc).

WDYT (attn: @dandi/dandiarchive and @dandi/dandi-cli ) ?

yarikoptic:
This is an initial interface which could be used to get a complete
digest for a file (serially ATM, but could be parallelized across parts),
just a list of part sizes, or be used iteratively to update a digest
e.g. during upload or download as more data comes in.

Closes #464 

TODOs
- [x] refine design/interface given the feedback
- [x] add copyright/license terms for borrowed piece of code
- [x] add tests for some sample files with known etags, mocking for some large ones

Later, after merge so could be used by dandi-api:
- use within `upload` and `download` for new dandi-api
- verify etag matching for some large files already uploaded to S3

### DIFF
--- a/dandi/core/digests/dandietag.py
+++ b/dandi/core/digests/dandietag.py
@@ -1,8 +1,9 @@
 # from https://github.com/girder/django-s3-file-field/blob/master/s3_file_field/_multipart.py
+from dataclasses import dataclass
 from hashlib import md5
 import math
 import os
-from typing import List, Optional, Tuple, Union
+from typing import Iterator, List, NamedTuple, Optional, Union
 
 
 def mb(bytes_size: int) -> int:
@@ -17,9 +18,17 @@ def tb(bytes_size: int) -> int:
     return bytes_size * 2 ** 40
 
 
-class DandiETag:
-    REGEX = r"[0-9a-f]{32}-\d{1,5}"
-    MAX_STR_LENGTH = 38
+class Part(NamedTuple):
+    number: int
+    offset: int
+    size: int
+
+
+@dataclass
+class PartGenerator:
+    part_qty: int
+    initial_part_size: int
+    final_part_size: int
 
     # S3 multipart limits: https://docs.aws.amazon.com/AmazonS3/latest/dev/qfacts.html
     # 10k is the maximum number of allowed parts allowed by S3
@@ -29,31 +38,9 @@ class DandiETag:
     # 5GB is the maximum part size allowed by S3
     MAX_PART_SIZE = gb(5)
 
-    def __init__(self, file_size: int):
-        self._file_size: int = file_size
-        self._part_sizes: Optional[Tuple[int, ...]] = None
-        self._md5_digests: List[bytes] = []
-
-    @property
-    def part_sizes(self) -> Tuple[int, ...]:
-        if self._part_sizes is None:
-            self._part_sizes = self.gen_part_sizes(self._file_size)
-        return self._part_sizes
-
-    def as_str(self) -> str:
-        if len(self._md5_digests) != len(self._part_sizes):
-            raise ValueError(
-                f"Cannot compute ETag with only {len(self._md5_digests)} out of"
-                f" {len(self._part_sizes)} parts collected"
-            )
-        parts_digest = md5(b"".join(self._md5_digests)).hexdigest()
-        return f"{parts_digest}-{len(self._md5_digests)}"
-
     # from https://github.com/girder/django-s3-file-field/blob/master/s3_file_field/_multipart.py
-    # but removing yielding sequential index. Could be wrapped
-    # with enumerate where needed
     @classmethod
-    def gen_part_sizes(cls, file_size: int) -> Tuple[int, ...]:
+    def for_file_size(cls, file_size: int) -> "PartGenerator":
         """Method to calculate sequential part sizes given a file size"""
         part_size = mb(64)
 
@@ -69,11 +56,73 @@ class DandiETag:
         if part_size > cls.MAX_PART_SIZE:
             part_size = cls.MAX_PART_SIZE
 
-        d, m = divmod(file_size, part_size)
-        sizes = [part_size] * d
-        if m:
-            sizes.append(m)
-        return tuple(sizes)
+        part_qty, final_part_size = divmod(file_size, part_size)
+        if final_part_size == 0:
+            final_part_size = part_size
+        else:
+            part_qty += 1
+        if part_qty == 1:
+            part_size = final_part_size
+        return cls(part_qty, part_size, final_part_size)
+
+    def __len__(self) -> int:
+        return self.part_qty
+
+    def __getitem__(self, index: int) -> Part:
+        if 1 <= index < self.part_qty:
+            return Part(
+                index, self.initial_part_size * (index - 1), self.initial_part_size
+            )
+        elif index == self.part_qty:
+            return Part(
+                index, self.initial_part_size * (index - 1), self.final_part_size
+            )
+        else:
+            raise IndexError(index)
+
+    def __iter__(self) -> Iterator[Part]:
+        offset = 0
+        for number in range(1, self.part_qty):
+            yield Part(number, offset, self.initial_part_size)
+            offset += self.initial_part_size
+        yield Part(self.part_qty, offset, self.final_part_size)
+
+
+class DandiETag:
+    REGEX = r"[0-9a-f]{32}-\d{1,5}"
+    MAX_STR_LENGTH = 38
+
+    def __init__(self, file_size: int):
+        self._part_gen: PartGenerator = PartGenerator.for_file_size(file_size)
+        self._md5_digests: List[Optional[bytes]] = [None] * len(self._part_gen)
+        self._next_index: int = 0
+
+    @property
+    def part_qty(self) -> int:
+        return len(self._part_gen)
+
+    @property
+    def complete(self) -> bool:
+        return self._next_index == self.part_qty
+
+    def get_parts(self) -> Iterator[Part]:
+        return iter(self._part_gen)
+
+    def next_part(self) -> Optional[Part]:
+        if self._next_index < self.part_qty:
+            return self._part_gen[self._next_index + 1]
+        else:
+            return None
+
+    def as_str(self) -> str:
+        if not self.complete:
+            raise ValueError("Not all part hashes submitted")
+        blob = b""
+        for d in self._md5_digests:
+            assert d is not None
+            blob += d
+        parts_digest = md5(blob).hexdigest()
+        return f"{parts_digest}-{len(self._md5_digests)}"
 
     @classmethod
     def from_file(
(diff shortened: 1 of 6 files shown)


dandi/core/digests/dandietag.py
```diff
@@ -0,0 +1,108 @@
...
+    def update(self, block):
+        """Update etag with the new block of data"""
+        if len(self._md5_digests) == self.part_sizes:
+            raise RuntimeError(
+                "Trying to update {self} with a new block having already processed {len(self._md5_digests)"
+            )
+        self._md5_digests.append(md5(block).digest())
+
+
+if __name__ == "__main__":
```
yarikoptic, line 103: this helper could be removed

yarikoptic:
@jwodder feel welcome to RF the way you see it being improved and deciding/following on feedback (when/if any would come in)

dandi/core/digests/dandietag.py
```diff
@@ -0,0 +1,103 @@
...
+
+def gb(bytes_size: int) -> int:
+    return bytes_size * 2 ** 30
+
+
+def tb(bytes_size: int) -> int:
+    return bytes_size * 2 ** 40
+
+
+class DANDIEtag:
```
dchiquito, line 21:
```suggestion
class DANDIETag:
```
ETag has two capital letters.

dchiquito, line 21:
```suggestion
class DandiETag:
```

This is how we've been naming classes in dandi-api (`DandiSettings`, `DandiPagination`, `DandisetViewSet`, etc). I think treating "Dandi" were a proper noun makes it much more wieldy, even if it is technically an acronym.

dandi/core/digests/dandietag.py
```diff
@@ -0,0 +1,101 @@
...
+    return bytes_size * 2 ** 30
+
+
+def tb(bytes_size: int) -> int:
+    return bytes_size * 2 ** 40
+
+
+class DandiETag:
+    REGEX = r"[0-9a-f]{32}-\d{1,4}"
+    MAX_STR_LENGTH = 37
```
jwodder, line 22: @yarikoptic What is this based on?  Wouldn't the etag suffixes go up through `-10000`, meaning that the regex should end with `\d{1,5}` and the `MAX_STR_LENGTH` should be 38?

dandi/core/digests/dandietag.py
yarikoptic: right -- and we have `DandiAPIClient` so let's make it `DandiETag` here.

dandi/core/digests/dandietag.py
yarikoptic: yes, I miscalculated I guess (was thinking about  `5,000`?) please fix accordingly

jwodder:
@yarikoptic 

* In order to support parallel computation, shouldn't this be able to take piece hashes in any order?
* Would it be useful to replace the return type of `gen_part_sizes()` with a class with methods for querying both the size as well as the offset of each piece?

jwodder:
@yarikoptic Also, I don't like the idea of using `__str__` to return the etag with a potential for failure; I'd rather there be a `get_etag()` method.  What do you think?

jwodder:
@yarikoptic This may be over-engineering things a bit, but I'm considering the following API for this:

* `Piece(NamedTuple)` — tells an external digester what part of the file to hash
    * `number: int` — starts at 1
    * `offset: int`
    * `size: int`
* `PieceGenerator` — generates `Piece`s
    * `classmethod for_file_size(size: int) -> PieceGenerator` — what is currently `DandiETag.gen_part_sizes()`
    * `__iter__() -> Iterator[Piece]`
    * `__len__() -> int`
    * `__getitem__(number: int) -> Piece` — numbers start at 1
    * Private(-ish) fields:
        * `piece_qty: int`
        * `initial_piece_size: int` — size of each piece other than the last
        * `final_piece_size: int` — size of the last piece
* `DandiETag`
    * `get_pieces() -> Iterator[Piece]`
    * `next_piece() -> Optional[Piece]` — returns the lowest-numbered `Piece` not yet hashed
    * `needed_pieces() -> Iterator[Piece]` — returns an iterator of all `Piece`s not yet hashed
    * `register_hash(p: Piece, hash: bytes) -> None` — errors if piece already hashed or piece number out of range
    * `register_next_hash(hash: bytes) -> none` — provide hash for lowest-numbered `Piece` not yet hashed; ensuring this is used correctly is the caller's job
    * `get_etag() -> str` — errors if not all piece hashes provided

yarikoptic:
> * In order to support parallel computation, shouldn't this be able to take piece hashes in any order?

YES!  I thought about it and forgot to add a note: it is partially why I cached/stored precomputed sizes, so we could initiate _md5_digests with None for the needed length.  Then I am thinking `update_part(data, index)` could be used to update any part.  For `update(data)` we would just keep an internal index to be incremented after each `update` and if we run into an already non-`None` value -- `raise RuntimeError`

> * Would it be useful to replace the return type of `gen_part_sizes()` with a class with methods for querying both the size as well as the offset of each piece?

right! -- for parallel computation we would need that indeed to not require "user code" to compute all the offsets.  So may be we should just make it `gen_parts` which would yield `(offset, size)` pairs with a "guarantee" that they do cover the full `file_size`.

> @yarikoptic Also, I don't like the idea of using `__str__` to return the etag with a potential for failure; I'd rather there be a `get_etag()` method. What do you think?

I was not certain about `__str__` myself primarily because it is "undefined" until all parts are processed. But I would prefer a more generic `as_str()` in favor of `get_etag()` since that one might mean returning some base non-str ETag instance or smth like that.

yarikoptic:
> @yarikoptic This may be over-engineering things a bit, but I'm considering the following API for this:

> * `Piece(NamedTuple)` — tells an external digester what part of the file to hash

I think having such a named tuple should not hurt and could indeed be nice.  Why `Piece` and not `Part` which would relate nicely  to "multi part uploads"?

I am not sure about generalizing `PieceGenerator` and somewhat complicating `DandiETag`.  

>   * `needed_pieces() -> Iterator[Piece]` — returns an iterator of all `Piece`s not yet hashed, in ascending order by number

sounds like might be useful but I do not see a need for it ATM so not sure if worth bothering with it ATM.

jwodder:
@yarikoptic 

> Why Piece and not Part which would relate nicely to "multi part uploads"?

No reason, though I see now that `Part` would be consistent with the rest of the names in the code.

> I am not sure about generalizing PieceGenerator and somewhat complicating DandiETag.

The Piece/PartGenerator would just be a private attribute of `DandiETag`, not exposed to any users, that would exist as a single class in order to keep related logic together.  The structure would also be a more efficient storage of part sizes than a lengthy tuple of ints, as it only needs to store the number of parts and the sizes of the initial & last parts.

yarikoptic:
I am ok with internal helpers if would help.  But not sure if I worth bothering with storage optimization for this or exposing that much of "pieces/parts" interfaces in `DandiETag` class.

dandi/core/digests/dandietag.py
yarikoptic: I would say even "shout be removed" in favor of a dedicated #477

jwodder:
@yarikoptic Do we want to cache `DandiETag.from_file()`?  Because I'm getting an error when combining `@classmethod` with `@memoize_path` (a different one depending on what order they're in), so we'd probably have to either move `from_file()` to a function or else make "etag" a digest method of `dandi.support.digests.get_digest()`.

yarikoptic:
what kind of error -- I feel like it might be something to fix on fscacher side? There are dangers about caching classmethod calls (since might rely on some class.attributes), but I think they aren't worse than those of closures, so I am ok to allow for that and leave it for documentation to explain/developer to decide.

jwodder:
@yarikoptic When I put the `@classmethod` above the `@memoize_path`, I get an error indicating that `path` [here](https://github.com/con/fscacher/blob/39bb67c67105a494e5185f565645a1d992196d54/src/fscacher/cache.py#L87-L91) is the `DandiETag` class instead of the path argument to the classmethod.  When I put the `@memoize_path` above the `@classmethod`, I get an error on [this line](https://github.com/con/fscacher/blob/39bb67c67105a494e5185f565645a1d992196d54/src/fscacher/cache.py#L108) saying that `'classmethod' object is not callable`.

jwodder:
@yarikoptic 
* So what should be done about caching this digest?
* Did I add the credit to the original source authors correctly?
* Any other changes desired?

yarikoptic:
> * So what should be done about caching this digest?

Let's keep `core.` unencumbered by fscache. Please extend dandi.support.digests.get_digest to use `DandiEtag` instead of `Digester` for `digest="dandi-etag"`

dandi/core/digests/dandietag.py
```diff
@@ -0,0 +1,169 @@
...
+    def from_file(
+        cls, path: Union[str, bytes, "os.PathLike[str]", "os.PathLike[bytes]"]
+    ) -> "DandiETag":
+        etag = cls(file_size=os.path.getsize(path))
+        with open(path, "rb") as f:
+            for part in etag.get_parts():
+                etag.update(f.read(part.size))
+        return etag
+
+    def add_digest(self, p: Part, part_digest: bytes) -> None:
```
yarikoptic, line 139: I would prefer to not expose in public interface methods exposing internal assumption (i.e. how `part_digest` is computed) and rather only operate on `block`s.  Please make `add_digest` and `add_next_digest` protected, and either add  `update_part(self, block: bytes, p: Part)` or make `update(self, block: bytes, p: Part=None)` so it would switch between the two depending either `p` is provided (that would also allow to centralize computing of md5() on a part).

dandi/core/digests/dandietag.py
```diff
@@ -0,0 +1,169 @@
+# Derived from <https://github.com/girder/django-s3-file-field/blob/master/
+# s3_file_field/_multipart.py>, copyright Kitware, Inc. <kitware@kitware.com>
+# under the Apache 2.0 license
```
yarikoptic, line 3: Please also add to `## 3rd party components included` of README.md where we have similar mentions ATM (we might want to move them to LICENSE later on)

yarikoptic (commented):
Thank you @jwodder!  Left the comments (can't request changes since I initiated the PR): more tests, slight change in API

dandi/support/digests.py
```diff
@@ -77,4 +77,9 @@ def __call__(self, fpath):
 
 @checksums.memoize_path
 def get_digest(filepath, digest="sha256"):
-    return Digester([digest])(filepath)[digest]
+    if digest == "dandi-etag":
+        from ..core.digests.dandietag import DandiETag
```
yarikoptic, line 81: any reason to not have it at the top of the file?

dandi/core/digests/dandietag.py
```diff
@@ -0,0 +1,169 @@
...
+            raise ValueError("File is larger than the S3 maximum object size.")
+
+        if math.ceil(file_size / part_size) >= cls.MAX_PARTS:
+            part_size = math.ceil(file_size / cls.MAX_PARTS)
+
+        if part_size < cls.MIN_PART_SIZE:
+            part_size = cls.MIN_PART_SIZE
+
+        if part_size > cls.MAX_PART_SIZE:
+            part_size = cls.MAX_PART_SIZE
```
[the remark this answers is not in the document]
jwodder: The above two conditions are actually unreachable.  `part_size` will always be at least 64 MiB, and the only way for `MAX_PART_SIZE` to be exceeded is for the file size to be greater than `MAX_PART_SIZE * MAX_PARTS`, which is greater than 5 TiB, the maximum file size.  I believe these lines are only covered in the original code because the tests monkey-patch the starting value of `part_size`.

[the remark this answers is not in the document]
jwodder: How do you recommend I do a sample file upload, seeing as the CLI hasn't been updated for the new upload API yet?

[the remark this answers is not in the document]
yarikoptic: indeed gets tricky/chicken-egg.  Could postpone and get those while RFing `upload` or just initiating upload manually via API directly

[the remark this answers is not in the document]
jwodder: I looked at the ETags for some zero-byte entries in s3://dandiarchive.  The only such entries were directories, but they both had ETags of `d41d8cd98f00b204e9800998ecf8427e`, which is the expected hash but without the part quantity (which I think S3 leaves off if there's only one part).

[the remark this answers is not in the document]
yarikoptic: re 0 or one part uploads: I would not rely on what you would see in `dandiarchive` bucket since it was uploaded possibly using a different strategy (thus might not even be "always in parts").  A fresh upload should be done to figure it out "for sure".

re unreachable code/conditions -- I would then either remove them or just convert into asserts (to make sure we do not somehow beef up part_size and then fail to acknowlede that down in the code or smth like that)

jwodder:
@yarikoptic Can this PR be merged now?  I'll add the zero-byte etag test later.

yarikoptic:
Sure, let's proceed

PR #480, opened 2021-03-22
Add "digest" command

### LINKED ISSUE #477: digest command
Primarily to facilitate user's access to `dandi-etag` digest which is to be introduced in https://github.com/dandi/dandi-cli/pull/474 .  Could be folded into #474 or developed independently after/on top.  Should by default provide our `dandi-etag` but allow for estimation of other digests we have supported in Digester as well, in particular md5 and sha256.

jwodder:
Closes #477.

### DIFF
--- a/dandi/cli/cmd_digest.py
+++ b/dandi/cli/cmd_digest.py
@@ -1,0 +1,24 @@
+import click
+
+from .base import map_to_click_exceptions
+from ..support.digests import get_digest
+
+
+@click.command()
+@click.option(
+    "-d",
+    "--digest",
+    "digest_alg",
+    type=click.Choice(
+        ["dandi-etag", "md5", "sha1", "sha256", "sha512"], case_sensitive=False
+    ),
+    default="dandi-etag",
+    help="Digest algorithm to use",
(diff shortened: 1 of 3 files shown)


jwodder:
The test failures are due to #479 not being included in this PR.

dandi/cli/cmd_digest.py
```diff
@@ -0,0 +1,25 @@
...
+    show_default=True,
+)
+@click.argument("paths", nargs=-1, type=click.Path(exists=True, dir_okay=False))
+@map_to_click_exceptions
+def digest(paths, digest_alg):
+    """ Calculate file digests """
+    from ..support.digests import get_digest
+
+    for p in paths:
+        print(f"{p}:", get_digest(p, digest=digest_alg))
```
yarikoptic, line 25: please add a rudimentary unit test similar to the ones we have for other `cmd_` code

jwodder: Done.

yarikoptic:
Thank you! (I will assume that all CI fails are due to upload RFing...)

Issue #478, opened 2021-03-18
RF/BF: adjust dandi-api `upload` for changes in the API


yarikoptic:
https://github.com/dandi/dandi-api/pull/156 which followed https://github.com/dandi/dandi-api/pull/150 was merged and deployed, DB wiped clean, blank dandisets reuploaded. 

- `dandi-cli upload` needs to be adjusted to reflect API changes.
- would need to sit on top of #474 since we would need to compute/provide dandi-etag to the API
- adjusted dandi-cli should be tried on uploads of small and large files across available dandisets

jwodder:
@yarikoptic @dchiquito So the changes that affect the CLI side are:

* The payload to & response from the `/uploads/initialize/` endpoint have changed; it now returns a UUID which needs to be included in the paths of the `/uploads/complete/` and `/uploads/validate/` endpoints.
    * The payload now needs to include a hash of the file; what algorithms are accepted here?
* The payload to the `/uploads/validate/` endpoint is gone.
* An ETag digest for the uploaded file needs to be provided, but where?  In the metadata or in the payload to `/uploads/initialize`?
* The `/uploads/validations/{filehash}/` endpoint is gone; should the related functionality just be removed from the CLI?
* Currently, the CLI starts out an upload by making a request to `/uploads/validate/` with the sha256 digest of the file in order to check whether the blob has already been uploaded to the server; what is the equivalent way to do this now?
* Anything else?

dchiquito:
@jwodder
* `/uploads/initialize/` should take a body like this:
```
{
  "contentSize": 12345,
  "digest": {
    "algorithm": "dandi:dandi-etag",
    "value": "..."
  }
}
```
The value should be calculated using the `DandiETag` thingy that's being added to dandi.core. This value is the ETag digest for the file being uploaded.

* The background processes no longer block the CLI, so yes, there is no reason for the CLI to wait for them to finish. After calling `POST /dandisets/{versions__dandiset__pk}/versions/{versions__version}/assets/` the CLI is done with the upload.

* You can call `POST /blobs/digest/` with `{"algorithm": "dandi:dandi-etag", "value": "..." }`. Alternatively, if you try to start the upload with `/uploads/initialize`, it should throw an error 409 with the `Location` header set to the UUID of an existing asset blob, if an asset blob with a matching etag exists.

dchiquito:
[...]

PR #481, opened 2021-03-22
Add dandi_etag digest type to schema

jwodder:
See [here](https://github.com/dandi/dandi-cli/pull/479#discussion_r598897361) and related conversation.

Note that I changed the name of the enum from `dandi-etag` to `dandi_etag` so that it would be a valid Python identifier.

### DIFF
--- a/dandi/consts.py
+++ b/dandi/consts.py
@@ -156,4 +156,4 @@ class routes(object):
     dandiset_draft = "{dandi_instance.redirector}/dandiset/{dandiset[identifier]}/draft"
 
 
-DANDI_SCHEMA_VERSION = "0.1.0"
+DANDI_SCHEMA_VERSION = "0.1.1"
--- a/dandi/model_types.py
+++ b/dandi/model_types.py
@@ -62,7 +62,7 @@ DigestTypeDict = {
             "rdfs:label": "SHA-512",
         },
         {
-            "@id": "dandi:dandi_etag",
+            "@id": "dandi:dandi-etag",
             "@type": "dandi:DigestType",
             "rdfs:comment": "S3-style ETag",
             "rdfs:label": "DANDI ETag",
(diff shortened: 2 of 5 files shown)


yarikoptic:
> Note that I changed the name of the enum from `dandi-etag` to `dandi_etag` so that it would be a valid Python identifier.

does it need to be a proper Python identifier?

jwodder:
@yarikoptic The name gets used as the name of the enum instance, so if we ever want to use the enum as an enum, yes, it needs to be a valid identifier.

yarikoptic:
I see that we also have e.g. `dandi:CCBYNC40` (which imho should better be `dandi:CC-BY-NC-40` or alike) and wonder if we haven't already catered to some pydantic/python inflicted limitation a bit too much.

I think it matters only for `klass` name generation, so may be we should just `klass = klass.replace('-', '_')` and thus allow `@id` to contain `-` and thus remain more readable etc.  Or there is more to it, @satra?

I see also in that code of `create_enum` that we use
```python
            if key in items:
                key = item["@id"].replace(":", "_")
```
but I do not think we use this one anywhere, thus could be removed?

If we remove, then I think we could even have 1-to-1 mapping from Enum klass name back into original `@id` value if so desired, so we keep `_` in Python classes and `-` in json schema values etc.  I think it might be useful in the long run.

yarikoptic:
@jwodder if you do not see immediate danger in replacing `-` with `_` for Enum `klass` name minting, to not keep upload RFing #478 blocked, let's introduce that mapping, while keeping it `dandi-etag` so it is aligned with what API expects ATM.

jwodder:
@yarikoptic Done.

satra:
feel free to go improving identifier names. i wouldn't go crazy but anything allowed in a URL is fine.

yarikoptic:
ok, let's proceed with this

PR #486, opened 2021-03-23
Remove "current" column from upload progress display

jwodder:
When uploading to the new API, when the code actually reaches the point at which the upload proper happens, a "CURRENT" column appears at the right side of the pyout display containing the current number of bytes uploaded per file.  As this column is not declared in the list passed to the pyout table constructor, I assume it is not supposed to be displayed in the first place.  This PR removes the "current" field from the upload progress dict before yielding it to pyout, which should get rid of the column.

### DIFF
--- a/dandi/upload.py
+++ b/dandi/upload.py
@@ -829,8 +829,7 @@ def _new_upload(
             validating = False
             for r in client.iter_upload(ds_identifier, "draft", metadata, str(path)):
                 if r["status"] == "uploading":
-                    uploaded_paths[str(path)]["size"] = r["current"]
-                    r.pop("current")
+                    uploaded_paths[str(path)]["size"] = r.pop("current")
                     yield r
                 elif r["status"] == "post-validating":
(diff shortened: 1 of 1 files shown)


dandi/upload.py
```diff
@@ -830,6 +830,7 @@ def process_path(path, relpath):
             for r in client.iter_upload(ds_identifier, "draft", metadata, str(path)):
                 if r["status"] == "uploading":
                     uploaded_paths[str(path)]["size"] = r["current"]
+                    r.pop("current")
```
yarikoptic, line 833:
```suggestion
                    uploaded_paths[str(path)]["size"] = r.pop("current")
```
?

yarikoptic (approved):
THANK YOU!  I kept annoyed with it myself, not yet sure if there would be any side-effect. Just recommended a little change

yarikoptic:
Thank you @jwodder for the PR and a detailed description. Let's release!

PR #489, opened 2021-03-24
dandi ls: Error if --schema is given with remote resource of different version

### LINKED ISSUE #485: schema option has no effect
i don't believe the schema flag has any effect here. i also tried 0.1.0.

`$ dandi ls --format  yaml --schema 0.1.1 https://dandiarchive.org/dandiset/000037/draft`

jwodder:
Closes #485.


yarikoptic:
Thanks, let's proceed

PR #491, opened 2021-03-24
Don't hardcode DANDI_SCHEMA_VERSION value in tests

### LINKED ISSUE #490: do not "hard code" DANDI_SCHEMA_VERSION in the tests
ATM
```
dandi/tests/test_metadata.py:  "schemaVersion": "0.1.1",
dandi/tests/test_metadata.py:  "schemaVersion": "0.1.1",
dandi/tests/test_metadata.py:  "schemaVersion": "0.1.1",
```
which requires changing the test code every time we would beef up version even if no changes to the tests are needed.  I think we should just use `DANDI_SCHEMA_VERSION` for this

jwodder:
Closes #490.

### DIFF
--- a/dandi/tests/test_metadata.py
+++ b/dandi/tests/test_metadata.py
@@ -4,6 +4,7 @@ import json
 from dateutil.tz import tzutc
 import pytest
 
+from ..consts import DANDI_SCHEMA_VERSION
 from ..metadata import (
     metadata2asset,
     migrate2newschema,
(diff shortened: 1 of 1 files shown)


dandi/tests/test_metadata.py
```diff
@@ -105,7 +106,7 @@ def test_metadata2asset(schema_dir):
     # data.json(exclude_unset=True, exclude_none=True, indent=2)
     json_data = """{
   "identifier": "0b0a1a0b-e3ea-4cf6-be94-e02c830d54be",
-  "schemaVersion": "0.1.1",
+  "schemaVersion": null,
```
yarikoptic, line 109: I think it would be best to validate that it corresponds to current `DANDI_SCHEMA_VERSION` or do you think it is not worth that?

jwodder: I believe that's taken care of by the `validate_asset_json()` call.

yarikoptic: ok then let's proceed, thanks

PR #493, opened 2021-03-24
Lowercase "sha256"

### LINKED ISSUE #482: `dandi:SHA256` -> `dandi:sha256`
spotted while reviewing #481 although saw it before in isolation from others thus it didn't trigger my attention: 
I didn't check how it came about capitalized but we have `dandi:sha1` and `dandi:sha512`etc, and I do not see why SHA256 should be capitalized.  I think we should keep them all lower case for consistency etc.

NB I also thought to say that we should align them to https://github.com/multiformats/multicodec/blob/master/table.csv  but I see now that it is more specific there and then should be `sha2-256` ... not sure if we should jump to align with the multicodec right away .

jwodder:
Closes #482.

### DIFF
--- a/dandi/model_types.py
+++ b/dandi/model_types.py
@@ -50,7 +50,7 @@ DigestTypeDict = {
             "rdfs:label": "SHA-1",
         },
         {
-            "@id": "dandi:SHA256",
+            "@id": "dandi:sha256",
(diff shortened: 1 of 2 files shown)


yarikoptic:
Great, thanks!

PR #495, opened 2021-03-24
Validate uploads before digesting

### LINKED ISSUE #492: upload: move digestion after validation
ATM user might wait for an hour of digestion only to later error out due to validation error.  I think we should delay digestion until right before the upload

PS does not need to change for girder implementation - to go away soon anyways

jwodder:
Closes #492.

### DIFF
--- a/dandi/upload.py
+++ b/dandi/upload.py
@@ -700,6 +700,45 @@ def _new_upload(
                 return
 
             #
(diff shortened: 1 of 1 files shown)


yarikoptic:
Looks kosher, thank you @jwodder !

PR #496, opened 2021-03-24
Check for already-uploaded blobs via /uploads/initialize/ instead of /blobs/digest/

### LINKED ISSUE #494: upload: avoid an explicit check for blob existence, just catch /initialize 409
There is no longer need for an explicit check for blob existence:
```python
        try:
            resp = self.post(
                "/blobs/digest/",
                json={"algorithm": "dandi:dandi-etag", "value": filetag},
            )
        except requests.HTTPError as e:
            if e.response.status_code == 404:
                lgr.debug("%s: Blob does not already exist on server", asset_path)
                blob_exists = False
            else:
                raise
```
please just proceed to `/uploads/initialize/` and react to 409 to decide that blob is already there .  That would make code more robust and avoid race condition if the same blob is uploaded meanwhile and then `initialize` would return 409.

Also please add `yield {"status": "initiating upload"}` right before then since otherwise client would show `{'status': 'calculating etag'}` for awhile although already proceeding to upload

jwodder:
Closes #494.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -416,24 +416,10 @@ class DandiAPIClient(RESTFullAPIClient):
                         f" {digest['value']} but is now {filetag}"
                     )
                 break
+        yield {"status": "initiating upload"}
+        lgr.debug("%s: Beginning upload", asset_path)
+        total_size = os.path.getsize(filepath)
         try:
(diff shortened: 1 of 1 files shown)


yarikoptic:
Great thank you @jwodder !

PR #499, opened 2021-03-24
Upload file parts in parallel

### LINKED ISSUE #498: upload: (dandi-api) parallelize upload of parts
I see upload speeds jumping between 1-50 MBps from drogon with dandi-cli reporting 22 MB/s average ATM.  I do not think it is network bottlnecked really, and neither it seems to be CPU bound.

With all the multiple upload urls we should be able to handle blocks (including digests per block) in parallel with benefit even if simply using threads. With original girder based upload I think there were no way to parallelize individual file upload, so we resorted to parallization across files.  I think we might reconsider that (although it would still benefit initial digestion and validation) or just compliment with parallelization within file (across parts) upload.

jwodder:
Closes #498.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -366,7 +366,7 @@ class DandiAPIClient(RESTFullAPIClient):
             dandiset["metadata"] = dandiset_metadata.pop("dandiset")
         return dandiset
 
-    def upload(self, dandiset_id, version_id, asset_metadata, filepath):
+    def upload(self, dandiset_id, version_id, asset_metadata, filepath, jobs=None):
         """
         Parameters
         ----------
@@ -381,10 +381,12 @@ class DandiAPIClient(RESTFullAPIClient):
         filepath: str or PathLike
           the path to the local file to upload
         """
-        for _ in self.iter_upload(dandiset_id, version_id, asset_metadata, filepath):
+        for _ in self.iter_upload(
+            dandiset_id, version_id, asset_metadata, filepath, jobs=jobs
+        ):
             pass
 
-    def iter_upload(self, dandiset_id, version_id, asset_metadata, filepath):
+    def iter_upload(self, dandiset_id, version_id, asset_metadata, filepath, jobs=None):
         """
         Parameters
         ----------
@@ -452,7 +454,7 @@ class DandiAPIClient(RESTFullAPIClient):
             lgr.debug("Uploading %s in %d parts", filepath, len(parts))
             with storage.session():
                 with open(filepath, "rb") as fp:
-                    with ThreadPoolExecutor() as executor:
+                    with ThreadPoolExecutor(max_workers=jobs or 5) as executor:
                         lock = Lock()
                         futures = [
                             executor.submit(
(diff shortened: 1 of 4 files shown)


dandi/dandiapi.py
```diff
@@ -620,3 +598,45 @@ def delete_asset_bypath(self, dandiset_id, version_id, asset_path):
...
+        raise RuntimeError(
+            f"Server and client disagree on size of upload part"
+            f" {part['part_number']}; server says {part['size']},"
+            f" client says {etag_part.size}"
+        )
+    with lock:
+        fp.seek(etag_part.offset)
+        chunk = fp.read(part["size"])
+    if len(chunk) != part["size"]:
+        raise RuntimeError(f"End of file {fp.name} reached unexpectedly early")
```
yarikoptic, line 615: while at it (although just an older code moved):
```suggestion
    if len(chunk) != part["size"]:
        raise RuntimeError(f"End of file {fp.name} reached unexpectedly early: read {len(chunk)} from expected {part['size']}")
```

dandi/dandiapi.py
```diff
@@ -450,53 +452,29 @@ def iter_upload(self, dandiset_id, version_id, asset_metadata, filepath):
             lgr.debug("Uploading %s in %d parts", filepath, len(parts))
             with storage.session():
                 with open(filepath, "rb") as fp:
-                    for part, etag_part in zip(parts, etagger.get_parts()):
-                        if part["size"] != etag_part.size:
-                            raise RuntimeError(
-                                f"Server and client disagree on size of upload part"
-                                f" {part['part_number']}; server says"
-                                f" {part['size']}, client says {etag_part.size}"
+                    with ThreadPoolExecutor() as executor:
```
yarikoptic, line 455: I think that default behavior of "If max_workers is None or not given, it will default to the number of processors on the machine, multiplied by 5" would be way too much on many contemporary boxes - we do not want to start 300 parallel uploads for a file.  I think we should
- add `jobs=5` kwarg here 
- percolate its specification to the `dandi upload`.  apparently we do not even have `--jobs` for upload (only for `download`) and rely on `pyout`'s default instead of specifying `max_workers` to it.  So we need to add `--jobs` to pass max_workers to `LogSafeTabular` but also somehow allow for specification of number of jobs per file upload.  Could be separate `--jobs-per-file` (or `--threads-per-file`?) or make `--jobs` to  support `[jobs][:threads]` specification so that user could specify e.g. `--jobs 2:10` which would instruct 2 files processed in parallel with 10 threads for upload per file.

WDYT @jwodder would be the best way to expose this "multiplicity" of `--jobs`?

yarikoptic (changes requested):
need to expose number of jobs (files)/threads (per file) for `upload`

dandi/dandiapi.py

dandi/dandiapi.py
jwodder: @yarikoptic I prefer the `--jobs 2:10` syntax.

jwodder: @yarikoptic `--jobs` option added.

yarikoptic:
Looks kosher and let's try it in the wild (on that 3.6TB reupload, will redo due to https://github.com/dandi/dandi-api/issues/184 with 10 threaded jobs, let's see how that goes)

PR #500, opened 2021-03-25
Add option for only running new Dandi API tests

jwodder:
Part of https://github.com/dandi/dandi-api/issues/42.

Note that this PR will need to be part of a release of dandi-cli before the dandi-api+dandi-cli integration tests can use the released version of dandi-cli.  If you are OK with releasing immediately, add the "release" tag to this PR before merging.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,11 +23,18 @@ jobs:
           - 3.7
           - 3.8
           - 3.9
+        dandi_api:
+          - false
         dev_deps:
           - false
         include:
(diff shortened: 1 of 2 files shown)


yarikoptic (changes requested):
I will add release tag

Let's add a matrix run on our CI here to ensure we don't break handling this option

jwodder:
@yarikoptic CI run added & passed.

yarikoptic:
GREAT! Thank you @jwodder !

PR #505, opened 2021-03-26
Fixes conversion of existing dandiset metadata with sub-object validation

### LINKED ISSUE #503: metadata conversion should not create empty objects
when we convert metadata from dandi/girder to the new schema, we should not create empty objects like contributor in the following:

```
{"name": "foo",
 "description": "foo",
 "contributor": [{"schemaKey": "Person"}]}
```

it should either include the required fields of that object (e.g., name)

```
{"name": "foo",
 "description": "foo",
 "contributor": [{"name": "Last, First", "schemaKey": "Person"}]}
```

or no object

```
{"name": "foo",
 "description": "foo",
 "contributor": []}
```

or no property

```
{"name": "foo",
 "description": "foo"}
```

satra:
closes #503 

Current dandiset metadata conversion was generating invalid values for some properties (Person, Resources, etc.,.). This updates the migration code to address these issues.

Also the current assets metadata uses modality, which is being renamed to approach.

### DIFF
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -424,7 +424,6 @@ def toContributor(value, contrib_type):
         value = [value]
     out = []
     for item in value:
-        print(item)
         if item == {"orcid": "", "roles": []}:
             continue
         contrib = {}
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -521,7 +521,6 @@ def convertv1(data):
                                 item["relation"] = models.RelationType.IsDerivedFrom
                         out.append(models.Resource(**item))
                     elif not any(item in val.dict().values() for val in out):
-                        print(item)
                         out.append(
                             models.Resource(
                                 url=item, relation=models.RelationType.IsDescribedBy
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -452,10 +452,10 @@ def toContributor(value, contrib_type):
             #    contrib["identifier"] = models.PropertyValue()
             del item["orcid"]
         if "affiliation" in item:
-            item["affiliation"] = [models.Organization(**{"name": item["affiliation"]})]
+            item["affiliation"] = [models.Organization(name=item["affiliation"])]
         if "affiliations" in item:
             item["affiliation"] = [
-                models.Organization(**{"name": affiliate})
+                models.Organization(name=affiliate)
                 for affiliate in item["affiliations"]
             ]
             del item["affiliations"]
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -424,6 +424,9 @@ def toContributor(value, contrib_type):
         value = [value]
     out = []
     for item in value:
+        print(item)
+        if item == {"orcid": "", "roles": []}:
+            continue
         contrib = {}
         if "name" in item:
             name = item["name"].split()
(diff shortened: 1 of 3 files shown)


satra:
when this is merged, dandiset.yaml should be regenerated and pushed to server.

satra:
@jwodder - how are we doing schema updates? is it automatic? or do i need to do something in the PR?

jwodder:
@satra Do you mean publication to dandi/schema?  That happens whenever a commit is made to master (including merging a PR) that touches `dandi/model*.py` or `tools/pubschemata.py`.

satra:
this is ready for review and potentially merging if things are ok. as i said, we should regenerate the dandi-api-datasets and also push the new metadata to the api.

dandi/metadata.py
```diff
@@ -449,18 +451,21 @@ def toContributor(value, contrib_type):
             # else:
             #    contrib["identifier"] = models.PropertyValue()
             del item["orcid"]
+        if "affiliation" in item:
+            item["affiliation"] = [models.Organization(**{"name": item["affiliation"]})]
```
jwodder, line 455: `models.Organization(**{"name": item["affiliation"]})` can be written as just `models.Organization(name=item["affiliation"])`.

yarikoptic:
Needs our schemaVersion const boost

satra:
should the const upgrade be done here? or separately? and is that constant used to release the schema?

dandi/metadata.py
satra: updated to reflect this.

yarikoptic:
> should the const upgrade be done here? or separately? 

here since it is the PR changing the schema

> and is that constant used to release the schema?

AFAIK yes.

yarikoptic:
> as i said, we should regenerate the dandi-api-datasets and also push the new metadata to the api.

That one should get automagically updated and we use master version of the dandi-cli (https://github.com/dandi/dandi-api-datasets/blob/master/tools/validate-all.sh#L62) so should be all set whenever we merge it (might take awhile for cron job to complete though but we run it every hour unless there is already one still running)

yarikoptic:
> push the new metadata to the api.

for that the easiest (since `dandi-cli` is not updating dandiset metadata) might be to delete/recreate all dandisets on beta instance.  We could do that after merge -- but may be only for 26 and 27 only for now?

satra:
> might be to delete/recreate all dandisets on beta instance. We could do that after merge -- but may be only for 26 and 27 only for now?

can't you just PUT metadata changes to dandiset? instead of having to recreate things?

yarikoptic:
> > might be to delete/recreate all dandisets on beta instance. We could do that after merge -- but may be only for 26 and 27 only for now?
> 
> can't you just PUT metadata changes to dandiset? instead of having to recreate things?

It is possible via API but we were told to not do that in client :-P

yarikoptic:
We're have a script to recreate, so it is just faster

satra:
> It is possible via API but we were told to not do that in client :-P

in this case i authorize you!

satra:
this is ready for merge.

yarikoptic:
> > It is possible via API but we were told to not do that in client :-P
> 
> in this case i authorize you!

I don't think it is worth/easier to  do it in `dandi upload` ATM, so let's have migration script tuned: #510

PR #506, opened 2021-03-26
Run "provision" container in the foreground

jwodder:
This PR causes the Girder "provision" container to be run in the foreground when spinning up the Docker Compose setup.  This allows us to catch provisioning failures early and also causes the logs for the container to be displayed, allowing debugging.  Both are relevant to fixing the failures currently seen on https://github.com/dandi/dandi-api/pull/183.

In case you're wondering why a Girder container is being spun up for dandi-api tests: The dandi-api container requires that a Dandi Girder setup be available for some reason, so we have to spin up all containers.

### DIFF
--- a/dandi/tests/fixtures.py
+++ b/dandi/tests/fixtures.py
@@ -189,14 +189,17 @@ def docker_compose_setup():
 
     if create:
         run(
-            ["docker-compose", "up", "-d", "provision"],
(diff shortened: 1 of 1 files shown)


yarikoptic (changes requested):
CI isn't happy

jwodder:
@yarikoptic That's because of the problem in #507.  Give me a minute to rebase.

PR #509, opened 2021-03-26
Add "delete" command

### LINKED ISSUE #501: delete command
By popular request of @satra and to reflect that web ui would never become as flexible as shell globbing, we need `dandi delete`.

- should be pretty much a trimmed down copy of `download` in terms of interface/handling inputs (paths or URLs) and just instead of `download` to `delete` of local and remote files
  - may be worth adding an option `--remote-only` to not bother with possibly present local files
- before jumping to the action, should ask user confirmation alike "Do you want to delete {len(assets)} from the {dandiset} on the server and {len(files)} local files?"  (later we might want to add a cmdline option to avoid this interaction, ok without for now since we better make it uniform with other commands like `upload` etc)
- tests should test against an instance of fixtured dandi-api that
  - deletion is forbidden by API server for assets in released versions
  - deletion in the draft does not delete assets (by downloading so we are certain that blobs are not removed from keystore) present in prior released versions
- no need for `girder` support - just raise NotImplementedError
 
Just ideas for later
- we might like to inform user on deletion of how many assets on remote server would cause those assets to no longer be present in the archive (i.e. not present in any other prior version of the dandiset)

jwodder:
Closes #501.

### DIFF
--- a/dandi/delete.py
+++ b/dandi/delete.py
@@ -1,6 +1,7 @@
 from dataclasses import dataclass, field
+from operator import attrgetter
 from pathlib import Path, PurePosixPath
-from typing import Iterable, Iterator, List, NamedTuple, Optional, Tuple
+from typing import Iterable, Iterator, NamedTuple, Optional, Set, Tuple
 
 import click
 import requests
@@ -24,7 +25,7 @@ class Deleter:
     client: Optional[DandiAPIClient] = None
     dandiset_id: Optional[str] = None
     deleting_dandiset: bool = False
-    remote_assets: List[RemoteAsset] = field(default_factory=list)
+    remote_assets: Set[RemoteAsset] = field(default_factory=set)
 
     def __bool__(self) -> bool:
         return self.deleting_dandiset or bool(self.remote_assets)
@@ -61,7 +62,7 @@ class Deleter:
             raise NotFoundError(
                 f"Asset at path {asset_path!r} not found in Dandiset {dandiset_id}"
             )
-        self.remote_assets.append(
+        self.remote_assets.add(
             RemoteAsset(dandiset_id, version_id, asset["asset_id"], asset["path"])
         )
 
@@ -75,7 +76,7 @@ class Deleter:
         for asset in self.client.get_dandiset_assets(
             dandiset_id, version_id, path=folder_path
         ):
-            self.remote_assets.append(
+            self.remote_assets.add(
                 RemoteAsset(dandiset_id, version_id, asset["asset_id"], asset["path"])
             )
             any_assets = True
@@ -152,7 +153,7 @@ class Deleter:
             else:
                 yield {"status": "Deleted"}
 
-        for asset in self.remote_assets:
+        for asset in sorted(self.remote_assets, key=attrgetter("path")):
             yield process(asset)
 
 
--- a/dandi/delete.py
+++ b/dandi/delete.py
@@ -39,7 +39,7 @@ class Deleter:
                 self.client.get_dandiset(dandiset_id, "draft")
             except requests.HTTPError as e:
                 if e.response.status_code == 404:
-                    raise RuntimeError(f"Dandiset {dandiset_id} not found on server")
+                    raise NotFoundError(f"Dandiset {dandiset_id} not found on server")
                 else:
                     raise
         elif not is_same_url(self.client.api_url, api_url):
--- a/dandi/delete.py
+++ b/dandi/delete.py
@@ -1,0 +1,204 @@
+from dataclasses import dataclass, field
+from pathlib import Path, PurePosixPath
+from typing import Iterable, Iterator, List, NamedTuple, Optional, Tuple
+
+import click
+import requests
+
+from .consts import known_instances
+from .dandiapi import DandiAPIClient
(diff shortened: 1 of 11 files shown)


jwodder:
@yarikoptic I'm unclear on how to properly invoke pyout.  My current attempt seems to result in pyout not consuming the iterators of progress dicts, and thus nothing is getting deleted.  (It works with `--devel-debug`, though.)

jwodder:
@yarikoptic @dchiquito I'm writing tests for the `delete()` function, but when running the entire test suite, it keeps failing because Django is erroring out with "psycopg2.OperationalError: FATAL: sorry, too many clients already".  Putting various things in sessions doesn't help.

yarikoptic:
@jwodder - is that for an asset delete or a dandiset delete?  smells like an issue to file against `dandi-api` since doesn't feel like anything to fix in dandi-cli.

yarikoptic:
@jwodder also does it work when you try DELETE on the api.dandiarchive.org or specific to docker-compose instance?

jwodder:
@yarikoptic I've only tested asset deletions so far, but the error is actually coming later in the test suite, after the deletion tests run — specifically, at `test_new_upload_extant_bad_existing` or adjacent.

jwodder:
@yarikoptic I was able to get the tests to pass again by making most dandi-api tests use the same `Session` instance, but I'm not convinced we won't run into this problem again later.

yarikoptic:
@jwodder please try it on the live instance.  Can do on any dandiset really besides that monster https://gui-beta-dandiarchive-org.netlify.app/#/dandiset/010000 which I am still uploading

jwodder:
@yarikoptic Try what on the live instance, exactly?  I already tried deleting assets from the live instance, and it worked.

yarikoptic:
I meant to try deleting (lots of?) assets on the live instance to see if similar "too many clients already" message would emerge in the dandi-api logs.

dandi/delete.py
```diff
@@ -0,0 +1,220 @@
...
+    version_id: str
+    asset_id: str
+    path: str
+
+
+@dataclass
+class Deleter:
+    client: Optional[DandiAPIClient] = None
+    dandiset_id: Optional[str] = None
+    deleting_dandiset: bool = False
```
yarikoptic, line 27: please add a comment on what is the role of this `bool`

dandi/delete.py
```diff
@@ -0,0 +1,220 @@
...
+
+class RemoteAsset(NamedTuple):
+    dandiset_id: str
+    version_id: str
+    asset_id: str
+    path: str
+
+
+@dataclass
+class Deleter:
```
yarikoptic, line 24: please add a docstring describing the purpose for this class

dandi/delete.py
```diff
@@ -0,0 +1,220 @@
...
+            )
+        self.remote_assets.add(
+            RemoteAsset(dandiset_id, version_id, asset["asset_id"], asset["path"])
+        )
+
+    def register_asset_folder(
+        self, api_url: str, dandiset_id: str, version_id: str, folder_path: str
+    ) -> None:
+        if version_id != "draft":
+            raise ValueError("Cannot delete assets from published versions")
```
yarikoptic, line 73: should we just rely on dandi-api to disallow such action and client to not care?  sorry if I forgot if we ran into some stumbling point on that or what -- if to keep, a comment should be added on why we need this client side logic

dandi/delete.py
```diff
@@ -0,0 +1,220 @@
...
+                asset_id["dandiset_id"],
+                asset_id.get("version") or "draft",
+                asset_id["location"],
+            )
+        else:
+            raise RuntimeError(f"Unexpected asset type for {url}: {asset_type}")
+
+    def register_local_path_equivalent(self, instance_name: str, filepath: str) -> None:
+        instance = known_instances[instance_name]
+        if instance.metadata_version == 0:
```
yarikoptic, line 116:
```suggestion
        if instance.girder:
```
to make it more explicit than some possible relationship between having metadata version and girder

dandi/delete.py
```diff
@@ -0,0 +1,220 @@
...
+            from .support import pyout as pyouts
+
+            pyout_style = pyouts.get_style(hide_if_missing=False)
+            rec_fields = ("path", "status", "message")
+            out = pyouts.LogSafeTabular(
+                style=pyout_style, columns=rec_fields, max_workers=jobs
+            )
+            with out:
+                for gen in deleter.process_assets_pyout():
+                    out(gen)
```
yarikoptic, line 199: please add at least a basic smoke test run with pyout as a backend to ensure that  no gotchas with pyout rendering etc,so we do not miss some breakage

yarikoptic (changes requested):
Looks great, thank you @jwodder .  I left just minor wishlists, the primary of which is to add a test run with pyout so we do not miss/spill some obvious errors into user faces.

jwodder:
@yarikoptic Review comments addressed.

jwodder:
@yarikoptic I deleted all 1328 assets in Dandiset 000008 with a single command, and there were no errors.

yarikoptic:
Thank you @jwodder !

PR #512, opened 2021-03-26
fix: remove unset fields to enable schemaKey

satra:
SchemaKey provides type information for disambiguation of certain objects. This is being removed from the generated information. This PR removes the unset keyword which allows the const/Literal fields to be exported.

### DIFF
--- a/dandi/models.py
+++ b/dandi/models.py
@@ -735,10 +735,10 @@ class CommonModel(DandiBaseModel):
     def json_dict(self):
         """
         Recursively convert the instance to a `dict` of JSONable values,
-        including converting enum values to strings.  Unset and `None` fields
+        including converting enum values to strings.  `None` fields
         are omitted.
         """
-        return json.loads(self.json(exclude_unset=True, exclude_none=True))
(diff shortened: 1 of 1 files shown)


yarikoptic:
I think it would be worth assessing the effect of that by doing the entire metadata conversion in dandi-api-dandisets using this patch (although some random uuids and dates would make it tricky ATM, https://github.com/dandi/dandi-api-datasets/issues/25 is yet TODO) ...
I wonder if it is only about schemaVersion - wouldn't it be just better to ensure that it is set or there are other fields which would be effected?

satra:
it's not about `schemaVersion` (not a literal), but `schemaKey` which is. and i'm trusting our minimal tests at this point. i haven't dug into assets yet. but currently meditor is somewhat broken because schemaKey is not included in the dandiset metadata.

yarikoptic:
ok, let's just proceed then and see what it would do us ;)

yarikoptic:
updated metadata for all dandisets.  Beware of needing to select "draft" version if you would like to see the result (ref: https://github.com/dandi/dandi-api/issues/188)

PR #516, opened 2021-03-27
[wip] adding to_datacite

djarecka:
Adding first version of `to_datacite`

@satra - I'm not sure about many of the fields. I've added questions to the code, please take a look.


satra:
@djarecka - this is a good starting point for a PR. if you haven't seen this, it may help (i believe we have the metadata for almost all the pieces discussed in the following url): 

https://github.com/UB-LMU/DataCite_BestPracticeGuide/blob/master/BestPracticeGuide.md#creator

also if you wanted to quickly manually try out things, you could use this: https://dhvlab.gwi.uni-muenchen.de/datacite-generator/ (which is related to the above guide)

some notes:

1. use the first contributor who is a Person as the creator. add the rest as contributors
1. it may be helpful to look at the converted metadata (under the new schema) for a couple of dandisets. you can look at the dandiset.yaml files in https://github.com/dandi/dandi-api-datasets to start with.
1. it would be good to add related resources and funding information when available.
1. for subject matter, i don't know how best to map to this, since most datasets have keywords and not the study targets or other info.

djarecka:
what to do with datasets that do not have `contributor`? (e.g. 20), `creator` is a mandatory field for datacite

satra:
that would be considered an invalid metadata. for checking whether a doi can be minted you should run the dandiset metadata through the validator. at present, migrate2newschema returns an unvalidated metadata.

djarecka:
I understand that you mean a different validator than `dandi validate path_to_dandiset`, since `dandi validate` does not say anything about lack od contributors

satra:
just do `Dandiset(**meta.json_dict())` where meta is the output of migrate2newschema. indeed validate does not validate the metadata for publishing.

djarecka:
This is how `newmeta.json_dict()` looks:
```
{'description': 'A 10 minute Neuropixels recording simulated using the MEArec Python package (Buccino et al., 2020). The recording contains the spiking activity of 250 biophysically detailed neurons (200 excitatory and 50 inhibitory cells from the Neocortical Micro Circuit Portal with independent Poisson firing patterns, and additive Gaussian noise with 10uV standard deviation.', 'identifier': 'DANDI:000028', 'name': 'Simulated cortical Neuropixels recording with ground truth', 'schemaVersion': '0.1.2', 'access': [{'status': 'dandi:Open'}], 'repository': 'https://dandiarchive.org/'}
```
`Dandiset.__init__` will just return bunch of errors...

satra:
@djarecka - those errors would mean the metadata are not complete/valid, and tell the user what is missing/erroneous. and hence they would not be able to publish/mint a doi without fixing it.

djarecka:
`Dandiset` takes `path` only: https://github.com/dandi/dandi-cli/blob/master/dandi/dandiset.py#L18

satra:
sorry i meant `DandisetMeta`

djarecka:
oh, ok, thanks! There are definitely dandisets with metadata that fails this. Will create a list

djarecka:
@satra - I've updated the function, I still have questions - some of them I've included in the function (I'd be happy to clean it, but I thought that it would be easier to review). 

And in addition below are the comments and questions based on working with dandiset `04` and `08`:
- not sure where to add `newmeta.access` (it looks like: `[AccessRequirements(status=<AccessType.Open: 'dandi:Open'>, contactPoint=None, description=None, embargoedUntil=None)]`)
- `newmeta.relatedResource`: couldn't find a place to put `name` and `url`
- these keys were empty for 04 and 08, and I'm not sure where they would go: `studyTarget`, `acknowledgement`, `wasGeneratedBy` ( maybe in `created` ??), `citation`, `assetsSummary`, `manifestLocation`, `protocol`
- `about` was empty for 08, but for 04 has `[TypeModel(identifier=None, name='Medial Temporal Lobe', schemaKey='GenericType')]"` -- still not sure where I would add it
- `url` is empty for 04 and 08 and not sure what you expect there, datacite has `url` and `contentUrl` (this is not described in the pdf documentation, not sure what they mean)
- `newmeta.repository`: I believe this could go to the Publisher, but we set to Dandi anyway 
- `newmeta.version` is empty and I'm not completely sure what version you expect here
- `newmeta.doi` is empty and not sure what is expected to be here, I thought we are just creating doi...

Note, that the resources I was using to figure out are in disagreement with each other, so sometimes I was guessing

satra:
> not sure where to add newmeta.access (it looks like: [AccessRequirements(status=<AccessType.Open: 'dandi:Open'>, contactPoint=None, description=None, embargoedUntil=None)])

you can skip this.

> newmeta.relatedResource: couldn't find a place to put name and url

url would need to be parsed to see if it's an identifier like doi and put into the appropriate place. other urls could be used just directly as identifiers (i think)

> these keys were empty for 04 and 08, and I'm not sure where they would go: studyTarget, acknowledgement, wasGeneratedBy ( maybe in created ??), citation, assetsSummary, manifestLocation, protocol

citation when ready could go to the bibtex property. the rest you can ignore for now.

> about was empty for 08, but for 04 has [TypeModel(identifier=None, name='Medial Temporal Lobe', schemaKey='GenericType')]" -- still not sure where I would add it

about should go into subject

> url is empty for 04 and 08 and not sure what you expect there, datacite has url and contentUrl (this is not described in the pdf documentation, not sure what they mean)

url should be the dandiarchive url for that dataset version. this should come when the version info has been added by the publish action.

> newmeta.repository: I believe this could go to the Publisher, but we set to Dandi anyway

yes

> newmeta.version is empty and I'm not completely sure what version you expect here

this should not be empty when this function will receive the metadata. the version info will be added by the server. 

> newmeta.doi is empty and not sure what is expected to be here, I thought we are just creating doi...

and this should contain the doi that should be used to mint

djarecka:
so my function should expect that `newmeta` has both `version` and `doi` - is that what you mean? It's a bit a chicken egg problem for me...

satra:
yes. btw, this PR needs to be resolved relative to master.

djarecka:
please see the comments/questions in the code. I can open a cleaner PR once it's confirmed.

Also, should I also test for other datasets? are they fixed?

dandi/models.py
```diff
@@ -113,6 +113,104 @@ def diff_models(model1, model2):
...
+    meta = dandiset.metadata
+    newmeta = migrate2newschema(meta)
+
+    dandiset_id = dandiset.identifier
+
+    attributes = {}
+    attributes["identifiers"] = [
+        {"identifier": newmeta.doi, "identifierType": "DOI"},
+        # not sure if I can add dandi_id in the PDF documentation the only option is DOI...
+        {"identifier": dandiset_id, "identifierType": "Dandi"},
```
satra, line 128: indeed they seem to only allow DOIs for identifiers, but you can add the other to `relatedIdentifiers`

dandi/models.py
```diff
@@ -113,6 +113,104 @@ def diff_models(model1, model2):
...
+    ]
+    attributes["titles"] = [{"title": newmeta.name}]
+    attributes["descriptions"] = [
+        # not sure what should be the type os use Other
+        # options are: Abstract, Methods, SeriesInformation,
+        # TableOfContents, TechnicalInfo, Other
+        {"description": newmeta.description, "descriptionType": "Other"}
```
satra, line 135: you can use Abstract

dandi/models.py
```diff
@@ -113,6 +113,104 @@ def diff_models(model1, model2):
...
+        {
+            "schemeURI": "https://spdx.org/licenses/",
+            "rightsIdentifierScheme": "SPDX",
+            "rightsIdentifier": el.name,
+        }
+        for el in newmeta.license
+    ]
+    # not sure if these is correct schema or should I provide newmeta.schemaVersion
+    # if not here, i'm not sure where newmeta.schemaVersion should go
+    attributes["schemaVersion"] = "http://datacite.org/schema/kernel-4"
```
satra, line 154: i would download the info from datacite for a few dois to see what this looks like.

dandi/models.py
```diff
@@ -113,6 +113,104 @@ def diff_models(model1, model2):
...
+    attributes["types"] = {"resourceType": "NWB", "resourceTypeGeneral": "Dataset"}
+    # newmeta has also attribute url, but it often empty
+    attributes["url"] = newmeta.url
+    # assuming that all licenses are from SPDX?
+    attributes["rightsList"] = [
+        {
+            "schemeURI": "https://spdx.org/licenses/",
+            "rightsIdentifierScheme": "SPDX",
+            "rightsIdentifier": el.name,
+        }
```
satra, line 149: this looks good - i'm going to update the license types in dandi to directly use this.

dandi/models.py
```diff
@@ -113,6 +113,104 @@ def diff_models(model1, model2):
...
+                )
+        elif isinstance(contr_el, Organization):
+            contr_dict["nameType"] = "Organizational"
+
+        if RoleType("dandi:Author") in getattr(contr_el, "roleName"):
+            creators.append(contr_dict)
+        # should creators  be also contributors if they have more roles
+        # I don't see a way to add more roles to the Contributor
+        else:
+            contr_dict["contributorType"] = getattr(contr_el, "roleName")
```
satra, line 179: i would suggest adding a comma separated string.

satra:
you should test this out by minting and updating some dois using the test server.

dandi/models.py
djarecka: but the question is if the `creator` should be also a `contributor`. I do not have "contributorType" field for a `creator`, so if I add someone to a contributor list only, I don't see a way to keep other roles

satra: so any person who is an author is a creator and any creator who has other roles should be added as contributor.

dandi/models.py
djarecka: many examples I checked have no `schemaVersion`, but some have `"http://datacite.org/schema/kernel-4"` or `"http://datacite.org/schema/kernel-3"`, so perhaps `"http://datacite.org/schema/kernel-4"` can stay

dandi/models.py
djarecka: I might be wrong, but I understood that `relatedIdentifiers` are for other resources that are related to this asset (and not other identifiers for the asset), that's why I added `newmeta.relatedResources` there

djarecka: i will just remove the second identifier for now

satra: you can use the relation `IsIdenticalTo`  when adding the DANDI identifier to relatedIdentifiers

djarecka: but in theory I should use a specific list for `relatedIdentifierType`, where there is no "Dandi". I could always try to add and see what happens...

satra: you can use URL and put `https://identifiers.org/DANDI:<datasetid>/version`

djarecka: ok

djarecka:
`newmeta.about` is always  `TypeModel`, if that's the case, I could read schema etc.

djarecka:
@satra - my tests that are using `secret.DATACITE_DEV_PASSWORD` are failing. Is it possible that it's not properly set? or am  I doing something wrong when using it?

djarecka:
moving to #596

PR #517, opened 2021-03-28
ENH: Enhancements to metadata schema and compatibility with JSONLD

### LINKED ISSUE #483: dandi:sha-X -> dandi:sha-2-X
more of future proofing I guess... I just paid attention that in https://github.com/multiformats/multicodec/blob/master/table.csv it is `sha-2-256` and `sha-3-256`... so I wondered if we should right away to switch away from ambigous `sha-X` (where X not a version like in `sha-1`)?

related: #482 and https://github.com/dandi/dandi-cli/pull/481/files#issuecomment-804347832 (on use of `-` in `@id` itself)

satra:
This PR contains a few enhancements to the schema as well as:

- generation of a context file for use with the dandiset and asset metadata to convert to JSONLD/Turtle/etc.,.
- it cleans up digests as per issues discussed in #483 (closes #483)
- It makes the asset digest a property only accessible in the json version of the schema. this will disappear if the schema is converted to any other graph format and the nested keys will become top level keys associated with the asset.
- For the moment a participant mixes identity and state. It is still kept with a lot of the properties that are likely to vary over time (age). This needs to be discussed and handled better. 
- At present, I am using the reproschema python library to test conversions to jsonld or turtle. I would suggest we play with this during the hackathon to generate a searchable graph and then finalize whatever components should be brought into dandi-cli.


satra:
for discussion on enums and keys in pydantic (which we are using on digest) see: https://github.com/samuelcolvin/pydantic/issues/2278

dandi/metadata.py
```diff
@@ -616,12 +606,67 @@ def migrate2newschema(meta):
     return dandimeta
 
 
+def generate_context():
+    import pydantic
+
+    fields = {
+        "@version": 1.1,
```
yarikoptic, line 613: What is it a version of?

yarikoptic (commented):
Too actively changing ATM, so I will have a look later, but left few questions

dandi/metadata.py
satra: jsonld

satra:
@jwodder / @yarikoptic - i will need some help here (see the failing tests on ubuntu). i tried to fix digest assuming it should be json_dict of AssetMeta, but looks like digest is a string for some reason not a dict. i suspect this is using some other form of asset metadata. my reading of the code suggests it should work, but obviously not.

yarikoptic:
You switched from list of dicts to a dict, so in that

```
for digest in asset_metadata.get("digest", {}):
            lgr.debug("Digest: %s", str(digest.items()))
```

`digest` is the key (the babe of the digest)... You no longer need a loop since can just check for desired dandi-etag key

satra:
thank you @yarikoptic

satra:
this should be good now. please review and merge if no changes are required.

dandi/model_types.py
```diff
@@ -46,20 +46,32 @@
...
+            "rdfs:label": "SHA3-256",
+        },
+        {
+            "@id": "dandi:blake2b-256",
+            "@type": "dandi:DigestType",
+            "rdfs:comment": "BLAKE2B-256 checksum",
+            "rdfs:label": "BLAKE2B-256",
+        },
+        {
+            "@id": "dandi:blake3-256",
```
yarikoptic, line 71:
```suggestion
            "@id": "dandi:blake3",
```
if we are to align to that https://github.com/multiformats/multicodec/blob/master/table.csv#L21 where they use just `blake3` for the one with 32 bytes (==256bits).  And according to https://en.wikipedia.org/wiki/BLAKE_(hash_function)#BLAKE3  I do not see any other size possible.  
Probably should be reflected also in the rdfs:label as well (i.e. drop -256)

dandi/models.py
```diff
@@ -598,71 +621,54 @@ class Participant(DandiBaseModel):
...
+        "rdfs:subClassOf": ["schema:Person", "prov:Agent"],
         "rdfs:label": "Information about the participant.",
         "nskey": "dandi",
     }
 
 
-class Activity(DandiBaseModel):
-    """Information about the Project activity"""
+class BioSample(DandiBaseModel):
+    """Description about the sample that was studied"""
```
yarikoptic, line 638:
```suggestion
    """Description of the sample that was studied"""
```

dandi/models.py
```diff
@@ -598,71 +621,54 @@ class Participant(DandiBaseModel):
...
-        max_length=150,
-        nskey="schema",
-    )
-    description: Optional[str] = Field(
-        None, description="A brief description of the project.", nskey="schema"
-    )
-    schemaKey: Literal["Project"] = Field("Project", readOnly=True)
+    _ldmeta = {
+        "rdfs:subClassOf": ["schema:Thing", "prov:Entity"],
+        "rdfs:label": "Information about the biosample.",
```
yarikoptic, line 666: Should we strive to have those consistent with the `__doc__` and thus may be just automagically copied from the class's `__doc__` for every `_ldmeta` (and raise an error if there is no `__doc__`)?  I just don't see why it should differ

yarikoptic (changes requested):
Minor comments to unify just a bit more

dandi/models.py
satra: let's do that in a separate PR. i worry that the doc may end up containing more things than we would want to put into the rdfs:label.

jwodder:
@yarikoptic After this is merged, https://github.com/dandi/dandi-api-datasets/pull/28 will need to be merged as well.

yarikoptic:
Thank you @satra and @jwodder, merging here and there (a failed run is unrelated...)

PR #521, opened 2021-03-29
Retry requests that fail with 503

### LINKED ISSUE #520: make client more robust than the server: reissue request to dandi-api service if errors out with 503
ATM happened only for `/initialize` (https://github.com/dandi/dandi-api/issues/185) but I wonder if we should make it available to other invocations as well.

jwodder:
Closes #520.


yarikoptic:
looks kosher, let's see if quality of life improves ;) thank you @jwodder !

PR #528, opened 2021-03-30
Fix retrying 503's

### LINKED ISSUE #527: doesn't retry on 503?
Followup to #520 
Trying again on that  huge file with 0.12.1+61.g08a14ff, and quickly (thus unlikely any retry is happening) get 503

log from dandi-cli is on drogon under /tmp/20210330142039Z-17279.log

jwodder:
Fixes #527.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -7,6 +7,7 @@ from threading import Lock
 from xml.etree.ElementTree import fromstring
 
 import requests
+import tenacity
 
 from .consts import MAX_CHUNK_SIZE, known_instances_rev
 from .girder import keyring_lookup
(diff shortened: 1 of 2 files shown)


yarikoptic:
Is there some logging from tenacity we could hook to our handlers eg if we are logging at debug level or below?

jwodder:
@yarikoptic [This line](https://github.com/dandi/dandi-cli/blob/master/dandi/utils.py#L680) is supposed to make tenacity log at `WARNING` (the log level of `try_multiple()` before it was rewritten to use tenacity) every time it retries.

yarikoptic:
ah ok, let's proceed and try, thank you!

PR #535, opened 2021-04-06
Support parsing & navigating asset download URLs

### LINKED ISSUE #533: ls: provide ability to list individual asset from a download url
ATM:
```
$> dandi ls -f json_pp https://api.dandiarchive.org/api/dandisets/000003/versions/draft/assets/0a748f90-d497-4a9c-822e-9c63811db412/download/
2021-04-06 12:20:16,926 [    INFO] Logs saved in /home/yoh/.cache/dandi-cli/log/20210406162015Z-1963932.log
Error: Response for getting https://api.dandiarchive.org/api/dandisets/000003/versions/draft/assets/0a748f90-d497-4a9c-822e-9c63811db412/download/ to redirect returned 403. Please verify that it is a URL related to dandiarchive and supported by dandi client

```
although URL is valid and wget follows redirect nicely.  Might be that dandi-api does not handle HEAD request correctly while we are trying to get redirect url (so something to file and issue or submit a PR about to dandi-api).  But I do not think in this case we should even try to follow any redirect since all pertinent information is in the URL and we just need to parse it and request metadata record for the asset

jwodder:
Closes #533.

### DIFF
--- a/dandi/dandiarchive.py
+++ b/dandi/dandiarchive.py
@@ -51,7 +51,7 @@ def navigate_url(url):
             asset_id["version"] = r["most_recent_version"]["version"]
         args = (asset_id["dandiset_id"], asset_id["version"])
         kwargs["include_metadata"] = True
-        if asset_id.get("location"):
+        if asset_id.get("location") or asset_id.get("asset_id"):
             with client.session():
                 dandiset = client.get_dandiset(*args)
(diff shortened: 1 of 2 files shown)


yarikoptic:
thank you! will test in action (one use case seems to be not covered)

PR #536, opened 2021-04-07
Give `ls` a `--metadata` option

### LINKED ISSUE #534: ls: -r is "insufficient" to get detailed information about assets in a folder
ATM
```
$> dandi ls -f json_pp -r 'https://gui-beta-dandiarchive-org.netlify.app/#/dandiset/000003/draft/files?location=sub-YutaMouse20%2F'       
{
  "asset_count": 101,
  "created": "2021-04-05T15:25:03.799071Z",
  "metadata": {
 ...
  "path": "000003",
  "size": 2559248010229,
  "version": "draft"
}
{
  "asset_id": "ca90f128-4e55-4d5b-8f0e-647bd2cf78b2",
  "created": "2021-04-05T15:29:02.570155Z",
  "modified": "2021-04-05T15:29:02.570211Z",
  "path": "sub-YutaMouse20/sub-YutaMouse20_ses-YutaMouse20-140327_behavior+ecephys.nwb",
  "size": 8417855886
}
...
```
so we do provide detailed record on the dandiset but not for each individual asset.  I think we should provide a way to make individual asset metadata returned as well. 

Obviously then a single `-r` cannot tell between the two desires (just a list of assets from end point as API returns or query/include metadata per asset).  We could have made it into an option `--recursive `None`,"auto","assets"`  (we tend to go for such style in datalad more these days acknowledging that flags eventually become insufficient) but I think it would make unnecessarily harder to split `-r|--recursive` from positional arguments (would require `--` in cases where value is not provided).

So let's go for an explicit `--metadata api,all,dandiset,assets` ?  `api` -- just return what API returns for a given case, and that would be current behavior. `all`/`dandiset`/`assets` -- would query metadata (for the dandiset or an asset) if no `metadata` in the corresponding record. 

I am open to other ideas as well. WDYT @jwodder @satra ?

jwodder:
Closes #534.

### DIFF
--- a/dandi/cli/cmd_ls.py
+++ b/dandi/cli/cmd_ls.py
@@ -41,17 +41,23 @@ from ..utils import is_url
     show_default=True,
 )
 @click.option(
+    "--metadata",
+    type=click.Choice(["api", "all", "assets"]),
+    default="api",
+)
+@click.option(
     "--schema",
     help="Convert metadata to new schema version",
     metavar="VERSION",
 )
 @click.argument("paths", nargs=-1, type=click.Path(exists=False, dir_okay=True))
 @map_to_click_exceptions
-def ls(paths, schema, fields=None, format="auto", recursive=False, jobs=6):
+def ls(paths, schema, metadata, fields=None, format="auto", recursive=False, jobs=6):
     """List .nwb files and dandisets metadata."""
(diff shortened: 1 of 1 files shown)


yarikoptic:
Thank you, I guess we will just test it in action... proper python interface is anyway yet to come

PR #545, opened 2021-04-09
Log dandi, hdmf, h5py, and pynwb versions to log file

### LINKED ISSUE #543: Write a log message with dandi-cli  into the log file
We do announce at INFO level whenever version of dandi-cli is not the same as released, but otherwise do not show it.  I think we should always add a record to the log file which would disclose the version of the dandi-cli.  
Ideally it should also state versions of hdmf, pynwb, and h5py.  Although for some commands (e.g. `download`) they would not even be used, I think it would still be useful even if would require their import right at the beginning of the execution: alternatively, it could be a log message at the end of the log and those versions reported only if were already imported (thus in `sys.modules`).  Somewhat relates to #57, since if `wtf` would produce already relevant record, it would be just a matter of logging it.

jwodder:
Closes #543.

### DIFF
--- a/dandi/cli/command.py
+++ b/dandi/cli/command.py
@@ -106,6 +106,14 @@ def main(ctx, log_level, pdb=False):
     handler.setFormatter(fmter)
     root.addHandler(handler)
 
+    import h5py
+    import hdmf
+    import pynwb
+
+    lgr.info("dandi v%s", __version__, extra={"file_only": True})
+    lgr.info("hdmf v%s", hdmf.__version__, extra={"file_only": True})
+    lgr.info("pynwb v%s", pynwb.__version__, extra={"file_only": True})
+    lgr.info("h5py v%s", h5py.__version__, extra={"file_only": True})
     lgr.info("sys.argv = %r", sys.argv, extra={"file_only": True})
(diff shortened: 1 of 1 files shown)


dandi/cli/command.py
```diff
@@ -106,6 +106,14 @@ def main(ctx, log_level, pdb=False):
...
+    import h5py
+    import hdmf
+    import pynwb
+
+    lgr.info("dandi v%s", __version__, extra={"file_only": True})
+    lgr.info("hdmf v%s", hdmf.__version__, extra={"file_only": True})
+    lgr.info("pynwb v%s", pynwb.__version__, extra={"file_only": True})
+    lgr.info("h5py v%s", h5py.__version__, extra={"file_only": True})
     lgr.info("sys.argv = %r", sys.argv, extra={"file_only": True})
     lgr.info("os.getcwd() = %s", os.getcwd(), extra={"file_only": True})
```
yarikoptic, line 118: Let's log all versions in a single line please.  And now we can use `get_module_version` (I merged that PR) so it would not crash due to some older `hdmf`

jwodder: Done.

yarikoptic:
Great, let's proceed, thank you @jwodder

PR #549, opened 2021-04-09
small fix of extract_sex

djarecka:
assuming that `U` or `u`  value can be treated as `Unknown`


yarikoptic:
Great, thank you @djarecka !

PR #553, opened 2021-04-12
Get hdmf, pynwb, h5py versions without importing

jwodder:
This cuts off about a second and a half from the dandi command's load time.

### DIFF
--- a/dandi/cli/command.py
+++ b/dandi/cli/command.py
@@ -107,16 +107,12 @@ def main(ctx, log_level, pdb=False):
     handler.setFormatter(fmter)
     root.addHandler(handler)
 
-    import h5py
-    import hdmf
-    import pynwb
-
     lgr.info(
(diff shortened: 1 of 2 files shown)


yarikoptic:
Let's do it this way, thank you.  My only concern is that in cases where it was `pip install -e`-ed and then worktree switched versions without rerunning `pip install -e`, we might report wrong version: then one which was installed, not actually used:
```shell
$> python -c 'import importlib.metadata; print(importlib.metadata.version("dandi"))'
0.12.1+64.gcebdf86.dirty

$> git co 0.11.0
Note: switching to '0.11.0'.

...
$> python -c 'import importlib.metadata; print(importlib.metadata.version("dandi"))'
0.12.1+64.gcebdf86.dirty

$> python -c 'import dandi; print(dandi.__version__)'                                     
0.11.0
```

but I guess it would be specific only to dev environments, and we just need to keep that in mind

PR #560, opened 2021-04-13
Models: define id, add various additional types (genotype, etc), boost model version to 0.3.0

satra:
- [x] id + identifier
- [x] genotype info
~- [ ] still decide on participant.~


dandi/tests/test_metadata.py
```diff
@@ -127,7 +126,7 @@ def test_metadata2asset_simple1(schema_dir):
             "nwb_version": "2.2.5",
             "experiment_description": "experiment_description1",
             "experimenter": ("experimenter1",),
-            "identifier": "bfc23fb6192b41c083a7257e09a3702b",
+            "id": "urn:uuid:bfc23fb6192b41c083a7257e09a3702b",
```
yarikoptic, line 129: seems to be not a uuid here.

But I also thought that a brief discussion on monday just boiled down to having `identifier` with actual value and `id` been either
- minted/populated by dandi-api when returning an adjusted metadata record.  Thus having both, and thus resulting in duplication.  
- or just produced while preparing json-ld view of the metadata

without that -- for any use of this `id` as an `asset_id` in the API endpoints a programmer would need to first strip this `urn:uuid:` prefix (we do not want `:` in the path - would need to be url encoded), we would loose clear mapping from an `asset_id` to its location in the key store . Altogether sounds like more pain IMHO for no gain.
Thus I still have original question on what problem (exactly) you are trying to solve with such `identifier` -> `id` replacement?  
if it is just to produce json-ld ready metadata, wouldn't such records be tuned anyways and accompanied with context etc?  or
do you see us using some other than `urn:uuid:` prefix for the assets?

satra: that is correct. i haven't added any validation here. 

regarding id vs identifier this is mimicing the schema.org setup. `id` maps to a URI (i.e. an identity for the object), while `identifier` is not necessarily a URI (it could be simply a text. when producing a context, i need to be able to map properties of a class to types. previously i was mapping identifier to id and therefore a URI, but that meant wherever i see identifier (which was in every class, it would automatically be turned into a uri, which did not make sense for certain things like orcid).

in this formulation, we explicitly ensure that value of id for an asset is a urn, which means know expansion needs to take place. it's an identifier that can be used in jsonld, rdf, etc.,. whereas if it was a uuid, it would require being turned into a uri. 

regarding parsing by library, i don't think that's a pain: this is a fully formed uri so urlparse and other parsers will work out of the box. also i would really want the cli to simply use the models when dealing with metadata from the api, so technically that should work as well.

satra:
@yarikoptic - i would love to get this merged if there are no objections at this point. 

this would then trigger a schema update for api and web-ui which should help as well. nothing should technically be different at this point with `id`, since it is an optional field.

dandi/metadata.py
```diff
@@ -603,6 +603,10 @@ def convertv1(data):
 
 def migrate2newschema(meta):
     newmeta = convertv1(meta)
+    if "version" in newmeta:
+        newmeta["id"] = f"{newmeta['identifier']}/{newmeta['version']}"
+    else:
+        newmeta["id"] = f"{newmeta['identifier']}/draft"
```
yarikoptic, line 609: `identifier` is made optional in `Identifiable` (and was/is optional for others such as `RelatedParticipant`) in this PR, so I guess it could fail or get a bogus `None` here?  may be no `id` should be assigned `if not identifier`?

dandi/metadata.py
```diff
@@ -613,6 +617,7 @@ def generate_context():
     fields = {
         "@version": 1.1,
         "dandi": "http://schema.dandiarchive.org/",
+        "dandiasset": "http://iri.dandiarchive.org/",
```
yarikoptic, line 620: is `-` allowed? then I would have preferred
```suggestion
        "dandi-asset": "http://iri.dandiarchive.org/",
```

dandi/models.py
```diff
@@ -823,6 +880,17 @@ class PublishedDandisetMeta(DandisetMeta):
         nskey="dandi",
     )  # TODO: formalize "publish" activity to at least the Actor
     datePublished: date = Field(readOnly=True, nskey="schema")
```
yarikoptic, line 882: note: we need to streamline (and probably adjust descriptions here here) the life cycle of all the `date*` markers, as relating to how `dandi-api` would treat/tune them.
ATM it is not immediately clear to me how this one would be different e.g. from `dateModified` since we should not modify published dandisets:
- I guess if `dateModified` would be "filled in" by `dandi-api` and reflect the date for the database row entry, and we adjust any columns which are not in the `metadata` (stats?) that would lead to `dateModified` change
- but still not clear how it would be different from `dateCreated`?  If dateCreated would be date of creation of dandiset (not filled in from DB row table), then `datePublished` would have purpose.  But if `dateCreated` is for the date of when published row record was created -- they should/would be identical. 

Meanwhile:
```suggestion
    datePublished: date = Field(readOnly=True,  title="Publication date and time", nskey="schema")
```

dandi/models.py
```diff
@@ -888,19 +956,23 @@ class BareAssetMeta(CommonModel):
 class AssetMeta(BareAssetMeta, Identifiable):
     """Metadata used to describe an asset on the server."""
 
-    identifier: UUID4 = Field(readOnly=True, nskey="schema")
-
+    id: Optional[str] = Field(readOnly=True, description="Dandi asset id")
```
yarikoptic, line 959:
```suggestion
    id: Optional[str] = Field(readOnly=True, description="Uniform resource identifier")
```
to stay inline with how we have it for DandisetMeta? 

initial attention was triggered by use of `"Dandi ..."`.  We do use `Dandi` in class names, but shouldn't it be `DANDI` if referring in text, and wouldn't everything be `Dandi` here?

dandi/models.py
```diff
@@ -888,19 +956,23 @@ class BareAssetMeta(CommonModel):
...
+    identifier: Optional[UUID4] = Field(readOnly=True, nskey="schema")
     # on publish or set by server
     contentUrl: Optional[List[HttpUrl]] = Field(None, readOnly=True, nskey="schema")
 
 
 class PublishedAssetMeta(AssetMeta):
+    id: str = Field(readOnly=True, description="Dandi asset id")
```
yarikoptic, line 966:
```suggestion
    id: str = Field(readOnly=True, description="Uniform resource identifier")
```
per similar above

dandi/models.py
```diff
@@ -888,19 +956,23 @@ class BareAssetMeta(CommonModel):
...
     publishedBy: HttpUrl = Field(
         description="The URL should contain the provenance of the publishing process.",
         readOnly=True,
         nskey="dandi",
     )  # TODO: formalize "publish" activity to at least the Actor
     datePublished: date = Field(readOnly=True, nskey="schema")
+    url: HttpUrl = Field(
+        None, readOnly=True, description="permalink to the item", nskey="schema"
+    )
```
yarikoptic, line 975: I think we could just add a mix-in `Published` (similar to `Identifiable`) which would provide `publishedBy`, `datePublished` and `url` and possibly even `id` -- if we can just use generic descriptions.  That would remove code duplication and make it easier to keep those attributes consistent between Dandiset and Asset.

dandi/tests/data/metadata/metadata2asset.json
```diff
@@ -1,5 +1,5 @@
 {
-  "identifier": "0b0a1a0b-e3ea-4cf6-be94-e02c830d54be",
+  "id": "dandiasset:0b0a1a0b-e3ea-4cf6-be94-e02c830d54be",
```
yarikoptic, line 2: hm, I thought that with the recent changes we still retain `identifier` and only provide `id` at the level of `dandi-api` queries (and may be only for json-ld serialization), thus minting it in addition/based on the `identifier`.  but even if "not only for json-ld" -- shouldn't original `identifier` also appear here?

again - one of my arguments was/is - with just an `id` users would need to "manually" split it inside of just using the value as `asset_id` for the API queries.  So by catering to the LD we are further sacrificing ease of use of this metadata for "more pragmatic" interactions ;-)

dandi/tests/test_models.py
```diff
@@ -116,9 +116,9 @@ def test_asset():
         (
             LicenseType,
             {
-                "CC0": "dandi:CC0",
-                "CCBY40": "dandi:CCBY40",
-                "CCBYNC40": "dandi:CCBYNC40",
+                "CC0_10": "dandi:CC0-1.0",
+                "CC_BY_40": "dandi:CC-BY-4.0",
+                "CC_BY_NC_40": "dandi:CC-BY-NC-4.0",
```
yarikoptic, line 121: BTW! should we may be define `spdx` prefix and point to https://spdx.org/licenses/ for this purpose?  just tested on a random one: https://spdx.org/licenses/CC-BY-3.0 works

yarikoptic (changes requested):
I am still a bit unsure about disappearance of `identifier` from test records.  Left various questions/recommendations throughout.

dandi/metadata.py
satra: this is a converter and this function will go away as soon as we migrate.

dandi/metadata.py
satra: not a big fan of `-` in prefixes, but i won't complain if you really insist.

dandi/models.py
satra: datePublished could be 5 days after last modified, unless we say the publishing action is a modification action on the dandiset. this really becomes a notion of what we expect modified to mean.

all dates are filled in from database table.

dandi/models.py
satra: we could, but i don't think we are using this for anything else.

dandi/tests/data/metadata/metadata2asset.json
satra: i don't think this affects anything in the CLI now, because this is about conversion and the CLI explicitly ignores identifier.

dandi/tests/test_models.py
satra: we could use spdx directly. i'll make the change

dandi/models.py
satra: done.

dandi/tests/data/metadata/metadata2asset.json
yarikoptic: rright -- I think we are relying on API to provide us `asset_id`s in its listing of assets for a path, so indeed should not get broken. But my comment was not about breakage but rather about changing metadata so there is no 1-to-1 immediate correspondence between those records and API (`asset_id`).

dandi/metadata.py
yarikoptic: I do not have too strong of a preference to insist. oh well, since we have `dandiset`, I guess `dandiasset` kinda follows .  I will let it be

dandi/metadata.py
yarikoptic: that is good, still don't see why it should be "not robust by design" meanwhile, but ok.  I will assign you an issue if I run into one ;-)

yarikoptic:
anyways -- I do  not want to hold the progress here, since I might still be just missing on some aspects of identifier-vs-id situation.  Time will show. Let's proceed and brace for an impact -- I will also release fresh dandi-cli with this as we just merged another PR which might be good to have for future proofing.

PR #563, opened 2021-04-13
Error with a decent message when trying to delete() a path not in a Dandiset

jwodder:


yarikoptic:
Looks great, thank you @jwodder !

PR #564, opened 2021-04-13
Fix a typo in the display string for one of the known URL patterns

jwodder:


dandi/dandiarchive.py
```diff
@@ -175,7 +175,7 @@ class _dandi_url_parser:
                 rf"(/files(\?location=(?P<location>.*)?)?)?"
             ),
             {"server_type": "api"},
-            "https://gui-beta-dandiarchive-org.netflif.app/#/dandiset"
+            "https://gui-beta-dandiarchive-org.netlify.app/#/dandiset"
```
yarikoptic, line 178: bugaga -- so this regex was of no value and it still worked?  may be we should just get rid of this record altogether?

jwodder: It's not a regex; it's part of the string displayed to humans when none of the regexes match.

yarikoptic: d'oh - right.  I will blame the narrow context of the diff for my slip. Thanks!

PR #565, opened 2021-04-13
Switch default dandi instance to dandi-api based on redirector

### LINKED ISSUE #541: automate switching to `-i dandi-api` via `dandiarchive.org/server-info` information
ATM that one reports
```json
{
"version": "1.0.0",
"cli-minimal-version": "0.6.0",
"cli-bad-versions": [],
"services": {
"girder": {
"url": "https://girder.dandiarchive.org"
},
"webui": {
"url": "https://gui.dandiarchive.org"
},
"api": {
"url": "https://publish.dandiarchive.org/api"
},
"jupyterhub": {
"url": "https://hub.dandiarchive.org"
}
}
}
```

I think  we should 
- ~~change redirector it to remove entry for `"api"` since that service is long gone (and boost `version` to e.g. `1.1.0`)~~. that was a bad idea, replaced with
  -  [x] make "api" empty (null == None)
- [ ] double check that `dandi-cli` would refuse to interact with the archive if its version is below minimal one specified in that `server-info`
- [ ] adjust `dandi-cli` so if there is a url for `"api"`, we would use `dandi-api` instance by default and not the girder one. (alternative - do that if there is no "girder")
- [ ] release `dandi-cli` 0.14.0

Upon migration
- in redirector - adjust minimal version of the client to become `0.14.0` (assuming that is the one which it would be with default switch) and add `"url"` for the `"api"` and remove "girder", so released client would automagically switch to use `dandi-api` instance 

any flaw in above plan/logic @satra ?

jwodder:
Closes #541.

### DIFF
--- a/dandi/tests/test_utils.py
+++ b/dandi/tests/test_utils.py
@@ -168,6 +168,7 @@ def test_get_instance_dandi():
             "services": {
                 "girder": {"url": "https://girder.dandi"},
                 "webui": {"url": "https://gui.dandi"},
+                "api": None,
                 "jupyterhub": {"url": "https://hub.dandi"},
             },
         },
@@ -191,6 +192,7 @@ def test_get_instance_dandi_with_api():
             "cli-minimal-version": "0.5.0",
             "cli-bad-versions": [],
             "services": {
+                "girder": None,
                 "webui": {"url": "https://gui.dandi"},
                 "api": {"url": "https://api.dandi"},
                 "jupyterhub": {"url": "https://hub.dandi"},
@@ -218,6 +220,7 @@ def test_get_instance_url():
             "services": {
                 "girder": {"url": "https://girder.dandi"},
                 "webui": {"url": "https://gui.dandi"},
+                "api": None,
                 "jupyterhub": {"url": "https://hub.dandi"},
             },
         },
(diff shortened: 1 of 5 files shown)


dandi/tests/test_utils.py
```diff
@@ -218,7 +241,6 @@ def test_get_instance_cli_version_too_old():
             "services": {
                 "girder": {"url": "https://girder.dandi"},
                 "webui": {"url": "https://gui.dandi"},
-                "api": {"url": "https://publish.dandi/api"},
```
yarikoptic, line 221: note that redirector now will return `"api": None`

jwodder: Updated.

dandi/utils.py
```diff
@@ -618,14 +618,27 @@ def get_instance(dandi_instance_id):
...
+            api=None,
+        )
+    elif server_info["services"].get("api"):
+        return dandi_instance(
+            metadata_version=1,
+            girder=None,
+            gui=server_info["services"]["webui"]["url"],
+            redirector=redirector_url,
+            api=server_info["services"]["api"]["url"],
+        )
```
yarikoptic, line 636: could we please retain robustness of `.get` to provide `None`? or what would be the benefit of client crashing here if redirector again just drops a field instead of explicitly annotating with `null`?

yarikoptic:
besides one comment (I would prefer `.get` over `[]` unless there is a good reason), looks good to me.  Should be taken out of the  Draft and to be merged

dandi/utils.py
jwodder: Done.

yarikoptic: Thank you! next one to touch this piece of code is to reduce duplication.  Let's proceed

PR #571, opened 2021-04-14
Add further tests of get_instance() and server-info

### LINKED ISSUE #569: add test for get_instance obtaining information from redirector's server-info
A follow up to https://github.com/dandi/redirector/issues/42#issuecomment-819487756 . Apparently we dropped covering that part of code testing against a "live" redirector whenever we removed (now gone) support for downloading multiple files given a URL from web ui (used to be supported in girder instance).

edit: test should smoke test that code and also test that provided version is greater than current one used, and that either girder or api URL is provided (but not both)

jwodder:
Closes #569.

Note that #565 must be merged before the redirector is updated to point solely to dandi-api or else the `test_get_instance_actual_dandi` test will fail.

### DIFF
--- a/dandi/tests/test_utils.py
+++ b/dandi/tests/test_utils.py
@@ -3,7 +3,9 @@ import os.path as op
 import time
 
 import pytest
+import requests
 import responses
+from semantic_version import Version
 
 from ..consts import dandi_instance, known_instances
 from ..exceptions import BadCliVersionError, CliVersionTooOldError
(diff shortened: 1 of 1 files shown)


yarikoptic:
looks good to me, let's proceed

yarikoptic:
re #565 - let's have it finished/merged then, it already accumulated conflicts.

jwodder:
@yarikoptic Then we need to finish the discussion in #541 about version-checking and GUI URLs.

PR #575, opened 2021-04-16
Update for the version /info endpoint in dandi-api

dchiquito:
Required by https://github.com/dandi/dandi-api/pull/236

As described in https://github.com/dandi/dandi-api/pull/236, the
original `.../versions/{version_id}/` has been updated to only return
the version metadata. Because clients are still relying on the Django
serialized information about the version, that data can now be found at
`.../versions/{version_id}/info/`.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -275,7 +275,7 @@ class DandiAPIClient(RESTFullAPIClient):
 
     def get_dandiset(self, dandiset_id, version):
         return self._migrate_dandiset_metadata(
-            self.get(f"/dandisets/{dandiset_id}/versions/{version}/")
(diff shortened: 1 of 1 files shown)


satra:
@dchiquito - is this temporary for now?

dchiquito:
@satra Yes, it is there to avoid rewriting substantial chunks of both clients. With this in place we can refactor at our leisure.

yarikoptic:
ok, merging and releasing. Note: as tests here fail since we do not use dandi-api from that PR, future RFing of API would be trickier since dandi-cli will be released/in production, so we would need to provide "deprecation" cycle of some kind for the `/info` endpoint if it is to be gone

PR #581, opened 2021-04-19
Support specifying the dandi redirector via an env var

jwodder:
For https://github.com/dandi/redirector/issues/42.

PR #586, opened 2021-04-21
Support specifying the dandi redirector via an env var (take 2)

jwodder:
For https://github.com/dandi/redirector/issues/42.

### DIFF
--- a/dandi/tests/test_utils.py
+++ b/dandi/tests/test_utils.py
@@ -1,4 +1,5 @@
 import inspect
+import os
 import os.path as op
 import time
 
@@ -364,7 +365,12 @@ def test_get_instance_actual_dandi():
 
 
 def test_server_info():
-    r = requests.get(known_instances["dandi"].redirector.rstrip("/") + "/server-info")
+    r = requests.get(
+        os.environ.get(
+            "DANDI_REDIRECTOR_BASE", known_instances["dandi"].redirector
+        ).rstrip("/")
+        + "/server-info"
+    )
     r.raise_for_status()
     data = r.json()
     assert "version" in data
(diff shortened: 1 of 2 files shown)


dandi/tests/test_utils.py
```diff
@@ -363,8 +364,14 @@ def test_get_instance_actual_dandi():
         assert inst.api is not None
 
 
+@pytest.mark.redirector
 def test_server_info():
-    r = requests.get(known_instances["dandi"].redirector.rstrip("/") + "/server-info")
+    r = requests.get(
+        os.environ.get(
+            "DANDI_REDIRECTOR_BASE", known_instances["dandi"].redirector
```
yarikoptic, line 371: why don't we override it directly in the `dandi` of the `known_instances`?  this would be the only way to test not only server_info but also any other interaction with the redirector

tox.ini
```diff
@@ -22,7 +22,9 @@ commands =
 
 [pytest]
 addopts = --tb=short --durations=10
-markers = integration
+markers =
+    integration
```
yarikoptic, line 26: I think it would be ok to just run a full sweep of `integration` tests for redirector testing

jwodder: The `integration` marker appears to be used for something else.  Specifically, [just this test](https://github.com/dandi/dandi-cli/blob/5a4f24c518b9fda2e2343afce5f2fc88fc7c20da/dandi/tests/test_organize.py#L105).

dandi/tests/test_utils.py
jwodder: PR closed in favor of switching back to #581.

jwodder:
PR closed in favor of #581.

PR #591, opened 2021-04-22
BF: "girder" record might be there but "url" might be None

yarikoptic:
minimal patch to get `download` working after migration and girder being gone.
While at it, trying to fix up some tests as well which did not survive girder being gone


yarikoptic:
oh, you are green! THANK YOU!

PR #596, opened 2021-04-26
adding to_datacite method

djarecka:
this is a continuation of #516 (I had issues with using the secrets, so testing this from the dandi branch)

Adding a new method `to_dandicite` that creates a datacite dictionary from the metadata object


djarecka:
@satra -  I'm testing metadata that I took from dandisets 04 and 08. I could modify dataset 27 and add to the test, but can also keep adding various metadata to [test_dantimeta_datacite](https://github.com/dandi/dandi-cli/pull/596/files#diff-4b008975a2df2bb2ddf1030178e22a3dc3585fabaa159d35f9b74d1b394acdf9R306)

satra:
for the moment you can use a local test, but we should also add a test which downloads from the API, augments it with the missing pieces (because draft metadata may not have doi, url, etc.,.) and then runs through your datacite test using the test server.

dandi/models.py
```diff
@@ -877,6 +1056,21 @@ def check_data(cls, values):
     }
 
 
+class PublishedDandisetMeta(DandisetMeta):
```
satra, line 1059: this seems off and does not seem to take into account the inheritance in the current model.

dandi/tests/test_models.py
```diff
@@ -154,3 +166,192 @@ def test_autogenerated_titles():
...
+    rp = requests.post(
+        "https://api.test.datacite.org/dois",
+        json=datacite,
+        headers={"Content-Type": "application/vnd.api+json"},
+        auth=("DARTLIB.DANDI", os.environ["DATACITE_DEV_PASSWORD"]),
+    )
+
+    # checking if i'm able to get the url
+    rg = requests.get(
+        url=f"https://api.test.datacite.org/dois/{doi.replace('/','%2F')}/activities"
```
satra, line 190: why activities and not compare to the posted metadata?

satra: also we shouldn't have to do the `doi.replace`

dandi/tests/test_models.py
```diff
@@ -154,3 +166,192 @@ def test_autogenerated_titles():
...
+
+    # cleaning url
+    _clean_doi(doi)
+    return rp.status_code, rg.status_code
+
+
+def _clean_doi(doi):
+    """removing doi, ignoring the status code"""
+    requests.delete(
+        f"https://api.test.datacite.org/dois/{doi.replace('/', '%2F')}",
```
satra, line 201: also should not need this.

dandi/tests/test_models.py
```diff
@@ -154,3 +166,192 @@ def test_autogenerated_titles():
...
+        "description": "testing",
+        "contributor": [
+            {
+                "name": "A_last, A_first",
+                "roleName": [RoleType("dandi:ContactPerson")],
+            }
+        ],
+        "license": [LicenseType("spdx:CC-BY-4.0")],
+        "publishedBy": "https://doi.test.datacite.org/dois",
+        "datePublished": str(datetime.now().year),
```
satra, line 328: i think i include datetime in datePublished in the model, and should be in isoformat with timezone info.

dandi/models.py
djarecka: inheriting from `DandisetMeta` seems reasonable to me, please let me know what you would like to change

satra: please see the current model. i think you changed with merging master.

dandi/tests/test_models.py
djarecka: indeed, the replacement is not needed. But the form with "activities" I took from datacite examples and it doesn't work without

dandi/tests/test_models.py
djarecka: so I should be using `Publishable` class or some version of `DandisetMeta`, e.g. `PublishedDandisetMeta`?

dandi/models.py
djarecka: oh, i see, the merge was not perfect..

djarecka:
> for the moment you can use a local test, but we should also add a test which downloads from the API, augments it with the missing pieces (because draft metadata may not have doi, url, etc.,.) and then runs through your datacite test using the test server.

And do we have any good way of getting metadata without downloading datasets? If we want to tests existing datasets starting from API, I'd have to somehow get the metadata without downloading the content, and not sure what is the best way.

Right now, I use datalad to get repo (without the content) and then I use `Dandiset(path_to_repo)`.

yarikoptic:
Yes, api. https://github.com/dandi/dandi-cli/blob/master/dandi/dandiapi.py#L301 provides python interface

satra:
right now dataset metadata is embedded in the returned json.

djarecka:
@satra @yarikoptic - thanks! I've used `DandiAPIClient("https://api.dandiarchive.org/api/")` without any tokens, and `version=draft`. Let me know if I should use something else

djarecka:
@satra @yarikoptic - please let me know if you have any suggestions for the changes or the tests

dandi/tests/test_models.py
yarikoptic: attn @djarecka

dandi/tests/test_models.py
```diff
@@ -154,3 +166,265 @@ def test_autogenerated_titles():
...
+    _clean_doi(doi)
+    return rp.status_code, rg.status_code
+
+
+def _clean_doi(doi):
+    """removing doi, ignoring the status code"""
+    requests.delete(
+        f"https://api.test.datacite.org/dois/{doi}",
+        auth=("DARTLIB.DANDI", os.environ["DATACITE_DEV_PASSWORD"]),
+    )
```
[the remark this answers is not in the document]
yarikoptic: may be there is some alternative route as well, e.g. - may be they have a schema (or some validator) to validate against without pestering API?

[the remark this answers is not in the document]
satra: only their xml schema is up to standard. their json schema is some bits of work in progress. i think a CI test is fine. we will also do this as part of the api validation check.

dandi/tests/test_models.py
```diff
@@ -154,3 +166,265 @@ def test_autogenerated_titles():
...
+        "https://raw.githubusercontent.com/datacite/schema/master/source/"
+        "json/kernel-4.3/datacite_4.3_schema.json"
+    )
+    schema = sr.json()
+    return schema
+
+
+@pytest.fixture(scope="module")
+def client():
+    return DandiAPIClient("https://api.dandiarchive.org/api/")
```
yarikoptic, line 216: I think we could use the one we have in the fixture here without relying on the live one and existence of specific dandisets there.

only 000027 (well and 000027) is the dandiset we have under "our control" and the ones which could be used in the tests.

if tests talk to real server, I think we should annotate them somehow, probably with `@pytest.mark.integration`.

@jwodder please review and provide comments and suggestions to get this PR landed.

dandi/models.py
```diff
@@ -113,6 +113,185 @@ def diff_models(model1, model2):
...
+        },
+    ]
+
+    attributes["doi"] = doi
+    attributes["titles"] = [{"title": meta.name}]
+    attributes["descriptions"] = [
+        {"description": meta.description, "descriptionType": "Abstract"}
+    ]
+    attributes["publisher"] = "DANDI Archive"
+    attributes["publicationYear"] = str(meta.datePublished)
```
jwodder, line 185: Why is this called `publicationYear` when it's being set to a timestamp?  Is that intentional?

dandi/models.py
```diff
@@ -113,6 +113,185 @@ def diff_models(model1, model2):
             print(f"{field} is different")
 
 
+DATACITE_CONTRTYPE = [
```
jwodder, line 116: As this is only used for checking containment, it may be more efficient to make it a set.

dandi/models.py
```diff
@@ -113,6 +113,185 @@ def diff_models(model1, model2):
...
+    if getattr(meta, "relatedResource"):
+        attributes["relatedIdentifiers"] = []
+        for rel_el in meta.relatedResource:
+            ident = rel_el.identifier.split(":")
+            if len(ident) == 2:
+                ident_tp, ident_nr = ident
+                if ident_tp.lower() in DATACITE_MAP:
+                    ident_tp = DATACITE_MAP[ident_tp.lower()]
+                else:
+                    raise Exception(
```
jwodder, line 274: You should probably raise something more specific than an `Exception`, like a `ValueError`.

dandi/tests/test_models.py
```diff
@@ -154,3 +166,265 @@ def test_autogenerated_titles():
...
+
+
+@pytest.mark.parametrize("dandi_id", ["000004", "000008"])
+def test_datacite(dandi_id, schema, client):
```
jwodder, line 220: This test should be skipped if `DATACITE_DEV_PASSWORD` is not set so that the test suite can be run without needing that secret.

dandi/tests/test_models.py
```diff
@@ -154,3 +166,265 @@ def test_autogenerated_titles():
...
+    """removing doi, ignoring the status code"""
+    requests.delete(
+        f"https://api.test.datacite.org/dois/{doi}",
+        auth=("DARTLIB.DANDI", os.environ["DATACITE_DEV_PASSWORD"]),
+    )
+
+
+@pytest.fixture(scope="module")
+def schema():
+    sr = requests.get(
```
jwodder, line 206: You should call `sr.raise_for_status()` to ensure that the request was successful.

dandi/tests/test_models.py
```diff
@@ -154,3 +166,265 @@ def test_autogenerated_titles():
...
+    rg = requests.get(url=f"https://api.test.datacite.org/dois/{doi}/activities")
+
+    # cleaning url
+    _clean_doi(doi)
+    return rp.status_code, rg.status_code
+
+
+def _clean_doi(doi):
+    """removing doi, ignoring the status code"""
+    requests.delete(
```
jwodder, line 198: You should call `.raise_for_status()` to ensure that the request is successful.

dandi/tests/test_models.py
```diff
@@ -154,3 +166,265 @@ def test_autogenerated_titles():
...
+        ],
+        "license": [LicenseType("spdx:CC-BY-4.0")],
+    }
+
+    # should work for DandisetMeta but PublishedDandisetMeta should raise an error
+    DandisetMeta(**meta_dict)
+    with pytest.raises(Exception) as exc:
+        PublishedDandisetMeta(**meta_dict)
+
+    assert [el["msg"] == "field required" for el in exc.value.errors()]
```
jwodder, line 269: This does not do what you think it does.  It simply tests that the list is nonempty.  Presumably you meant to wrap this in `all()`?

dandi/tests/test_models.py
```diff
@@ -154,3 +166,265 @@ def test_autogenerated_titles():
...
+                "name": "last name, first name",
+                "roleName": [RoleType("dandi:ContactPerson")],
+            }
+        ],
+        "license": [LicenseType("spdx:CC-BY-4.0")],
+    }
+
+    # should work for DandisetMeta but PublishedDandisetMeta should raise an error
+    DandisetMeta(**meta_dict)
+    with pytest.raises(Exception) as exc:
```
jwodder, line 266: Below, you seem to be assuming that the exception is a pydantic `ValidationError`.  This should thus be changed to `with pytest.raises(ValidationError) as exc:`.

dandi/tests/test_models.py
```diff
@@ -154,3 +166,265 @@ def test_autogenerated_titles():
...
+                "creators": (1, {"name": "A_last, A_first"}),
+                "contributors": (
+                    2,
+                    {"name": "A_last, A_first", "contributorType": "Other"},
+                ),
+            },
+        ),
+    ],
+)
+def test_dantimeta_datacite(schema, additional_meta, datacite_checks):
```
jwodder, line 373: This test should be skipped if `DATACITE_DEV_PASSWORD` is not set so that the test suite can be run without needing that secret.

dandi/tests/test_models.py
```diff
@@ -154,3 +166,265 @@ def test_autogenerated_titles():
...
+        headers={"Content-Type": "application/vnd.api+json"},
+        auth=("DARTLIB.DANDI", os.environ["DATACITE_DEV_PASSWORD"]),
+    )
+
+    # checking if i'm able to get the url
+    rg = requests.get(url=f"https://api.test.datacite.org/dois/{doi}/activities")
+
+    # cleaning url
+    _clean_doi(doi)
+    return rp.status_code, rg.status_code
```
jwodder, line 193: Is there a reason to assert on the status codes instead of calling `raise_for_status()` on the responses?

dandi/tests/test_models.py
```diff
@@ -8,10 +16,14 @@
     IdentifierType,
     LicenseType,
     ParticipantRelationType,
+    PublishedDandisetMeta,
     RelationType,
     RoleType,
+    to_datacite,
 )
 
+METADATA_DIR = Path(__file__).with_name("data") / "metadata"
```
jwodder, line 25: This appears to be unused.

jwodder (changes requested):

dandi/models.py
djarecka: yes, can move it to set

dandi/tests/test_models.py
djarecka: yes, i did mean to add run `all`, thanks!

dandi/tests/test_models.py
djarecka: this is sometimes run "just in case" before creating a new doi, so it can return an error

dandi/models.py
djarecka: moved to `meta.datePublished.year`

dandi/tests/test_models.py
djarecka: probably not, changed it

dandi/tests/test_models.py
djarecka: @satra - sorry, forgot to answer, but looks like `str(datetime.now().year)` works fine - do you expect that the error should be raised?

dandi/tests/test_models.py
djarecka: I also could come back to the previous version where I used metadata from real dandiset as examples, but I saved the metadata as a json file in the repo

yarikoptic:
Please merge master or rebase -- "Test populate_dandiset_yaml.py / test (push) " should be gone now.

yarikoptic:
didn't look through all details (we should put con/tinuous to work here, filed #639) but those failing which are saw are of the 
```
dandi/tests/test_models.py:232: in test_datacite
    newmeta = PublishedDandisetMeta(**newmeta_js)
pydantic/main.py:406: in pydantic.main.BaseModel.__init__
    ???
E   pydantic.error_wrappers.ValidationError: 1 validation error for PublishedDandisetMeta
E   assetsSummary
E     field required (type=value_error.missing)
```
nature

djarecka:
the biggest problem is that `assetsSummary` was added to `PublishedDandisetMeta`. Do you have any function to create it?

satra:
that would be a good function to right. this function should iterate over all assets in a dandiset to create a summary of unique values for those fields.

however, this is also what yarik was going to add to the asset. for example approach/technique are not being populated at the asset level right now. we can relax some of the components to optional, but i think species, approach, technique should really be there.

djarecka:
i'm a bit confused since it looks like a combination of information that should be provided by the author and info that could be calculated automatically...

satra:
all of that info is calculated from asset metadata, hence the name. the authors would provide this when creating nwb files (and they generally do). we have to do a bit more magic for bids datasets, so those would likely not be publishable till we have figured out some pieces.

djarecka:
ok, so for testing, I guess it's ok for me to use some random entry

satra:
> I guess it's ok for me to use some random entry

yes

djarecka:
for now, I just saved meta from 000004 and 00008 in `tests/data/metadata` for testing. I could work on dandiset 000027 and additional dandisets later. Today I was not really able to use `https://gui-staging.dandiarchive.org/` (doesn't load) and I still can't edit `000027`

djarecka:
not sure what to do with `Test model schemata / publish` test

yarikoptic:
well, if you look at that fail run fail you will see
```
diff --git a/releases/0.3.1/published-dandiset.json b/releases/0.3.1/published-dandiset.json
index 14b909b..3f03ea5 100644
--- a/releases/0.3.1/published-dandiset.json
+++ b/releases/0.3.1/published-dandiset.json
@@ -270,7 +270,8 @@
     "contributor",
     "license",
     "assetsSummary",
-    "version"
+    "version",
+    "doi"
   ],
   "definitions": {
     "RoleType": {
Error:  Existing schema files modified instead of creating a new version
Error: Process completed with exit code 1.
```
so this PR modifies schema by adding `doi` (I am just surprised we did not have it for published-dandiset yet).  I guess it is intended, so let's just boost schema version to 0.3.2 and merge.  Or @satra -- you want to mutate 0.3.1? change is minor but not sure if would not cause some ripples in dandi-api / webui.

satra:
@yarikoptic - the doi was already there in the model, and should also be non-optional. this mutates it by removing a default value, which i think has an impact on whether or not it got exported to JSON schema. so this was a bug in the schema generation which this PR fixes.

dandi/tests/test_models.py
yarikoptic: shouldn't it matter what kind of error it is?  e.g. if we disconnected an not removed DOI -- we should try again.  `grep` for `tenacity` to see how we orchestrate retries ATM.

Note: we now have 57 DOIs in test. instance of datacite, some of them seems to be for the same dataset(s).  I wonder if some of those are due to not carrying if cleanup succeeded or not, and shouldn't we actually cleanup most if not all of them? ;-)

dandi/models.py
```diff
@@ -116,6 +116,185 @@ def diff_models(model1, model2):
...
+                ident_tp, ident_nr = ident
+                if ident_tp.lower() in DATACITE_MAP:
+                    ident_tp = DATACITE_MAP[ident_tp.lower()]
+                else:
+                    raise ValueError(
+                        f"identifier has to be from the list: {DATACITE_IDENTYPE}, "
+                        f"but {ident_tp} provided"
+                    )
+            else:
+                raise Exception("identifier is expected to be type:number")
```
yarikoptic, line 282: seems to be also of `ValueError` kind and we better inform the value on which we choked:

```suggestion
                raise ValueError(f"identifier is expected to be type:number, got {rel_el.identifier}")
```

yarikoptic:
ok, left minor comments but could be handled post-merge/migration to dedicated `dandischema`. 
So I guess we could just wait for @dchiquito to give blessing to mutate or could boost the .patch and proceed with the merge.

dandi/models.py
yarikoptic: Let  me just proceed with this invaluable contribution of mine -- hopefully it doesn't break anything

dandi/tests/test_models.py
yarikoptic: yeah -- we should not rely on testing on real arbitrary datasets out of our control, at least until we know that they are metadata records are as kosher as they could be and that we have safeguards that nobody could make them invalid ;)   but ok, let's proceed as is for now and adjust later.

djarecka: this has been removed after our discussion on Monday

yarikoptic:
Ok, @dchiquito gave blessing in slack, I fixed lint issue I introduced, let's proceed.  Anything remaining should be done against `dandischema`

dandi/tests/test_models.py
yarikoptic: Thank you @djarecka ! I must had been looking at the outdated diff while checking on this before merge :-/ Great!

PR #602, opened 2021-04-28
Recognize "DANDI:<identifier>" strings as URL-likes

### LINKED ISSUE #600: `dandi ls` should support dandi identifiers
currently `dandi ls` does not support dandi identifiers such as below.

```
$ dandi ls DANDI:000008
$ dandi ls DANDI:000008/draft
$ dandi ls DANDI:000008/<version>
```

unfortunately, identifiers.org is not case sensitive, so the above identifiers also work in lowercase.

jwodder:
Closes #600.

### DIFF
--- a/dandi/utils.py
+++ b/dandi/utils.py
@@ -716,7 +716,8 @@ def is_url(s):
 
     TODO: redo
     """
-    return s.lower().startswith(("http://", "https://", "dandi://"))
+    return s.lower().startswith(("http://", "https://", "dandi:"))
+    # Slashes are not required after "dandi:" so as to support "DANDI:<id>"
 
 
 def get_module_version(module: Union[str, types.ModuleType]) -> Optional[str]:


dandi/utils.py
```diff
@@ -716,7 +716,8 @@ def is_url(s):
 
     TODO: redo
     """
-    return s.lower().startswith(("http://", "https://", "dandi://"))
+    return s.lower().startswith(("http://", "https://", "dandi:"))
+    # Slashes are not required after "dandi:" so as to support "DANDI:<id>"
```
yarikoptic, line 720: Indeed, Wikipedia says

```
URI = scheme:[//authority]path[?query][#fragment]
```

yarikoptic: Just then spotted it says URI not URL, so remains unclear but https://url.spec.whatwg.org/ suggests that could be used interchangeably... So whatever, let it be this way

PR #603, opened 2021-04-29
Remove numpy pre-pinning in test.yml

jwodder:
This started out as a test to see what happened, but with the adjustment of pytest's warning settings, I've gotten the change to work.


yarikoptic:
yikes.. well -- CI is green, so I am happy. Let's proceed!

PR #604, opened 2021-04-29
Remove unused functions (and other cleanups)

jwodder:
Most of this consists of deleting unused code found with [dead](https://pypi.org/project/dead/) (which is not currently suitable for use as an automated linter).


yarikoptic:
ok, I don't mind - unlikely someone relied on those as part of the `dandi` Python package API surface.  
Tests seems to be happy, so let's proceed.
But may be adjust pull request title (it is the one to become CHANGELOG entry right?) to signal that some unused functions/methods were removed since not used internally, and make it `minor` release (again - due to "API" purge)

yarikoptic (approved):

jwodder:
@yarikoptic Done.

PR #605, opened 2021-04-29
Restructure parse_dandi_url() return type

### LINKED ISSUE #554: Restructure parse_dandi_url() return type
As `parse_dandi_url()` grows to support a greater variety of URLs that require fetching assets in different ways, I think it's past time we change the function to return object instances with methods for retrieving a URL's client, Dandiset, and assets.

Specifically, I propose the following type hierarchy:

* The return type of `parse_dandi_url()` will be an instance of the abstract class `DandiURL`, whose attributes & methods are:
    * `api_url: str` — the base URL of the API
    * `server_type: Literal["girder", "api"]`
    * `dandiset_id: str`
    * `version_id: Optional[str]`
    * `get_client()` — Returns either a Girder or DandiAPIClient instance, as appropriate
    * `get_dandiset(client) -> dict`
    * `get_version_id(client) -> str` — Returns `version_id` if it is set or else determines it based on the Dandiset's state
    * `abstractmethod get_assets(client, include_metadata=False) -> Iterator[dict]`
        * Note that some subclasses may always include metadata regardless of the value of `include_metadata`.

* The concrete subclasses of `DandiURL` include:
    * `DandisetURL` — for URLs that only refer to a Dandiset (possibly with a version); `get_assets()` returns all assets in the Dandiset
    * `AssetIDURL` — for URLs that identify an asset by ID
    * `AssetPathPrefixURL` — for URLs that identify a collection of assets by path prefix
    * `AssetItemURL` — for URLs that identify a specific asset by path
    * `AssetFolderURL` — for URLs that identify a collection of assets by folder path

@yarikoptic Thoughts?

### LINKED ISSUE #598: ls -r dandi:// url with /path lists only the dandiset  and not assets
```
$> dandi ls -F path -r "dandi://000026"             
[]
2021-04-27 15:38:45,191 [    INFO] Logs saved in /home/yoh/.cache/dandi-cli/log/20210427193845Z-2089678.log
Error: We do not know how to map URL dandi://000026 to our servers.
Patterns for known setups:
 - DANDI:<dandiset id>
 - https://dandiarchive.org/...
 - https://identifiers.org/DANDI:<dandiset id>
 - https://gui-beta-dandiarchive-org.netlify.app/#/dandiset/<dandiset id>[/<version>][/files[?location=<path>]]
 - https://*dandiarchive-org.netflify.app/...
 - https://<server>[/api]/dandisets/<dandiset id>[/versions[/<version>]]
 - https://<server>[/api]/dandisets/<dandiset id>/versions/<version>/assets/<asset id>[/download]
 - https://<server>[/api]/dandisets/<dandiset id>/versions/<version>/assets/?path=<path>
 - https://<server>[/api]#/dandiset/<dandiset id>[/draft][/files[?_id=<id>[&_modelType=folder]]]
 - dandi://<instance name>/<dandiset id>[@<version>][/<path>]
 - https://<server>/...
```

says that we support `dandi://<instance name>/<dandiset id>[@<version>][/<path>]`  URLs. But if `/<path>` is provided it seems to not how assets:

```shell
$> dandi ls -F path -r "dandi://dandi/000026/sub-I46"   
- asset_count: 1196
  created: '2020-06-24T18:34:50.022000Z'
  metadata:
... 
```
so it uses yaml since a single entry (dandiset) and no assets from under sub-I46 are listed.

Also, whenever some path is provided (even if just the trailing `/`) no dandiset record should probably be shown.

jwodder:
Closes #554.

Closes #598.

### DIFF
--- a/dandi/cli/cmd_download.py
+++ b/dandi/cli/cmd_download.py
@@ -101,8 +101,8 @@ def download(
     if dandi_instance is not None:
         if url:
             for u in url:
-                _, server_url, _, _ = parse_dandi_url(u)
-                if known_instances_rev.get(server_url.rstrip("/")) != dandi_instance:
+                parsed_url = parse_dandi_url(u)
+                if known_instances_rev.get(parsed_url.api_url) != dandi_instance:
                     raise click.UsageError(
                         f"{u} does not point to {dandi_instance!r} instance"
                     )
--- a/dandi/cli/cmd_ls.py
+++ b/dandi/cli/cmd_ls.py
@@ -114,7 +114,7 @@ def ls(paths, schema, metadata, fields=None, format="auto", recursive=False, job
                 with parsed_url.navigate(
                     include_metadata=metadata in ("all", "assets")
                 ) as (client, dandiset, assets):
-                    if isinstance(parsed_url, DandisetURL) or not recursive:
+                    if isinstance(parsed_url, DandisetURL):
                         rec = {
                             "path": dandiset.pop("dandiset", {}).get(
                                 "identifier", "ERR#%s" % id(dandiset)
@@ -124,7 +124,7 @@ def ls(paths, schema, metadata, fields=None, format="auto", recursive=False, job
                         # rec.update(dandiset.get('metadata', {}))
                         rec.update(dandiset)
                         yield rec
-                    else:
+                    if not isinstance(parsed_url, DandisetURL) or recursive:
                         yield from assets
             else:
                 # For now we support only individual files
--- a/dandi/cli/cmd_ls.py
+++ b/dandi/cli/cmd_ls.py
@@ -5,7 +5,7 @@ import os.path as op
 import click
 
 from .base import lgr, map_to_click_exceptions
-from ..dandiarchive import _dandi_url_parser
+from ..dandiarchive import DandisetURL, _dandi_url_parser, parse_dandi_url
 from ..utils import is_url
 
 # TODO: all the recursion options etc
@@ -110,11 +110,11 @@ def ls(paths, schema, metadata, fields=None, format="auto", recursive=False, job
     def assets_gen():
         for path in paths:
             if is_url(path):
-                from ..dandiarchive import navigate_url
-
-                with navigate_url(path) as (client, dandiset, assets):
-                    if dandiset:
-                        dandiset_id = dandiset.get("dandiset", {}).get("identifier")
+                parsed_url = parse_dandi_url(path)
+                with parsed_url.navigate(
+                    include_metadata=metadata in ("all", "assets")
+                ) as (client, dandiset, assets):
+                    if isinstance(parsed_url, DandisetURL) or not recursive:
                         rec = {
                             "path": dandiset.pop("dandiset", {}).get(
                                 "identifier", "ERR#%s" % id(dandiset)
@@ -125,19 +125,7 @@ def ls(paths, schema, metadata, fields=None, format="auto", recursive=False, job
                         rec.update(dandiset)
                         yield rec
                     else:
-                        dandiset_id = None
-                    if recursive and assets:
-                        if metadata in ("all", "assets"):
-                            for a in assets:
-                                if "metadata" not in a:
-                                    a["metadata"] = client.get_asset(
-                                        dandiset_id,
-                                        dandiset["version"],
-                                        a["asset_id"],
-                                    )
-                                yield a
-                        else:
-                            yield from assets
+                        yield from assets
             else:
                 # For now we support only individual files
                 yield path
--- a/dandi/dandiarchive.py
+++ b/dandi/dandiarchive.py
@@ -1,7 +1,7 @@
 from abc import ABC, abstractmethod
 from contextlib import contextmanager
 import re
-from typing import Iterator, Optional
+from typing import Iterator, Optional, Tuple
 from urllib.parse import unquote as urlunquote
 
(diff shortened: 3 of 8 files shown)


dandi/dandiarchive.py
```diff
@@ -22,8 +16,200 @@
...
+        return client.get_dandiset(self.dandiset_id, self.get_version_id(client))
+
+    def get_version_id(self, client: DandiAPIClient) -> str:
+        if self.version_id is None:
+            r = client.get(f"/dandisets/{self.dandiset_id}/")
+            published_version = r["most_recent_published_version"]
+            if published_version:
+                return published_version["version"]
+            else:
+                return r["draft_version"]["version"]
```
yarikoptic, line 41:
```suggestion
            version = r["most_recent_published_version"] or r["draft_version"]
            return version["version"]
```
so we centralize a bit happen we need to later make it into `version.get("version")` etc

dandi/dandiarchive.py
```diff
@@ -1,19 +1,13 @@
...
-Eventually it is largely to be "dissolved" whenever we stop talking to girder.
-"""
-
+from abc import ABC, abstractmethod
 from contextlib import contextmanager
 import re
+from typing import Iterator, Optional, Tuple
 from urllib.parse import unquote as urlunquote
 
+from pydantic import AnyHttpUrl, BaseModel, parse_obj_as, validator
```
yarikoptic, line 7: just a note: likely because of this module RF and extra imports (but didn't investigate), code of this PR adds ~40 msec to `dandi --help` (to the total of 800ms, so about 5%)... at some point we might review import paths etc if we keep growing - since waiting seconds for `--help` might be a bit too much, and would be impediment if we add some shell completions support (IIRC I did not find any ready to use for `click` such as argcomplete which was odd)

dandi/dandiarchive.py
```diff
@@ -22,8 +16,200 @@
...
+        ...
+
+    def get_asset_ids(self, client: DandiAPIClient) -> Iterator[str]:
+        for a in self.get_assets(client):
+            yield a["asset_id"]
+
+    @contextmanager
+    def navigate(
+        self, include_metadata: bool = True
+    ) -> Iterator[Tuple[DandiAPIClient, dict, Iterator[dict]]]:
```
yarikoptic, line 58: please add a docstring describing the purpose and  what it would be yielding in those tuples?

Also since it is the only method which doesn't demand `client` - document that it establishes the session for the client while navigating. 
But if that is how all the other methods are expected to be used, why not to make `.client` to be the lazy property and this one start/stopping the session?

dandi/dandiarchive.py
```diff
@@ -22,8 +16,200 @@
 lgr = get_logger()
 
 
+class ParsedDandiURL(ABC, BaseModel):
```
yarikoptic, line 19: Please add high level description for this base class and what to expect subclasses to override (seems to be `get_assets`?) and may be that all methods but `get_client` and `navigate` expect client to be passed ?

yarikoptic (changes requested):
minor comments.  I like how changes look in tests -- much more self-descriptive code ;)

dandi/dandiarchive.py
jwodder: FYI, click has command completion built-in: https://click.palletsprojects.com/en/7.x/bashcomplete/ (though I think some details may be changing in Click 8).

dandi/dandiarchive.py
jwodder: I chose to make `client` an argument in order to avoid proliferation of clients in cases where the calling code has its own client instance.

dandi/cli/cmd_ls.py
```diff
@@ -110,11 +110,11 @@ def ls(paths, schema, metadata, fields=None, format="auto", recursive=False, job
     def assets_gen():
         for path in paths:
             if is_url(path):
-                from ..dandiarchive import navigate_url
```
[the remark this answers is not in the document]
jwodder: Should the test just be a smoke test that calls `ls` with a URL (pointing to dandiarchive.org or the Docker instance?), or should it also expect something from the output?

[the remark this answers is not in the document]
yarikoptic: I guess with `-f json_pp` you could expect nicely formatted record you could loads and if ran on a dandiset with assets -- could check that there is at least one asset (without getting into details)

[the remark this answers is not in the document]
yarikoptic: by `-f json_pp` I didn't mean "run in command line"... `format="json_pp"` ;)

dandi/cli/cmd_download.py
```diff
@@ -101,8 +101,8 @@ def download(
     if dandi_instance is not None:
         if url:
```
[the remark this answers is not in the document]
jwodder: I made a separate PR for that: https://github.com/dandi/dandi-cli/pull/607

dandi/dandiarchive.py
yarikoptic: cool, filed https://github.com/dandi/dandi-cli/issues/608 . Feel welcome to self-assign but it is of no urgency so I didn't

dandi/dandiarchive.py
jwodder: Done.

dandi/dandiarchive.py
jwodder: Docstring added.

PR #609, opened 2021-04-30
DOC: Extend description for delete to point that it could be URL etc

yarikoptic:
isn't that right @jwodder ?

### DIFF
--- a/dandi/cli/cmd_delete.py
+++ b/dandi/cli/cmd_delete.py
@@ -10,7 +10,11 @@ from .base import devel_debug_option, instance_option, map_to_click_exceptions
 @devel_debug_option()
 @map_to_click_exceptions
 def delete(paths, skip_missing, dandi_instance="dandi", devel_debug=False):
-    """ Delete Dandisets and assets from the server """
+    """Delete dandisets and assets from the server.
+
+    PATH could be a local path or a URL to an asset, directory, or an entire
+    dandiset.
+    """
     from ..delete import delete
 
     delete(


dandi/cli/cmd_delete.py
```diff
@@ -10,7 +10,11 @@
 @devel_debug_option()
 @map_to_click_exceptions
 def delete(paths, skip_missing, dandi_instance="dandi", devel_debug=False):
-    """ Delete Dandisets and assets from the server """
+    """Delete Dandisets and assets from the server
```
jwodder, line 13: There needs to be a period at the end of this line to terminate the short description shown in `dandi --help`.

dandi/cli/cmd_delete.py
```diff
@@ -10,7 +10,11 @@
...
+
+    PATH could be a local path or a URL to an asset, directory, or an entire
+    dandiset.
```
jwodder, line 16: "Dandiset" should be capitalized, shouldn't it?

jwodder (changes requested):

dandi/cli/cmd_delete.py
yarikoptic: THANKS!

dandi/cli/cmd_delete.py
yarikoptic: For consistency with prior summary one -- sure... but it seems that we do not have consistency in general, but other commands use it lowercased.  So I will make it lower cased here as well and it would look like

```
Usage: dandi [OPTIONS] COMMAND [ARGS]...

  A client to support interactions with DANDI archive
  (http://dandiarchive.org).

  To see help for a specific command, run

      dandi COMMAND --help

  e.g. dandi upload --help

Options:
  --version
  -l, --log-level [DEBUG|INFO|WARNING|ERROR|CRITICAL]
                                  Log level (case insensitive).  May be
                                  specified as an integer.  [default: INFO]

  --pdb                           Fall into pdb if errors out
  --help                          Show this message and exit.

Commands:
  delete    Delete dandisets and assets from the server.
  digest    Calculate file digests
  download  Download a file or entire folder from DANDI
  ls        List .nwb files and dandisets metadata.
  organize  (Re)organize files according to the metadata.
  upload    Upload dandiset (files) to DANDI archive.
  validate  Validate files for NWB (and DANDI) compliance.

```

yarikoptic:
I have addressed the comments, let's proceed.  Thanks @jwodder for the review!

PR #610, opened 2021-05-03
Cut down on some code duplication in delete.py

### LINKED ISSUE #606: delete: RF Deleter to avoid code duplication
Looking at repetitive pattern of 
```python
        try:
            self.set_dandiset(parsed_url.api_url, parsed_url.dandiset_id)
        except NotFoundError:
            if self.skip_missing:
                return
            else:
                raise
```
across all methods it begs the question on either repetition could be avoided one way or another.... E.g. if we limit Deleter to be able to operate only on a single dandiset instance (so url and id must stay the same), `set_dandiset` could be invoked in constructor... yet to recall what use cases we want to fulfill, but ran into it while reviewing https://github.com/dandi/dandi-cli/pull/605/files#

jwodder:
Closes #606.

### DIFF
--- a/dandi/delete.py
+++ b/dandi/delete.py
@@ -36,7 +36,11 @@ class Deleter:
     def __bool__(self) -> bool:
         return self.deleting_dandiset or bool(self.remote_assets)
 
-    def set_dandiset(self, api_url: str, dandiset_id: str) -> None:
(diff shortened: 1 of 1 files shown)


yarikoptic:
Thank you! looks legit

PR #612, opened 2021-05-04
Make test_server_info use Docker fixture if DANDI_REDIRECTOR_BASE is set

jwodder:
This is needed for https://github.com/dandi/redirector/pull/47.

### DIFF
--- a/dandi/tests/test_utils.py
+++ b/dandi/tests/test_utils.py
@@ -1,4 +1,5 @@
 import inspect
+import os
 import os.path as op
 import time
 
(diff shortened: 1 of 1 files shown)


yarikoptic:
one run consistently (restarted it) errors out: https://github.com/dandi/dandi-cli/pull/612/checks?check_run_id=2501911680  is that expected (e.g. requires that change in redirector) or a "bug"?

we will need to also release dandi-cli for the purpose of https://github.com/dandi/redirector/pull/47 right? (then add `release` label)

jwodder:
@yarikoptic The errors are occurring because the connection to Dandiarchive keeps failing for some reason.

Yes, a release is required, unless we want to change the redirector tests to clone dandi-cli from GitHub.

yarikoptic:
ok, let's just proceed -- it doesn't fail on pull_request one which is what matters AFAIK

PR #613, opened 2021-05-04
add Dataset neurodata_types

### LINKED ISSUE #611: pynwb metadata: get neurodata types
This function assumes that neurodata_types can only be Groups

https://github.com/dandi/dandi-cli/blob/5a4f24c518b9fda2e2343afce5f2fc88fc7c20da/dandi/pynwb_utils.py#L171-L178

Although it is less common, Datasets can also have a neurodata_type.

bendichter:
fix #611

### DIFF
--- a/dandi/pynwb_utils.py
+++ b/dandi/pynwb_utils.py
@@ -173,6 +173,9 @@ def _scan_neurodata_types(grp):
     if "neurodata_type" in grp.attrs:
         out.append((grp.attrs["neurodata_type"], grp.attrs.get("description", None)))
     for v in list(grp.values()):
+        if isinstance(v, h5py._hl.dataset.Dataset):
+            if "neurodata_type" in v.attrs:
+                out.append((v.attrs["neurodata_type"], v.attrs.get("description", None)))
         if isinstance(v, h5py._hl.group.Group):
             out += _scan_neurodata_types(v)
     return out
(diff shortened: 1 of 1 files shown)


dandi/pynwb_utils.py
```diff
@@ -173,6 +173,9 @@ def _scan_neurodata_types(grp):
     if "neurodata_type" in grp.attrs:
         out.append((grp.attrs["neurodata_type"], grp.attrs.get("description", None)))
     for v in list(grp.values()):
+        if isinstance(v, h5py._hl.dataset.Dataset):
+            if "neurodata_type" in v.attrs:
+                out.append((v.attrs["neurodata_type"], v.attrs.get("description", None)))
         if isinstance(v, h5py._hl.group.Group):
```
yarikoptic, line 179:
```suggestion
        elif isinstance(v, h5py._hl.group.Group):
```

yarikoptic (commented):
do you think it would be feasible to come up with a unittest for this?  we could either have some tiny sample file (under `dandi/tests/data/sample_files`) or have a fixture (which is our current preferred way but we might then not spot fresh hdmf/pynwb "load issues" if every come back) which generates such a file.

yarikoptic:
ping @bendichter , see my question above

bendichter:
@yarikoptic could you point me to an example fixture? I could easily add to it to create a Dataset-based neurodata_type

yarikoptic:
@bendichter should we finish up this respectfully senior in age PR? ;)

Issue #614, opened 2021-05-05
Session has disappeared from filename construction


satra:
The file in dandiset 69 is a file from another dandiset. When you see the metadata, session is there. But it is not in the name. https://api.dandiarchive.org/api/dandisets/000069/versions/draft/assets/93b8e5ef-69fd-4ac1-863f-5ff1f9210554/

This was also reported by cody when fixing one of the dandisets.

If this is intentional, it should be made clear. I have no problems with filenames being as short as possible with changes made only when disambiguating. However I don't think the cli is ready for that yet. For example, adding a new file that runs into a clash. It may also be disruptive in the short run in terms of human expectations.

yarikoptic:
uff... first I thought that it was just a single asset among many and it didn't have session while others had: which could have happened is that it was an incremental `organize` of more files (from the same session) into an existing dandiset. Since session is optional (relates to prior discussion in #69) and was the same among multiple files, it was not added.

But apparently it is just a single asset in that dandiset: https://gui.dandiarchive.org/#/dandiset/000069/draft/files?location=sub-anm106211%2F  and it is exactly the discussion we had in #69 and "optional" nature of the session field -- we do not populate it unless it would help to resolve conflicts.  So it did not "disappear", it was never in the filename to  begin with.

satra:
ok if it is optional, then indeed we should consider what partial updates means (like cody ran into): https://github.com/dandisets/000053/issues/1 

this seems to imply some logic needed beyond just naming locally.

yarikoptic:
for partial updates (also attn @CodyCBakerPhD), I think we should just instruct people ATM to follow up with a subsequent run of `dandi organize -f move` in that dandiset **after** "they are done" with all desired incremental additions.  The reason is that loading metadata does take time (although we cache it) so on a large dandiset it could take awhile, and thus loading metadata from all the files in dandiset should not be done "all the time" for incremental updates.

I have now tried on the 000053, where I first renamed two files into "without session":

```
drogon:/mnt/backup/dandi/trash/000053
*$> git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        deleted:    sub-npI1/sub-npI1_ses-20190415_behavior+ecephys.nwb
        deleted:    sub-npI1/sub-npI1_ses-20190415_behavior.nwb

Untracked files:
[...]

PR #615, opened 2021-05-06
RF: organize - should no longer alter dandiset.yaml

yarikoptic:
We will just rely on users to populate it online without providing them
with any assistance locally since we do not upload dandiset.yaml
anyways.

Discussion/observation which triggered this change: https://github.com/dandi/dandi-cli/issues/614#issuecomment-832921269


jwodder (approved):

yarikoptic:
thank you @jwodder for the review, let's proceed

PR #616, opened 2021-05-06
Add "sync" option for upload & download

### LINKED ISSUE #365: upload/download(/organize): --sync mode
Both `upload` and `download` should have an option to "sync" two copies of a dandiset in its entirety.  This functionality need was mentioned before: https://github.com/dandi/dandi-cli/issues/41#issuecomment-593183534 and https://github.com/dandi/dandi-cli/issues/48 but apparently there were no clear dedicated issue for that, so decided to file a new issue (although #48 is likely to be addressed along).

Should be 
- done only toward dandi-api (no girder) implementation . Thus at large is blocked ATM by #347
- smart about avoiding re-uploads  of files: files already uploaded must not be re-uploaded. I see multiple ways we could do that. At large specific implementation would depend on either we would like to "preserve" the identity/provenance of an already existing asset (if we keep asset UUID preserved while changing its metadata).
- smart about "deletion" of assets on the "remote" end (e.g. to not delete until a replacement file is obtained)

jwodder:
Closes #365.

Does not include renaming.

### DIFF
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -8,7 +8,6 @@ from shutil import rmtree
 import sys
 import time
 
-import click
 import humanize
 import requests
 
@@ -18,6 +17,7 @@ from .dandiset import Dandiset
 from . import get_logger
 from .support.pyout import naturalsize
 from .utils import (
+    abbrev_prompt,
     ensure_datetime,
     find_files,
     flattened,
@@ -132,11 +132,23 @@ def download(
                 a_path = a_path.replace("\\", "/")
             if a_path not in asset_paths:
                 to_delete.append(p)
-        if to_delete and click.confirm(
-            f"Delete {pluralize(len(to_delete), 'local asset')}?"
-        ):
-            for p in to_delete:
-                os.unlink(p)
+        if to_delete:
+            while True:
+                opt = abbrev_prompt(
+                    f"Delete {pluralize(len(to_delete), 'local asset')}?",
+                    "yes",
+                    "no",
+                    "list",
+                )
+                if opt == "list":
+                    for p in to_delete:
+                        print(p)
+                elif opt == "yes":
+                    for p in to_delete:
+                        os.unlink(p)
+                    break
+                else:
+                    break
 
 
 def download_generator(
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -17,7 +17,14 @@ from .dandiarchive import DandisetURL, MultiAssetURL, SingleAssetURL, parse_dand
 from .dandiset import Dandiset
 from . import get_logger
 from .support.pyout import naturalsize
-from .utils import ensure_datetime, find_files, flattened, is_same_time, on_windows
+from .utils import (
+    ensure_datetime,
+    find_files,
+    flattened,
+    is_same_time,
+    on_windows,
+    pluralize,
+)
 
 lgr = get_logger()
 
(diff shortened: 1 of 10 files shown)


dandi/download.py
```diff
@@ -81,10 +101,38 @@ def download(
...
+        to_delete = []
+        for p in find_files(".*", download_dir, exclude_datalad=True):
+            if p == op.join(output_path, dandiset_metadata_file):
+                continue
+            a_path = op.normpath(op.join(prefix, op.relpath(p, download_dir)))
+            if on_windows:
+                a_path = a_path.replace("\\", "/")
+            if a_path not in asset_paths:
+                to_delete.append(p)
+        if to_delete and click.confirm(f"Delete {len(to_delete)} local assets?"):
```
yarikoptic, line 128:
```suggestion
        if to_delete and click.confirm(f"Delete {len(to_delete)} local asset{'s' if len(to_delete) > 1 else ''}?"):
```

dandi/tests/test_download.py
```diff
@@ -180,3 +180,43 @@ def test_download_asset_id(local_dandi_api, text_dandiset, tmp_path):
...
+    dspath = tmp_path / text_dandiset["dandiset_id"]
+    os.rename(text_dandiset["dspath"], dspath)
+    confirm_mock = mocker.patch("click.confirm", return_value=confirm)
+    download(
+        f"dandi://{local_dandi_api['instance_id']}/{text_dandiset['dandiset_id']}",
+        tmp_path,
+        existing="overwrite",
+        sync=True,
+    )
+    confirm_mock.assert_called_with("Delete 1 local assets?")
```
yarikoptic, line 199:
```suggestion
    confirm_mock.assert_called_with("Delete 1 local asset?")
```
if patch above taken

dandi/tests/test_download.py
```diff
@@ -180,3 +180,43 @@ def test_download_asset_id(local_dandi_api, text_dandiset, tmp_path):
...
+        text_dandiset["dandiset_id"], "draft", "subdir2/banana.txt"
+    )
+    confirm_mock = mocker.patch("click.confirm", return_value=True)
+    download(
+        f"dandi://{local_dandi_api['instance_id']}/{text_dandiset['dandiset_id']}/subdir2/",
+        text_dandiset["dspath"],
+        existing="overwrite",
+        sync=True,
+    )
+    confirm_mock.assert_called_with("Delete 1 local assets?")
```
yarikoptic, line 220:
```suggestion
    confirm_mock.assert_called_with("Delete 1 local asset?")
```

dandi/download.py
yarikoptic: Let's improve UX right away, and instead of plain confirm give a choice for "[l]ist" or alike and if asked, list all the assets to be deleted so user could make an informed decision

yarikoptic: filed related https://github.com/dandi/dandi-cli/issues/620 on allowing for non-interactive, which might be done after or melded into this PR if we agree before it is ready

dandi/upload.py
```diff
@@ -335,3 +341,22 @@ def upload_agg(*ignored):
...
+        to_delete = []
+        for asset in client.get_dandiset_assets(
+            ds_identifier, "draft", path=path_prefix
+        ):
+            if (
+                any(p == "" or path_is_subpath(asset["path"], p) for p in relpaths)
+                and not Path(dandiset.path, asset["path"]).exists()
+            ):
+                to_delete.append(asset["asset_id"])
+        if to_delete and click.confirm(f"Delete {len(to_delete)} assets on server?"):
```
yarikoptic, line 360: do the same pluralism here as I suggested for download.  could even be made into a helper function in tools to avoid duplication

yarikoptic (changes requested):
From code review alone looks fine, yet to try. Thank you @jwodder.  Left minor suggestion and filed a dedicated issue about allowing for non-interactive use.

dandi/upload.py
jwodder: Done.

dandi/download.py
jwodder: Done.

yarikoptic:
Didn't try but looks great. Thank you @jwodder , let's proceed

PR #621, opened 2021-05-07
Add dandi-staging to known_instances

dchiquito:
There is now a staging API server set up at https://api-staging.dandiarchive.org, so it would be cool if dandi-cli could upload to it.

### DIFF
--- a/dandi/consts.py
+++ b/dandi/consts.py
@@ -86,6 +86,11 @@ known_instances = {
         None,
         None,
     ),
+    "dandi-staging": dandi_instance(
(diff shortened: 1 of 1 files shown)


jwodder (approved):

yarikoptic:
sweet, didn't try it but let's proceed.

BTW, we left
```
    "dandi-devel": dandi_instance(
        "https://gui-beta-dandiarchive-org.netlify.app",
        None,
        None,
    ),
```
in place -- is that web ui instance ever to come back or should we kill it?

PR #623, opened 2021-05-09
REF: updated model requirements

satra:
This PR improves the schema to:
- address some issues with identifiers (specific forms)
- adds publishing the publish schemas
- cleans up some of the descriptions and requirements
- remove metadata migration code

technically this could increase schema version, but also ok to just overwrite it i think.


yarikoptic:
> technically this could increase schema version, but also ok to just overwrite it i think.

Since we already migrated, if changes change effective schema (eg json schema gets changed), version must be increased or we would have no clue if assets have old or new metadata schema. I would even keep increasing at least patch if no effective changes to the actual schema records are expected.

satra:
this doesn't change any keys, just how they are validated or whether they are required. since we didn't run validation this would not directly impact any asset till we run validation. i agree we should get into the habit of updating, but we are at this point where this is only a validation update, rather than a model update. once we hit publish i absolutely think we will need to deal with different schema versions, but at present i was hoping to avoid updating every asset and dandiset metadata for a validation change.

satra:
one of the failing things is related to a tool in the snapshots repo. now that migration is over, do you want to keep it the same way?

dandi/models.py
```diff
@@ -775,7 +789,7 @@ class CommonModel(DandiBaseModel):
 
     # Linking to this dandiset or the larger thing
     access: List[AccessRequirements] = Field(
-        title="Access Type",
+        title="Access Information",
```
yarikoptic, line 792:
```suggestion
        title="Access information",
```
Alternative to above comment on casing "Ethics Approvals" - imho we should not upper case

yarikoptic:
> > if we ourselves do not adhere to at least basic versioning progress upon changes, how we could inspire others to do that?
> 
> because we haven't actually implemented validation. and we will need to update every single metadata json in the server. this is a hack but the only time we can do this hack before validation is implemented.

And how changing schema without version boost helps here? I only can see how it hurts (need for ad-hoc logic to figure out which are older records (and thus supposed to be valid according to previous constraints) and which are new; inability for server to reject metadata in older schema submitted by older versions of clients which still use older version of constraints etc)

yarikoptic:
hm -- I pushed the change to dandi-api-datasets which no longer migrates. But now it does fail to validate (like if it was validating all clean with migration but fails due to this PR changes)

yarikoptic:
I am a little confused why I see more errors reported in the github actions workflow, but I don't see them locally... 

> yes, version isn't being returned right now. let me add that to @dchiquito's PR.

so will we return some adhoc 'draft' as a version for draft dandisets?  IMHO it is better to discriminate published vs non at the level of schema and allow for version to be empty if there is no version. (although I do comprehend that we do have "draft" as a version in API)

> also this dandiset should validate: https://gui.dandiarchive.org/#/dandiset/000029

I will rerun now the "backup" script so possibly to fetch any fresh metadata updates into datalad datasets (that is what the workflow operates on)

satra:
> I will rerun now the "backup" script so possibly to fetch any fresh metadata updates into datalad datasets

it will validate only if id, version are injected. won't before that. so you could temporarily inject in that metadata.

satra:
> I am a little confused why I see more errors reported in the github actions workflow, but I don't see them locally...

are you using the updated models in this PR locally?

yarikoptic:
> > I am a little confused why I see more errors reported in the github actions workflow, but I don't see them locally...
> 
> are you using the updated models in this PR locally?

AFAIK yes. BTW

<details>
<summary>if with master version where we tried to convert schema if version doesn't match (not that we should keep doing that), we got "Could not find about" error</summary> 

```shell
$> DANDI_CACHE=ignore DANDI_DEVEL=1 dandi validate --devel-debug dandisets/000029/dandiset.yaml
2021-05-10 16:47:11,095 [    INFO] Logs saved in /home/yoh/.cache/dandi-cli/log/20210510204710Z-3993951.log
  File "/home/yoh/proj/dandi/dandi-cli-master/venvs/dev3/bin/dandi", line 33, in <module>
    sys.exit(load_entry_point('dandi', 'console_scripts', 'dandi')())
  File "/usr/lib/python3/dist-packages/click/core.py", line 829, in __call__
  File "/usr/lib/python3/dist-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3/dist-packages/click/core.py", line 1259, in invoke
  File "/usr/lib/python3/dist-packages/click/core.py", line 1066, in invoke
  File "/usr/lib/python3/dist-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/cli/base.py", line 81, in wrapper
    return f(*args, **kwargs)
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/cli/cmd_validate.py", line 39, in validate
    for path, errors in validate_(
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/validate.py", line 32, in validate
    errors = validate_file(
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/validate.py", line 40, in validate_file
    return validate_dandiset_yaml(
  File "/home/yoh/proj/fscacher/src/fscacher/cache.py", line 128, in fingerprinter
    ret = fingerprinted(path, *args, **kwargs_)
  File "/home/yoh/proj/fscacher/src/fscacher/cache.py", line 84, in fingerprinted
    return f(path, *args, **kwargs)
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/validate.py", line 70, in validate_dandiset_yaml
    new_meta = migrate2newschema(meta)
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/metadata.py", line 608, in migrate2newschema
    newmeta = convertv1(meta)
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/metadata.py", line 471, in convertv1
    raise KeyError(f"Could not find {oldkey}")
KeyError: 'Could not find about'

```
</details>

With this version (without master merge, so plain 0.16.0-9-g4fc934c) 

<details>
<summary>I get "'dict' object has no attribute 'dict'" </summary> 

```shell
$> DANDI_CACHE=ignore DANDI_DEVEL=1 dandi validate --devel-debug /tmp/000029/dandiset.yaml     
2021-05-10 16:50:38,537 [    INFO] Logs saved in /home/yoh/.cache/dandi-cli/log/20210510205037Z-3994598.log
  File "/home/yoh/proj/dandi/dandi-cli-master/venvs/dev3/bin/dandi", line 33, in <module>
    sys.exit(load_entry_point('dandi', 'console_scripts', 'dandi')())
  File "/usr/lib/python3/dist-packages/click/core.py", line 829, in __call__
  File "/usr/lib/python3/dist-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3/dist-packages/click/core.py", line 1259, in invoke
  File "/usr/lib/python3/dist-packages/click/core.py", line 1066, in invoke
  File "/usr/lib/python3/dist-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/cli/base.py", line 81, in wrapper
    return f(*args, **kwargs)
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/cli/cmd_validate.py", line 39, in validate
    for path, errors in validate_(
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/validate.py", line 32, in validate
    errors = validate_file(
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/validate.py", line 40, in validate_file
    return validate_dandiset_yaml(
  File "/home/yoh/proj/fscacher/src/fscacher/cache.py", line 128, in fingerprinter
    ret = fingerprinted(path, *args, **kwargs_)
  File "/home/yoh/proj/fscacher/src/fscacher/cache.py", line 84, in fingerprinted
    return f(path, *args, **kwargs)
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/validate.py", line 69, in validate_dandiset_yaml
    DandisetMeta(**meta.dict())
AttributeError: 'dict' object has no attribute 'dict'

```
</details>

The fix seems trivial
```
diff --git a/dandi/validate.py b/dandi/validate.py
index de0d2ea..dd207e8 100644
--- a/dandi/validate.py
+++ b/dandi/validate.py
@@ -66,7 +66,7 @@ def validate_dandiset_yaml(filepath, schema_version=None, devel_debug=False):
                 f"Unsupported schema version: {schema_version}; expected {current_version}"
             )
         try:
-            DandisetMeta(**meta.dict())
+            DandisetMeta(**meta)
         except ValidationError as e:
             if devel_debug:
                 raise
```
but the fact that no unit tests failed gives me worries - may be I have some pydantic  (I have 1.8.1) where I don't need that explicit `.dict()`?

<details>
<summary>with that fix I get</summary> 

```shell
$> DANDI_CACHE=ignore dandi validate  /tmp/000029/dandiset.yaml      
2021-05-10 16:55:25,201 [   ERROR] /tmp/000029/dandiset.yaml: 1 error(s)
2021-05-10 16:55:25,201 [   ERROR]   2 validation errors for DandisetMeta
contributor
  At least one contributor must have role ContactPerson (type=value_error)
version
  field required (type=value_error.missing)
Summary: Validation errors in 1 out of 1 files
2021-05-10 16:55:25,201 [    INFO] Logs saved in /home/yoh/.cache/dandi-cli/log/20210510205523Z-3995534.log
```
</details>

NB need to ignore cache since somehow branches jumping without `pip install -e .` I guess confused it all :-/

satra:
now that age is merged, let's put some effort into all the other asset pieces so that we can merge this with a version upgrade.

satra:
ok - i don't know why the schemata publish workflow is failing. this PR does upgrade the schema.

satra:
i reverted this from merging age as well to keep it cleaner.

satra:
@yarikoptic - do you want to merge this? or do you want to upgrade version on this one?

yarikoptic:
yeah, still think that we better beef up the .patch if you expect that no actual metadata "model" changes are due and only validation. api/web-ui should then be able to take 0.3.0 records and handle them with this new 0.3.1 schema

yarikoptic:
Let me try to figure it out/fix

yarikoptic:
well, it is green here now.  version is boosted, so let's proceed.  I probably would delay releasing `dandi-cli` until we test that it would not cause disturbance anywhere and dandi-api / web-ui can handle new 0.3.1?

PR #626, opened 2021-05-10
Error on PRs that modify existing schemata instead of creating a new version

### LINKED ISSUE #624: publish-schemata.yml: fail for PR if schema export changes without version boost
Or could be a dedicated workflow.

The goal: prevent json schema changes without version boost

jwodder:
Closes #624.

### DIFF
--- a/.github/workflows/test-schema.yml
+++ b/.github/workflows/test-schema.yml
@@ -1,0 +1,49 @@
+name: Test model schemata
+
+on:
+  push:
+    paths:
+      - dandi/model*.py
+      - tools/pubschemata.py
+  pull_request:
+    paths:
+      - dandi/model*.py
+      - tools/pubschemata.py
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout this repository
+        uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-schema.yml
```diff
@@ -0,0 +1,49 @@
...
+        with:
+          repository: dandi/schema
+          path: schema
+
+      - name: Generate model schemata
+        run: python dandi/tools/pubschemata.py schema/releases
+
+      - name: Commit changes
+        run: |
+          if git status --porcelain | grep -q "^.M"
```
yarikoptic, line 43: what if new files get  added, removed or renamed - I believe this would not catch it. I think it should boil down to ensuring no changes whatsoever to any of the previously existed folders, so getting a list of folders and running `git status --porcelain` on all of them and then grep'ing for any indicator.

yarikoptic (changes requested):
robustify the check

.github/workflows/test-schema.yml
jwodder: Updated.

yarikoptic:
Great, looks ok to me - thank you!  I will merge, @jwodder  please submit a TEMP PR with changes to schema to verify that it would trigger a fail

PR #628, opened 2021-05-10
Modify metadata schema without a version change

jwodder:
This is a test of #626.


yarikoptic:
and it seems we got this PR resulting in schema change! https://github.com/dandi/schema/commit/a0e4ca85c39c6c2f69a695708717dbd61cc9707b

we must limit to the master branch... reverting that commit as well for now

Issue #101, opened 2020-05-12
validate should ensure dandi requirements are met


satra:
currently validate does not enforce requirements beyond subject id. we should validate for other fields as well.

yarikoptic:
yes. Let's initiate may be `dandi_nwb.json` (to complement `dandiset_nwb.json`). https://github.com/dandi/dandiarchive/tree/master/web/src/assets/schema is the origin to https://github.com/dandi/dandi-core/tree/master/assets/schema ATM, but with git powers I think I will be able to still sync with changes to other files there in dandiarchive which we probably want to keep as origin for a bit longer, while all of those schemas related to web ui are still in flux.

What specific fields you had in mind? if you could express that in json schema - would be great ;-)

satra:
for nwb level, we will likely be using this: https://nih-cfde.org/product/cfde-c2m2/ but we need to map these fields onto nwb fields or functions that can operate on nwb files.

yarikoptic:
Now schema being developed in https://github.com/dandi/schema .  There are schemas for json-validator but also pydantic (original) models which could be used for validation in https://github.com/dandi/schema/blob/master/scripts/models.py . So the plan would be

1. adjust our metadata extraction to extract relevant metadata in the desired structure (this function is the "entry" point to extract metadatahttps://github.com/dandi/dandi-cli/blob/master/dandi/metadata.py#L17:5).  If desired metadata is missing - we need to identify which metadata within nwb is the one to get: This is the function which "interrogates" .nwb file and should also harmonize metadata https://github.com/dandi/dandi-cli/blob/master/dandi/pynwb_utils.py#L170 .

2. use `pydantic` to validate (or would you @satra recommend using produced json schemas?)

For inclusion of schema: lets wait for https://github.com/dandi/schema/issues/9 and proceed accordingly (module, git submodule, or git subtree as in https://github.com/dandi/dandi-core#assets)

References:
-  https://github.com/NeurodataWithoutBorders/nwb-schema

satra:
@yarikoptic - i think all the python code for models/validation should move into dandi-cli (i.e. that will be our dandi python library). the schema repo should be used for making releases and as such it may have a couple of python scripts for processing but not a library.

the closest example at this point is [reproschema](https://github.com/ReproNim/reproschema) - but there we use jsonld and shacl directly instead of pydantic. i think for dandi pydantic makes sense, especially since we may be able to deliver it as the models for the api as well.

[...]

PR #633, opened 2021-05-11
updating parse age to cover more formats

djarecka:


dandi/metadata.py
```diff
@@ -95,9 +95,100 @@ def get_metadata(path):
     return meta
 
 
+def _parse_iso8601(age):
```
satra, line 98: here are the regex patterns we use in reproschema. for at least iso8601, we should cover the formal patterns for duration definition.

https://github.com/ReproNim/reproschema/blob/master/validation/reproschema-shacl.ttl#L327

satra: i'm checking the regex as it seems to fail for something specific, but in general the following age patterns should be supported:

https://en.wikipedia.org/wiki/ISO_8601#Durations

```
<duration>  # This would be pure age
<start>/<duration>  # This is for encoding date of birth/gestation
```

satra: sorry the error was weird, but this should help for the duration pattern:

```python
In [20]: pat = "^P(?!$)(\\d+(?:\\.\\d+)?Y)?(\\d+(?:\\.\\d+)?M)?(\\d+(?:\\.\\d+)?W)?(\\d+(?:\\.\\d
    ...: +)?D)?(T(?=\\d)(\\d+(?:\\.\\d+)?H)?(\\d+(?:\\.\\d+)?M)?(\\d+(?:\\.\\d+)?S)?)?$"

In [21]: re.match(pat, "P149D")
Out[21]: <re.Match object; span=(0, 5), match='P149D'>

In [22]: re.findall(pat, "P149D")
Out[22]: [('', '', '', '149D', '', '', '', '')]

In [26]: re.findall(pat, "P0Y2M1W149DT23H3M6S")
Out[26]: [('0Y', '2M', '1W', '149D', 'T23H3M6S', '23H', '3M', '6S')]
```

djarecka: ok, can move to this, thanks!

djarecka: it doesn't cover "P149.D" - is that ok?

djarecka: it also doesn't change "P149.0D" to "P149D", but that's perhaps is better

satra: it's not valid iso8601 so that should be ok. i think anything that is close but not ok, could be converted and then verified with the regex. i think my above comment was mostly for ensuring that the age format we use corresponds to one of those two iso8601 cases.

1. if input is valid for either case:
   add to age as is
2. try converting and repeat step 1.
3. if all conversions fail, raise an error and ask user to fix.

i also suspect that we need to somehow indicate in nwb the age reference (right now we are using gestat* to determine alternate or assigning birth reference). the ideal will be when we don't need to modify at all, just validate iso8601 + reference.

djarecka: I would disagree that the pattern you posted covers all valid ISO8601 duration values and only valid ones. At least not how I understand the wikipedia description, e.g. "P0,5Y" is a valid value that is not covered, and "P0.6Y2D" is not a valid value that would work with the pattern.

djarecka: It's just not clear to me should I be fixing values like this "P0.6Y2D" or we just keep them

yarikoptic:
what is the plan for this one @satra and @djarecka ?

satra:
we should include this. @djarecka - where does this stand? at least this should merge master now.

djarecka:
I think I was checked this on some of the dandisets, but not all of the dandiset. I could still do it, but general comments are welcome. I included tests that can explain what and  how is converted

dandi/metadata.py
```diff
@@ -90,9 +90,86 @@ def get_metadata(path):
...
+        # checking pattern with "unit" word
+        pattern_unit = rf"(\d+\.?\d*)\s*units?:?\s*({pat_un}s?)"
+        matchstr = re.match(pattern_unit, age, flags=re.I)
+        if matchstr is None:
+            return age, None
+
+    if "." in matchstr.group(1):
+        qty = float(matchstr.group(1))
+        if int(qty) == qty:
+            qty = int(qty)
```
yarikoptic, line 147: this code block of `if` isn't tests covered.  I've not read code to detail yet -- do you @djarecka see a quick test case to  add to tests to catch this one as well, just to be on a safe side? (code looks logical to me)

dandi/metadata.py
```diff
@@ -90,9 +90,86 @@ def get_metadata(path):
...
+    )
+
+    matchstr = re.match(pattern, age, flags=re.I)
+    if matchstr:
+        age_rem = age.replace(matchstr.group(0), "")
+        if age_rem:
+            raise ValueError(
+                f"ISO 8601 expected, but {age} was received,"
+                f"no rules to parse {age_rem}"
+            )
```
yarikoptic, line 109: if you could add a test to cover this exception raise so whenever we change code and rename some variables exception raising doesn't raise some formatting exception?

dandi/metadata.py
```diff
@@ -90,9 +90,86 @@ def get_metadata(path):
...
+    if "." in matchstr.group(1):
+        qty = float(matchstr.group(1))
+        if int(qty) == qty:
+            qty = int(qty)
+    else:
+        qty = int(matchstr.group(1))
+    age_rem = age.replace(matchstr.group(0), "")
+    age_rem = age_rem.strip()
+    if age_rem and age_rem[0] in [",", ";", "."]:
+        age_rem = age_rem[1:]
```
yarikoptic, line 153: is there an easy test case to add to cover this line?

yarikoptic (commented):
a testing zealot me just added a few comments, otherwise, given already a nice tests coverage seems to be LGTM

djarecka:
@yarikoptic - I will improve the test coverage. But you (or others) can also see what "age format" should be convertible, I tried to take my cases from the current dandisets.

satra:
@yarikoptic - i think this also relates to the #101 . in the ideal world we should not have to extract metadata, simply map directly and make them fix at source. but i think this kind of validation is a bigger issue for existing datasets. so we should perhaps identify issues and work with providers to fix. also age can be computed from date of birth.

@djarecka - is there a notion of this could not convert and will raise an exception?

dandi/metadata.py
djarecka: at the end I'm not sure what case I was trying to cover here... I guess all wrong values go to the next exception (that is tested)

dandi/metadata.py
djarecka: added

dandi/metadata.py
djarecka: i think this is already covered somewhere else, so no if needed here

djarecka:
@satra - yes, I believe everything what is not converted will rasie an exception, but there are multiple exceptions that I added, so it won't be always the same, see [here](https://github.com/dandi/dandi-cli/pull/633/files#diff-4c262edb282dc9bba6d8570deae3527ec4b7bdd15eda261a5b3b707192673d1fR90)

btw. I'm still a bit confused about the usage of floats in the standard. What was the document you were reading during one of the meeting, can't find it, and don't see an answer in the summaries

satra:
here is the doc: https://www.loc.gov/standards/datetime/iso-tc154-wg5_n0038_iso_wd_8601-1_2016-02-16.pdf

relevant section is 4.4.3.2

djarecka:
@satra - thanks! this document says that the decimal part can be in the lowest order component (without specifying that it has to be in the Time part), so it is consistent with wikipedia. I've added an extra check, so should be fine

satra:
are you allowing the "comma is preferred" mode in addition to decimals?

djarecka:
@satra - I allow for comma in the iso format (test "P2DT10,5H" added), but I do convert it to the format with dot, so would return "P2DT10.5H".
I do not allow comma in the more natural human format, e.g. "3,5 weeks" is not allowed. Let me know if you want me to change it

satra:
i would say for now that is fine.

PR #634, opened 2021-05-11
move test_get_metaadata to test_metadata.py

bendichter:


yarikoptic:
even linters are happy so let's move forward, thank you @bendichter !

Test populate_dandiset_yaml.py -- will be removed for now in #287 (but may be I will send a dedicated PR if it lingers too long)

PR #637, opened 2021-05-12
Make the "#/" in GUI URLs optional

### LINKED ISSUE #632: support dandiarchive.org urls without fragment (#)
a satellite for https://github.com/dandi/dandiarchive/pull/612 . From the preview https://deploy-preview-612--gui-dandiarchive-org.netlify.app/dandiset/000003/draft/files they look like current ones, just without `#` ;)  but looking forward, I think we should just handle any `(gui\.|)dandiarchive\.org/dandiset/...` (not only gui.dandiarchive.org) to be ready for the ultimate ultimate proxy setup, right @satra and @waxlamp ?

jwodder:
Closes #632.

### DIFF
--- a/dandi/dandiarchive.py
+++ b/dandi/dandiarchive.py
@@ -320,12 +320,12 @@ class _dandi_url_parser:
         ),
         (
             re.compile(
-                rf"{server_grp}#/(?P<asset_type>dandiset)/{dandiset_id_grp}"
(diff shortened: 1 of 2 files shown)


yarikoptic:
Looks good.  I have got stumbled on `(#/)?` at first but since `{server_grp}` should end with `/` we are good. Thank you @jwodder !

PR #640, opened 2021-05-18
Add "shell-completion" command

### LINKED ISSUE #608: Add shell completion support and a helper to "activate"
e.g. in datalad we added
```
  shell-completion
      Display shell script for enabling shell completion for DataLad
```
command so I could just do `source <(datalad shell-completion)` wherever datalad is installed

@jwodder pointed to https://click.palletsprojects.com/en/7.x/bashcomplete/ for click, so might be the one to try/use.

jwodder:
Closes #608.

### DIFF
--- a/dandi/cli/cmd_shell_completion.py
+++ b/dandi/cli/cmd_shell_completion.py
@@ -1,0 +1,37 @@
+import os
+from os.path import basename
+
+import click
+from packaging.version import Version
+
+SHELLS = ["bash", "zsh", "fish"]
+
+
+@click.command("shell-completion")
+@click.option(
+    "-s",
+    "--shell",
+    type=click.Choice(["auto"] + SHELLS),
+    default="auto",
+    help="The shell for which to generate completion code",
+)
+def shell_completion(shell):
+    """ Emit command completion activation code """
+    if shell == "auto":
+        try:
+            shell = basename(os.environ["SHELL"])
+        except KeyError:
+            raise click.UsageError(
+                "Could not determine running shell: SHELL environment variable not set"
+            )
+        if shell not in SHELLS:
+            raise click.UsageError(f"Unsupported/unrecognized shell {shell!r}")
+    if Version(click.__version__) < Version("8.0.0"):
+        varfmt = "source_{shell}"
+    else:
+        varfmt = "{shell}_source"
+    os.environ["_DANDI_COMPLETE"] = varfmt.format(shell=shell)
+
+    from .command import main
+
+    main.main(args=[])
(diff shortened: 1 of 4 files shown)


dandi/cli/cmd_shell_completion.py
```diff
@@ -0,0 +1,37 @@
...
+@click.command("shell-completion")
+@click.option(
+    "-s",
+    "--shell",
+    type=click.Choice(["auto"] + SHELLS),
+    default="auto",
+    help="The shell for which to generate completion code",
+)
+def shell_completion(shell):
+    """ Emit command completion activation code """
```
yarikoptic, line 19: add instructions alike

```
$> datalad shell-completion --help-np   
Usage: datalad shell-completion [-h] [--version]

Display shell script for enabling shell completion for DataLad.

Output of this command should be "sourced" by the bash or zsh to enable
shell completions provided by argcomplete.

Example:

    $ source <(datalad shell-completion)
    $ datalad --<PRESS TAB to display available option>
```

dandi/cli/cmd_shell_completion.py
```diff
@@ -0,0 +1,37 @@
...
+    if shell == "auto":
```
yarikoptic, line 20: please add the click runner smoke test run so at least not everything in this code block looks red according to coverage

yarikoptic (changes requested):
minor desires. Otherwise -- seems to work, nice

yarikoptic:
I guess just skip the tests on windows or make them conditioned on `bash -version` to run cleanly and have GNU in the output

dandi/cli/cmd_shell_completion.py
jwodder: What exactly do you mean?

dandi/cli/cmd_shell_completion.py
jwodder: Done.

dandi/cli/cmd_shell_completion.py
yarikoptic: via `from click.testing import CliRunner` one -- that it would contribute to coverage as is happening for other `cmd_` click interfaces code like  https://sourcegraph.com/github.com/dandi/dandi-cli/-/blob/dandi/cli/cmd_download.py#L96

jwodder: Due to how Click implements command completion, running the command that way wouldn't work unless the command messes with `sys.argv`, which may not be such a good idea.

yarikoptic: hm, ok, will trust you on that and let's just proceed

PR #643, opened 2021-05-20
Move schema code to dandischema

### LINKED ISSUE #638: Move metadata functionality to a dedicated dandischema
A dedicated follow up to #631.

After #596 is merged we should split away all schema components into an independent python package. Rationale:

- dandi-cli depends on heavy h5py/hdmf/libhdf5 - complicates considering it for use by dandi-api which wouldn't need those
- we do not have a clean path toward staging metadata schema for a release -- we always release from master as soon as it gets a new version

So the decision was made to split away all the metadata schema (pydantic models, validation of them) into `dandischema`.  All metadata extraction/harmonization from files (nwb and later bids etc) will remain in dandi-cli, but we would validate using functionality moved to `dandischema`.

Attn @satra , @jwodder (not yet assigning since we probably want to have #596 merged first or WDYT?)

jwodder:
Closes #638.

This PR is pending a release of dandischema (now done).


yarikoptic (approved):
What a meticulous amputation, nicely done @jwodder  ;-) Haven't spotted any gotcha

yarikoptic:
Let's proceed and hope for the bright future!

PR #644, opened 2021-05-21
ENH: more metadata to reconstruct filename, upgrade to use dandischema 0.2.3 (schema 0.4.0)

### LINKED ISSUE #642: asset updates: where things should go in metadata
"subject_id" -> in current wasAttributedTo
"session_id"  -> in current wasGeneratedBy
"tissue_sample_id"  -> in current wasDerivedFrom
"slice_id"  -> in current wasDerivedFrom
"cell_id" -> in current wasDerivedFrom
"probe_ids" -> ???
"obj_id" -> add to "wasDerivedFrom" (this could be derived from an NWB file with identifier obj id?)
"modalities" -> see below
"extension" -  in current contentType

NB edited by @yarikoptic for uniformity and to reflect discussion below

we do need to fill both approach and measurementTechnique.  @bendichter started this: https://docs.google.com/spreadsheets/d/1x6aI5e7uWwyUJf_S1RyYkmOW-OzodLcRserZVq4DL2A/edit#gid=0

we will need to curate a list of named approaches and techniques from tom's list (i would use dandi readable prefixes that map via sameas to tom's list, unless tom's already has readable prefixes.

anything missing?

cc: @yarikoptic @bendichter @tgbugs

we need to finalize this and rerun modification of asset metadata before we can generate assetsSummary. so this is a bit urgent to get going.

### LINKED ISSUE #658: upgrade to dandischema 0.2.1
spotted after adding dandischema to depends of dandi conda feedstock
```
...
2021-06-07T19:42:11.7740941Z     from .metadata import get_metadata
2021-06-07T19:42:11.7742066Z ../_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_/lib/python3.8/site-packages/dandi/metadata.py:307: in <module>
2021-06-07T19:42:11.7743071Z     ) -> models.BareAssetMeta:
2021-06-07T19:42:11.7743869Z E   AttributeError: module 'dandischema.models' has no attribute 'BareAssetMeta'
```

and indeed we are still at 0.1.0:
```
setup.cfg:    dandischema ~= 0.1.0
```
Let's progress forward to `~=0.2.1` and release

satra:
here is a first pass to doing this for NWB files.

closes #642 
closes #658


dandi/metadata.py
```diff
@@ -270,9 +270,25 @@ def extract_wasDerivedFrom(metadata):
...
+
+def extract_session(metadata):
+    probe_ids = metadata.get("probe_ids", [])
+    if isinstance(probe_ids, str):
+        probe_ids = [probe_ids]
+    probes = []
+    for val in probe_ids:
+        probes.append(models.Agent(identifier=f"probe:{val}", name="Ecephys Probe"))
+    if len(probes) == 0:
+        probes = None
```
yarikoptic, line 282:
```suggestion
    probes = probes or None
```
or 
```suggestion
    if not probes:
        probes = None
```
imho would be more pythonic

dandi/metadata.py
```diff
@@ -270,9 +270,25 @@ def extract_wasDerivedFrom(metadata):
...
+
+    return [
+        models.Session(
+            identifier=metadata.get("session_id"),
+            name=metadata.get("session_id"),
+            description=metadata.get("session_description", None),
```
yarikoptic, line 288:
```suggestion
            description=metadata.get("session_description"),
```
is as good AFAIK ;-)

dandi/metadata.py
```diff
@@ -270,9 +270,25 @@ def extract_wasDerivedFrom(metadata):
     models.Participant, "identifier", "subject_id", id=...
 )
 
-extract_wasGeneratedBy = extract_model_list(
-    models.Session, "name", "session_id", id=...
-)
+
+def extract_session(metadata):
```
yarikoptic, line 274: is that a
```suggestion
def extract_session(metadata: dict) -> list:
```
?
I should myself get into a habit to use type annotations

dandi/metadata.py
```diff
@@ -302,6 +318,216 @@ def extract_field(field, metadata):
         return metadata.get(field, ...)
 
 
+nd_list = {
```
yarikoptic, line 321: is this to be the canonical source of this listing? 
I think we should keep that google spreadsheet by @bendichter for now and auto-convert it to .tsv or .json and carry as a file to be loaded here.  Dumping all these data structures which can change here makes harder to navigate the code IMHO

it seems the key here is identical to `neurodata_type`, so should not be "hardcoded" but mapped at load time to avoid duplication (the evil of large % of bugs)

yarikoptic, line 321: also `nd_list` is a bad name for a dictionary ;)  so the file might get named `neurodata_types.json` or alike, and here `neurodata_types`

yarikoptic (commented):
Some comments for now... main aspect -- we shouldn't hardcode data structures in `.py` code.

dandi/metadata.py
```diff
@@ -270,9 +270,25 @@ def extract_wasDerivedFrom(metadata):
...
 )
 
-extract_wasGeneratedBy = extract_model_list(
-    models.Session, "name", "session_id", id=...
-)
+
+def extract_session(metadata):
+    probe_ids = metadata.get("probe_ids", [])
+    if isinstance(probe_ids, str):
+        probe_ids = [probe_ids]
```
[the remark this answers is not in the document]
satra: i think i have such a function in every library, well not quite. some libraries only require it once :) i leave it to you here whether to include in dandi or not. in this case probe_ids are either just a string or a list as returned by the metadata extractor.

dandi/metadata.py
satra: i don't know where this will live eventually. i also need to add identifiers to this and so it may move to dandischema (at least some pieces of it).

satra:
@yarikoptic - the reason the test is failing is that the basic setup doesn't compute dandi-etag for an nwb file. i think we should compute that instead of sha1. perhaps that's an easy fix.

satra:
@yarikoptic - we need a place to add the nwb object identifier.

yarikoptic:
I wonder if a digest would be too wrong of a place. nwb-object-id . Fast to compute ;-)

satra:
you can put it in digest. it would only be available for nwb files, but a known place and is a digest/checksum. we would need to add digests to the schema and release though because the validator checks for keys. that doesn't stop us, but just saying.

yarikoptic:
note: for some reason only the LGTM checks are reported in the last status update.  I think we need some commit to trigger full sweep and see where are we standing... will do

satra:
shall we at least get this into master?

satra:
just changed the pin here to 0.2.3 - this is the next thing to merge and, potentially, release.

satra:
can we merge this now and then figure out what else is needed to release?

yarikoptic:
skimmed through, haven't found anything to whine about. 
Added it closing #658  and a `minor` label.
So we are postponing dealing with the NWB object-id for the purpose of this PR. I adjusted the title to be a bit more descriptive of changes.  Filed a dedicated https://github.com/dandi/dandi-cli/issues/659 as specifically for `cmd-organize`, where mentioned object-id aspect.

yarikoptic:
not releasing right away so I guess we could try it out etc.

PR #647, opened 2021-05-24
Warn on ignored symlinks to directories

### LINKED ISSUE #641: upload using `.` to indicate present directory did not work
in dandiset 105 to move the data, the following command was used for trying upload

`DANDI_DEVEL=1 dandi upload --allow-any-path --validation skip .`

the above did not send any files but the following command did do the appropriate thing.

`DANDI_DEVEL=1 dandi upload --allow-any-path --validation skip sub-mgh191021520`

jwodder:
Closes #641.


yarikoptic:
I think this should suffice for now. thank you @jwodder !

PR #648, opened 2021-05-25
Include CHANGELOG.md and tox.ini in sdists

jwodder:


yarikoptic:
there is no rush on this to get released, right?

jwodder:
@yarikoptic No, no rush.

PR #649, opened 2021-05-26
Set `auto` author to "DANDI Bot"

jwodder:
As instructed [here](https://github.com/dandi/dandischema/pull/17#discussion_r640013204).


yarikoptic:
CRAP -- should have merged this before #650 heh heh. there is always the next release

PR #650, opened 2021-05-26
Pin dandischema to compatible 0.minor version

yarikoptic:
since we are not at luxury to "redeploy" dandi-cli rapidly (installed by
users), I see no other way than this.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,7 +30,7 @@ install_requires =
     appdirs
     click
     click-didyoumean
-    dandischema
+    dandischema ~= 0.1.0
     etelemetry >= 0.2.0
     fasteners
     fscacher


setup.cfg
```diff
@@ -30,7 +30,7 @@ install_requires =
     appdirs
     click
     click-didyoumean
-    dandischema
+    dandischema ~= 0.1.0
```
jwodder, line 33: This is equivalent to `>= 0.1.0, <0.2.0`.  Is that what you meant, or did you mean `~= 0.1` (which is `>= 0.1, <1.0`)?

yarikoptic: 0.2.0 broke it.  so I guess we will need to resort for now for this level.

yarikoptic: but I filed https://github.com/dandi/dandischema/issues/22 which could be assisted with by https://github.com/dandi/dandischema/issues/21 ;-)

yarikoptic:
Green! lets fix the world

PR #654, opened 2021-06-02
dandi ls: Make json and json_pp formats output arrays; add json_lines for old json format

### LINKED ISSUE #652: dandi ls -f json_pp doesn't produce a valid json
we wrapping into a list. ATM `-f json` produces legit "json lines" having each record on a separate line.  But I think we better make it into a  explicit `json_lines` or alike.

`json_pp` takes those lines and renders them across multiple lines thus making overall illegit.  For both `json` and `json_pp` let's place all the records into a list and output that one

note: -f yaml seems to do the right thing and makes a list

jwodder:
Closes #652.

### DIFF
--- a/dandi/cli/formatter.py
+++ b/dandi/cli/formatter.py
@@ -22,21 +22,57 @@ class JSONFormatter(Formatter):
     def __init__(self, indent=None, out=None):
         self.out = out or sys.stdout
         self.indent = indent
+        self.first = True
 
     @staticmethod
     def _serializer(o):
         if isinstance(o, datetime.datetime):
-            return o.__str__()
+            return str(o)
+        return o
+
+    def __enter__(self):
+        print("[", end="", file=self.out)
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        if not self.first:
+            print(file=self.out)
+        print("]", file=self.out)
+
+    def __call__(self, rec):
+        import json
+        from textwrap import indent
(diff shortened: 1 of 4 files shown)


dandi/cli/formatter.py
```diff
@@ -22,21 +22,57 @@ class JSONFormatter(Formatter):
...
+
+    def __call__(self, rec):
+        import json
+        from textwrap import indent
+
+        if self.first:
+            print(file=self.out)
+            self.first = False
+        else:
+            print(",", file=self.out)
```
yarikoptic, line 49: would you be so kind to add a proper unittest which would operate over some very simple datastructures but with at least 2 items so we could ensure that we are formatting multi-item output correctly?

yarikoptic (changes requested):
Looks good, but I am asking for some new extra test to finally ensure that we are doing it correctly now

dandi/cli/formatter.py
jwodder: Just for `JSONFormatter` or also for `JSONLinesFormatter`?

yarikoptic: thanks for the tests, please add one for `JSONLinesFormatter` as well while at it -- would only help

jwodder: Done.

yarikoptic:

PR #655, opened 2021-06-04
Reprompt or fail on attempt to use an invalid API token

### LINKED ISSUE #653: improve API key check
could the cli try to use the API-key to `GET` the `/auth/token` and if it returns the same value the key is valid, if the GET fails due to an authorization error, then the key is invalid. in the latter case,  ask the user to refresh their key and then provide it to the cli.

valid
```shell
$ curl --request GET --header "Authorization: token ${DANDI_API_KEY}" --url \
https://api.dandiarchive.org/api/auth/token/

<DANDI_API_KEY>
```

invalid

```shell
$ curl --request GET --header "Authorization: token ${DANDI_API_KEY}bad" --url \
 https://api.dandiarchive.org/api/auth/token/

{"detail":"Invalid token."}
```

jwodder:
Closes #653.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -260,6 +260,12 @@ class DandiAPIClient(RESTFullAPIClient):
 
     def authenticate(self, token):
         self._headers["Authorization"] = f"token {token}"
+        try:
+            # Fails if token is invalid:
+            self.get("/auth/token")
+        except requests.HTTPError:
+            del self._headers["Authorization"]
+            raise
 
     def dandi_authenticate(self):
         # Shortcut for advanced folks
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -276,24 +276,26 @@ class DandiAPIClient(RESTFullAPIClient):
             raise NotImplementedError("TODO client name derivation for keyring")
         app_id = f"dandi-api-{client_name}"
         keyring_backend, api_key = keyring_lookup(app_id, "key")
-        if api_key:
-            self.authenticate(api_key)
-        else:
-            while True:
+        key_from_keyring = api_key is not None
+        while True:
+            if not api_key:
                 api_key = input(f"Please provide API Key for {client_name}: ")
-                try:
-                    self.authenticate(api_key)
-                except requests.HTTPError:
-                    if is_interactive() and click.confirm(
-                        "API key is invalid; enter another?"
-                    ):
-                        continue
-                    else:
-                        raise
+                key_from_keyring = False
+            try:
+                self.authenticate(api_key)
+            except requests.HTTPError:
+                if is_interactive() and click.confirm(
+                    "API key is invalid; enter another?"
+                ):
+                    api_key = None
+                    continue
                 else:
+                    raise
+            else:
+                if not key_from_keyring:
                     keyring_backend.set_password(app_id, "key", api_key)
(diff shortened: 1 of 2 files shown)


dandi/dandiapi.py
```diff
@@ -260,6 +260,12 @@ def __init__(self, api_url, token=None):
 
     def authenticate(self, token):
         self._headers["Authorization"] = f"token {token}"
+        try:
+            # Fails if token is invalid:
+            self.get("/auth/token")
+        except requests.HTTPError:
+            del self._headers["Authorization"]
+            raise
```
yarikoptic, line 268: would it possible to instrument some test for this to ensure that we do prompt user to enter a new token?

jwodder: We don't prompt the user for a new token.  Failure has the same result as it did for Girder: an exception is raised, and the program exits.

yarikoptic: well -- we should at least here in `dandi_authenticate`: https://github.com/dandi/dandi-cli/blob/HEAD/dandi/dandiapi.py#L276 where we first ask for the key if not known and then proceed to self.authenticate.  We should make it into a loop which would try to authenticate, and if fails - tell user that the key is not good, would they like to enter a new one -- if say 'y' -- go back into the loop prompting for `Please provide API Key ...` and if 'n' -- reraise

yarikoptic: clarification: loop/ask only if `dandi/utils.py:def is_interactive():`

dandi/dandiapi.py
```diff
@@ -273,11 +276,24 @@ def dandi_authenticate(self):
             raise NotImplementedError("TODO client name derivation for keyring")
         app_id = f"dandi-api-{client_name}"
         keyring_backend, api_key = keyring_lookup(app_id, "key")
-        if not api_key:
-            api_key = input(f"Please provide API Key for {client_name}: ")
-            keyring_backend.set_password(app_id, "key", api_key)
-            lgr.debug("Stored key in keyring")
-        self.authenticate(api_key)
+        if api_key:
+            self.authenticate(api_key)
```
yarikoptic, line 280: not quite... since if there is a known one, we would just try to auth and fail without going through "the loop".

just remove this `if api_key` here, make loop unconditional, ask to provide a key right there only `if not api_key`, set `api_key = None` if auth fails before `continue`.

dandi/dandiapi.py
```diff
@@ -273,11 +276,26 @@ def dandi_authenticate(self):
...
+            except requests.HTTPError:
+                if is_interactive() and click.confirm(
+                    "API key is invalid; enter another?"
+                ):
+                    api_key = None
+                    continue
+                else:
+                    raise
+            else:
+                if not key_from_keyring:
```
yarikoptic, line 295: why do you think it is worth to not store a new key if we got it from a keyring but it is no longer valid?  I would just always store it

jwodder: If a keyring key is invalid, `key_from_keyring` will be set to `False` after the user enters a new key, so the new key will be stored in the keyring.

yarikoptic: oh right -- so we store only if a new key was entered, got it, thanks

yarikoptic:
Let's proceed and I also marked it for `release` so we hopefully make some users' lives easier. Thank you @jwodder !

PR #656, opened 2021-06-07
Update pre-commit repo versions and configure isort to properly handle "from . import" lines

jwodder:


yarikoptic:
Looks good to me.  Even LGTM seems to be not non-useful ;-)  Thanks @jwodder, let's proceed

PR #657, opened 2021-06-07
Address LGTM alerts

jwodder:
Some alerts are false positives and thus are not addressed by this PR.


dandi/_version.py
```diff
@@ -504,7 +504,7 @@ def get_versions():
         # versionfile_source is the relative path from the top of the source
         # tree (where the .git directory might live) to this file. Invert
         # this to find the root from __file__.
-        for i in cfg.versionfile_source.split('/'):
+        for _ in cfg.versionfile_source.split('/'):
```
yarikoptic, line 507: this is versioneer one, can we just skip it for LGTM?

meanwhile checked -- current master in versioneer already has this:
```
$> git grep -A1 'in cfg.versionfile_source.split'
src/git/long_get_versions.py:        for _ in cfg.versionfile_source.split('/'):
src/git/long_get_versions.py-            root = os.path.dirname(root)
```
so in principle we could just proceed, but still I think we should just skip this file

tools/validate-api-against-girder.py
```diff
@@ -1,16 +1,12 @@
 #!/usr/bin/env python3
 import click
-import requests
 
 from dandi.dandiapi import DandiAPIClient
-from dandi.dandiset import APIDandiset
 from dandi.girder import GirderCli
```
yarikoptic, line 5: let's kill this file altogether -- girder is gone gone

dandi/_version.py
jwodder: I don't know how ignoring alerts works in LGTM, so I can't answer that.

tools/validate-api-against-girder.py
jwodder: Should `tools/instantiate-dandisets.py` (which also uses Girder) be deleted as well?

yarikoptic: sure, why to drag it now

dandi/_version.py
yarikoptic: google has all the answers: https://lgtm.com/help/lgtm/customizing-file-classification

dandi/due.py
```diff
@@ -55,7 +55,7 @@ def _donothing_func(*args, **kwargs):
 
 
 try:
-    from duecredit import due, BibTeX, Doi, Url, Text
+    from duecredit import due, Doi  # lgtm [py/unused-import]
```
yarikoptic, line 58: Let's skip this file as well from lgtm checks - it is a stub which comes from duecredit and better be left unmodified

jwodder: Do you want me to revert the changes to our local `due.py`?

yarikoptic: yes, to due.py and `_version.py` and list them to be excluded in to be added lgtm.yml per https://lgtm.com/help/lgtm/customizing-file-classification

yarikoptic: oh I see you added already lgtm.yml -- great

jwodder: Done.

yarikoptic:
Thank you @jwodder !  Let's proceed.

I feel we might like to pacify this LGTM bot to not add comments -- adding to the checks list is IMHO sufficient.

PR #660, opened 2021-06-09
Python API rewrite, part 1

### LINKED ISSUE #673: Default to the dandi instance API url
ATM (as as of #660 0.19.0-17-g7d4541f) DandiAPIClient requires API URL to be passed into constructor.  Since the majority (if not 99%) of the use cases would be against our main instance, I think we should make not necessary to explicitly specify API server endpoint.  

But then may be we should still allow to specify the "default instance" for a run.  grep'ing through the codebase I see
```shell
(git)lena:~/proj/dandi/dandi-cli-master[api-01]git
$> git grep DANDI.*INST
DEVELOPMENT.md:- `DANDI_INSTANCEHOST` -- defaults to `localhost`. Point to host/IP which hosts
dandi/consts.py:instancehost = os.environ.get("DANDI_INSTANCEHOST", "localhost")
doc/demos/basic-workflow1.sh:: "${DANDI_INSTANCE:=local-docker}"
doc/demos/basic-workflow1.sh:dandi register -i "$DANDI_INSTANCE" -n "$(basename $SRCPATH)" -D "description"
doc/demos/basic-workflow1.sh:dandi upload -i "$DANDI_INSTANCE"

$> git grep instancehost
dandi/consts.py:instancehost = os.environ.get("DANDI_INSTANCEHOST", "localhost")
dandi/consts.py:        None, None, f"http://{instancehost}:8000/api"
```
but from a brief git log archaeological expedition I don't think we ever supported `DANDI_INSTANCE` env variable to point to the default instance to use.  I think it might be worth bringing it in, and defaulting to `"dandi"` instance, and using its "api" end point.

jwodder:
Closes #673 (default to `dandi` instance)

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -253,6 +253,12 @@ class DandiAPIClient(RESTFullAPIClient):
     def get_dandiset(
         self, dandiset_id: str, version_id: Optional[str] = None
     ) -> "RemoteDandiset":
+        """
+        Fetches the Dandiset with the given ``dandiset_id``.  If ``version_id``
+        is not specified, the `RemoteDandiset`'s version is set to the most
+        recent published version if there is one, otherwise to the draft
+        version.
+        """
         d = RemoteDandiset._make(self, self.get(f"/dandisets/{dandiset_id}/"))
         if version_id is not None and version_id != d.version_id:
             if version_id == "draft":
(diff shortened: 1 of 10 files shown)


yarikoptic:
magnificent.  I am still reading it ;) would you be so kind to merge master or rebase so we also get fresh dandischema for most "coherent" test-driving?

jwodder:
@yarikoptic Rebased.

yarikoptic:
looking at the diff, from a CLI user perspective, no changes should affect (negatively) the experience, right? API is getting changed, but that is intended... my point is -- should be ok to just merge to master, and proceed even if we are to cut a new dandi-cli api release to fix up a bug even if "part 2" is coming, right?

dandi/dandiapi.py
```diff
@@ -98,9 +70,9 @@ def send_request(
...
             Note that the path string should not begin or end with the path  separator, '/'.
         :type path: str
-        :param parameters: A dictionary mapping strings to strings, to be used
+        :param params: A dictionary mapping strings to strings, to be used
             as the key/value pairs in the request parameters.
-        :type parameters: dict
+        :type params: dict
         :param data: A dictionary, bytes or file-like object to send in the body.
         :param files: A dictionary of 'name' => file-like-objects for multipart encoding upload.
         :type files: dict
```
yarikoptic, line 78: I guess we could get rid of those `:type` doc descriptions after it gets properly typed, right?

Also in other spots, I have used numpy docstrings format, not pydoc (or whatever that original is).  we better stay consistent one way or another.

jwodder:
@yarikoptic I believe so, yes.

yarikoptic (approved):
Great. Tested from within CLI a bit -- still worked nice, so no side-effects detected.

PR #663, opened 2021-06-10
Create validation.md

satra:
@yarikoptic and @bendichter - i'd like to use this to figure out what we use to validate NWB files for DANDI. consider this the list of things we can enforce. we may have to grandfather in existing dandisets, but moving forward i want to be stricter about collecting information at the user side.


satra:
@bendichter - also for NWB, perhaps you can list the location of such info in NWB.

doc/design/validation.md
```diff
@@ -0,0 +1,25 @@
...
+- subject id
+- age in iso8601 interval format or datetime of birth
+- age reference 
+- sex
+
+### NWB organization disambiguation requirements
+
+At least one of the following should be present. in case of having to disambiguate
+- session
+- sample
```
yarikoptic, line 16: I am not sure if there is a generic notion of a "sample" in NWB, so I guess it is for the `tissue_sample_id` [*].

[*] forgot that it is coming probably from `ndx-dandi-icephys` and wasted time to search it up in the main nwb schema, uff

yarikoptic: and what about `slice_id`?  might have both I guess (multiple slices from the same sample).

I guess in our metadata schema they both (tissue and slice) would be `BioSample` but of different  `sampleType`, right?

yarikoptic (commented):
initial comments/questions

doc/design/validation.md
```diff
@@ -0,0 +1,25 @@
+# Requirements for being a valid DANDI asset
+
+## General
+1. All assets must have a non-zero file size.
+
+## NWB files
+- subject id
```
bendichter, line 7:
```suggestion
- subject id (nwb: NWBFile.subject.subject_id)
```

doc/design/validation.md
```diff
@@ -0,0 +1,25 @@
...
+- age in iso8601 interval format or datetime of birth
```
bendichter, line 8:
```suggestion
- age in iso8601 interval format or datetime of birth (nwb.subject.age or nwb.subject.date_of_birth)
```

doc/design/validation.md
```diff
@@ -0,0 +1,25 @@
...
+- age reference 
+- sex
+
+### NWB organization disambiguation requirements
+
+At least one of the following should be present. in case of having to disambiguate
+- session
+- sample
+- cell
```
bendichter, line 17: we don't have a notion of cell in core either

doc/design/validation.md
```diff
@@ -0,0 +1,25 @@
...
+- probe
```
bendichter, line 18: what do you mean by this? Probe type, or specific probe serial number?

satra: probe identifier - i don't think we found a place to get probe type information but if that exists, let us know.

doc/design/validation.md
satra: correct.

yarikoptic: was that is the answer to the last question, or to both? ;-)

satra: the last question (biosample of different sampletype) as implemented currently in the extractor.

doc/design/validation.md
yarikoptic: I think @bendichter 's idea to add explicit nwb schema paths is a nice one.  But I see only that @satra got them resolved without comments/changes.   Was that intentional?

doc/design/validation.md
yarikoptic: is it in `ndx-dandi-icephys` or some other extension?

doc/design/validation.md
satra: no i think i exposed a weird github bug. i added them to batch, but something refreshed and i hit refresh and my batch disappeared.

satra:
```suggestion
- subject id (nwb.subject.subject_id)
```

satra: closing this - they have been merged into a commit. so not a bug.

yarikoptic:
moved back to draft since never finalized. I think it was waiting your responses @satra and @bendichter but may be we should reconsider it in full again since there were considerable number of changes to validation since then.

yarikoptic:
it was never followed up on, and I am not sure how much of it is actually reflective of current/desired state, and either it should be here at all, or be a part of some user facing documentation (e.g. handbook) where we outline what we support already. So I will close for now, but if you feel different and plan to finish it up @satra -- feel welcome to reopen.

PR #664, opened 2021-06-10
fix: pin to range

satra:
lower bound should move only as we finalize the lowest schema version supported. for now allow for ranges of schema

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,7 +30,7 @@ install_requires =
     appdirs
     click
     click-didyoumean
-    dandischema ~= 0.2.3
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -30,7 +30,7 @@ install_requires =
     appdirs
     click
     click-didyoumean
-    dandischema ~= 0.2.3
+    dandischema >= 0.2.3,<0.3.0
```
jwodder, line 33: These mean the same thing.

satra: thanks @jwodder

PR #669, opened 2021-06-13
fix: raise error when unable to parse age and set session name properly

### LINKED ISSUE #668: age parsing silently ignores value
example file: https://gui.dandiarchive.org/#/dandiset/000004/draft/files?location=sub-P10HMH%2F

`'age': '37',`

but after metadata extraction there is no age value for participant. this patch should at least raise an error if not being interpreted.

```python
diff --git a/dandi/metadata.py b/dandi/metadata.py
index 831ca54..b791a9a 100644
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -250,6 +250,8 @@ def extract_age(metadata):
         try:
             duration = parse_age(metadata["age"])
         except (KeyError, TypeError, ValueError):
+            if metadata.get("age"):
+                raise
             return ...
     else:
         duration = timedelta2duration(start - dob)
```

satra:
closes #668 

also addresses validation of session errors because dictionary gets only return provided default when key is missing in original dictionary.


dandi/metadata.py
```diff
@@ -408,13 +410,16 @@ def extract_session(metadata: dict) -> list:
     for val in probe_ids:
         probes.append(models.Equipment(identifier=f"probe:{val}", name="Ecephys Probe"))
     probes = probes or None
-    if (metadata.get("session_id") or probes) is None:
+    if (
+        metadata.get("session_id") or metadata.get("session_start_time") or probes
+    ) is None:
```
yarikoptic, line 415: Afaik session id 0 is as legit as any other. I think such shortcut would get us into trouble eventually .

dandi/metadata.py
```diff
@@ -250,6 +250,8 @@ def extract_age(metadata):
         try:
             duration = parse_age(metadata["age"])
         except (KeyError, TypeError, ValueError):
+            if metadata.get("age"):
```
yarikoptic, line 253: Yet another implicit one.

dandi/metadata.py
satra: this shouldn't evaluate against `None` it's a very specific check to see if the boolean combination is None.

dandi/tests/data/metadata/metadata2asset.json
```diff
@@ -37,55 +39,54 @@
...
+      ]
     }
   ],
   "wasAttributedTo": [
     {
+      "schemaKey": "Participant",
       "identifier": "a1b2c3",
       "age": {
+        "schemaKey": "PropertyValue",
         "unitText": "Years from birth",
```
satra, line 76: @djarecka - this `unitText` says `Years from birth` when the value is in days.

dandi/metadata.py
satra: this one i'll change

dandi/tests/data/metadata/metadata2asset.json
djarecka: i'm not sure how this file was created. I've changed this file in one of my PR: https://github.com/dandi/dandi-cli/pull/450/files

dandi/metadata.py
yarikoptic: there is no explicit test for `None` for anything here but `probes` here pretty much. So if session_id is set and == 0, we would get `None` if probes is None.
```
$> python3 -c 'ses = 0; probes = None; print((ses or probes) is None)'
True
```

dandi/metadata.py
```diff
@@ -408,13 +410,16 @@ def extract_session(metadata: dict) -> list:
...
-    if (metadata.get("session_id") or probes) is None:
+    if (
+        metadata.get("session_id") or metadata.get("session_start_time") or probes
+    ) is None:
         return None
     return [
         models.Session(
             identifier=metadata.get("session_id"),
-            name=metadata.get("session_id", "Acquisition session"),
+            name=metadata.get("session_id") or "Acquisition session",
```
yarikoptic, line 420: again -- no session_id 0 is allowed ?

dandi/tests/data/metadata/metadata2asset.json
satra: this is created in the test itself: https://github.com/dandi/dandi-cli/blob/193723c96cba70a65d5a1f7481564534348e1ae1/dandi/tests/test_metadata.py#L143

so this is not a real file (and age does not match session - date of birth). but the conversion of session - dob, should generate the right units as well, which doesn't look like the case.

djarecka: the file is not created in the test, but it's used in the test, but I guess that's what you meant. 
In #450 I've changed [age_parse](https://github.com/dandi/dandi-cli/pull/450/files#diff-efe33fb24af3f1d5572d20580fa315cc3b2a8d6a69992c260fd900f643f2e924R132) so it does use differ `unitText` and I've changed the file used in the test to reflect the changes. For some reason  this is gone from the master...:( 
I will try to figure out what went wrong, perhaps I have to create the same PR again...

dandi/metadata.py
```diff
@@ -250,6 +250,8 @@ def extract_age(metadata):
         try:
             duration = parse_age(metadata["age"])
         except (KeyError, TypeError, ValueError):
+            if "age" in metadata and metadata["age"] is not None:
+                raise
             return ...
```
yarikoptic, line 255: I think such code would be cleaner/easier to digest, or am I missing some other intention here?
```suggestion
        if metadata.get("age") is not None:
            duration = parse_age(metadata["age"])
        else:
            return ...
```

dandi/metadata.py
```diff
@@ -408,13 +410,17 @@ def extract_session(metadata: dict) -> list:
     for val in probe_ids:
         probes.append(models.Equipment(identifier=f"probe:{val}", name="Ecephys Probe"))
     probes = probes or None
-    if (metadata.get("session_id") or probes) is None:
+    session_id = None
+    if "session_id" in metadata and metadata["session_id"] is not None:
```
yarikoptic, line 414:
```suggestion
    if metadata.get("session_id") is not None:
```
?

dandi/tests/data/metadata/metadata2asset.json
satra: did you send a PR, or can you cherrypick your changes? i would like to merge this in tomorrow if possible.

yarikoptic: If I see it right -- are we waiting on this discussion to be resolved to have this merged? meanwhile adding `patch` label

djarecka: @satra - you commented in #450 that you created PR, but I've just realized that your question in this thread is newer. do you want me to open #450 again?

djarecka: never mind, my mistake, I've misread your comment, will create pr

dandi/metadata.py
```diff
@@ -361,7 +367,10 @@ def extract_species(metadata):
             value_id = value
             value = None
         else:
-            raise ValueError(f"Cannot interpret species field: {value}")
+            raise ValueError(
+                f"Cannot interpret species field: {value}. Please "
+                "contact help@dandiarchive.org to add your species"
```
yarikoptic, line 372: note: I see emails from Roni testing that this address works, so should be all good.  Adding a trailing `.` to trigger tests rerun etc:
```suggestion
                "contact help@dandiarchive.org to add your species."
```

dandi/metadata.py
```diff
@@ -435,13 +444,17 @@ def extract_session(metadata: dict) -> list:
     for val in probe_ids:
         probes.append(models.Equipment(identifier=f"probe:{val}", name="Ecephys Probe"))
     probes = probes or None
-    if (metadata.get("session_id") or probes) is None:
+    session_id = None
+    if metadata.get("session_id") is not None:
+        session_id = str(metadata["session_id"])
+    if (session_id or metadata.get("session_start_time") or probes) is None:
         return None
```
yarikoptic, line 451: filed a dedicated https://github.com/dandi/dandi-cli/issues/698

dandi/metadata.py
```diff
@@ -347,7 +347,13 @@ def extract_species(metadata):
...
         elif "norvegicus" in value:
             value_id = "http://purl.obolibrary.org/obo/NCBITaxon_10116"
-            value = "Brown rat"
+            value = "Rat; Norway rat; rats; Brown rat"
+        elif "rattus rattus" in value:
+            value_id = "http://purl.obolibrary.org/obo/NCBITaxon_10117"
+            value = "Black rat; Roof rat; House rat"
+        elif "rat" in value:
+            value_id = "http://purl.obolibrary.org/obo/NCBITaxon_10116"
+            value = "Rat; Norway rat; rats; Brown rat"
```
yarikoptic, line 356: I guess check for `norvegicus` could have moved down here and have a single block which would be for both `norvegicus` and `rat` as the "last resort".

yarikoptic:
there is no end to perfection and I guess we will get to this code later anyways, so since @satra wanted a release with this one, let's just do it! ;)

PR #676, opened 2021-06-21
Python API rewrite, part 2

jwodder:


dandi/dandiapi.py
```diff
@@ -699,13 +802,45 @@ def api_path(self) -> str:
...
+            raise NotFoundError(f"No {digest_type} digest found in metadata")
+
+    def set_metadata(self, metadata: models.Asset) -> None:
+        """Set the metadata for the asset to the given value"""
+        self.set_raw_metadata(metadata.json_dict())
+
+    def set_raw_metadata(self, metadata: Dict[str, Any]) -> None:
+        """Set the metadata for the asset to the given value"""
+        self.client.put(
+            self.api_path, json={"metadata": metadata, "blob_id": self.identifier}
```
yarikoptic, line 840: hm, why asset's id is provided as `blob_id` here? isn't it an asset identifier?

yarikoptic: note: discussion is ongoing with @dchiquito in slack internal about this blob_id etc

dandi/dandiapi.py
```diff
@@ -729,13 +832,61 @@ def api_path(self) -> str:
...
+        """
+        if isinstance(digest_type, models.DigestType):
+            digest_type = digest_type.value
+        metadata = self.get_raw_metadata()
+        try:
+            return metadata["digest"][digest_type]
+        except KeyError:
+            raise NotFoundError(f"No {digest_type} digest found in metadata")
+
+    def set_metadata(self, metadata: models.Asset) -> None:
```
yarikoptic, line 863: I think such a name would be quite a bit off with our "Copy-on-Write" implementation of dandi-api: metadata cannot really be modified, but rather we provide a new metadata and get a new asset (id).  So `set_raw_metadata` seems to be doing that below and returns a new `RemoteAsset`.  I wonder if we should just promote it into `get_modified_asset` or something like that and not bother with `raw` vs non-raw separation (which is a bit confusing to me at least)?

jwodder: Further bikeshedding of the method name: `update_metadata()`? `modify_metadata()`?

I think the `raw` vs. non-`raw` separation is useful for when you're working with non-model metadata, in which case converting it to a model and back would just be a waste of cycles.  Or are you suggesting one method that takes both raw and non-raw and converts the latter into the former?

dandi/dandiapi.py
```diff
@@ -325,24 +349,67 @@ class Version(APIBase):
...
+        data: Optional[Dict[str, Any]] = None,
+    ) -> None:
+        self.client = client
+        self.identifier = identifier
+        self._version: Optional[Version]
+        if isinstance(version, str):
+            self.version_id = version
+            self._version: Optional[Version] = None
+        else:
+            self.version_id = version.identifier
```
yarikoptic, line 372: tried on code which works on master but crashes here:
```
$> git describe
0.22.0-12-g53c5c5e

$> python -c 'from dandi.dandiapi import DandiAPIClient as C; c=C(); ds = c.get_dandiset("000055"); print(ds)'
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/dandiapi.py", line 302, in get_dandiset
    return RemoteDandiset(self, dandiset_id, version_id)
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/dandiapi.py", line 372, in __init__
    self.version_id = version.identifier
AttributeError: 'NoneType' object has no attribute 'identifier'
```

jwodder: Fixed; tests added.

jwodder:
The tests are currently failing because of [this problem with dandi-api](https://github.com/dandi/dandi-api/issues/393#issuecomment-874890385).

dandi/dandiapi.py
yarikoptic: `.update*` and `.modify*` suggest in-place operations for me, which isn't what happening here, unless we mutate the Asset instance with returned new id etc, which I don't think we should do.

ok, let's keep `raw` if you find it useful.

jwodder: How about `apply_metadata()`?

yarikoptic: also suggests mutating an existing object.  I think following all the get/set, could be `get_updated`

jwodder: That suggests to me that it's only mutating the local object, not the data on the server.  I really think `set_metadata()` is the best choice.

yarikoptic: but new metadata produces a new asset, so it would need to mutate the `RemoteAsset` instance then to replace everything (id, metadata) with new metadata as on the remote server.

jwodder: @satra We need another opinion on what to name this method.

jwodder: @yarikoptic I've decided to make the method update the `RemoteAsset` in place and stick with the `set_{raw,}_metadata()` name.

yarikoptic: Ok, I see advantages to that from UX perspective.  Thank you. Let's see where it leads us ;-)

waxlamp: For a further viewpoint: libraries such as [ImmutableJS](https://github.com/immutable-js/immutable-js) have data structure APIs with methods named `update()`, `set_*()`, et al. that actually manufacture new instances and return them, leaving the old instance unchanged. The use of mutational verbs for these actions isn't a bad thing if it's understood that you're working with immutable data. In fact, it's a nice abstraction to be able to say "set the value to X" and hide the fact that nothing is really being set at all because it continues the semantics that we are accustomed to in earlier, non-immutable data structure libraries.

Following an argument like that, I think using mutational names here is perfectly fine. (This argument also applies to using `PUT` to apparently "modify" these resources in the REST API.)

dandi/dandiapi.py
```diff
@@ -269,21 +287,27 @@ def dandi_authenticate(self):
...
+        if lazy:
+            return RemoteDandiset(self, dandiset_id, version_id)
+        else:
+            d = RemoteDandiset._make(self, self.get(f"/dandisets/{dandiset_id}/"))
+            if version_id is not None and version_id != d.version_id:
+                if version_id == DRAFT:
+                    d.version = d.draft_version
+                else:
+                    return d.for_version(version_id)
+            return d
```
yarikoptic, line 310: coverage is odd here but suggests that we do not cover non-lazy code block.  Please add a dedicated test

dandi/dandiapi.py
```diff
@@ -230,6 +237,17 @@ def __init__(self, api_url=None, token=None):
         if token is not None:
             self.authenticate(token)
 
+    @classmethod
+    def for_dandi_instance(
```
yarikoptic, line 241: `from_` methods remain a bit odd in naming for me, but I think I am growing into "ok"ing having them, although difference from `get_` or `from_` still remains somewhat opaque to me.

dandi/dandiapi.py
```diff
@@ -269,21 +287,27 @@ def dandi_authenticate(self):
...
+            d = RemoteDandiset._make(self, self.get(f"/dandisets/{dandiset_id}/"))
+            if version_id is not None and version_id != d.version_id:
+                if version_id == DRAFT:
+                    d.version = d.draft_version
+                else:
+                    return d.for_version(version_id)
+            return d
 
     def get_dandisets(self) -> Iterator["RemoteDandiset"]:
         for data in self.paginate("/dandisets/"):
```
yarikoptic, line 313: seems needing test

dandi/dandiapi.py
```diff
@@ -325,24 +349,82 @@ class Version(APIBase):
...
+        identifier: str,
+        version: Union[str, Version, None],
+        data: Optional[Dict[str, Any]] = None,
+    ) -> None:
+        self.client = client
+        self.identifier = identifier
+        self._version_id: Optional[str]
+        self._version: Optional[Version]
+        if version is None:
+            self._version_id = None
```
yarikoptic, line 370: needs a test since I am not sure what would happen whenever I try to instantiate without providing a version - would it be a draft one? last released? or some error? if I start performing some actions.
only looking below saw it being lazily populated in `@version_id` -- worth a comment here on that.

dandi/dandiapi.py
```diff
@@ -325,24 +349,82 @@ class Version(APIBase):
...
+            self._version = None
+        elif isinstance(version, str):
+            self._version_id = version
+            self._version = None
+        else:
+            self._version_id = version.identifier
+            self._version = version
+        self._data = data
```
yarikoptic, line 378: given that those are "datasets", seeing `._data` suggests some notion of actual "data" (in our terms "assets") of the dataset, not its ... well, not "metadata" but may be , `._remote_rec` for "remote record" or just `._rec`?

dandi/dandiapi.py
```diff
@@ -325,24 +349,82 @@ class Version(APIBase):
...
+            if self._version_id is None:
+                self._get_data()
+            if self._data is not None:
+                for vattr in ["most_recent_published_version", "draft_version"]:
+                    vdict = self._data.get(vattr)
+                    if vdict and (
+                        self._version_id is None or vdict["version"] == self.version_id
+                    ):
+                        self._version = Version.parse_obj(vdict)
+                        return self._version
```
yarikoptic, line 404: looks like needing a test coverage too

dandi/dandiapi.py
```diff
@@ -729,13 +847,63 @@ def api_path(self) -> str:
...
+        return models.Asset.parse_obj(self.get_raw_metadata())
+
     def get_raw_metadata(self) -> Dict[str, Any]:
         """Fetch the metadata for the asset as an unprocessed `dict`"""
         if self._metadata is not None:
             return self._metadata
         else:
             return cast(Dict[str, Any], self.client.get(self.api_path))
 
+    def get_digest(self, digest_type: Union[str, models.DigestType]) -> str:
```
yarikoptic, line 864: Let's default to our `dandi-etag` may be?

dandi/dandiapi.py
yarikoptic: Thank you @waxlamp for more input/examples.  It still seats not entirely well with me, but I guess let's proceed the "mutable" way and see where it brings us ;-)

yarikoptic (changes requested):
Left minor comments.  Overall I think it is a good way forward so let's bring it to a mergeable state -- I have not looked into what tests failures are but I guess they are is what contributed to many code paths not being covered (please ignore those comments I guess unless they remain non-covered after tests are happy).

Also @jwodder - do you expect breakage in any of the downstream use-cases? (dandi-api, etc)  if so -- any simple ways to provide compatibility shims/deprecation cycle?

dandi/dandiapi.py

dandi/dandiapi.py
jwodder: This is `for_`, not `from_`.

dandi/dandiapi.py
jwodder: Done.

dandi/dandiapi.py
jwodder: First of all, end users shouldn't be calling the constructors for any of the API resource classes.  If no version is specified, it follows the same rules as documented under `DandiAPIClient.get_dandiset()`.

dandi/dandiapi.py
jwodder: Do you mean tests for `get_dandisets()`?  How should I test that?  Should the test just consume everything in the iterator and check that the elements are `RemoteDandiset`s?

dandi/dandiapi.py
jwodder: Covered.

dandi/dandiapi.py
jwodder: It's data about the Dandiset.  I think `_data` is perfectly fine.

dandi/dandiapi.py
jwodder: Covered by tests.

jwodder:
Failing tests are caused by https://github.com/dandi/dandi-api/issues/428.

dandi/dandiapi.py
yarikoptic: we have tests which create/populate dandisets, so it would be just a matter of adding to one of those tests that `get_dandisets` would return a list which includes the created dandiset

dandi/dandiapi.py
yarikoptic: well, it is also in kwargs
```
 data: Optional[Dict[str, Any]] = None,
```
of a public interface and even without any documentation on what that `data` is.  So even if I agree that non-public `._data` is easy to understand the purpose of (which still isn't the case for me), having `data` in public interface is really confusing.

Also given that kwarg `data` is probably to be used only for internal purposes, why not to remove it from the interface and just make those internal "constructors" assign `._data` or have some internal `._set_data()` method?  Do you foresee use-case for a user to provide `data`?

dandi/dandiapi.py
```diff
@@ -325,24 +349,82 @@ class Version(APIBase):
...
     def version_id(self) -> str:
-        return self.version.identifier
+        if self._version_id is None:
+            self._version_id = self.version.identifier
+        return self._version_id
+
+    @property
+    def version(self) -> Version:
+        """The version in question of the Dandiset"""
+        if self._version is None:
```
yarikoptic, line 394: should we may be default `version` to None in `__init__` if we can handle it?

dandi/dandiapi.py
```diff
@@ -325,24 +349,82 @@ class Version(APIBase):
...
-    modified: datetime
-    #: The version in question of the Dandiset
-    version: Version
-    most_recent_published_version: Optional[Version]
-    draft_version: Version
+    def __init__(
+        self,
+        client: DandiAPIClient,
+        identifier: str,
+        version: Union[str, Version, None],
```
yarikoptic, line 362: there should be a docstring which would describe what happens if `version=None`

yarikoptic: oh, from right below https://github.com/dandi/dandi-cli/pull/676/files#r673408547 I see that intention is to have these classes not really to be instantiated by "users" (who might be developers) directly... hm... need to digest

dandi/dandiapi.py
yarikoptic: > end users shouldn't be calling the constructors for any of the API resource classes

filed a dedicated https://github.com/dandi/dandi-cli/issues/731 since such separation was not clear to me at all

yarikoptic (changes requested):
Some comments, many are about tests, in particular about new functionality added but without being accompanied by tests.  I filed dedicated issues for possibly improving organization/documentation of the API elements so it might be easier to get using it without needing my of RTFM.

dandi/dandiapi.py
jwodder: Done.

dandi/dandiapi.py
jwodder: The constructors are not public interfaces.  None of them should be called by end users.

dandi/dandiapi.py
jwodder: Done.

dandi/dandiapi.py
```diff
@@ -325,24 +349,82 @@ class Version(APIBase):
     modified: datetime
 
 
-class RemoteDandiset(APIBase):
+class RemoteDandiset:
```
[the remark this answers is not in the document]
jwodder: The laziness of `RemoteDandiset` is incompatible with pydantic.  I do not foresee adding any laziness to `RemoteAsset`.

dandi/dandiapi.py
```diff
@@ -455,15 +577,11 @@ def publish(self) -> "RemoteDandiset":
...
-            }
+        return self.for_version(
+            Version.parse_obj(self.client.post(f"{self.version_api_path}publish/"))
         )
 
-    def get_assets(self, path=None) -> Iterator["RemoteAsset"]:
+    def get_assets(self) -> Iterator["RemoteAsset"]:
         """Returns an iterator of all assets in this version of the Dandiset"""
         for a in self.client.paginate(f"{self.version_api_path}assets/"):
             yield self._mkasset(a)
```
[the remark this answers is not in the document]
jwodder: `get_assets_under_path()` gets all assets whose paths start with `path`.  Why would we want to enforce a terminating slash?  Fetching "partial" paths is a perfectly cromulent thing to do.

dandi/dandiapi.py
```diff
@@ -385,6 +469,24 @@ def _mkasset_from_metadata(self, metadata: Dict[str, Any]) -> "RemoteAsset":
             _metadata=metadata,
         )
 
+    def json_dict(self) -> Dict[str, Any]:
+        """
+        Convert to a JSONable `dict`, omitting the ``client`` attribute and
+        using the same field names as in the API
+        """
+        return {**self._get_data(), "version": self.version.json_dict()}
```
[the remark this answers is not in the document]

dandi/dandiapi.py
```diff
@@ -385,6 +469,24 @@ def _mkasset_from_metadata(self, metadata: Dict[str, Any]) -> "RemoteAsset":
...
+    def refresh(self) -> None:
+        """Refetch the data for this Dandiset & version"""
+        self._data = self.client.get(f"/dandisets/{self.identifier}/")
+        for vattr in ["most_recent_published_version", "draft_version"]:
+            vdict = self._data.get(vattr)
+            if vdict and vdict["version"] == self.version_id:
+                self._version = Version.parse_obj(vdict)
+                break
+        else:
+            self._version = self.get_version(self.version_id)
```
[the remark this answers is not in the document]
jwodder: `refresh()` updates the `_data` and also updates the attributes of the `RemoteDandiset`'s `version` attribute, in case anything there has changed.  The version continues to point at the same version as before.  We loop through the version attributes in order to avoid an API call in the event that the `RemoteDandiset`'s version is one of them; if it is neither, we update the version data by refetching it.

dandi/dandiapi.py
```diff
@@ -385,6 +469,24 @@ def _mkasset_from_metadata(self, metadata: Dict[str, Any]) -> "RemoteAsset":
...
+
```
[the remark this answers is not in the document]
jwodder: Done.

dandi/dandiapi.py
```diff
@@ -407,19 +511,37 @@ def for_version(self, version_id: Union[str, Version]) -> "RemoteDandiset":
...
         Fetch the metadata for this version of the Dandiset as an unprocessed
         `dict`
         """
         return cast(Dict[str, Any], self.client.get(self.version_api_path))
 
+    def set_metadata(self, metadata: models.Dandiset) -> None:
+        """
+        Set the metadata for this version of the Dandiset to the given value
+        """
+        self.set_raw_metadata(metadata.json_dict())
```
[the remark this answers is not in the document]
jwodder: Done.

dandi/dandiapi.py
```diff
@@ -729,13 +847,65 @@ def api_path(self) -> str:
     def download_url(self) -> str:
         return self.client.get_url(f"{self.api_path}download/")
 
+    def get_metadata(self) -> models.Asset:
+        """
+        Fetch the metadata for the asset as a `dandischema.models.Asset`
+        instance
+        """
+        return models.Asset.parse_obj(self.get_raw_metadata())
```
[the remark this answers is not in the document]
jwodder: Done.

dandi/dandiapi.py
```diff
@@ -729,13 +847,65 @@ def api_path(self) -> str:
...
 
+    def get_digest(
+        self, digest_type: Union[str, models.DigestType] = models.DigestType.dandi_etag
+    ) -> str:
+        """
+        Retrieves the value of the given type of digest from the asset's
+        metadata.  Raises `NotFoundError` if there is no entry for the given
+        digest type.
+        """
+        if isinstance(digest_type, models.DigestType):
```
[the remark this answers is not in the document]
jwodder: Done.

dandi/dandiapi.py
```diff
@@ -407,19 +511,37 @@ def for_version(self, version_id: Union[str, Version]) -> "RemoteDandiset":
...
-        return self.copy(update={"version": version_id})
+        return type(self)(
+            client=self.client,
+            identifier=self.identifier,
+            version=version_id,
+            data=self._data,
+        )
 
     def delete(self) -> None:
         """Delete the Dandiset"""
```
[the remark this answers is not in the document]

dandi/dandiapi.py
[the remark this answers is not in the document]
yarikoptic: oh -- I mixed up with `assets/paths/` endpoint reading the name `..._under_path`, and we do not even interface that `assets/paths/` endpoint yet (so far no need).

OOS for this PR, so filed a dedicated https://github.com/dandi/dandi-cli/issues/738

dandi/dandiapi.py
yarikoptic: Let's continue on that in https://github.com/dandi/dandi-cli/issues/731 since from current state it is not obvious on what is intended to be public and not

dandi/dandiapi.py
```diff
@@ -385,6 +469,24 @@ def _mkasset_from_metadata(self, metadata: Dict[str, Any]) -> "RemoteAsset":
...
+    def json_dict(self) -> Dict[str, Any]:
+        """
+        Convert to a JSONable `dict`, omitting the ``client`` attribute and
+        using the same field names as in the API
+        """
+        return {**self._get_data(), "version": self.version.json_dict()}
+
+    def refresh(self) -> None:
+        """Refetch the data for this Dandiset & version"""
+        self._data = self.client.get(f"/dandisets/{self.identifier}/")
```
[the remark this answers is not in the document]
yarikoptic: filed a dedicated https://github.com/dandi/dandi-cli/issues/739 since want to proceed with current state of PR for now

yarikoptic:
thank you @jwodder ! Let's proceed and see where it brings us.  I will even mark this PR for release so it (and thus all accumulated changes) get released and we get a "wider" chance to test it out.

PR #680, opened 2021-06-23
Fix a failing test on Windows on conda-forge

jwodder:
dandi's conda-forge build is failing on Windows in part because `SHELL` isn't set for some reason when passing the shell completion command to Bash.  This PR fixes that by always setting `SHELL` explicitly.


yarikoptic:
thank you!

yarikoptic:
Let's release, even though it includes a piecemeal API RF, I think it was not too drastic and one issue was resolved after that... let's see in the wild.

PR #681, opened 2021-06-24
Fix a failing test on Windows on conda-forge (again)

jwodder:
dandi's conda-forge build is failing on Windows in part because the bash executable is named `bash.EXE`.  This PR fixes that by stripping off any ".exe" suffix (case insensitive) from the SHELL value before using it.

### DIFF
--- a/dandi/cli/cmd_shell_completion.py
+++ b/dandi/cli/cmd_shell_completion.py
@@ -36,6 +36,8 @@ def shell_completion(shell):
             raise click.UsageError(
                 "Could not determine running shell: SHELL environment variable not set"
             )
+        if shell.lower().endswith(".exe"):
+            shell = shell[:-4]
         if shell not in SHELLS:
             raise click.UsageError(f"Unsupported/unrecognized shell {shell!r}")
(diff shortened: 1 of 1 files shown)


dandi/cli/cmd_shell_completion.py
```diff
@@ -36,6 +36,8 @@ def shell_completion(shell):
             raise click.UsageError(
                 "Could not determine running shell: SHELL environment variable not set"
             )
+        if shell.lower().endswith(".exe"):
```
yarikoptic, line 39: I wouldn't rely on it be just .exe -- couldn't it be `.com` (IIRC some "small executable fiting into RAM, e.g. a "driver" ;)) or some `.bat` shim? may be do some `splitext` and check if any of those?

jwodder: Done.

yarikoptic:
I guess we will just release some time soon and "test in production". Thank you @jwodder

yarikoptic:
restarted some workflows, green now, merging

PR #684, opened 2021-06-28
Remove references to "dandi register" command

jwodder:
The command was removed when the Girder code was removed.


yarikoptic:
Thank you @jwodder !

Issue #625, opened 2021-05-10
Different ways of validating dandiset metadata


satra:
i'm doing optional injections only to show other types of errors when known issues are taken into account.

```python
import requests
import json
import jsonschema
def get_dandisets():
   url = f"https://api.dandiarchive.org/api/dandisets"
   querystring = {"page_size":"100"}
   headers = {"Accept": "application/json"}
   resp = requests.request("GET", url, headers=headers, params=querystring).json()
   results = resp["results"]
   while resp["next"]:
       resp = requests.request("GET", resp["next"], headers=headers).json()
       results.extend(resp["results"])
   assert len(results) == resp["count"]
   return results
dandisets = get_dandisets()
metadata = {}
for dandiset in dandisets:
    url = f"https://api.dandiarchive.org/api/dandisets/{dandiset['identifier']}/versions/draft"
    headers = {"Accept": "application/json"}
    resp = requests.request("GET", url, headers=headers).json()
    metadata[dandiset["identifier"]] = resp

with open("../tmp/schema/0.3.0/dandiset.json") as fp:
    schema = json.load(fp)

from dandi.models import DandisetMeta
inject = False  # doesn't include some known missing fields

# using model
for val in metadata.values:
    base = {}
    if inject:
        base = dict(version="draft", license=["spdx:CC0-1.0"], contributor=[{"schemaKey": "Person", "name": "Last, First", "roleName": ['dandi:ContactPerson']}], id="fancyid")
    base.update(**val)
    try:
        DandisetMeta(**base)
    except ValidationError as e:
        print(val["identifier"], str(e))

# using json schema
for val in metadata.values():
    base = {}
    if inject:
        base = dict(version="draft", license=["spdx:CC0-1.0"], contributor=[{"schemaKey": "Person", "name": "Last, First", "roleName": ['dandi:ContactPerson']}], id="fancyid")
    base.update(**val)
    try:
        jsonschema.validate(base, schema)
    except jsonschema.ValidationError as e:
        print(val["identifier"], e.message)

```

satra:
jsonschema will give the first error only and stop. pydantic will give multiple errors, but not being used on the server. this was simply an exercise to try and get to understand what errors exist. 

so we may want to indicate to the dataset creator through different means about validity of the metadata and on different interfaces tell them what to fix.

yarikoptic:
ok, so we should 

[...]

PR #685, opened 2021-06-28
Use dandischema.metadata.validate() instead of _validate_*_json()

jwodder:
Judging by the underscores at the start of the function names, we shouldn't be using them.  This PR changes to an interface intended for public use.

### DIFF
--- a/dandi/tests/test_metadata.py
+++ b/dandi/tests/test_metadata.py
@@ -4,11 +4,7 @@ import json
 from pathlib import Path
 
 from dandischema.consts import DANDI_SCHEMA_VERSION
-from dandischema.metadata import (
-    _validate_asset_json,
-    _validate_dandiset_json,
-    publish_model_schemata,
-)
+from dandischema.metadata import validate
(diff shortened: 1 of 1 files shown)


yarikoptic:
Those two functions validate against/using jsonschema, not pydantic models.  We indeed should stop away from using jsonschema for validating there (well -- in the tests).  Eventually we might want to get back to @satra 's #625 and provide a dedicated interface similar to `metadata.validate` to validate against jsonschema if such needs arises.
But then - should we just remove  those functions may be? they seems to be not used anywhere besides tests
```
$> git grep '_validate.*json('
dandi/tests/test_metadata.py:    _validate_asset_json(data_as_dict, schema_dir)
dandi/tests/test_metadata.py:    _validate_asset_json(data_as_dict, schema_dir)
dandi/tests/test_metadata.py:    _validate_dandiset_json(data_as_dict, schema_dir)

```

yarikoptic (approved):
Feel welcome to proceed and merge, we can remove / refactor for jsonschema-based validation later

PR #686, opened 2021-06-28
changing unitText

djarecka:
changes of `unitText` in `extract_age`
a subset of changes from #450

### DIFF
--- a/dandi/tests/data/metadata/metadata2asset.json
+++ b/dandi/tests/data/metadata/metadata2asset.json
@@ -69,7 +69,7 @@
     {
       "identifier": "a1b2c3",
       "age": {
-        "unitText": "Years from birth",
+        "unitText": "ISO-8601 duration",
         "value": "P170DT12212S",
         "schemaKey": "PropertyValue"
       },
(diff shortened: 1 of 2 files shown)


dandi/tests/data/metadata/metadata2asset.json
```diff
@@ -69,7 +69,7 @@
     {
       "identifier": "a1b2c3",
       "age": {
-        "unitText": "Years from birth",
+        "unitText": "ISO-8601 duration",
```
yarikoptic, line 72: Should it still be 

```suggestion
        "unitText": "ISO-8601 duration since birth",
```

until we introduce an explicit field for the point of reference (`birth`, `gestation`)?  @satra @bendichter ?  since otherwise it is not clear for what "age" it refers to?

BTW -- does ISO-8601 allow for "negative" durations -- it is not clear from https://en.wikipedia.org/wiki/ISO_8601 if it does.  I.e. how would we encode "age" as "1 month before birth" if someone e.g. does some recordings within embryo in vivo, and post-referencing all the dates to the date of birth.

djarecka: my understanding is that it does not to have to be "since birth", and that's why we wanted to have various option in `AgeReferenceType`: https://github.com/dandi/dandischema/pull/50/files

yarikoptic: rright . I lost track of PRs.  CI is green, let's proceed.

PR #687, opened 2021-06-29
Increase minimum dandischema version to 0.2.9

jwodder:
As requested by @satra on Slack.

PR #691, opened 2021-06-30
Adjust RemoteAsset.json_dict()

### LINKED ISSUE #690: API design: tune up asset structure/records
further tuning up backups2datalad.py to make it work, noted that assets records now contain 
```
$> python -c 'import dandi; print(dandi.__version__); from dandi.dandiapi import DandiAPIClient; ds = DandiAPIClient().get_dandiset("000027", "draft"); print([a.json_dict() for a in ds.get_assets()])'         
0.21.0+14.g0a2903f
[{'dandiset_id': '000027', 'version_id': 'draft', 'asset_id': 'ff453f4c-a435-4a5d-a48b-128abca5ec47', 'path': 'sub-RAT123/sub-RAT123.nwb', 'size': 18792, 'modified': datetime.datetime(2021, 5, 18, 13, 13, 56, 247883, tzinfo=datetime.timezone.utc)}]
```

1. information about dandiset (dandiset_id, version_id) - I think we should aim to have asset decoupled from a dandiset. dandi-api doesn't provide dandiset information upon `/assets`, and eventually assets might get shared by multiple dandisets
2. `modified` is datetime (not str), despite `json_dict` somewhat suggesting that it would be json-digestable dict, but datetime causes json to puke upon dumps.  I think for `json_dict` it should be serialized into str.

jwodder:
Closes #690.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -1,11 +1,22 @@
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from datetime import datetime
+import json
 import os.path
 from pathlib import Path
 import re
 from threading import Lock
 from time import sleep
-from typing import Any, Callable, Dict, Iterator, Optional, Union, cast
+from typing import (
+    Any,
+    Callable,
+    ClassVar,
+    Dict,
+    FrozenSet,
+    Iterator,
+    Optional,
+    Union,
+    cast,
+)
 from urllib.parse import urlparse, urlunparse
 from xml.etree.ElementTree import fromstring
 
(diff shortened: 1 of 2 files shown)


jwodder:
The tests are failing because of https://github.com/dandi/dandi-api/issues/380.

yarikoptic:
WDYT about `datetime` (the `2.` of #690)?

jwodder:
@yarikoptic The field should be a string in ISO 8601 format now.

yarikoptic:
oh, I missed that check in the tests, thanks!

yarikoptic:
Let's proceed despite our shameful red colors ;-) and I will release with #687 since we would need newer schema

PR #693, opened 2021-06-30
Add get_asset_metadata() function

### LINKED ISSUE #692: centralize logic for file2asset(path, ds)
ATM upload does
```
            yield {"status": "extracting metadata"}
            try:
                asset_metadata = nwb2asset(
                    path, digest=file_etag, digest_type="dandi_etag"
                )
            except Exception as exc:
                lgr.exception("Failed to extract metadata from %s", path)
                if allow_any_path:
                    yield {"status": "failed to extract metadata"}
                    asset_metadata = get_default_metadata(
                        path, digest=file_etag, digest_type="dandi_etag"
                    )
                else:
                    yield skip_file("failed to extract metadata: %s" % str(exc))
                    return
            metadata = asset_metadata.json_dict()
            metadata["path"] = str(relpath)
```
so besides fall-back to `get_default_metadata` it adds a logic to convert to json_dict (could be omitted) but also fixes up the path to be relative to the dataset.  For the purpose of metadata migration/updates etc,  would be useful to centralize that logic so we do not need to (possibly incompletely) duplicate it within backups2datalad.py etc.

Could be a helper function or a class method on Asset, e.g. `from_file` then (IIRC has idea about dandiset) may be even `ds` does not need to be provided

jwodder:
Closes #692.

### DIFF
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -660,6 +660,28 @@ def process_ndtypes(asset, nd_types):
     return asset
 
 
+def get_asset_metadata(
+    filepath, relpath, digest=None, digest_type=None, allow_any_path=True
+) -> models.BareAsset:
+    try:
+        metadata = nwb2asset(filepath, digest=digest, digest_type=digest_type)
+    except Exception as e:
+        lgr.info(
+            "Failed to extract NWB metadata from %s: %s: %s",
+            filepath,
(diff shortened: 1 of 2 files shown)


yarikoptic:
Looks good to me and even green, so let's proceed. Thank you @jwodder !

PR #694, opened 2021-07-01
fix: set CLI version to align with schema base version

satra:
The current schema minimum version is `0.4.4`. This is only available as of CLI release `0.22.0`, since it pins dandischema to `0.2.9`. To make sure people are putting in the appropriate metadata and seeing the proper metadata locally via `dandi ls` we should prevent earlier CLI usage.

after the age PR(s) is(are) merged and released we should update this again.


yarikoptic:
failing test is known: https://github.com/dandi/dandi-api/issues/393 let's proceed to encourage client updates

PR #702, opened 2021-07-08
Skip shell completion test entirely on Windows

jwodder:
It just doesn't work on conda-forge.


yarikoptic:
ok, let's please conda's anti-windows Gods and try with new release

PR #705, opened 2021-07-08
BF+RF(TST): populate contentUrl to satisfy Asset requirement in 0.5.0

### LINKED ISSUE #704: contentUrl is now mandatory for Asset in dandischema 0.5.0
See https://github.com/dandi/dandischema/pull/52#discussion_r666427566 for discussion etc.  Since `BareAsset`s are "casted" into plain `Asset`s via `schemaKey`, `validate` would fail if contentUrl is not specified even through originally it was a BareAsset.

yarikoptic:
While at it, I RFed the code to avoid duplicate logic for
two nearly identical tests

Not sure if that is the best way to proceed ATM, but should get tests into green again even when schema-0.5.0 gets released

Closes #704

### DIFF
--- a/dandi/tests/test_metadata.py
+++ b/dandi/tests/test_metadata.py
@@ -11,6 +11,7 @@ from dandischema.models import Dandiset as DandisetMeta
 from dandischema.models import PropertyValue
 from dateutil.tz import tzutc
 import pytest
+from semantic_version import Version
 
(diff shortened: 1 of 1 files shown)


jwodder (approved):

satra (approved):
LGTM

PR #706, opened 2021-07-09
Use timed wait in publish test

dchiquito:
I think the flaky tests we have been seeing are being caused by multiple
sequential celery tasks. Creating a new dandiset and uploading an asset
both trigger the validation task, but the first one is guaranteed to
fail. I think the test was noticing that first failure and quitting
immediately, instead of waiting the extra time required for the second
task to validate correctly.

Should fix dandi/dandi-api#393

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -429,28 +429,33 @@ class RemoteDandiset(APIBase):
             json={"metadata": metadata, "name": metadata.get("name", "")},
         )
 
+    def wait_until_valid(self, min_time=20):
+        """
+        Wait for a Dandiset to be valid.  Validation is a background celery
+        task which runs asynchronously, so we need to wait for it to complete.
+        """
+        lgr.debug("Waiting for Dandiset %s to complete validation ...", self.identifier)
+        start = time()
+        while time() - start < min_time:
+            r = self.client.get(f"{self.version_api_path}info/")
+            if "status" not in r:
+                # Running against older version of dandi-api that doesn't
+                # validate
+                return
+            if r["status"] == "Valid":
+                return
+            sleep(0.5)
+        else:
+            raise ValueError(
+                f"Dandiset {self.identifier} is {r['status']}: {r['validation_error']}"
+            )
+
     def publish(self) -> "RemoteDandiset":
         """
         Publish this version of the Dandiset.  Returns a copy of the
         `RemoteDandiset` with the `version` attribute set to the new published
         `Version`.
         """
-        lgr.debug("Waiting for Dandiset %s to complete validation ...", self.identifier)
-        start = time()
-        # Wait 20s for celery tasks to finish
-        while time() - start < 20:
-            r = self.client.get(f"{self.version_api_path}info/")
-            if "status" not in r:
-                # Running against older version of dandi-api that doesn't
-                # validate
-                break
-            if r["status"] == "Valid":
-                break
-            sleep(0.5)
-        else:
-            raise ValueError(
-                f"Dandiset {self.identifier} is {r['status']}: {r['validation_error']}"
-            )
         return self.copy(
             update={
                 "version": Version.parse_obj(
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -5,7 +5,7 @@ import os.path
 from pathlib import Path
 import re
 from threading import Lock
-from time import sleep
+from time import sleep, time
 from typing import (
     Any,
     Callable,
@@ -436,7 +436,9 @@ class RemoteDandiset(APIBase):
         `Version`.
         """
         lgr.debug("Waiting for Dandiset %s to complete validation ...", self.identifier)
-        while True:
+        start = time()
+        # Wait 20s for celery tasks to finish
+        while time() - start < 20:
             r = self.client.get(f"{self.version_api_path}info/")
             if "status" not in r:
                 # Running against older version of dandi-api that doesn't
@@ -444,11 +446,11 @@ class RemoteDandiset(APIBase):
                 break
             if r["status"] == "Valid":
                 break
-            elif r["status"] == "Invalid":
-                raise ValueError(
-                    f"Dandiset {self.identifier} is invalid: {r['validation_error']}"
-                )
             sleep(0.5)
+        else:
+            raise ValueError(
+                f"Dandiset {self.identifier} is {r['status']}: {r['validation_error']}"
+            )
         return self.copy(
             update={
                 "version": Version.parse_obj(
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -445,10 +445,9 @@ class RemoteDandiset(APIBase):
             if r["status"] == "Valid":
                 return
             sleep(0.5)
-        else:
-            raise ValueError(
-                f"Dandiset {self.identifier} is {r['status']}: {r['validation_error']}"
-            )
+        raise ValueError(
+            f"Dandiset {self.identifier} is {r['status']}: {r['validation_error']}"
+        )
 
     def publish(self) -> "RemoteDandiset":
         """
--- a/dandi/tests/test_dandiapi.py
+++ b/dandi/tests/test_dandiapi.py
@@ -64,6 +64,7 @@ def test_publish_and_manipulate(local_dandi_api, monkeypatch, tmp_path):
         validation="skip",
     )
 
+    d.wait_until_valid()
     version_id = d.publish().version.identifier
 
     download_dir = tmp_path / "download"


jwodder:
@dchiquito The problem (aside from the lack of a discernable reason why the metadata would fail validation) is that the dandi-api server is returning a status of "Invalid" when the validation is still partially in progress and thus the error message is unhelpful.  I believe dandi-api should be modified to either return the validation errors seen so far or else continue reporting the status as "Pending" until all validations have completed.

dandi/dandiapi.py
```diff
@@ -436,19 +436,21 @@ def publish(self) -> "RemoteDandiset":
...
+        # Wait 20s for celery tasks to finish
+        while time() - start < 20:
             r = self.client.get(f"{self.version_api_path}info/")
             if "status" not in r:
                 # Running against older version of dandi-api that doesn't
                 # validate
                 break
             if r["status"] == "Valid":
                 break
-            elif r["status"] == "Invalid":
```
jwodder, line 447: While a time limit is useful here, I think that this check should be kept so that Dandisets deemed invalid are reported as such quickly.  The only reason this check is a problem is because dandi-api is setting the status to "Invalid" without a helpful `validation_error`.

jwodder (changes requested):

dandi/dandiapi.py
```diff
@@ -436,19 +436,21 @@ def publish(self) -> "RemoteDandiset":
...
-                raise ValueError(
-                    f"Dandiset {self.identifier} is invalid: {r['validation_error']}"
-                )
             sleep(0.5)
+        else:
+            raise ValueError(
+                f"Dandiset {self.identifier} is {r['status']}: {r['validation_error']}"
+            )
```
yarikoptic, line 453: should it ideally result in e.g. 503 with `Retry-After` set for e.g. a few minutes if it is a matter of celery job just not completing within 20 seconds and otherwise not clear status ?

dchiquito:
The issue is that two atomic operations are being done that trigger validation: saving the initial dandiset, and uploading the asset. These are two different API requests, and two different validation tasks.
The server has no way of knowing that the asset upload is coming, so it simply validates what it has and records the dandiset as `Invalid`, which is correct. The second task would then mark it as `Valid` after validating it with the uploaded asset, but at this point the test has already failed.

jwodder:
@dchiquito Can't the server mark the Dandiset as "Pending" on new uploads?

dandi/dandiapi.py
dchiquito: The server doesn't necessarily know that another validation is in the queue, so it might not be appropriate to always respond with a 503. RabbitMQ does not allow non-destructive access to the message queue, so there's no real way to tell that another validation is coming soon.

dchiquito:
>Can't the server mark the Dandiset as "Pending" on new uploads?
Sadly no, not really. When an upload is started the server has no notion of which dandiset it would be added to, so it can't mark that dandiset as Pending, plus of course an upload might fail.
Even if we did mark the dandiset as "Pending" after the upload was added to the dandiset as an asset, that window while uploading is still open, so the test still has the potential to fail.

I think this change is the simplest possible solution that will allow us to continue working without having to deal with spurious CI errors. The validation system is quite complicated and moving bits of it around is much easier said than done.

dandi/dandiapi.py
```diff
@@ -436,19 +436,21 @@ def publish(self) -> "RemoteDandiset":
         `Version`.
         """
         lgr.debug("Waiting for Dandiset %s to complete validation ...", self.identifier)
-        while True:
+        start = time()
+        # Wait 20s for celery tasks to finish
+        while time() - start < 20:
```
satra, line 441: why is sleeping in this api file. this should only depend on whether the dataset is valid or not. the process of sleeping should be left up to any client depending on the tasks it does.

satra: shouldn't any wait be in the test related to publish?

dchiquito: This wait is currently only used before publishing a dandiset, so it lives in the publish function. We could move it out into a separate `wait_until_valid` function @jwodder ?

yarikoptic:
FWIW -- that single flaky dandi-cli test, sent out a PR https://github.com/dandi/dandi-cli/pull/707

dandi/dandiapi.py
yarikoptic: @jwodder why do you say that it is only the absence of validation_error which is the issue?
```
_________________________ test_publish_and_manipulate __________________________
/home/runner/work/dandi-cli/dandi-cli/dandi/tests/test_dandiapi.py:69: in test_publish_and_manipulate
    version_id = d.publish().version.identifier
/home/runner/work/dandi-cli/dandi-cli/dandi/dandiapi.py:551: in publish
    f"Dandiset {self.identifier} is invalid: {r['validation_error']}"
E   ValueError: Dandiset 000002 is invalid:
```
so it does `format` the string ok, and it seems a matter of receiving `"Invalid"` status whenever it should not be "Invalid" yet.  

There must be some explicit way API would say that status is "Pending" or "Unknown".  It must not return "Invalid" if it just "has not decided yet", which is what I guess happens here.

dandi/dandiapi.py
yarikoptic: @dchiquito what values "state" could be -- only "Valid" and "Invalid" (docs in https://api.dandiarchive.org/swagger/ tell nothing about any of expected values)?  What is the "life-cycle" for/between those two states?

jwodder:
@dchiquito Could the server update the status as part of responding to the `POST .../assets/` or `PUT .../assets/{asset_id}/` request at the end of an upload?  Exactly what sort of requests trigger an asset validation job, anyway?

dandi/dandiapi.py
jwodder: There is a "Pending" status, but @dchiquito is saying that there's no decent way to set the status back to that during the validation lifecycle.

dandi/dandiapi.py
satra: but why wait and for how long (i mean practically in a real dataset scenario) ? a dataset could have multiple 5TB objects whose checksum computation could take hours. that's why in the context of a test, one should know what a potential time window is and wait for that time in the test itself, not here in the publish function.

dchiquito:
> Could the server update the status as part of responding to the POST .../assets/ or PUT .../assets/{asset_id}/ request at the end of an upload?

Yes, it could. I just filed https://github.com/dandi/dandi-api/issues/407 to address that.
Sadly, that will only mitigate this flaky test, not solve it permanently. When uploading the file, the server still does not know which dandiset is should mark as "Pending", so there is still that window where the dandiset could temporarily be "Invalid".

> Exactly what sort of requests trigger an asset validation job, anyway?

Any PUT or POST on a `/assets` endpoint will eventually trigger asset validation.

Completely avoiding a premature "Invalid" state on the server side is not something that can be just tacked on; it would take a substantial overhaul to provide that guarantee. If this solution is unpalatable, please close the PR and we can begin the process of hashing out a better validation system next week.

yarikoptic:
> Completely avoiding a premature "Invalid" state on the server side is not something that can be just tacked on; it would take a substantial overhaul to provide that guarantee. If this solution is unpalatable, please close the PR and we can begin the process of hashing out a better validation system next week.

I guess we are doomed to have this PR (or alike, I will leave final word to @jwodder)  merged, **and**  "hash[ing] out a better validation system next week".

dandi/dandiapi.py
jwodder: I agree that splitting the waiting into a separate `wait_until_valid()` method  (perhaps with a `min_time` parameter) makes sense.

dandi/dandiapi.py
```diff
@@ -429,28 +429,33 @@ def set_raw_metadata(self, metadata: Dict[str, Any]) -> None:
...
         start = time()
-        # Wait 20s for celery tasks to finish
-        while time() - start < 20:
+        while time() - start < min_time:
             r = self.client.get(f"{self.version_api_path}info/")
             if "status" not in r:
                 # Running against older version of dandi-api that doesn't
                 # validate
-                break
+                return
```
yarikoptic, line 444: hm, I am confused how it makes it "wait" if it returns as soon as `status` not in the returned value.

Also, note that LGTM legitimately raises a concern that `else:` below is then not reachable if all we have is `return` and not `break`

yarikoptic:
@dchiquito please let me and @jwodder know if you need help to finalize this PR.

dandi/dandiapi.py
dchiquito: Normally `status` is a string like `"Pending"` or `"Validating"` or `"Invalid"`. It will always be present in recent versions of dandi-api. The only time it is not present is an internal server error or a very old version of dandi-api.

PR #707, opened 2021-07-09
BF(TST): Allow for IteratorWithAggregation to get nothing if reraise_immediately

yarikoptic:
with all the threading etc, I guess there is no way to guarantee it always
getting at least that single one before exception.  From one of the runs
in dandi-api:
https://github.com/dandi/dandi-cli/pull/706/checks?check_run_id=3030777195

	_________________________ test_IteratorWithAggregation _________________________
	dandi/support/tests/test_iterators.py:72: in test_IteratorWithAggregation
		assert got == [0]
	E   assert [] == [0]
	E     Right contains one more item: 0
	E     Full diff:
	E     - [0]
	E     ?  -
	E     + []

### DIFF
--- a/dandi/support/tests/test_iterators.py
+++ b/dandi/support/tests/test_iterators.py
@@ -59,7 +59,7 @@ def test_IteratorWithAggregation():
     assert got == [0, 1, 2]
     assert it.finished
 
-    # If there is an exception thrown, it would be raised only by the end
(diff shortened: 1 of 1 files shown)


yarikoptic:
fail is unrelated and the topic of heated discussion in #706 and elsewhere, so I will proceed with merge

Issue #713, opened 2021-07-15
Deploy sphinx-produced docs to RTD


yarikoptic:
- [x] by default should provide the one for the most recent release, with `master` to provide the version in master
- [x] should also do preview for the PRs, example e.g. https://github.com/con/open-brain-consent/pulls 
- [x] https://github.com/dandi/dandi-cli/issues/717

follow up to https://github.com/dandi/dandi-cli/pull/712

jwodder:
@yarikoptic Do you alreay have a user account on Readthedocs that I can add as a maintainer?  After I add you, you need to [set up the webhook](https://readthedocs.org/dashboard/dandi/integrations/160420/), which I apparently don't have permission to do.

yarikoptic:
- `site-readthedocs@onerussian.com`
- you are the Admin now for dandi-cli

jwodder:
@yarikoptic Added you and set up the webhook.  #715 just needs to be merged now, and then the site can build.

yarikoptic:
`latest` (master) fails to build, details from https://readthedocs.org/projects/dandi/builds/14260074/:

<details>
<summary></summary> 

```shell
Running Sphinx v4.1.1
loading translations [en]... done
making output directory... done
[autosummary] generating autosummary for: cmdline/delete.rst, cmdline/digest.rst, cmdline/download.rst, cmdline/index.rst, cmdline/ls.rst, cmdline/organize.rst, cmdline/shell-completion.rst, cmdline/upload.rst, cmdline/validate.rst, index.rst, modref/consts.rst, modref/dandiapi.rst, modref/dandiarchive.rst, modref/index.rst, modref/support.digests.rst, modref/utils.rst
/home/docs/checkouts/readthedocs.org/user_builds/dandi/envs/latest/lib/python3.7/importlib/_bootstrap.py:219: RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 80 from C header, got 88 from PyObject
  return f(*args, **kwds)
WARNING: [autosummary] failed to import 'dandi.tests.fixtures': no module named dandi.tests.fixtures
WARNING: [autosummary] failed to import 'dandi.tests.skip': no module named dandi.tests.skip
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 16 source files that are out of date
updating environment: [new config] 16 added, 0 changed, 0 removed
reading sources... [  6%] cmdline/delete
reading sources... [ 12%] cmdline/digest
[...]

PR #715, opened 2021-07-16
Add Readthedocs config

jwodder:
Part of #713.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,11 @@
+version: 2
+formats: all
+python:
+    version: 3
+    install:
+        - requirements: docs/requirements.txt
+        - method: pip
+          path: .
+sphinx:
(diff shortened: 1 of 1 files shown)


satra:
@yarikoptic - any reason to not just use gh-pages? if you do you would see this document at: www.dandiarchive.org/dandi-cli

yarikoptic:
> @yarikoptic - any reason to not just use gh-pages? if you do you would see this document at: [www.dandiarchive.org/dandi-cli](http://www.dandiarchive.org/dandi-cli)

- well, I guess we can use that as the top url on RTD as well, or assign/use dandi-cli.dandiarchive.org
- I like how it is possible to pre-view the PRs , see e.g. https://github.com/con/open-brain-consent/pull/100
- no need for custom actions -- we still don't know what hard-reset master on the handbook has happened ;)

PR #724, opened 2021-07-21
Make upload() fail if client & server schema versions are not in sync

### LINKED ISSUE #722: upload should fail if client uses outdated schema
I was blindly sure that we consult API about the version of the schema it expects, but apparently I was wrong since I cannot find any trace of that!?

`/info` endpoint, e.g. https://api-staging.dandiarchive.org/api/info/?format=json provides `schema_version`. `dandi-cli` should not even attempt upload if the schema version it uses doesn't match exactly ATM (I would prefer allowing for higher .patch being allowed, but I think neither dandischema nor dandi-api support that ATM, ref: e.g. https://github.com/dandi/dandischema/issues/22  )

I think absence of such prevention mechanism is what resulted in e.g. #719

jwodder:
Closes #722.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -299,11 +299,18 @@ class DandiAPIClient(RESTFullAPIClient):
     def match_schema_version(self, schema_version: Optional[str] = None) -> None:
         if schema_version is None:
             schema_version = get_schema_version()
-        server_schema_version = self.get("/info/")["schema_version"]
+        server_info = self.get("/info/")
+        server_schema_version = server_info.get("schema_version")
+        if not server_schema_version:
+            raise RuntimeError(
+                "Server did not provide schema_version in /info/;"
+                f" returned {server_info!r}"
+            )
         if server_schema_version != schema_version:
             raise SchemaVersionError(
                 f"Server requires schema version {server_schema_version};"
-                f" client only has {schema_version}"
+                f" client only supports {schema_version}.  You may need to"
+                " upgrade dandi and/or dandischema."
             )
 
 
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -296,7 +296,7 @@ class DandiAPIClient(RESTFullAPIClient):
             self, self.post("/dandisets/", json={"name": name, "metadata": metadata})
         )
 
-    def match_schema_version(self, schema_version: Optional[str] = None) -> None:
+    def check_schema_version(self, schema_version: Optional[str] = None) -> None:
         if schema_version is None:
             schema_version = get_schema_version()
         server_info = self.get("/info/")
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -21,13 +21,14 @@ from urllib.parse import urlparse, urlunparse
 from xml.etree.ElementTree import fromstring
 
 import click
+from dandischema.models import get_schema_version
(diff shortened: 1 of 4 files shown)


dandi/dandiapi.py
```diff
@@ -295,6 +296,16 @@ def create_dandiset(self, name: str, metadata: Dict[str, Any]) -> "RemoteDandise
             self, self.post("/dandisets/", json={"name": name, "metadata": metadata})
         )
 
+    def match_schema_version(self, schema_version: Optional[str] = None) -> None:
```
yarikoptic, line 299: I am hopeful in a bright(er) future, could we make it a

```suggestion
    def check_schema_version(self, schema_version: Optional[str] = None) -> None:
```
?

dandi/dandiapi.py
```diff
@@ -295,6 +296,16 @@ def create_dandiset(self, name: str, metadata: Dict[str, Any]) -> "RemoteDandise
...
+        if schema_version is None:
+            schema_version = get_schema_version()
+        server_schema_version = self.get("/info/")["schema_version"]
```
yarikoptic, line 302: you trust dandi-api a lot ;-)  could you make it into something like
```suggestion
        server_info = self.get("/info/")
        server_schema_version = server_info.get("schema_version")
        if not server_schema_version:
                raise RuntimeError(f"Server did not provide schema_version in its /info output: {server_info}")
```
so if api server malfunctions or changes its "API", users would get some informative message which could help us to debug

dandi/dandiapi.py
```diff
@@ -295,6 +296,16 @@ def create_dandiset(self, name: str, metadata: Dict[str, Any]) -> "RemoteDandise
...
+        if server_schema_version != schema_version:
+            raise SchemaVersionError(
+                f"Server requires schema version {server_schema_version};"
+                f" client only has {schema_version}"
```
yarikoptic, line 306:
```suggestion
                f" client supports {schema_version}. Upgrade of the dandischema and/or dandi client might be needed."
```
to provide some guidance to the user on what to do... 
(may be here we should remind with etelemetry, but we do not even track dandischema with it ATM... so let's not bother for this one)

dandi/upload.py
```diff
@@ -41,6 +41,7 @@ def upload(
     api_url = instance.api
 
     client = DandiAPIClient(api_url)
+    client.match_schema_version()
```
yarikoptic, line 44: +1 on doing it only in upload... I thought we might want it more generic (e.g. at DandiAPIClient level) but indeed there is no need.

yarikoptic (changes requested):
minor changes request

dandi/dandiapi.py
jwodder: Are you asking for any behavior change or just a name change?

dandi/dandiapi.py

dandi/dandiapi.py

dandi/dandiapi.py
yarikoptic: just the name.  ATM it is equality (AKA "match") check but later could become more sophisticated, and thus better make the name more descriptive at once.
Later, we might need to adjust it anyways: filed https://github.com/dandi/dandi-api/issues/437

yarikoptic:
Thank you @jwodder!

PR #725, opened 2021-07-22
ENH: exit with non-0 when "bad_version" of dandi-cli is used

yarikoptic:
BadVersion was added in 0.2.2 thus boosting etelemetry version

### DIFF
--- a/dandi/cli/command.py
+++ b/dandi/cli/command.py
@@ -128,7 +128,13 @@ def main(ctx, log_level, pdb=False):
     try:
         import etelemetry
 
-        etelemetry.check_available_version("dandi/dandi-cli", __version__, lgr=lgr)
+        try:
+            etelemetry.check_available_version(
+                "dandi/dandi-cli", __version__, lgr=lgr, raise_exception=True
+            )
+        except etelemetry.client.BadVersionError as exc:
+            # note: SystemExit is based of BaseException, so is not Exception
+            raise SystemExit(str(exc))
     except Exception as exc:
         lgr.warning(
             "Failed to check for a more recent version available with etelemetry: %s",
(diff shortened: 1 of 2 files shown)


jwodder (approved):

dandi/cli/command.py
```diff
@@ -128,7 +128,13 @@ def main(ctx, log_level, pdb=False):
...
         import etelemetry
 
-        etelemetry.check_available_version("dandi/dandi-cli", __version__, lgr=lgr)
+        try:
+            etelemetry.check_available_version(
+                "dandi/dandi-cli", __version__, lgr=lgr, raise_exception=True
+            )
+        except etelemetry.client.BadVersionError as exc:
+            # note: SystemExit is based of BaseException, so is not Exception
+            raise SystemExit(str(exc))
```
satra, line 137: Do you want to provide a more DANDI CLI specific message here? I think the generic message from etelemetry may scare some users. Something like:

DANDI CLI has detected that you are using a version that is known to contain bugs, is incompatible with our current data archive, or has other significant performance limitations. To continue uploading data to DANDI, please upgrade your dandi client to a newer version using `pip install --upgrade dandi`. If you have any issues, please contact the DANDI helpdesk at https://github.com/dandi/helpdesk/issues/new/choose .

yarikoptic: message now looks like
```
$> dandi ls
2021-07-22 10:22:41,825 [ WARNING] A newer version (0.23.2) of dandi/dandi-cli is available. You are using 0.21.0
You are using a version of dandi/dandi-cli with a critical bug. Please use a different version.
```
it might have been installed not via pip (`conda`) so we should not suggest specific command... indeed we can update message a bit here though so choosing your version with just slight tune ups.  pushed

satra (approved):

yarikoptic:
2 approvals and linter still isn't happy, damn it! force pushed the fix

yarikoptic:
for one failed run filed https://github.com/dandi/dandi-cli/issues/727 .  I think we can proceed now

PR #728, opened 2021-07-22
Invoke etelemetry when constructing a DandiAPIClient; honor DANDI_NO_ET

### LINKED ISSUE #726: Move etelemetry logic to become in effect also when used as a Python library
ATM we have etelemetry under `cli/command.py:main` so it is only in effect for CLI.  As we will use it now to prevent "bad" versions (https://github.com/dandi/dandi-cli/pull/725), we should make it used consistently between CLI and when used as a Python module.

Not yet sure how to not bother with etelemetry if user does use CLI but merely for `--help` -- I guess it is ok to do all etelemetry checks/etc, but if we are to fail on bad version, we should not bother with etelemetry for "every import".

Since the main rationale is for "interactions with dandiarchive" I think we should 

- make etelemetry check helper a "run once" (so even if invoked multiple times - checks only once)
- keep it at CLI level where it is
- add to top level `DandiAPIClient.__init__` 

@satra - do you foresee any other aspects I forgot, or locations where to do the check? e.g. `organize` in principle also a worthy location to perform the check, but since used most likely through CLI -- should be covered.

jwodder:
Closes #726.

### DIFF
--- a/dandi/cli/command.py
+++ b/dandi/cli/command.py
@@ -122,10 +122,13 @@ def main(ctx, log_level, pdb=False):
 
     if pdb:
         map_to_click_exceptions._do_map = False
-        from ..utils import check_dandi_version, setup_exceptionhook
+        from ..utils import setup_exceptionhook
 
         setup_exceptionhook()
-        check_dandi_version()
+
+    from ..utils import check_dandi_version
+
+    check_dandi_version()
 
 
 #
--- a/dandi/cli/command.py
+++ b/dandi/cli/command.py
@@ -122,32 +122,10 @@ def main(ctx, log_level, pdb=False):
 
     if pdb:
         map_to_click_exceptions._do_map = False
-        from ..utils import setup_exceptionhook
+        from ..utils import check_dandi_version, setup_exceptionhook
 
         setup_exceptionhook()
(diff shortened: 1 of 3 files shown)


dandi/cli/command.py
```diff
@@ -122,32 +122,10 @@ def main(ctx, log_level, pdb=False):
...
-                "version (e.g., using pip install --upgrade dandi if you installed using pip). "
-                "If you have any issues, please contact the DANDI "
-                "helpdesk at https://github.com/dandi/helpdesk/issues/new/choose ."
-            )
-    except Exception as exc:
-        lgr.warning(
-            "Failed to check for a more recent version available with etelemetry: %s",
-            exc,
-        )
+        check_dandi_version()
```
yarikoptic, line 128: I think indentation is wrong and it falls under `if pdb:`

yarikoptic: yeap, import makes it obvious... shouldn't be just in pdb mode

jwodder: Fixed.

yarikoptic:
Thank you @jwodder -- let's proceed and see where it brings us

PR #733, opened 2021-07-22
RF: attempt nwb metadata extraction only on .nwb, if fails -- warning

### LINKED ISSUE #730: adjust nwb metadata extraction only for nwb files
using encodingFormat only do metadata extraction for nwb files. right now, for non-nwb datasets, the logs are filled with errors about unable to open nwb file. to the user that seems unnecessary.

yarikoptic:
I think time to assume that all NWB files would have .nwb extension I think.

This PR is maintaining prior behavior of re-raising the exception if allow_any_path is
False.  Otherwise -- and regardless if allow_any_path is True or not (thus
would not re-raise for non .nwb files even if not allow_any_path) - provide default
metadata record.

For upload it should be fine AFAIK even with the change  of behavior since we first
use allow_any_path to select .nwb files.

Closes #730

### DIFF
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -711,8 +711,8 @@ def get_asset_metadata(
                 type(e).__name__,
                 str(e),
             )
-        if not allow_any_path:
-            raise
+            if not allow_any_path:
+                raise
     if metadata is None:
         metadata = get_default_metadata(
             filepath, digest=digest, digest_type=digest_type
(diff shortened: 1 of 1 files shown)


dandi/metadata.py
```diff
@@ -700,21 +700,23 @@ def process_ndtypes(asset, nd_types):
...
+            metadata = nwb2asset(filepath, digest=digest, digest_type=digest_type)
+        except Exception as e:
+            lgr.warning(
+                "Failed to extract NWB metadata from %s: %s: %s",
+                filepath,
+                type(e).__name__,
+                str(e),
             )
-        else:
+        if not allow_any_path:
```
jwodder, line 714: This needs to be indented one level.

jwodder (changes requested):

dandi/metadata.py
satra: +1 to this. otherwise LGTM

jwodder (approved):

PR #734, opened 2021-07-23
Retry following redirects that return 404

### LINKED ISSUE #727: retry redirect if 400
from https://github.com/dandi/dandi-cli/pull/725/checks?check_run_id=3135699904

```
_ test_download_000027_metadata_only[https://dandiarchive.org/dandiset/000027/draft] _
395
dandi/tests/test_download.py:65: in test_download_000027_metadata_only
396
    ret = download(url, tmpdir, get_assets=False)
397
dandi/download.py:59: in download
398
    parsed_url = parse_dandi_url(urls[0])
399
dandi/dandiarchive.py:378: in parse
400
    new_url = cls.follow_redirect(url)
401
dandi/dandiarchive.py:482: in follow_redirect
402
    f"Response for getting {url} to redirect returned {r.status_code}."
403
E   dandi.exceptions.FailedToConnectError: Response for getting https://dandiarchive.org/dandiset/000027/draft to redirect returned 400. Please verify that it is a URL related to dandiarchive and supported by dandi client
```

attn @satra -- may be smth to be tuned up on redirector.  I think this is not first/unique incident

jwodder:
Closes #727.

### DIFF
--- a/dandi/dandiarchive.py
+++ b/dandi/dandiarchive.py
@@ -1,6 +1,7 @@
 from abc import ABC, abstractmethod
 from contextlib import contextmanager
 import re
+from time import sleep
 from typing import Iterator, Optional, Tuple
 from urllib.parse import unquote as urlunquote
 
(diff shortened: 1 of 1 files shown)


dandi/dandiarchive.py
```diff
@@ -474,18 +475,25 @@ def parse(cls, url, *, map_instance=True):
...
-            raise FailedToConnectError(
-                f"Response for getting {url} to redirect returned {r.status_code}."
-                f" Please verify that it is a URL related to dandiarchive and"
-                f" supported by dandi client"
-            )
-        elif r.url != url:
-            return r.url
-        return url
+        i = 0
+        while True:
```
yarikoptic, line 479: why custom loop, I thought that we have `try_multiple` using `tenacity` for such purposes?

yarikoptic:
note: restarted tests since some were failing for some `git` reasons

dandi/dandiarchive.py
jwodder: `try_multiple()` doesn't support the 0.1, 1, 10 sequence of sleep times that you requested.

yarikoptic: it could be any reasonable sequence it supports, I do not care about precise timing, just spread it so it tries for up to 10ish seconds

jwodder: I feel that, in this case, the code is simpler without involving `try_multiple()`.

yarikoptic: ok, I will resort to your judgement, lets proceed, thank you!

PR #741, opened 2021-07-26
Rename get_assets_under_path() to get_assets_with_path_prefix()

### LINKED ISSUE #738: rename get_assets_under_path ?
inspired by my confusion while reviewing https://github.com/dandi/dandi-cli/pull/676/files#r675083336 .

"under path" for me sounds like "under a directory" not "starting with the path prefix", as even docstring says
```
    def get_assets_under_path(self, path: str) -> Iterator["RemoteAsset"]:
        """
        Returns an iterator of all assets in this version of the Dandiset whose
        `~RemoteAsset.path` attributes start with ``path``
        """
```

and actual endpoint performs:

```
$> curl -X GET "https://api.dandiarchive.org/api/dandisets/000027/versions/draft/assets/?path=sub" -H  "accept: application/json"
{"count":1,"next":null,"previous":null,"results":[{"asset_id":"1c095f5f-d1e2-45db-b807-fdcfea08c6de","path":"sub-RAT123/sub-RAT123.nwb","size":18792,"created":"2021-06-29T22:59:29.678450Z","modified":"2021-06-29T22:59:29.746868Z"}]}%        
```

so I wonder if there could/should be a clearer name or may be even do not bother with a separate method but rather add a `path` (or `path_prefix`) kwarg to `get_assets` with some docstring to describe it?

jwodder:
Closes #738.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -638,7 +638,7 @@ class RemoteDandiset:
             self.client.get(f"{self.version_api_path}assets/{asset_id}/")
         )
 
-    def get_assets_under_path(self, path: str) -> Iterator["RemoteAsset"]:
+    def get_assets_with_path_prefix(self, path: str) -> Iterator["RemoteAsset"]:
         """
         Returns an iterator of all assets in this version of the Dandiset whose
         `~RemoteAsset.path` attributes start with ``path``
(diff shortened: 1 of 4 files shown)


yarikoptic:
found no hits on it being used elsewhere with https://github.com/search?q=org%3Adandi+get_assets_under_path&type=code so I guess should be "safe" to do the rename.  
I do like the more explicit name but I guess decision on https://github.com/dandi/dandi-api/issues/444 , e.g. if we eventually add more options etc, might need to be reflected in dandi-cli, without causing breeding dedicated functions for merely parametrizations of the same `get_assets` (or some `get_assets_filtered` or `get_assets_by_path`) invocation.
But since `_under_path` is indeed too misleading let's just proceed with this rename for now.

thank you @jwodder

PR #742, opened 2021-07-26
Correctly set User-Agent for client requests

jwodder:
`requests.Session` objects start out with a `User-Agent` header already set, and so trying to set the header with `setdefault()` (which was chosen in order to not override a header set by the user on creating the Session object or set in `headers`) does nothing.


yarikoptic:
looks LGTM, thank you @jwodder

PR #744, opened 2021-07-27
Pin rabbitmq version

dchiquito:
RabbitMQ just released version 3.9.0, and it's not playing nice with
Celery. To get tests passing again, pin to version 3.8.

### DIFF
--- a/dandi/tests/data/dandiarchive-docker/docker-compose.yml
+++ b/dandi/tests/data/dandiarchive-docker/docker-compose.yml
@@ -110,6 +110,6 @@ services:
       retries: 5
 
   rabbitmq:
-    image: rabbitmq:management
+    image: rabbitmq:3.8-management
     ports:
       - "5672:5672"


jwodder (approved):

yarikoptic:
oh hoh! nice digging @dchiquito ! we should recommend downstream testing against dandi-cli to rabbitmq as well ;-)

yarikoptic:
@dchiquito 
Any pinning needed/done at dandi-api?
Is issue known/reported to rabbitmq?

dchiquito:
Heroku uses a managed deployment of rabbitmq, so we don't have to sweat about the deployments. If anyone reports problems with dev or test environments then we can fix it, but I haven't seen anything so I vote not doing anything there.

I couldn't find any existing issues, and I haven't had time yet to write up a proper bug report myself.

dchiquito:
Heroku uses a managed deployment of rabbitmq, so we don't have to sweat about the deployments. If anyone reports problems with dev or test environments then we can fix it, but I haven't seen anything so I vote not doing anything there.

I couldn't find any existing issues, and I haven't had time yet to write up a proper bug report myself.

yarikoptic:
> Heroku uses a managed deployment of rabbitmq, so we don't have to sweat about the deployments.

which version does it use?  wouldn't we end up in the same situation after they decide to upgrade it, and shouldn't we prevent that (now that we know that there is an issue)?

dchiquito:
The cloudamqp managed version is 3.8.8, which is way behind the final 3.8 release 3.8.18. If they decide to bump the version and it breaks, we are paying them so we can expect reasonable support.

PR #746, opened 2021-07-29
Test against rc/2.0.0 branch of pynwb

### LINKED ISSUE #745: Establish a CI run testing against rc/2.0.0 branch of pynwb
as asked for by @satra in slack, so we get no surprises whenever new release comes out (oh well, ideal use case for https://github.com/dandi/nwb-healthstatus but we haven't finished that one yet)

I think whatever fixes in the code we might need to do, should not require a dedicated branch and should be done in the `master` to accomodate old/newer version of pynwb.

jwodder:
Closes #745.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,6 +37,9 @@ jobs:
             mode: dev-deps
           - os: ubuntu-18.04
             python: 3.7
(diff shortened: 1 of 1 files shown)


yarikoptic:
I think it does the drill. Thank you @jwodder !

PR #753, opened 2021-08-20
Give API resource classes `__str__` methods

### LINKED ISSUE #751: provide user friendly __str__ for new API classes
An extract from larger #731 to break it apart.

ATM `print(dandiset)` instance results in `<dandi.dandiapi.RemoteDandiset object at 0x7fae2329e1f0>` which gives user only information that it is a remote dandiset, not clear from where and which dataset it is really.
 
May be make it into e.g. f`{id}` from metadata so would look like `DANDI:000003/draft` for one  coming from DANDI.  Re `DANDI-STAGING:` etc -- let's address at dandischema level first (https://github.com/dandi/dandischema/issues/76).

Similar things for assets etc.

jwodder:
Closes #751.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -293,6 +293,22 @@ class DandiAPIClient(RESTFullAPIClient):
                     lgr.debug("Stored key in keyring")
                 break
 
+    @property
+    def instance_name(self) -> str:
+        url = self.api_url.rstrip("/")
+        try:
+            return known_instances_rev[url]
+        except KeyError:
+            return url
+
+    @property
+    def upper_instance_name(self) -> str:
+        url = self.api_url.rstrip("/")
+        try:
+            return known_instances_rev[url].upper()
+        except KeyError:
+            return url
+
     def get_dandiset(
         self, dandiset_id: str, version_id: Optional[str] = None, lazy: bool = True
     ) -> "RemoteDandiset":
@@ -408,7 +424,7 @@ class RemoteDandiset:
         self._data = data
 
     def __str__(self) -> str:
-        return f"DANDI:{self.identifier}/{self.version_id}"
+        return f"{self.client.upper_instance_name}:{self.identifier}/{self.version_id}"
 
     def _get_data(self) -> Dict[str, Any]:
         if self._data is None:
@@ -892,7 +908,7 @@ class BaseRemoteAsset(APIBase):
         self._metadata = data.get("metadata", data.get("_metadata"))
 
     def __str__(self) -> str:
-        return self.identifier
+        return f"{self.client.upper_instance_name}:assets/{self.identifier}"
 
     @classmethod
     def _from_metadata(
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -294,15 +294,7 @@ class DandiAPIClient(RESTFullAPIClient):
                 break
 
     @property
-    def instance_name(self) -> str:
-        url = self.api_url.rstrip("/")
-        try:
-            return known_instances_rev[url]
-        except KeyError:
-            return url
-
-    @property
-    def upper_instance_name(self) -> str:
+    def _instance_id(self) -> str:
         url = self.api_url.rstrip("/")
         try:
             return known_instances_rev[url].upper()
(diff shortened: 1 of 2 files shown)


yarikoptic:
Please add some basic tests to ensure we get what we expect and don't kaboom while simply str'ing those objects

jwodder:
@yarikoptic Done.

dandi/tests/test_dandiapi.py
```diff
@@ -59,6 +59,7 @@ def test_publish_and_manipulate(local_dandi_api, monkeypatch, tmp_path):
         },
     )
     dandiset_id = d.identifier
+    assert str(d) == f"DANDI {dandiset_id}/draft"
```
yarikoptic, line 62: hm, how do we get a space there?  I thought it would be just what we have in metadata
```
$> curl --silent https://api.dandiarchive.org/api/dandisets/000003/versions/0.210812.1448/ | jq .identifier               
"DANDI:000003"
$> curl --silent https://api.dandiarchive.org/api/dandisets/000003/versions/draft/ | jq .identifier     
"DANDI:000003"
```
where the divergence comes from?

dandi/dandiapi.py
```diff
@@ -404,6 +407,9 @@ def __init__(
             self._version = version
         self._data = data
 
+    def __str__(self) -> str:
+        return f"DANDI {self.identifier}/{self.version_id}"
```
yarikoptic, line 411:
```suggestion
        return f"DANDI:{self.identifier}/{self.version_id}"
```
to mimic
```
$> curl --silent https://api.dandiarchive.org/api/dandisets/000003/versions/draft/ | jq .identifier     
"DANDI:000003"
```
but if we are to mint it ourselves (I guess to avoid fetching metadata which should have that `identifier`) -- do we have access to the instance name? then do not hardcode `DANDI` but use `instance_name.upper()`

dandi/dandiapi.py
```diff
@@ -885,6 +891,9 @@ def __init__(self, **data: Any) -> None:
         # underscores, so we have to do it ourselves.
         self._metadata = data.get("metadata", data.get("_metadata"))
 
+    def __str__(self) -> str:
+        return self.identifier
```
yarikoptic, line 895: if we have access to instance name via `.client`, please make it `{instance_name.upper()}:assets/{self.identifier}`

jwodder: @yarikoptic The client only stores the API URL.  While `known_instances_rev` can be used to look up a corresponding instance name, the URL could be anything, not just a known URL.

yarikoptic: If unknown url, just use entire url for now and the name. At least this way we/user would know exactly about what that thing is from

yarikoptic: For all pragmatic use cases I expect us to get a nice and meaningful name

dandi/dandiapi.py

dandi/dandiapi.py
```diff
@@ -293,6 +293,22 @@ def dandi_authenticate(self):
...
+        except KeyError:
+            return url
+
+    @property
+    def upper_instance_name(self) -> str:
+        url = self.api_url.rstrip("/")
+        try:
+            return known_instances_rev[url].upper()
+        except KeyError:
+            return url
```
yarikoptic, line 310:
```suggestion
    @property
    def _instance_id(self) -> str:
        url = self.api_url.rstrip("/")
        try:
            return known_instances_rev[url].upper()
        except KeyError:
            return url
```
and use that protected `_instance_id`  . Rationale:

- we use `instance_name`. Here it might also be a url, so a different beast, let's not occlude
- do not see a need for it to be "public" for now, since we might want to change what it is later on etc

yarikoptic:
looks good

failing test
```
dandi/dandiarchive.py:532: in follow_redirect
    raise FailedToConnectError(
E   dandi.exceptions.FailedToConnectError: Response for getting https://bit.ly/dandi12 to redirect returned 400. Please verify that it is a URL related to dandiarchive and supported by dandi client
```
is unrelated ... we are retrying on 404 (only) there ... I wonder if we should retry on 400 ?

PR #754, opened 2021-08-20
Remove test against rc/2.0.0 branch of pynwb

jwodder:
pynwb 2.0.0 was released on the 13th, and the rc/2.0.0 branch is no more.


yarikoptic:
Thank you!

PR #756, opened 2021-08-25
Add module docstring to dandiapi.py

### LINKED ISSUE #755: provide main modules docstring describing their purpose, giving some example
Part of #731

example could be a "poor man search + get" along the lines of
- go through all dandisets
- list assets which have `two-photon` in their `metadata.measurementTechnique[].name`
- download those assets (under the path `{dandiset}/{path_within_dandiset}`

jwodder:
Closes #755.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -10,12 +10,9 @@ Example code for downloading all assets from all published Dandisets that have
 
     with DandiAPIClient.for_dandi_instance("dandi") as client:
         for dandiset in client.get_dandisets():
-            published_versions = [
-                v for v in dandiset.get_versions() if v.identifier != "draft"
-            ]
-            if not published_versions:
+            latest_version = dandiset.most_recent_published_version
+            if latest_version is None:
                 continue
-            latest_version = max(published_versions, key=attrgetter("created"))
             latest_dandiset = dandiset.for_version(latest_version)
             for asset in latest_dandiset.get_assets():
                 metadata = asset.get_metadata()
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -8,28 +8,33 @@ objects can, in turn, be used to retrieve `RemoteAsset` objects (representing
 assets associated with Dandisets).  Aside from `DandiAPIClient`, none of these
 classes should be instantiated directly by the user.
 
-Example code for downloading all assets from all published Dandisets that have
-"two-photon" in their ``metadata.measurementTechnique[].name``:
+Example code for printing the metadata of all assets with "two-photon" in their
+``metadata.measurementTechnique[].name`` for the latest published version of
+every Dandiset:
 
-.. code:: python
-
-    from pathlib import Path
-    from dandi.dandiapi import DandiAPIClient
-
-    with DandiAPIClient.for_dandi_instance("dandi") as client:
-        for dandiset in client.get_dandisets():
-            if dandiset.most_recent_published_version is None:
-                continue
-            latest_dandiset = dandiset.for_version(
-                dandiset.most_recent_published_version
-            )
-            for asset in latest_dandiset.get_assets():
-                metadata = asset.get_metadata()
-                if any(
-                    mtt is not None and "two-photon" in mtt
-                    for mtt in (metadata.measurementTechnique or [])
-                ):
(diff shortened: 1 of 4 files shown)


dandi/dandiapi.py
```diff
@@ -1,3 +1,27 @@
...
+
+.. code:: python
+
+    from pathlib import Path
+    from dandi.dandiapi import DandiAPIClient
+
+    with DandiAPIClient.for_dandi_instance("dandi") as client:
+        for dandiset in client.get_dandisets():
+            for version in dandiset.get_versions():
+                ds = dandiset.for_version(version)
```
yarikoptic, line 13: let's not go through all versions, but rather consider only released dandisets and take the latest released version and not include it in the resultant download path -- user would not need all versions (possibly identical) of the same file

dandi/dandiapi.py
```diff
@@ -374,7 +402,11 @@ class Config:
 
 
 class Version(APIBase):
-    """The version information for a Dandiset retrieved from the API"""
+    """
+    The version information for a Dandiset retrieved from the API.
+
+    This class should not be instantiated directly.
```
yarikoptic, line 408: please extend with an example on how should it be instantiated

dandi/dandiapi.py
```diff
@@ -1,3 +1,31 @@
...
+    from dandi.dandiapi import DandiAPIClient
+
+    with DandiAPIClient.for_dandi_instance("dandi") as client:
+        for dandiset in client.get_dandisets():
+            published_versions = [
+                v for v in dandiset.get_versions() if v.identifier != "draft"
+            ]
+            if not published_versions:
+                continue
+            latest_version = max(published_versions, key=attrgetter("created"))
```
yarikoptic, line 18: don't we expose somehow the `most_recent_published_version` API provides for the dandiset record?

dandi/dandiapi.py
jwodder: End-users should only retrieve instances from the properties & methods that return `Version` objects; do you want me to list every single one?

dandi/dandiapi.py
jwodder: Yes; forgot about it.

dandi/dandiapi.py
yarikoptic: No, but at least one with a description of which class instances have methods to get instances

dandi/dandiapi.py
```diff
@@ -1,3 +1,29 @@
+"""
+Example code for downloading all assets from all published Dandisets that have
+"two-photon" in their ``metadata.measurementTechnique[].name``:
+
+.. code:: python
+
+    from operator import attrgetter
```
yarikoptic, line 7:
```suggestion
```
since no longer used.

Please make this code snippet/docstring  doc tested so we could ensure it stays valid.

With that let's finalize this PR and let's get it merged.

jwodder: @yarikoptic The code downloads assets.  How is that supposed to be safely doctested (to say nothing of how long it could take to run the code)?

yarikoptic: d'oh , right.  Lets may be comment out the `download` line (with a comment to use it if download is desired) and just print metadata?  I just feel that having at least majority of the docstring sample tested would be of great benefit long term

jwodder: @yarikoptic Doctests work by comparing the code's output to the output given in the doctest.  Aside from the fact that most Dandiset metadata is multiple lines long even when not formatted, we would have to update the doctest whenever a new version is published.

yarikoptic: could we use ` # doctest:+ELLIPSIS` and have `...` to swallow all output?

jwodder: @yarikoptic That works, but it adds about 50 seconds to the total test runtime.

yarikoptic: that is indeed notable.  Could we may be run such a full sweep with doc-tests only on cron (e.g. daily)?

jwodder: @yarikoptic Done.

yarikoptic: Looks good to me.  Time to undraft and merge?

yarikoptic:
Great, Thank you, let's proceed.

PR #760, opened 2021-08-31
Add dandischema into fscacher tokens for validate

yarikoptic:
So we do not use outdated results if dandischema was upgraded
without dandi-cli upgrade


jwodder (approved):

yarikoptic:
tests fails seems to be due to redirect fails (504, attn @satra), filed https://github.com/dandi/dandi-cli/issues/761 . Let's proceed

PR #762, opened 2021-08-31
Retry following redirects on 400 (not 404), 502, 503, & 504 up to 5 times

jwodder:
Closes #761.

PR #763, opened 2021-08-31
Use new published version of 000027 in tests

### LINKED ISSUE #261: Update to use api.dandiarchive.org instead of publish.dandiarchive.org
publish.dandiarchive.org was retired at some point recently, and now api.dandiarchive.org should be used instead.  This change caused several tests to fail; all of them are in `dandi/tests/test_download.py` and involve the URL <https://dandiarchive.org/dandiset/000027/0.200721.2222>.  For now, I have marked the failing tests with `pytest.mark.xfail(reason="publish.dandiarchive.org is gone")` so that they will not cause the test suite to fail.

jwodder:
Closes #261.

### DIFF
--- a/dandi/tests/test_download.py
+++ b/dandi/tests/test_download.py
@@ -16,10 +16,7 @@ from ..utils import find_files
 @pytest.mark.parametrize(
     "url",
     [  # Should go through API
-        pytest.param(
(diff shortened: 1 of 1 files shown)


yarikoptic:
Thank you! Failing test is 400 from redirect, #762 yet to be merged

PR #767, opened 2021-09-02
Stop uploading a file if no metadata can be extracted

jwodder:


yarikoptic:
Since quite surprising and "core", please add a unittests (probably a fixture to produce such broken nwb) to ensure that we don't even try to upload then, even upon rerun (thus mimicking scenario in original report which isn't yet troubleshooted into full clarity)

jwodder:
@yarikoptic Test added.  It only calls `upload()` once, though, as I really don't think calling it twice is ever going to make a difference.

yarikoptic:
thank you @jwodder!  I thought to toy with it a bit (still hopeful to replicate) but since it fixes a bug, lets proceed with merge and release!

PR #775, opened 2021-09-27
Update dandischema version

jwodder:


jwodder:
The tests are failing because the dandi-api code doesn't yet support the new schema version, and so attempts to publish Dandisets are failing.

yarikoptic:
ufff... these tight dependencies keep making it "too rigid"... clearly we don't want to merge/release until dandi-api main deployment updates to newer version.  In the longer run I think we should start supporting multiple (two) "minor" versions of dandischema in dandi-cli, have matrix runs while testing against both, and then skip tests against "next" dandischema version whenever dandi-api is not support it.  But I guess we should not bother with that just yet.

yarikoptic:
per your https://github.com/dandi/dandischema/pull/88#issuecomment-928149691 -- and since we should not merge/release until main (non-staging) instance progresses forward to the next dandischema version, should we at least merge https://github.com/dandi/dandi-cli/commit/024b93be00e98c454063a5de6945d97f2fd77b82 and https://github.com/dandi/dandi-cli/commit/45a4215089699366c4b067d86417b86cd5005c50 interim and keep this PR clearly only boosting dandischema version?

jwodder:
@yarikoptic Commits split off into PRs #776 and #777.

setup.cfg
```diff
@@ -30,7 +30,7 @@ install_requires =
     appdirs
     click
     click-didyoumean
-    dandischema ~= 0.3.1
+    dandischema ~= 0.4.0
```
satra, line 33:
```suggestion
    dandischema ~= 0.4.1
```

yarikoptic:
@satra - merge to release ;-)

PR #776, opened 2021-09-27
Fix a test

jwodder:
Split off from https://github.com/dandi/dandi-cli/pull/775

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,7 +9,7 @@ commands =
     # Using pytest-cov instead of using coverage directly leaves a bunch of
     # .coverage.$HOSTNAME.#.# files lying around for some reason
     coverage erase
-    coverage run -m pytest -v {posargs} --doctest-modules dandi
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -9,7 +9,7 @@ commands =
     # Using pytest-cov instead of using coverage directly leaves a bunch of
     # .coverage.$HOSTNAME.#.# files lying around for some reason
     coverage erase
-    coverage run -m pytest -v {posargs} --doctest-modules dandi
+    coverage run -m pytest -v {posargs} dandi
```
yarikoptic, line 12: why no doctest modules any more?  that could leave that example you created for dandiapi without even smoke testing

jwodder: Because, due to the increased number of published Dandisets, it now takes six minutes to run the doctest.  We still test the doctest as part of the scheduled CI test.

yarikoptic: ah, ok -- fair enough. Thanks!

PR #777, opened 2021-09-27
Fix display of Dandiset validation errors

jwodder:
Split off from https://github.com/dandi/dandi-cli/pull/775


yarikoptic:
thank you!

PR #778, opened 2021-09-30
Specify DJANGO_DANDI_VALIDATION_JOB_INTERVAL

dchiquito:
https://github.com/dandi/dandi-api/pull/538 is adding a scheduled job
which checks for Versions that need validation and validates them. The
default wait interval is 60 seconds, but it is configurable through an
environment variable, `DJANGO_DANDI_VALIDATION_JOB_INTERVAL`. Set this
environment variable to `5` so that the validation job runs every 5
seconds, which should be quick enough for `wait_until_valid`.

### DIFF
--- a/dandi/tests/fixtures.py
+++ b/dandi/tests/fixtures.py
@@ -183,6 +183,7 @@ def docker_compose_setup():
     env["DJANGO_DANDI_GIRDER_API_URL"] = "http://localhost:8080/api/v1"
     env["DJANGO_DANDI_GIRDER_API_KEY"] = "abc123"
     env["DJANGO_DANDI_SCHEMA_VERSION"] = DANDI_SCHEMA_VERSION
+    env["DJANGO_DANDI_VALIDATION_JOB_INTERVAL"] = "5"
 
     try:
         if create:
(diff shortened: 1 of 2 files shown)


dandi/tests/fixtures.py
```diff
@@ -183,6 +183,7 @@ def docker_compose_setup():
     env["DJANGO_DANDI_GIRDER_API_URL"] = "http://localhost:8080/api/v1"
     env["DJANGO_DANDI_GIRDER_API_KEY"] = "abc123"
     env["DJANGO_DANDI_SCHEMA_VERSION"] = DANDI_SCHEMA_VERSION
+    env["DJANGO_DANDI_VALIDATION_JOB_INTERVAL"] = "5"
```
jwodder, line 186: There's no need for this change.  The purpose of the `env` dict here is to set environment variables that are left blank in `docker-compose.yml`.  Since `DJANGO_DANDI_VALIDATION_JOB_INTERVAL` is already being set in `docker-compose.yml`, don't set it here.

jwodder (changes requested):

jwodder (approved):

PR #779, opened 2021-09-30
Remove Girder envvars from Docker Compose setup

jwodder:
The Girder environment variables have been removed from dandi-api, so we no longer need to set them here.


yarikoptic:
Thank you @jwodder!

PR #781, opened 2021-10-01
Give `RemoteAsset` a `created` attribute

jwodder:
Motivated by https://github.com/dandi/dandisets/issues/84.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -552,18 +552,6 @@ class RemoteDandiset:
             **data,
         )
 
-    def _mkasset_from_metadata(self, metadata: Dict[str, Any]) -> "RemoteAsset":
-        return RemoteAsset(
-            client=self.client,
-            dandiset_id=self.identifier,
-            version_id=self.version_id,
-            identifier=metadata["identifier"],
-            path=metadata["path"],
-            size=metadata["contentSize"],
-            modified=metadata["dateModified"],
-            _metadata=metadata,
-        )
-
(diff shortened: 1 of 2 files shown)


dandi/dandiapi.py
```diff
@@ -1116,6 +1116,9 @@ class RemoteAsset(BaseRemoteAsset):
     #: The identifier for the version of the Dandiset to which the asset
     #: belongs
     version_id: str
+    #: The date at which the asset was created.  This is not set if the asset
+    #: was returned by `RemoteDandiset.get_asset()`.
```
yarikoptic, line 1120: when is it set?

yarikoptic: should it become some lazy property so it could be queried if not known but requetsed?

jwodder: @yarikoptic It's set when it's returned by any of the other methods that return `RemoteAsset`s.  The only way to query it if not known is by iterating through all the assets.

yarikoptic: There is path argument to dandisets_versions_assets_list (didn't try) - do may be could be limited to just that asset?

jwodder: @yarikoptic The `get_asset()` method now makes a second call to dandiset_versions_assets_list in order to get the value for the `created` field.  It also uses this call to populate the `modified` field, which has been removed from the `BaseRemoteAsset` class, as there it was derived from the metadata, which need not agree with the asset listing value.

PR #782, opened 2021-10-04
Fix an AttributeError raised when a URL is used as a species

jwodder:


yarikoptic:
Thank you @jwodder! Could you please adjust the title to summarize actual change/fix so it could be self sufficient in CHANGELOG etc.

jwodder:
@yarikoptic Adjusted.

yarikoptic:
cool, thank you! Let's proceed. I will not rush to release ATM

PR #783, opened 2021-10-04
Give `validate` command an `--allow-any-path` option

### LINKED ISSUE #780: NWB metadata errors not caught by validate
Using version 0.27.2
We've gotten several NWB metadata related errors on upload that are not caught by running dandi validate locally beforehand:
AttributeError: 'NoneType' object has no attribute 'capitalize'
ValueError: ISO 8601 expected, but P was received
We've tracked these down or followed up on them separately, but the fundamental issue here is that `dandi validate` doesn't seem to actually run all of the validation that is run on upload, and so doesn't catch many errors.

jwodder:
Closes #780.

### DIFF
--- a/dandi/validate.py
+++ b/dandi/validate.py
@@ -5,7 +5,7 @@ from .consts import dandiset_metadata_file
 from .metadata import get_metadata
 from .pynwb_utils import validate as pynwb_validate
 from .pynwb_utils import validate_cache
-from .utils import find_dandi_files, yaml_load
+from .utils import find_dandi_files, find_files, yaml_load
 
 lgr = get_logger()
 
@@ -15,7 +15,7 @@ _required_nwb_metadata_fields = ["subject_id"]
 
 
 # TODO: provide our own "errors" records, which would also include warnings etc
-def validate(paths, schema_version=None, devel_debug=False):
+def validate(paths, schema_version=None, devel_debug=False, allow_any_path=False):
     """Validate content
 
     Parameters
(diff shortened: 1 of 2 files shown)


dandi/validate.py
```diff
@@ -28,7 +28,8 @@ def validate(paths, schema_version=None, devel_debug=False):
     path, errors
       errors for a path
     """
-    for path in find_dandi_files(paths):
+    filepaths = find_files(".*", paths) if allow_any_path else find_dandi_files(paths)
```
yarikoptic, line 31: FWIW, I thought to suggest adding `exclude_datalad=True` but since we already default to `exclude_dotdirs = True` I guess that doesn't matter.

yarikoptic:
need to think whether we would like to make `validate` consistent with `upload` and gain devel option `--allow-any-path` (so not just always do `allow_any_path=True`)... otherwise we are making `validate` and `upload` not behave analogously and that could confuse folks.  WDYT?

jwodder:
@yarikoptic Done.

yarikoptic:
Thank you @jwodder ... let's proceed with this and release to please dandischema testing with all the fixes we piled up.

yarikoptic:
note -- there is one failing test run (ubuntu-18.04, 3.7, dev-deps) with
```
=================================== FAILURES ===================================
_____________ test_download_000027_resume[0.210831.2033-<lambda>2] _____________
dandi/tests/test_download.py:90: in test_download_000027_resume
    download(url, tmp_path, get_metadata=False)
dandi/download.py:59: in download
    parsed_url = parse_dandi_url(urls[0])
dandi/dandiarchive.py:421: in parse
    new_url = cls.follow_redirect(url)
dandi/dandiarchive.py:534: in follow_redirect
    raise NotFoundError(url)
E   dandi.exceptions.NotFoundError: https://dandiarchive.org/dandiset/000027/0.210831.2033
------------------------------ Captured log call -------------------------------
DEBUG    dandi:dandiarchive.py:402 Parsing url https://dandiarchive.org/dandiset/000027/0.210831.2033
```
but it can't relate to this PR, so proceeding

PR #786, opened 2021-10-05
Log a message if a request succeeds after one or more failures

### LINKED ISSUE #784: INFO upon successful operation after earlier WARNING
Screenshot_20211005-122632

jwodder:
Tenacity was unable to do this, so I got rid of it and wrote the retrying code myself.

Closes #784.

### DIFF
--- a/dandi/utils.py
+++ b/dandi/utils.py
@@ -9,7 +9,6 @@ except ImportError:
 import inspect
 import io
 import itertools
-import logging
 from mimetypes import guess_type
 import os
 import os.path as op
@@ -20,13 +19,12 @@ import shutil
 import subprocess
 import sys
 import types
-from typing import Optional, Union
+from typing import Iterator, Optional, Union
 
 import dateutil.parser
 import requests
 import ruamel.yaml
 from semantic_version import Version
-import tenacity
 
 #
 # Additional handlers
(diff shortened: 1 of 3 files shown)


dandi/utils.py
```diff
@@ -616,23 +614,18 @@ def get_instance(dandi_instance_id):
         )
 
 
-def try_multiple(ntrials, retry, base):
```
yarikoptic, line 619: we didn't use this in any downstream code -- may be better should go through a round of deprecation?

yarikoptic (commented):
> Tenacity was unable to do this, so I got rid of it and wrote the retrying code myself.

according to https://tenacity.readthedocs.io/en/latest/#retrying-code-block  `Retrying` could be used as an iterator, so it might be just a matter of wrapping that one up?

dandi/utils.py
```diff
@@ -616,23 +614,18 @@ def get_instance(dandi_instance_id):
...
-        stop=tenacity.stop_after_attempt(ntrials),
-        before_sleep=tenacity.before_sleep_log(lgr, logging.WARNING),
-        reraise=True,
-    )
+def exp_wait(
+    base: float = 1.25,
+    multiplier: Optional[float] = None,
+    attempts: Optional[int] = None,
+) -> Iterator[float]:
+    if multiplier is None:
```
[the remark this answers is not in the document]
jwodder: @yarikoptic The function call occurs, but because it's a generator, nothing is run unless something needs to be retried.  Documentation added.

dandi/utils.py
jwodder: @yarikoptic None of the other repositories that I have cloned use this function.  I don't see the point in deprecating it instead of just deleting it immediately, and I think we should consider `utils.py` to be entirely an implementation detail that can be modified at will.

yarikoptic:
> > according to https://tenacity.readthedocs.io/en/latest/#retrying-code-block `Retrying` could be used as an iterator, so it might be just a matter of wrapping that one up?
> 
> If we go that way, the log messages for the failed tries will all be of the form "Retrying None in x seconds as it raised ..." without even showing the name of the function that failed. In contrast, with this PR, the log messages will now include the failing URL.

I meant -- use tenacity simply as a "repeater" so that example elaborated for our case:

```python
from tenacity import Retrying, RetryError, stop_after_attempt

try:
    for attempt in Retrying(stop=stop_after_attempt(3)):
        with attempt:
            if attempt.retry_state.attempt_number == 1:
               print("warning: failing")
               raise Exception('My code is failing!')
            print("my call succeeded")

    if attempt.retry_state.attempt_number > 1:
        print("info: did succeed eventually")
except RetryError:
    pass
```

PR #787, opened 2021-10-05
Register all celery queues with dandiarchive env

dchiquito:
There are now multiple queues to consume from in dandi-api, so the test
environment celery worker needs to explicitly handle all of them.

Requires https://github.com/dandi/dandi-api/pull/551 to be approved first.


jwodder:
The test failures are due to a new version of joblib being released that broke fscacher.  I'm going to release a new version that limits the joblib dependency version and then rerun the tests.

yarikoptic:
added `release` label and will merge to unblock dandi-api

PR #790, opened 2021-10-06
Change docs to use "furo" theme

jwodder:
I don't like Alabaster's lack of spacing between consecutive classes, so I thought I'd suggest switching to the latest hotness, [furo](https://github.com/pradyunsg/furo).

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,2 +1,2 @@
-alabaster
+furo
 Sphinx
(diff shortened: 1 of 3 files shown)


yarikoptic (approved):
looks good to me

yarikoptic:
ok, let's proceed with docs tune up PRs. Thank you @jwodder

PR #792, opened 2021-10-06
Log a message if a request succeeds after one or more failures [take 2]

### LINKED ISSUE #784: INFO upon successful operation after earlier WARNING

jwodder:
This is an alternative to #786 that retains the use of tenacity.

Closes #784.


yarikoptic (changes requested):
it looks all good and kosher, and I think I like this one over take 1, but could you please add a test to at least smoke test this logic (or ideally also to verify that we do log that INFO whenever desired)?

PR #794, opened 2021-10-06
Test against Python 3.10

jwodder:
Python 3.10 was released on Monday, so we need to support it.

Note that the issue with installing h5py on 3.9 is gone, as pynwb and hdmf have relaxed their version requirements to work with the latest version of h5py, which provides wheels for 3.9.  However, at time of writing, the latest h5py does not provide wheels for 3.10, so I just changed the h5py handling from 3.9 to 3.10.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,8 +53,9 @@ jobs:
         # Fetch all commits so that versioneer will return something compatible
         # with semantic-version
         fetch-depth: 0
+
     - name: Set up Python ${{ matrix.python }}
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python }}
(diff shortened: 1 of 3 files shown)


yarikoptic:
3.10 tests seems to fail, so I guess there might be more todo

yarikoptic:
Then let's wait, may be make it into draft pr for now

jwodder:
@yarikoptic dandi's dependencies all seem to support Python 3.10 now.

yarikoptic:
woohoo -- coolio.  thanks! Just one conflict to resolve

jwodder:
@yarikoptic Conflict resolved.  The one failing job is due to an inability to connect to dandiarchive.org.

yarikoptic:
coolio, thanks!  Let's proceed then.

PR #797, opened 2021-10-11
Apply "skipif_no_network" to all relevant tests

### LINKED ISSUE #796: Network sandboxing for tests doesn't work, in spite of `DANDI_TESTS_NONETWORK=1`
I get the following failures while running the tests in the network sandbox: https://ppb.chymera.eu/8a10b2.log
This is with `DANDI_TESTS_NONETWORK=1` exported in the environment, but a lot of tests using `curl` etc. don't seem to be skipped.

In any case, not high priority, no tests fail if run as user now, but it would be good to have this sorted out so we can distribute dandi-cli it for Gentoo.

Maybe there's a better way to skip them (other than skipping them manually, I mean :) )?

jwodder:
Closes #796.

### DIFF
--- a/dandi/cli/tests/test_ls.py
+++ b/dandi/cli/tests/test_ls.py
@@ -5,6 +5,8 @@ from click.testing import CliRunner
 from dandischema.consts import DANDI_SCHEMA_VERSION
 import pytest
 
+from dandi.tests.skip import mark
+
 from ..command import ls
 from ...utils import yaml_load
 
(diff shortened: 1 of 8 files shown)


jwodder:
@yarikoptic Is there any reason not to retry on *every* 5xx status code?

yarikoptic:
> @yarikoptic Is there any reason not to retry on _every_ 5xx status code?

some of [them](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#5xx_server_errors) look like not worth retrying on, e.g. `501 Not Implemented`, `508 Loop Detected (WebDAV; RFC 5842)` (and probably `507 Insufficient Storage`) since I would not expect change in behavior upon retries. But indeed for some others it makes sense:
- `502 Bad Gateway` -- I believe we had those occasionally
- `503 Service Unavailable` -- might come back
- `504 Gateway Timeout` (or is that the one we saw and not 502?)

so lets retry on 500, 502, 503, 504 for now.

PR #799, opened 2021-10-11
Document & type-annotate all public values in `dandi.dandiapi`

jwodder:


dandi/dandiapi.py
```diff
@@ -485,6 +620,7 @@ def _get_data(self) -> Dict[str, Any]:
 
     @property
     def version_id(self) -> str:
+        """The identifier for the version in question of the Dandiset"""
```
yarikoptic, line 623:
```suggestion
        """The identifier for the Dandiset version"""
```
since they're is no question/parameter and other methods don't use such a form

yarikoptic:
Thanks! No outstanding PRs destined for merge soon, and we had no network fix merged after release, so let's release again ;-)

PR #801, opened 2021-10-13
dandi.dandiapi: Convert 404 responses to NotFoundErrors

jwodder:


dandi/dandiapi.py
```diff
@@ -514,11 +520,20 @@ def check_schema_version(self, schema_version: Optional[str] = None) -> None:
...
+        The returned object will not have any information about the Dandiset
+        associated with the asset; for that, the `RemoteDandiset.get_asset()`
+        method must be used instead.
         """
-        return BaseRemoteAsset._from_metadata(self, self.get(f"/assets/{asset_id}"))
+        try:
+            return BaseRemoteAsset._from_metadata(self, self.get(f"/assets/{asset_id}"))
+        except requests.HTTPError as e:
+            if e.response.status_code == 404:
+                raise NotFoundError(f"No such asset: {asset_id}")
```
yarikoptic, line 534: minor, I wonder why use `!r` for dandiset error message but not for this (asset) one?

jwodder:
@yarikoptic That couldn't be a function decorator, as the error message requires too much customization, and I was told on IRC that using a context manager for this would be a bad idea.

yarikoptic:
indeed -- decorator would not work, not sure why I thought it would ;) but what about centralizing it within `.request` itself like here: https://github.com/dandi/dandi-cli/commit/d60396d23db1d6451eb9bc5b74a0338c286e1cbf (untested etc)?

PR #803, opened 2021-10-14
Improve dandiarchive.py docs

jwodder:


dandi/dandiarchive.py
```diff
@@ -370,35 +407,65 @@ class _dandi_url_parser:
...
-        :rtype: ParsedDandiURL
+        - :samp:`https://{server}[/api]/dandisets/{dandiset-id}/versions/{version}/assets/{asset-id}[/download]`
+          — Refers to a single asset and is converted to an `AssetIDURL`
+
+        - :samp:`https://{server}[/api]/dandisets/{dandiset-id}/versions/{version}/assets/?path={path}`
+          — Refers to all assets in the given Dandiset whose paths begin with
+          the prefix ``path``; converted to an `AssetPathPrefixURL`
+
+        - :samp:`dandi://{instance-name}/{dandiset-id}[@{version}][/{path}]`,
+          where ``instance name`` is the name of a registered Dandi Archive
```
yarikoptic, line 458:
```suggestion
          where ``instance-name`` is the name of a registered Dandi Archive
```
?

but also wondered if it wouldn't be better to use python's `_` instead of `-` even though we might not have `dandiset_id` or `instance_name` variables?

yarikoptic (commented):
looks nice.  left only few minor comments for your consideration

PR #804, opened 2021-10-15
Make `--dandi-instance` public and add "instances" command

jwodder:
Seeing as we're planning on instructing users to do testing using the staging server, we should make it easier for them to use that server by eliminating the need to set the `DANDI_DEVEL` environment variable in order to use the `-i`/`--dandi-instance` option.  This PR also further helps out with that by adding an "instances" subcommand for dumping the known instances as YAML.


jwodder:
@yarikoptic Fixed.

PR #812, opened 2021-10-25
Stop using distutils

### LINKED ISSUE #522: Get rid of distutils
[Distutils is deprecated and slated for removal from the Python standard library.](https://www.python.org/dev/peps/pep-0632/)  Currently, dandi-cli uses distutils in one place: using `LooseVersion` to compare pynwb versions.  The recommended replacement for `distutils.version` is [`packaging.version`](https://packaging.pypa.io/en/latest/version.html); however, its equivalent(?) of `LooseVersion`, [`LegacyVersion`](https://packaging.pypa.io/en/latest/version.html#packaging.version.LegacyVersion), is also deprecated.  Finding the proper version class to use depends on just how pynwb versions are structured.

@yarikoptic What do we know about the format of pynwb versions?

jwodder:
Closes #522.

Note that I still don't know what the actual version format is for the version of the NWB standard used by a file.  As long as it's always something of the form `N(.N)*`, this will work, but if things like incrementing alphabetic suffixes are used, this will break.

### DIFF
--- a/dandi/pynwb_utils.py
+++ b/dandi/pynwb_utils.py
@@ -1,5 +1,4 @@
 from collections import Counter
-from distutils.version import LooseVersion
 import os
 import os.path as op
 import re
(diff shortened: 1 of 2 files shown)


yarikoptic:

PR #829, opened 2021-11-05
Convert DANDI: resource identifiers to lowercase when prepending "https://identifiers.org/" and tighten regex

jwodder:
See https://github.com/dandi/helpdesk/issues/37#issuecomment-962037772

### DIFF
--- a/dandi/dandiarchive.py
+++ b/dandi/dandiarchive.py
@@ -363,9 +363,12 @@ class _dandi_url_parser:
         # TODO: Later should better conform to our API, so we could allow
         #       for not only "dandiarchive.org" URLs
         (
-            re.compile(r"DANDI:.*"),
+            re.compile(
+                fr"DANDI:{DANDISET_ID_REGEX}(?:/[0-9]+\.[0-9]+\.[0-9]+)?",
+                flags=re.I,
+            ),
             {"rewrite": lambda x: "https://identifiers.org/" + x.lower()},
-            "DANDI:<dandiset id>",
+            "DANDI:<dandiset id>[/<version id>]",
         ),
         (
             re.compile(r"https?://dandiarchive\.org/.*"),
@@ -373,9 +376,13 @@ class _dandi_url_parser:
             "https://dandiarchive.org/...",
         ),
         (
-            re.compile(r"https?://identifiers\.org/DANDI:.*"),
+            re.compile(
+                fr"https?://identifiers\.org/DANDI:{DANDISET_ID_REGEX}"
+                fr"(?:/[0-9]+\.[0-9]+\.[0-9]+)?",
+                flags=re.I,
+            ),
             {"handle_redirect": "pass"},
-            "https://identifiers.org/DANDI:<dandiset id>",
+            "https://identifiers.org/DANDI:<dandiset id>[/<version id>]",
         ),
         (
             re.compile(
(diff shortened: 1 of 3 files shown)


yarikoptic:
@satra -- do you think it would be a good/kosher thing to do in the long run too?

satra:
> do you think it would be a good/kosher thing to do in the long run too?

yes, unless identifiers.org flips around and creates the opposite error.

satra:
btw, this should trap for draft identifiers. those are not supported by identifiers.org.

yarikoptic:
> btw, this should trap for draft identifiers. those are not supported by identifiers.org.

decision from standup IIRC was: make regex for `DANDI:` identifiers match the one in dandischema which should make it stricter, and proceed.

jwodder:
@yarikoptic @satra When I went to restrict the format of DANDI: identifiers, I found that dandischema and dandi-cli use two different regexes for non-draft version identifiers: dandischema uses `\d+\.\d+\.\d+`, while dandi-cli uses `[.0-9]{5,}`.  Which one should be preferred?

satra:
technically we are using dates and times so ideally: `0\.\d{6}\.\d{4}`, but the current setup is made more general than that with just `\d+` so we don't have to be super strict. also the only thing that can generate this is the API-server on publish, and were reserving the right to alter it.

jwodder:
@satra I'm going to change the regex in dandi-cli to `[0-9]+\.[0-9]+\.[0-9]+`, OK?

satra:
> I'm going to change the regex in dandi-cli to [0-9]+\.[0-9]+\.[0-9]+, OK?

sounds ok but, perhaps just use `\d+`

jwodder:
@satra `\d` accepts Unicode digits outside of `0-9` that we don't want.

dandi/dandiarchive.py
```diff
@@ -363,19 +363,26 @@ class _dandi_url_parser:
         # TODO: Later should better conform to our API, so we could allow
         #       for not only "dandiarchive.org" URLs
         (
-            re.compile(r"DANDI:.*"),
-            {"rewrite": lambda x: "https://identifiers.org/" + x},
-            "DANDI:<dandiset id>",
+            re.compile(
+                fr"DANDI:{DANDISET_ID_REGEX}(?:/[0-9]+\.[0-9]+\.[0-9]+)?",
```
yarikoptic, line 367: Can't `VERSION_REGEX` be used here instead of duplicating regex?

jwodder: @yarikoptic We need to exclude `draft`, which `VERSION_REGEX` includes.

yarikoptic: I guess I would have made this one to not include it, and ANY_VERSION_REGEX or alike to include it?

jwodder: @yarikoptic Do you want me to do that, or are you concerned about breaking backwards compatibility?

yarikoptic: to keep compatible, lets go the other way: add `PUBLISHED_VERSION_REGEX = '[0-9]+\.[0-9]+\.[0-9]+'`?

jwodder: @yarikoptic Done.

yarikoptic:
sorry, forgot about this one. Thank you @jwodder.   Let's proceed -- I think it would make the world a better place ;)  I will even release with it

PR #831, opened 2021-11-09
Use sphinx-click to document commands

jwodder:
This is an experiment to test out the use of sphinx-click for this project.  Compare the results of this PR to the current documentation.

Reasons to use sphinx-click:

* New options are automatically added to the documentation without needing to remember to update the documentation
* Cuts down on duplication of help text

Reasons not to use sphinx-click:

* All help text must be valid reStructuredText, which leads to things like having to add a backslash to the `https://*dandiarchive-org.netflify.app/...` pattern shown in `dandi download --help`.
* We can't use Sphinx markup in command documentation without that markup also appearing in `--help` output, which can confuse users.
* `--help` output is meant to be concise, while online documentation is meant to be thorough.  These two requirements are sometimes at odds, and sphinx-click means we can't use different text for the two.


yarikoptic:
oh, I forgot that doc pages were not autogenerated and you manually produced them.  Sure thing we should automate, even if it would be a bit suboptimal.

- we might want to run doc building with DANDI_DEVEL=1 if we would like to add devel options (as listed now) .  I only wish we could also automagically annotate them somehow (may be the `@devel_option` could expand doc for the option to state that it is `[For development only]`)?
- I do dislike now present `\` in `download --help` .  May be we could just make those strings "verbatim markdown" and avoid needing to backslash?

jwodder:
@yarikoptic What exactly do you mean by "verbatim markdown"?  Keep in mind that, when using sphinx-click, any markup added to the docstrings for the sake of rendering in the Sphinx docs will also show up when running the commands with `--help`.

I should probably state now that I'd prefer not to use sphinx-click, and I'd prefer having the Sphinx docs contain only a single copy of the list of resource patterns, with the command docs linking to the list.

yarikoptic:
> @yarikoptic What exactly do you mean by "verbatim markdown"? Keep in mind that, when using sphinx-click, any markup added to the docstrings for the sake of rendering in the Sphinx docs will also show up when running the commands with `--help`.

E.g. just indenting that listing by four leading spaces. So it would look indented in CLI and for markdown it would be the "verbatim".

> ... with the command docs linking to the list.

not sure about that since it would 
- require user to go online to get some basic information
- that URL might migrate/change and some older client would point to a possibly outdated location

jwodder:
@yarikoptic I meant that the `--help` outputs would continue to include the resource patterns and the `dandi download`, `dandi ls`, and `parse_dandi_url()` docs on dandi.rtfd.io would all link to a dedicated page on dandi.rtfd.io listing the resource patterns.

jwodder:
@yarikoptic 

> E.g. just indenting that listing by four leading spaces. So it would look indented in CLI and for markdown it would be the "verbatim".

Sphinx uses reStructuredText, not Markdown.  Basic indentation of a block in reStructuredText creates a blockquote, while verbatim text requires a `::` marker.

Regardless, when sphinx-click processes the `\b` marker in docstrings that Click uses to disable word wrapping, it seems to disable any markup, and so there does not seem to be a way to get a block to be formatted as-is in both `--help` text and sphinx-click-processed output.

jwodder:
I'm going to close this issue and move forwards with improving the documentation without sphinx-click.

yarikoptic:
I think sooner than later automated generation will be very desired to keep them in sync.

PR #839, opened 2021-11-19
Convert metadata 404's when downloading into an "error" status

### LINKED ISSUE #821: Handle "gracefully" 404ing assets which were ok whenever listing was received
an extract from https://github.com/dandi/dandi-cli/issues/817#issuecomment-954002663 

> since URL could point to some draft version of a huge dandiset (e.g. 000108) which would take weeks to download, and then particular asset might get "GCed" if replaced/changed -- we still can encounter NotFoundError for an asset even though top level URL was Ok. Can't we make it handle such cases properly as well right away?

jwodder:
Based on my reading of the code, there are only two points at which a suddenly-missing asset could cause a 404 while downloading:

* [At the call to `get_raw_metadata()`](https://github.com/dandi/dandi-cli/blob/902f7fdb6b7816dbc9c5a4a806628550702a1fcb/dandi/download.py#L199) when first starting to process an asset; this is the case that this PR addresses.  A 404 here is not as unlikely as you might think at first, as assets are fetched from the API in pages of 25 each, so there could be an arbitrary delay between an asset being initially fetched from the API and its metadata being fetched.
* [When actually performing the download](https://github.com/dandi/dandi-cli/blob/902f7fdb6b7816dbc9c5a4a806628550702a1fcb/dandi/download.py#L549); a 404 at this point already causes the download to stop with status "error", so there is no need to change anything for this case.

Closes #821.

@yarikoptic In the event that you want a test for this, while it is feasible to mock a specific metadata endpoint so that requests to it return 404, exactly what should the tests test for after that?  Should the test just run `download()` with the mock in place and only fail if the download raises an exception?

### DIFF
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -15,6 +15,7 @@ from . import get_logger
 from .consts import RETRY_STATUSES, dandiset_metadata_file
 from .dandiarchive import DandisetURL, MultiAssetURL, SingleAssetURL, parse_dandi_url
 from .dandiset import Dandiset
+from .exceptions import NotFoundError
 from .support.digests import get_digest
 from .support.pyout import naturalsize
 from .utils import (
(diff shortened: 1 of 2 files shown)


yarikoptic:
> @yarikoptic In the event that you want a test for this, while it is feasible to mock a specific metadata endpoint so that requests to it return 404, exactly what should the tests test for after that? Should the test just run `download()` with the mock in place and only fail if the download raises an exception?

sounds like a viable way, but I guess we could just test the `download_generator`'s generator directly and ensure that it doesn't raise but yields error record, right?

jwodder:
@yarikoptic Test added.

yarikoptic:
Thank you!  We haven't released for awhile , so let's release while at it

Issue #840, opened 2021-11-23
detect external links


bendichter:
DANDI does not currently have a good way of handling external links. Eventually, we should find a good way to handle them. For now, we at least need a way to check whether they exist. Here is how you would do that. I don't know how you would want to integrate this into dandi-cli, but you can copy/paste this code:

Create an NWB file with an external link. Here, file4 has an external link to file1:
```python
from datetime import datetime
from dateutil.tz import tzlocal
from pynwb import NWBFile
from pynwb import TimeSeries
from pynwb import NWBHDF5IO
import numpy as np
import h5py

# Create the base data
start_time = datetime(2017, 4, 3, 11, tzinfo=tzlocal())
create_date = datetime(2017, 4, 15, 12, tzinfo=tzlocal())
data = np.arange(1000).reshape((100, 10))
timestamps = np.arange(100)
filename1 = 'external1_example.nwb'
filename4 = 'external_linkdataset_example.nwb'

# Create the first file
nwbfile1 = NWBFile(session_description='demonstrate external files',
                   identifier='NWBE1',
                   session_start_time=start_time,
                   file_create_date=create_date)
# Create the second file
test_ts1 = TimeSeries(name='test_timeseries1',
                      data=data,
                      unit='SIunit',
                      timestamps=timestamps)
nwbfile1.add_acquisition(test_ts1)
# Write the first file
io = NWBHDF5IO(filename1, 'w')
io.write(nwbfile1)
io.close()

# Create the second file
nwbfile2 = NWBFile(session_description='demonstrate external files',
                   identifier='NWBE2',
                   session_start_time=start_time,
                   file_create_date=create_date)

# Create the first file
nwbfile4 = NWBFile(
    session_description='demonstrate external files',
    identifier='NWBE4',
    session_start_time=start_time,
    file_create_date=create_date
)

# Get the first timeseries
io1 = NWBHDF5IO(filename1, 'r')
nwbfile1 = io1.read()
timeseries_1 = nwbfile1.get_acquisition('test_timeseries1')
timeseries_1_data = timeseries_1.data

# Create a new timeseries that links to our data
test_ts4 = TimeSeries(
    name='test_timeseries4',
    data=timeseries_1_data,   # <-------
    unit='SIunit',
    timestamps=timestamps
)
nwbfile4.add_acquisition(test_ts4)


io4 = NWBHDF5IO(filename4, 'w')
io4.write(nwbfile4, link_data=True)
io4.close()
```

[...]

Issue #70, opened 2020-04-03
organize: maintain mapping to original filenames and checksums


satra:
a request by labs to know which file was converted to the dandiset. perhaps organize can upload this with the file/item metadata and dandi cli can look this up from the server.

however in general perhaps we can add a simple provenance file:

<filename 1> wasDerivedFrom <old filename> .
<filename1> <sha512_or_some_such> <sha> .

this may be useful to create a manifest file later or for checking against amazon store on download.

at present organize doesn't change checksums. so that notion is simply a flag for now.

yarikoptic:
ATM organize indeed doesn't change files, and I hope it will never have to. We already upload object id from nwb which could be used to track which file was already uploaded. I guess there could be a dedicated function/command to check for each file if it is already known to an uploaded dandiset. Actually might just need to be an option to dandi ls.

Tracking provenance of changes is a much bigger issue, but since out tools don't change files yet, we aren't even in that realm yet, besides tracking uploads - #43

satra:
but a name change is still a change, and i think that's what the labs were looking for. till we get labs to use dandiset organization locally, it may be nice to provide such an option to `lookup` old name.

yarikoptic:
thanks for the clarification. Indeed knowing association would be helpful.
I am thinking that adding `dandi diff` command would make most sense here.  Filed a dedicated https://github.com/dandi/dandi-cli/issues/72

satra:
i think some of this is coming from the fact that @bendichter  did the conversion and uploaded, so only he has the symlinks to the original dataset. but in general this would be an issue with any member of the lab

satra:
that's a good starting point for now.  will the local path support a tree (or does it have to be flat).

yarikoptic:
the sky is the limit! since nothing is actually implemented ;) but it should indeed operate across subdirectories as well

yarikoptic:
It seems that users reporting issues which could be mitigated by having this issue addressed. I will slate it for 0.7.0 for now

yarikoptic:
FWIW
- #1024 is adding `dandi move` command to operate locally and/or remotely
- we still do not have any local "provenance tracking" so nowhere to record what was derived from/renamed to. Not even sure how to ensure such provenance adequacy since people would use `mv` directly... so not sure how feasible to do that "robustly" (without content tracking alike git-annex where to associate based on checksums etc...)

PR #841, opened 2021-11-24
Video files organize

### LINKED ISSUE #785: organizing video files using dandi organize
This is a continuation of https://github.com/dandi/helpdesk/discussions/30

It would be more disk space efficient to store behavioral video files outside of an nwbfile since the compression is not at par with the codecs for mp4 etc. If there is a standard to organize video files and then link those to the nwb file as `external_file` argument, it would help. 

If a user has a video file stored locally and an nwbfile that links to that local file on disc, uploading the nwb on DANDI would break the link. So we should come up with a design for organizing the nwb associated video files on DANDI and update the link stored in the nwbfile. 

 `dandi organize` for video files specifically should:
1. Rename the video files prior to uploading on DANDI.
2. Update the nwbfile in real time and change the external file link based on the new name of the video file. 

@satra @bendichter I have created a design draft for us to collaborate on
https://docs.google.com/document/d/1oclxWFRNbakjYBlrze4DiFWkh_v0yVhmGjNar09skx4/edit?usp=sharing

Saksham20:
fixes #785

This PR adds the functionality to upload on DANDI additional files of different extensions that are be linked to the nwbfile with the `external_file` argument. 

Prior to upload this requires editing the nwbfile's `external_file` argument to reflect the new location of the video in the DANDI re-organized folder structure as it would be on the cloud. This would keep external links consistent for another user that downloads that dandiset. This is a multi-step process: we need a strategy to organise the video files like `Dandi organise`, edit the nwbfile and then upload this.


dandi/organize.py
```diff
@@ -172,6 +172,23 @@ def create_unique_filenames_from_metadata(metadata):
     return metadata
 
 
+def create_external_file_names(metadata):
+    metadata = deepcopy(metadata)
+    import uuid
```
yarikoptic, line 177: no need to delay this import IMHO -- should go to the top of the module

dandi/organize.py
```diff
@@ -172,6 +172,23 @@ def create_unique_filenames_from_metadata(metadata):
     return metadata
 
 
+def create_external_file_names(metadata):
```
yarikoptic, line 175: needs a docstring on what it does and its purpose ...

dandi/consts.py
```diff
@@ -130,3 +130,6 @@
 #: HTTP response status codes that should always be retried (until we run out
 #: of retries)
 RETRY_STATUSES = (500, 502, 503, 504)
+
+external_file_types = [".mp4"]
```
yarikoptic, line 134: I would call it `external_file_extensions` to be clearer since "types" could mean different things

dandi/cli/cmd_organize.py
```diff
@@ -212,6 +214,11 @@ def _get_metadata(path):
 
     metadata = create_unique_filenames_from_metadata(metadata)
 
+    # create video file name and re write nwb file external files:
+    metadata = create_external_file_names(metadata)
```
yarikoptic, line 218: I think this "mode" should be optional (add `--rewrite` option with for now a single possible value `external-paths`) and be made "available" (otherwise - error) only in `files_mode='copy'` for now. In principle we could go another way and if such option is available we would document/assume that those files which need to be modified will be copied, and the rest (which do not need to be modified) -- acted upon according to `files_mode` setting.
WDYT @bendichter @satra @jwodder ?

Also relates to https://github.com/dandi/dandi-cli/issues/840 in that if it is not given, we should still do the check if there are external links (if such mode is not enabled) and error out in such a case.

yarikoptic (commented):
Very preliminary review to possible help align ongoing development. Extra overall

- I think we need docstrings for all those functions since even though name is descriptive somewhat, the purpose and what is returned is not clear
  - also please use type annotations 
- Need to come up with unittest(s)

dandi/consts.py
Saksham20: yeah thanks that is more descriptive

dandi/cli/cmd_organize.py
Saksham20: I don't understand why it would not work with `files_mode=move/symlink` also. In the current design the videos would be nested as follows:
```
sub-subjectid/
    sub-subjectid_ses-sessionid_ecephys.nwb
    sub-subjectid_ses-sessionid_ecephys/ # the videos sit beside the Dandi oganized nwb files
        uuid_external_file_0.mp4 # actual file or a symlink
```
Lets say we have created symlinks instead and have correctly renamed of nwbfiles' `external_file` arg. 
After that, would the Dandi upload operation not work with video symlinks and create the correct folder structure on the hosted dandiset?
I imagine if another user downloads that, the same folder would exist on the local drive and the `external_file` in the downloaded nwb files be valid again.

yarikoptic: I would assume that `external_file` field should be modified in the .nwb file itself, right? (then I guess ok for `move` but not for symlink or hardlink since would modify original file)

although I am confused how it is modified if file seems to be opened in `mode='r'` (read) mode:
https://github.com/dandi/dandi-cli/pull/841/files#diff-222d83ee68f525e4cf9e4b6b189b26401c17338b7b40c0d65b46e66968226c71R284

Saksham20: yes, sorry I meant if the video files exist as symlinks/hardlink instead of copy/move. In case of the nwbfile it makes sense since we are modifying the original in case of symlink/hardlink.

Saksham20:
todo:

- [ ]  add tests for `dandi organise`
- [ ] test dandi upload on https://gui-staging.dandiarchive.org/#/

satra:
@Saksham20 - have you considered the external name in NWB to be some form of a template possibly following a URI scheme. for example: `{host}:video/{basename}.mp4` this would take the current file location/name and replace the name with `video/{basename}.mp4`. the set of keywords allowed can be discussed. but this would allow creating nwb files without having to rewrite them as the name changes. and then the only consistency required is external. 

the potential advantage of such a scheme is that it could use either relative or absolute paths. an appropriate tool (e.g., pynwb, dandi cli) could also determine how to interpret host properly.

this does require interpreting the location, but even currently the user has to do that.

yarikoptic:
@satra (and attn @bendichter) In NWB-DANDI sync call we arrived at an agreement that the best way to proceed ATM would be 
- (immediate term) rely on (expensive) `copy` / `move` which would allow for rewrite of the nwb files with updated external filenames
- (shortish term) see sidecar support implemented. https://github.com/hdmf-dev/hdmf/pull/677#issuecomment-987139363 provides example "replace" operations to be performed on nwb file fields.

satra:
@yarikoptic - does that mean no possibility for templating?

yarikoptic:
I would say, no plan to support it ATM. Possibility can always exist! ;-) what particular extra use cases, in presence of sidecar file support, templating would allow for?

satra:
@yarikoptic - just a quick pointer to an old comment :) - https://github.com/dandi/dandi-cli/issues/70#issuecomment-608438295  - more seriously, external files are not interpreted by pynwb at present as anything other than a string right? as in there is no representation of seeking into a file, or that it is a file. i'm assuming that the link could be a url and pynwb won't complain. so having a template which is resolved by a reader would be fine without even any changes to pynwb (i think). and this would allow for external referencing of a related file and potentially even bytestream and offset without the need to rewrite any component of the file. i think this becomes a better standard operating procedure moving forward than a fixed file. if one wanted to encode specific version, one could optionally embed the checksum into the link as well. using a schema uri would offer a lot of benefits in the long run i think and would still support an absolute path in the short term.

oruebel:
> more seriously, external files are not interpreted by pynwb at present as anything other than a string right?

This is true only for the special case of [ImageSeries](https://nwb-schema.readthedocs.io/en/latest/format.html?highlight=external_file#imageseries) types which define an external reference to a non-HDF5 file. External links to HDF5 files are resolved automatically by HDF5. 

> so having a template which is resolved by a reader would be fine without even any changes to pynwb (i think)

In principle yes, but this is only because PyNWB currently does not check that the paths are valid (but it arguably should). In terms of the schema, I would argue that this sort of thing would require an update to the schema, as it explitly states this should be  "Paths to one or more external file(s)." 

> a template which is resolved by a reader

Could you provide an example of how such a template would like (i.e., what are the components of the template) and how you a reader (e.g., PyNWB) would be able to determine the parameters for the template?

Saksham20:
> @Saksham20 - have you considered the external name in NWB to be some form of a template possibly following a URI scheme. for example: `{host}:video/{basename}.mp4` this would take the current file location/name and replace the name with `video/{basename}.mp4`. the set of keywords allowed can be discussed. but this would allow creating nwb files without having to rewrite them as the name changes. and then the only consistency required is external.


I'm not sure what you mean, 

Do you mean templating of the video file name once its on DANDI or after the local renaming. Aren't we already applying a template agreed upon: video renamed = `<nwbfile_name/<ImageSeries.Objectid>_external_file_0.mp4`. This is how the video file is renamed after `dandi organise` with the copy/move argument.  

Do you suggest we rename it differently to identify it on the cloud once uploaded on dandi?

Saksham20:
> @Saksham20 - have you considered the external name in NWB to be some form of a template possibly following a URI scheme. for example: `{host}:video/{basename}.mp4` this would take the current file location/name and replace the name with `video/{basename}.mp4`. the set of keywords allowed can be discussed. but this would allow creating nwb files without having to rewrite them as the name changes. and then the only consistency required is external.

Wouldn't the URI change once uploaded on DANDI? Then we would have to rewrite the value in the nwbfile

satra:
@Saksham20 - the ideal scheme would require no renaming inside the nwb file at all. i'm not sure we can achieve ideal, but some examples are below.

> Could you provide an example of how such a template would look like 

sure @oruebel . it may be useful to read [this article on URIs](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier) and browse [this list of schemes](https://en.wikipedia.org/wiki/List_of_URI_schemes). if you look at the list of schemes, what a scheme is completely up to us to define and convey.

example 1:  `nwbext://{host}/{basedir}/video/{basename}-<objectid>.mp4` --> this would be the string stored in the imageseries object. 

locally pynwb or other reader would interpret this as `host=localhost; basedir=current file base directory; basename=name of current file; objectid=imageseries object id`. if this was being derived from `https://api.dandiarchive.org/api/dandisets/<dandiset>/versions/<version>/assets/<uuid>/download` with a content disposition of `name=somename.nwb` similar notions would hold. e.g., that the file is available in some path relative to this file. to truly make this work we would need to expose a path api ( a la github ) as opposed to the current path api. (but that's true of any current implementation as well). we would not allow the global assets endpoint to do path referencing. 

example 2: more bids like: `nwbext://{host}/{basedir}/{basebidsname}_video+<objectid>.mp4`

basebidsname would remove the suffix, which is about technique/assay type typically in bids.

example 3: add offset `nwbext://{host}/{basedir}/video/{basebidsname}_video+<objectid>.mp4;offset=4`

example 4: relative path: `nwbext://{host}/{basedir}/../up/down/video/megasession_video+<objectid>.mp4;offset=100`

example 5: add checksum: `nwbext://{host}/{basedir}/../up/down/video/megasession_video+<objectid>.mp4;offset=100;multihash=0x123456789`

my use of `;` simply indicates the notion that we could make this whatever. we could also use the more standard `?` and `&`

example 5alt: add checksum: `nwbext://{host}/{basedir}/../up/down/video/megasession_video+<objectid>.mp4?offset=100&multihash=0x123456789`

example 6: absolute file `nwbext:///some/path/to/abc.mp4` obviously this wouldn't work with dandi and the upload validator should complain, but would work for any local reader.

oruebel:
@satra thanks for those examples. It wasn't clear to me that you were thinking about URI schemes. It seems this is effectively means designing a REST API around ``ImageSeries``. If we do this, I think it may be worth considering whether we need to assign a neurodata_type for this kind of dataset. I'm not fully convinced yet if we really need a complex path scheme if we have sidecar file. @rly  @ajtritt thoughts?

> to truly make this work we would need to expose a path api

Does this mean, the user would need to know the original source of the file (e.g., that this file was orignially downloaded from DANDI) to interpret the path?

satra:
@oruebel - i'm not sure this would require a REST API. this is more like jinja templating. 

regarding the link between files for local or downloading, that's going to be the case independent of this suggestion. if you are local, the external file needs to be in a specific location relative to the current file (unless an absolute path is used). if the relative or absolute path changes, the link breaks. if you are remote same thing holds and something needs to know/represent where the other file is. 

in either case a reader can inform the user what the filename should look like, so in the worst case scenario, a user could then download the file and place it manually. 

> I'm not fully convinced yet if we really need a complex path scheme if we have sidecar file.

i don't quite follow what you mean by a sidecar file? are you talking about an external json file? or the video file itself. and the path scheme can be as simple as pointing to a file, but can be as complex as providing a template for a file.

oruebel:
> i don't quite follow what you mean by a sidecar file?

Sorry, I was referring to the ability to update datasets/attributes via a JSON file, that we are discussing here  https://github.com/hdmf-dev/hdmf/pull/677

dandi/tests/fixtures.py
```diff
@@ -312,3 +320,53 @@ def upload_dandiset(paths=None, **kwargs):
         "dandiset_id": dandiset_id,
         "reupload": upload_dandiset,
     }
+
+
+@pytest.fixture(scope="session")
+def download_video_files():
+    from datalad.api import Dataset, install
+    pt = Path(tempfile.mkdtemp())/"ophys_testing_data"
```
jwodder, line 328: Use pytest's [`tmp_path_factory`](https://docs.pytest.org/en/6.2.x/reference.html#tmp-path-factory) fixture instead of `tempfile`.

dandi/tests/fixtures.py
```diff
@@ -312,3 +320,53 @@ def upload_dandiset(paths=None, **kwargs):
         "dandiset_id": dandiset_id,
         "reupload": upload_dandiset,
     }
+
+
+@pytest.fixture(scope="session")
+def download_video_files():
+    from datalad.api import Dataset, install
```
jwodder, line 327: `datalad` will need to be added to the test requirements in `setup.cfg`.

setup.cfg
```diff
@@ -55,6 +55,7 @@ install_requires =
     semantic-version
     tenacity
     tqdm
+    datalad
```
jwodder, line 58: Datalad is only used while testing, so it should be added to the `test` extras under `[options.extras_require]`, not to `install_requires`.  Also, when adding it, please keep the list in alphabetical order.

dandi/cli/cmd_organize.py
```diff
@@ -34,6 +40,22 @@
     default="auto",
     show_default=True,
 )
+@click.option(
+    "-rw",
```
jwodder, line 44: Multicharacter options should begin with two hyphens, not one.

dandi/cli/cmd_organize.py
```diff
@@ -34,6 +40,22 @@
...
+    "--rewrite",
+    type=click.Choice(["external-file"]),
+    default=None,
+    help="if set to re-write attributes in an nwbfile. For example, if the value "
+         "is set to external-paths then the external_path attribute of nwbfiles'"
+         "ImageSeries will be written with renamed video paths according to set rules."
+)
+@click.option(
+    "-ef",
```
jwodder, line 53: Multicharacter options should begin with two hyphens, not one.

dandi/cli/cmd_organize.py
```diff
@@ -34,6 +40,22 @@
     default="auto",
     show_default=True,
 )
+@click.option(
+    "-rw",
+    "--rewrite",
+    type=click.Choice(["external-file"]),
+    default=None,
+    help="if set to re-write attributes in an nwbfile. For example, if the value "
```
jwodder, line 48: I think the initial "if" should be removed.

dandi/cli/cmd_organize.py
```diff
@@ -212,6 +239,26 @@ def _get_metadata(path):
 
     metadata = create_unique_filenames_from_metadata(metadata)
 
+    # update metadata with external_file information:
+    external_files_missing_metadata_bool = [len(m["external_file_objects"]) == 0 for m in metadata]
+    if all(external_files_missing_metadata_bool) and rewrite == "external-file":
+        lgr.warning("rewrite option specified as 'external_file' but no external_files found"
+                    f"linked to any nwbfile found in {paths}")
```
jwodder, line 246: The proper way to format a log message is with `%s`-style specifiers and arguments, e.g., `lgr.warning("... linked to any nwbfile found in %s", paths)`.

dandi/organize.py
```diff
@@ -9,13 +9,14 @@
...
 import re
+import uuid
 
 import numpy as np
 
 from . import get_logger
 from .exceptions import OrganizeImpossibleError
 from .pynwb_utils import get_neurodata_types_to_modalities_map, get_object_id
-from .utils import ensure_datetime, flattened, yaml_load
+from .utils import ensure_datetime, flattened, yaml_load, copy_file, move_file
```
jwodder, line 19: I can tell from the fact that the items here aren't alphabetized that you aren't using [pre-commit](https://github.com/pre-commit/pre-commit).  Please install pre-commit, run `pre-commit install` in your clone of the dandi-cli repository, and run `pre-commit run -a` to properly format all of your changes so far.

dandi/organize.py
```diff
@@ -172,6 +173,74 @@ def create_unique_filenames_from_metadata(metadata):
     return metadata
 
 
+def _create_external_file_names(metadata: dict):
```
jwodder, line 176: The docstring and function body indicate that `metadata` should actually be actually `List[dict]`.  Also, you need to add a return-type annotation of `-> List[dict]`.

dandi/organize.py
```diff
@@ -172,6 +173,74 @@ def create_unique_filenames_from_metadata(metadata):
...
+            renamed_path_list = []
+            uuid_str = ext_file_dict.get("id", str(uuid.uuid4()))
+            for no, ext_file in enumerate(ext_file_dict['external_files']):
+                renamed = op.join(nwb_folder_name,f'{uuid_str}_external_file_{no}{ext_file.suffix}')
+                renamed_path_list.append(str(renamed))
+            ext_file_dict['external_files_renamed'] = renamed_path_list
+    return metadata
+
+
+def organize_external_files(metadata: list, dandiset_path: str, files_mode: str):
```
jwodder, line 210: Please change `metadata`'s type annotation to `List[dict]` and add a `-> None` return type annotation.

dandi/pynwb_utils.py
```diff
@@ -227,9 +230,73 @@ def _get_pynwb_metadata(path):
...
+    for module_name in external_file_modules:
+        module_cont = getattr(nwb, module_name)
+        for name, ob in module_cont.items():
+            if isinstance(ob, pynwb.image.ImageSeries) and ob.external_file is not None:
+                out_dict = dict(id=ob.object_id, name=ob.name, external_files=[])
+                for ext_file in ob.external_file:
+                    if Path(ext_file).suffix in external_file_extensions:
+                        out_dict['external_files'].append(Path(ext_file))
+                    else:
+                        lgr.warning(f'external file {ext_file} should be one of: {external_file_extensions}')
```
jwodder, line 263: Use `%s` formatting for log messages.

jwodder (changes requested):

jwodder:
The use of datalad in tests is indirectly leading to a warning being raised by the `future` library, and pytest is failing as a result.

@Saksham20 Please add the following line to the `filterwarnings` list in the `[pytest]` section in `tox.ini`.  (I'd do it myself, but it seems I don't have push permission to the PR head.)

    ignore:the imp module is deprecated:DeprecationWarning

dandi/cli/cmd_organize.py
```diff
@@ -212,6 +235,40 @@ def _get_metadata(path):
...
+            "rewrite option specified as 'external_file' but no external_files found"
+            "linked to any nwbfile found in %s",
+            paths,
+        )
+
+    elif not all(external_files_missing_metadata_bool) and rewrite != "external-file":
+        raise ValueError(
+            "rewrite option not specified but found external video files linked to "
+            f"the nwbfiles "
+            f"""{[metadata[no]['path']
```
jwodder, line 253: Use `%s` log formatting here.

dandi/cli/cmd_organize.py
```diff
@@ -212,6 +235,40 @@ def _get_metadata(path):
...
+                  for no, a in enumerate(external_files_missing_metadata_bool)
+                  if not a]}, """
+            f"change option: -rewrite 'external_file'"
```
jwodder, line 256: `-rewrite` should be `--rewrite`.

dandi/consts.py
```diff
@@ -134,3 +134,6 @@
 #: HTTP response status codes that should always be retried (until we run out
 #: of retries)
 RETRY_STATUSES = (500, 502, 503, 504)
+
+external_file_extensions = [".mp4", ".avi", ".wmv", ".mov", ".flv"]
```
jwodder, line 138: Constant names should be in all-caps.

dandi/consts.py
```diff
@@ -134,3 +134,6 @@
...
+external_file_modules = ["processing", "acquisition"]
```
jwodder, line 139: This is only used once in the code.  Is there a need to define it in `consts.py`?

dandi/organize.py
jwodder: @Saksham20 You did not address this comment.

dandi/organize.py
```diff
@@ -172,6 +173,81 @@ def create_unique_filenames_from_metadata(metadata):
...
+    Returns
+    -------
+    metadata: list
+        updated list of metadata dictionaries
+    """
+    metadata = deepcopy(metadata)
+    for meta in metadata:
+        if "dandi_path" not in meta or "external_file_objects" not in meta:
+            continue
+        nwb_folder_name = op.basename(meta["dandi_path"]).split(".")[0]
```
jwodder, line 200: What if the basename begins with a period?

dandi/pynwb_utils.py
```diff
@@ -227,9 +230,95 @@ def _get_pynwb_metadata(path):
...
+        module_cont = getattr(nwb, module_name)
+        for name, ob in module_cont.items():
+            if isinstance(ob, pynwb.image.ImageSeries) and ob.external_file is not None:
+                out_dict = dict(id=ob.object_id, name=ob.name, external_files=[])
+                for ext_file in ob.external_file:
+                    if Path(ext_file).suffix in external_file_extensions:
+                        out_dict["external_files"].append(Path(ext_file))
+                    else:
+                        lgr.warning(
+                            "external file %s should be one of: %s}",
```
jwodder, line 267: Remove trailing `}`

dandi/pynwb_utils.py
```diff
@@ -227,9 +230,73 @@ def _get_pynwb_metadata(path):
             key = f[len("number_of_") :]
             out[f] = len(getattr(nwb, key, []) or [])
 
+        # get external_file data:
+        out["external_file_objects"] = _get_image_series(nwb)
+
     return out
 
 
+def _get_image_series(nwb: pynwb.NWBFile):
```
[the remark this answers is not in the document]
jwodder: @Saksham20 You did not address this comment.

dandi/tests/fixtures.py
```diff
@@ -312,3 +317,63 @@ def upload_dandiset(paths=None, **kwargs):
...
+    return Path(resp[0]["path"])
+
+
+@pytest.fixture()
+def create_video_nwbfiles(download_video_files, tmp_path):
+    base_nwb_path = tmp_path / "nwbfiles"
+    base_nwb_path.mkdir(parents=True, exist_ok=True)
+    base_vid_path = tmp_path / "video_files"
+    base_vid_path.mkdir(parents=True, exist_ok=True)
+    shutil.copytree(download_video_files, str(base_vid_path))
```
jwodder, line 340: `shutil.copytree()` can't copy to a directory that already exists.  Remove the `base_vid_path.mkdir(...)` line.

dandi/tests/test_organize.py
```diff
@@ -248,3 +249,45 @@ def error_link(src, dest):
     monkeypatch.setattr(os, "symlink", succeed_link if sym_success else error_link)
     monkeypatch.setattr(os, "link", succeed_link if hard_success else error_link)
     assert detect_link_type(tmp_path) == result
+
+
+@pytest.mark.filterwarnings("ignore:the imp module is deprecated:DeprecationWarning")
```
jwodder, line 254: This should be redundant.

setup.cfg
```diff
@@ -83,6 +83,7 @@ test =
     pytest-cov
     pytest-mock
     responses
+    datalad
```
jwodder, line 86: Put this in the list in alphabetical order.

jwodder (changes requested):

jwodder:
@yarikoptic Question: Do we want the file types listed in `external_file_extensions` to be uploaded and/or validated alongside NWB assets when running `dandi upload`/`dandi validate` without `--allow-any-path`?  If so, should that change be in this PR or a separate one?

dandi/consts.py
Saksham20: we should then be following this for all the constants defined above also which can go in another PR

dandi/cli/cmd_organize.py
Saksham20: Do you suggest we log this Exception before throwing it?
Or to use format method instead of f strings: `'string{}'.format(var)`

dandi/consts.py
Saksham20: we might want to use it again for other assisting datatypes that DANDI could host in the future. This just defines where in the nwbfile we want to restrict the lookup for external_files.

dandi/organize.py
Saksham20: the basename is the name of an nwbfile as renamed by dandi so its not expected to start with a period.

dandi/cli/cmd_organize.py
jwodder: Oh, sorry, for some reason I thought this was a logging call.  Never mind.

dandi/pynwb_utils.py
```diff
@@ -4,7 +4,7 @@
 from pathlib import Path
 import re
 import warnings
-
+from typing import List
```
jwodder, line 7: @Saksham20 I can tell that you're not using pre-commit properly.  You need to run `pre-commit install` in your local clone of the repository so that it will take effect automatically whenever you commit, and you also need to run `pre-commit run -a` again to enforce the changes that should have been made already.

Saksham20: oops sorry, I pushed todays changes from another computer which did not have pre-commit. Pushing now

dandi/pynwb_utils.py
```diff
@@ -227,9 +231,95 @@ def _get_pynwb_metadata(path):
...
+                for ext_file in ob.external_file:
+                    if Path(ext_file).suffix in external_file_extensions:
+                        out_dict["external_files"].append(Path(ext_file))
+                    else:
+                        lgr.warning(
+                            "external file %s should be one of: %s",
+                            ext_file,
+                            external_file_extensions,
+                        )
+                out.append(out_dict)
```
yarikoptic, line 272: could there be some sample file(s) created (or worse to worse - shipped along)  to be able to design a test which would cover this notable in size codeblock?

dandi/pynwb_utils.py
```diff
@@ -227,9 +231,95 @@ def _get_pynwb_metadata(path):
...
+                else:
+                    container = container_list[0]
+                # rename all external files:
+                for no, (name_old, name_new) in enumerate(
+                    zip(
+                        ext_file_dict["external_files"],
+                        ext_file_dict["external_files_renamed"],
+                    )
+                ):
+                    container.external_file[no] = str(name_new)
```
yarikoptic, line 320: that's yet another big block without tests. need to figure out to cover at least with smoke testing

dandi/organize.py
yarikoptic: not expected indeed, but i would not rely on that 100%.
```suggestion
        nwb_folder_name = op.splitext(op.basename(meta["dandi_path"]))[0]
```
?

dandi/pynwb_utils.py
yarikoptic: probably just disregard this comment for now since it seems work needs to be done to address failing tests first, and then coverage would get uploaded possibly invalidating my comment above ;)

Saksham20:
> @yarikoptic Question: Do we want the file types listed in `external_file_extensions` to be uploaded and/or validated alongside NWB assets when running `dandi upload`/`dandi validate` without `--allow-any-path`? If so, should that change be in this PR or a separate one?

@yarikoptic will the flag `--allow-any-path` allow any file type to be validated or any folder structure with only nwb/.yml files?
if I do:
```shell
DANDI_DEVEL=1 dandi validate --allow-any-path
```
OR the same without the flag, it does not validate any video files (in the master). In this branch after changing the `validate_file` function [here](https://github.com/dandi/dandi-cli/blob/a3ca6909cb60383093723b9235df94317f08b5d7/dandi/validate.py#L40) it validates correctly.

dandi/cli/cmd_organize.py
```diff
@@ -140,7 +163,7 @@ def act(func, *args, **kwargs):
                 "Only 'dry' or 'move' mode could be used to operate in-place "
                 "within a dandiset (no paths were provided)"
             )
-        lgr.info(f"We will organize {dandiset_path} in-place")
+        lgr.info("We will organize %s in-place", dandiset_path)
```
yarikoptic, line 166: I am ok either way here but note that it is generally preferable to use original `%` + args notation since then string is not even interpolated unless logging level is "enabled".  With `f""` strings you would always cause string interpolation (and possibly relatively expensive `str`) even if end message would not even be displayed

dandi/cli/cmd_organize.py
```diff
@@ -214,6 +237,40 @@ def _get_metadata(path):
 
     metadata = create_unique_filenames_from_metadata(metadata)
 
+    # update metadata with external_file information:
+    external_files_missing_metadata_bool = [
+        len(m["external_file_objects"]) == 0 for m in metadata
+    ]
+    if all(external_files_missing_metadata_bool) and rewrite == "external-file":
+        lgr.warning(
+            "rewrite option specified as 'external_file' but no external_files found"
```
yarikoptic, line 246:
```suggestion
            "rewrite option specified as 'external_file' but no external_files found "
```

dandi/cli/cmd_organize.py
jwodder: More importantly, if you use f-strings and one of the interpolated strings contains a `%`, the logger will still try to format it, resulting in an error.

yarikoptic:
> In this branch after changing the `validate_file` function [here](https://github.com/dandi/dandi-cli/blob/a3ca6909cb60383093723b9235df94317f08b5d7/dandi/validate.py#L40) it validates correctly.

well, that change just doesn't do anything for those files -- returns an empty list of errors.  I wouldn't call it "validates correctly" ;)  so, why to pretend that we are validating them?

dandi/tests/test_organize.py
```diff
@@ -249,3 +250,44 @@ def error_link(src, dest):
...
+        "external-file",
+        "--external-files-mode",
+        video_mode,
+        "-d",
+        str(dandi_organize_path),
+        str(create_video_nwbfiles),
+    ]
+    video_files_list = [
+        i for i in (create_video_nwbfiles.parent / "video_files").iterdir()
+    ]
```
yarikoptic, line 273:
```suggestion
    video_files_list = list((create_video_nwbfiles.parent / "video_files").iterdir())
```
?

dandi/tests/test_organize.py
```diff
@@ -249,3 +250,44 @@ def error_link(src, dest):
...
+        with NWBHDF5IO(str(nwbfile_name), "r", load_namespaces=True) as io:
+            nwbfile = io.read()
+            # get iamgeseries objects as dict(id=object_id, external_files=[])
+            ext_file_objects = _get_image_series(nwbfile)
+            for ext_file_ob in ext_file_objects:
+                filename = vid_folder.name + f'/{ext_file_ob["id"]}_external_file_'
+                for name in ext_file_ob["external_files"]:
+                    video_files_organized.append(name)
+                    # check if external_file arguments are correctly named according to convention:
+                    assert filename in str(name)
```
yarikoptic, line 289: please make it a bit more specific -- e.g. should `name` end with that filename? 
check that it is not an absolute path (right?)

dandi/validate.py
```diff
@@ -41,6 +42,8 @@ def validate_file(filepath, schema_version=None, devel_debug=False):
         return validate_dandiset_yaml(
             filepath, schema_version=None, devel_debug=devel_debug
         )
+    elif os.path.splitext(filepath)[-1] in EXTERNAL_FILE_EXTENSIONS:
+        return []
```
yarikoptic, line 46: IMHO we shouldn't just pretend that there are no errors if we are not validating such files.
Since NWB files AFAIK should always have `.nwb` extension by convention, let's just convert `else:` into `if os.path.splitext(filepath)[-1] == '.nwb'`?
even though in effect it might be the same (we return `None` instead of an empty list of errors), at least it would not pretend to validate them. WDYT?

yarikoptic (commented):
Some extra comments. But I guess the main goal should be ATM to bring CI testing back to green.

dandi/tests/test_organize.py
Saksham20: yes, this assertion is more general. I have made it more specific in this commit: https://github.com/dandi/dandi-cli/pull/841/commits/5a49a8a341b9204a46a74a2b951b5b92936a5a5d

.github/workflows/test.yml
```diff
@@ -108,6 +108,25 @@ jobs:
           -f dandi/tests/data/dandiarchive-docker/docker-compose.yml \
           down -v
 
+    - name: Install git-annex (ubuntu)
+      if: startsWith(matrix.os, 'ubuntu')
+      run: |
+        apt-get install git-annex
```
jwodder, line 114: This needs to be run with `sudo`.

jwodder:
@Saksham20 You put the "Install git-annex" steps in the wrong place in the workflow.  They need to go *before* the "Run all tests" step.

dandi/tests/fixtures.py
```diff
@@ -307,6 +312,65 @@ def upload_dandiset(paths=None, **kwargs):
     }
 
 
+@pytest.fixture(scope="session")
+def download_video_files(tmp_path_factory):
+    from datalad.api import install
```
jwodder, line 317: Add in here:

```python
if shutil.which("git-annex") is None:
    pytest.skip("git-annex required")
```

That way, tests requiring git-annex will be skipped if it's not installed.

setup.cfg
```diff
@@ -78,6 +78,8 @@ style =
 test =
     anys ~= 0.2
     coverage
+    datalad
+    pyfakefs ~= 4.0
```
jwodder, line 82: This line should not have been added back in.  Remove it.

dandi/tests/test_organize.py
```diff
@@ -249,3 +250,44 @@ def error_link(src, dest):
...
+def test_video_organize(video_mode, mode, create_video_nwbfiles):
+    dandi_organize_path = create_video_nwbfiles.parent / "dandi_organized"
+    dandi_organize_path.mkdir(parents=True, exist_ok=True)
+    cmd = [
+        "--files-mode",
+        mode,
+        "--rewrite",
+        "external-file",
+        "--external-files-mode",
+        video_mode,
```
jwodder, line 266: When `video_mode` is `None`, both it and the `"--external-files-mode"` string need to be omitted from the command.

dandi/cli/cmd_organize.py
```diff
@@ -34,6 +36,20 @@
...
+    default=None,
+    help="Use to re-write attributes in an nwbfile. For example, if the value "
+    "is set to 'external-paths' then the external_path attribute of nwbfiles' "
+    "ImageSeries will be written with renamed video paths according to set rules.",
+)
+@click.option(
+    "--external-files-mode",
+    type=click.Choice(file_operation_modes),
+    default=None,
+    help="check help for --files-mode option",
```
jwodder, line 51: Rewrite this help string to something like "Like `--files-mode` but [explain difference here]".

dandi/cli/cmd_organize.py
```diff
@@ -34,6 +36,20 @@
     default="auto",
     show_default=True,
 )
+@click.option(
+    "--rewrite",
+    type=click.Choice(["external-file"]),
```
jwodder, line 41: Why is this a Choice with one option?  Are there other rewrite methods that could be added in the future?

.github/workflows/test.yml
```diff
@@ -79,6 +79,22 @@ jobs:
       run: |
         pip install git+https://github.com/NeurodataWithoutBorders/pynwb
 
+    - name: Install git-annex (ubuntu)
+      if: startsWith(matrix.os, 'ubuntu')
+      run: |
+        sudo apt-get install git-annex
```
jwodder, line 85: The version of git-annex installed this way is out of date.  Change this to:

```
pip install datalad-installer
datalad-installer --sudo ok neurodebian git-annex -m neurodebian
```

dandi/cli/cmd_organize.py
Saksham20: I have made changes here https://github.com/dandi/dandi-cli/pull/841/commits/d37a9758a8600c2bcd3f0370850a99133c0a5c32

dandi/tests/test_organize.py
Saksham20: https://github.com/dandi/dandi-cli/pull/841/commits/37e951e6f02fd159ebbedee2c5d759603fce0a7e

.github/workflows/test.yml
Saksham20: https://github.com/dandi/dandi-cli/pull/841/commits/90978c7067e1c9c0d4f22789c117e5c1f11c8d64

dandi/cli/cmd_organize.py
jwodder: That's still not good enough.  The help strings should give the user an idea when they would want to use `--files-mode` versus `--external-files-mode`.  What does `--external-files-mode` do that `--files-mode` doesn't?

Secondly, if `--external-file-mode` can only be "move" or "copy", then its `click.Choice` values should be limited to those instead of being `file_operation_modes`.

dandi/cli/cmd_organize.py
Saksham20: I wanted to generalise this just in case we intend to add the functionality to change some other attributes of the nwb file before upload without change in any cmd options. 
But for now its only video files and having just one option is more user friendly. Check https://github.com/dandi/dandi-cli/pull/841/commits/d6f65220ef45180b18445bd5c4719bb01bc6c7ca

dandi/cli/cmd_organize.py
```diff
@@ -34,6 +36,18 @@
     default="auto",
     show_default=True,
 )
+@click.option(
+    "--rewrite-external-files",
+    type=click.Choice(["copy", "move"]),
+    default=None,
+    help="using this options will rewrite the 'external_file' argument of the "
```
bendichter, line 43:
```suggestion
    help="using this options will rewrite the 'external_file' field of the "
```

dandi/cli/cmd_organize.py
```diff
@@ -34,6 +36,18 @@
...
+    "ImageSeries in the nwb file. The value written will correspond to the"
```
bendichter, line 44:
```suggestion
    "ImageSeries in the nwb file to correspond to the"
```

dandi/cli/cmd_organize.py
```diff
@@ -34,6 +36,18 @@
...
+    " new location of the video files after being organized. This option will "
+    "work only if --files-mode is copy/move since the original nwb file is "
+    "modified Supply one of 'copy' or 'move' as the value. Copy will copy the "
+    "video file in the organized dandiset "
+    "(Copy option is not recommended for large video files)",
+)
```
bendichter, line 50: @Saksham20 

It seems that these options are already handled by "--files-mode" above.

Saksham20: (`--files-mode` , `--rewrite-external-files` ) can be any of permutation of (`copy`, `move`)

bendichter:
@Saksham20 I think this CLI interface could be more intuitive and featureful. "rewrite-external-files" seems like the external files (i.e. videos) are rewritten, which they are not, it is the external_files field of the nwb file that is rewritten. Also, I don't know why a user would want to copy or move the video files. These video files will not be altered, so why not create links? It seems odd to require a user to move or copy them.

How about this for options?

Two new options: modify-external-files-fields, media-files-mode

Any time "organize" is run, it checks if there are any "external_file" fields in the NWB files. If there are the user must also include the flag "--modify-external-files-fields" (which is boolean, defaulting to False) and the user must **not** have selected a "files-mode" of "symlink" or "hardlink". If the user does not set that flag, or if they chose an incompatible files-mode option, they receive an error that explains why their data requires another mode. Something like:

> This dandiset contains links to media files outside of the NWB files, and organizing it requires modifying the "external_files" fields of the NWB files to match the new file organization. Therefore, "symlink" and "hardlink" cannot be used. Please use "--file-mode" of "copy" or "move" instead. Also, please use the "--modify-external-files-fields" flag to allow modification of the "external_files" fields in the NWB files.

We can add an option "--media-files-mode" to handle how media files are organized, which should be optional and can be "symlink," "hardlink", "copy," or "move". If no value is provided (None), I'm not sure what should be the default. I could see having "symlink" be the default here, since that will be the best option for users in most cases, but I could also see an argument for defaulting to the value of "files-mode" so we don't confuse users by copying NWB files and symlinking media files.

dandi/pynwb_utils.py
```diff
@@ -230,9 +232,95 @@ def _get_pynwb_metadata(path: Union[str, Path]) -> Dict[str, Any]:
...
+
+    Parameters
+    ----------
+    metadata: List[dict]
+        list of dictionaries containing the metadata gathered from the nwbfile
+    dandiset_path: str
+        base path of dandiset
+    """
+    for meta in metadata:
+        if not np.all(
```
bendichter, line 292: why use numpy?

dandi/pynwb_utils.py
```diff
@@ -230,9 +232,95 @@ def _get_pynwb_metadata(path: Union[str, Path]) -> Dict[str, Any]:
...
+            [i in meta for i in ["path", "dandi_path", "external_file_objects"]]
+        ):
```
yarikoptic, line 294: indeed no reason for numpy here.  Moreover can save 2 more bytes for extra readability ;)

```suggestion
        if not all(
            i in meta for i in ["path", "dandi_path", "external_file_objects"]
        ):
```

jwodder:
@Saksham20 This PR previously contained a function for validating a movie file using cv2, which was removed when PR #853 was merged into this branch.  Was the removal intentional or accidental?

dandi/cli/cmd_organize.py
```diff
@@ -34,6 +34,22 @@
...
+    "ImageSeries in the nwb file. The value written will correspond to the"
+    " new location of the video files after being organized. This option will "
+    "work only if --files-mode is copy/move since the original nwb file is "
+    "modified",
+)
+@click.option(
+    "--media-files-mode",
+    type=click.Choice(["copy", "move", "symlink", "hardlink"]),
+    default=None,
+    help="Supply one of copy/move/symlink/hardlink as the value",
```
jwodder, line 51: This help string says nothing that isn't already conveyed by the option metavar.  Either make it informative or delete it.

dandi/cli/cmd_organize.py
```diff
@@ -34,6 +34,22 @@
     default="auto",
     show_default=True,
 )
+@click.option(
+    "--modify-external-file-fields",
+    is_flag=True,
+    default=False,
+    help="using this options will rewrite the 'external_file' argument of the "
```
jwodder, line 41: Suggested rewording: "Rewrite the 'external_file' arguments of ImageSeries in NWB files.  The new values will correspond to the new locations of the video files after being organized.  This option requires --files-mode to be 'copy' or 'move'."

dandi/cli/cmd_organize.py
```diff
@@ -34,6 +34,22 @@
     default="auto",
     show_default=True,
 )
+@click.option(
+    "--modify-external-file-fields",
```
jwodder, line 38: Suggestion: Rename to `--update-external-file-paths`.

dandi/cli/cmd_organize.py
```diff
@@ -104,6 +124,11 @@ def act(func, *args, **kwargs):
             lgr.debug("%s %s %s", func.__name__, args, kwargs)
             return func(*args, **kwargs)
 
+    if modify_external_file_fields and files_mode not in ["copy", "move"]:
+        raise ValueError(
+            "files_mode needs to be one of 'copy/move' for the rewrite option to work"
```
jwodder, line 129:
```suggestion
            "--files-mode needs to be 'copy' or 'move' when --modify-external-file-fields is given"
```

dandi/cli/cmd_organize.py
```diff
@@ -104,6 +124,11 @@ def act(func, *args, **kwargs):
             lgr.debug("%s %s %s", func.__name__, args, kwargs)
             return func(*args, **kwargs)
 
+    if modify_external_file_fields and files_mode not in ["copy", "move"]:
+        raise ValueError(
```
jwodder, line 128:
```suggestion
        raise click.UsageError(
```

dandi/cli/cmd_organize.py
```diff
@@ -214,6 +239,36 @@ def _get_metadata(path):
...
     metadata = create_unique_filenames_from_metadata(metadata)
 
+    # update metadata with external_file information:
+    external_files_missing_in_nwbfiles = [
+        len(m["external_file_objects"]) == 0 for m in metadata
+    ]
+
+    if all(external_files_missing_in_nwbfiles) and modify_external_file_fields:
+        lgr.warning(
+            "modify_external_file_fields specified but no external_files found "
```
jwodder, line 249:
```suggestion
            "--modify-external-file-fields specified but no external_files found "
```

dandi/cli/cmd_organize.py
```diff
@@ -214,6 +239,36 @@ def _get_metadata(path):
...
+            "linked to any nwbfile found in %s",
+            paths,
+        )
+    elif (
+        not all(external_files_missing_in_nwbfiles) and not modify_external_file_fields
+    ):
+        raise ValueError(
```
jwodder, line 256:
```suggestion
        raise click.UsageError(
```

dandi/cli/cmd_organize.py
```diff
@@ -214,6 +239,36 @@ def _get_metadata(path):
...
+            "modify_external_file_fields option not specified but found "
```
jwodder, line 257:
```suggestion
            "--modify-external-file-fields option not specified but found "
```

dandi/cli/cmd_organize.py
```diff
@@ -214,6 +239,36 @@ def _get_metadata(path):
...
+            "external video files linked to the nwbfiles "
+            f"""{[metadata[no]['path']
```
jwodder, line 259: Join the elements of the list together with `", "` instead of stringifying a list.

dandi/cli/cmd_organize.py
```diff
@@ -214,6 +239,36 @@ def _get_metadata(path):
...
+                  for no, a in enumerate(external_files_missing_in_nwbfiles)
+                  if not a]}"""
+        )
+
+    if modify_external_file_fields and media_files_mode is None:
+        media_files_mode = "symlink"
+        lgr.warning(
+            "media_files_mode not specified, setting to recommended mode: 'symlink' "
```
jwodder, line 267:
```suggestion
            "--media-files-mode not specified, setting to recommended mode: 'symlink'"
```

dandi/organize.py
```diff
@@ -172,6 +174,83 @@ def create_unique_filenames_from_metadata(metadata):
     return metadata
 
 
+def _create_external_file_names(metadata: list) -> List[dict]:
```
jwodder, line 177:
```suggestion
def _create_external_file_names(metadata: List[dict]) -> List[dict]:
```

bendichter:
> @Saksham20 This PR previously contained a function for validating a movie file using cv2, which was removed when PR https://github.com/dandi/dandi-cli/pull/853 was merged into this branch. Was the removal intentional or accidental?

@jwodder we decided to remove that in order to make this PR smaller and easier to integrate. That section of code added opencv as a requirement and we thought that might be too heavy for this package. We are happy to add that back in a future PR.

dandi/organize.py
```diff
@@ -172,6 +174,83 @@ def create_unique_filenames_from_metadata(metadata):
...
+    """
+    Renames the external_file attribute in an ImageSeries according to the rule:
```
jwodder, line 179: According to what rule?  It doesn't look like you finished this sentence.

dandi/organize.py
```diff
@@ -172,6 +174,83 @@ def create_unique_filenames_from_metadata(metadata):
...
+            continue
+        nwb_folder_name = op.splitext(op.basename(meta["dandi_path"]))[0]
+        for ext_file_dict in meta["external_file_objects"]:
+            renamed_path_list = []
+            uuid_str = ext_file_dict.get("id", str(uuid.uuid4()))
+            for no, ext_file in enumerate(ext_file_dict["external_files"]):
+                renamed = op.join(
+                    nwb_folder_name, f"{uuid_str}_external_file_{no}{ext_file.suffix}"
+                )
+                renamed_path_list.append(str(renamed))
```
jwodder, line 209: `renamed` is already a `str`.

dandi/organize.py
```diff
@@ -172,6 +174,83 @@ def create_unique_filenames_from_metadata(metadata):
...
+            ext_file_dict["external_files_renamed"] = renamed_path_list
+    return metadata
+
+
+def organize_external_files(
+    metadata: list, dandiset_path: str, files_mode: str
```
jwodder, line 215:
```suggestion
    metadata: List[dict], dandiset_path: str, files_mode: str
```

dandi/organize.py
```diff
@@ -172,6 +174,83 @@ def create_unique_filenames_from_metadata(metadata):
...
+            for no, (name_old, name_new) in enumerate(
+                zip(
+                    ext_file_dict["external_files"],
+                    ext_file_dict["external_files_renamed"],
+                )
+            ):
+                new_path = op.join(dandiset_path, op.dirname(e["dandi_path"]), name_new)
+                name_old_str = str(name_old)
+                if not op.exists(op.dirname(new_path)):
+                    os.makedirs(op.dirname(new_path))
```
jwodder, line 241: Pass `exist_ok=True` to `os.makedirs()` and get rid of the `if not op.exists()` check.

dandi/pynwb_utils.py
```diff
@@ -230,9 +232,93 @@ def _get_pynwb_metadata(path: Union[str, Path]) -> Dict[str, Any]:
...
+            if isinstance(ob, pynwb.image.ImageSeries) and ob.external_file is not None:
+                out_dict = dict(id=ob.object_id, name=ob.name, external_files=[])
+                for ext_file in ob.external_file:
+                    if Path(ext_file).suffix in VIDEO_FILE_EXTENSIONS:
+                        out_dict["external_files"].append(Path(ext_file))
+                    else:
+                        lgr.warning(
+                            "external file %s should be one of: %s",
+                            ext_file,
+                            VIDEO_FILE_EXTENSIONS,
```
jwodder, line 271: Join the elements of the list together with `", "` instead of stringifying a list.

dandi/pynwb_utils.py
```diff
@@ -230,9 +232,93 @@ def _get_pynwb_metadata(path: Union[str, Path]) -> Dict[str, Any]:
...
+                        )
+                out.append(out_dict)
     return out
 
 
+def rename_nwb_external_files(metadata: list, dandiset_path: str) -> None:
```
jwodder, line 277:
```suggestion
def rename_nwb_external_files(metadata: List[dict], dandiset_path: str) -> None:
```

dandi/tests/fixtures.py
```diff
@@ -390,6 +395,87 @@ def zarr_dandiset(new_dandiset: SampleDandiset) -> SampleDandiset:
...
 
 
+@pytest.fixture()
+def create_video_files(tmp_path):
+    video_paths = []
+    import cv2
+
+    video_path = tmp_path / "video_files"
+    video_path.mkdir()
+    for no, ext in enumerate([".avi", ".avi"]):
```
jwodder, line 405: Are these file extensions supposed to be the same?  If yes, just do `ext = ".avi"` and `for no in range(2):`.

dandi/tests/fixtures.py
```diff
@@ -390,6 +395,87 @@ def zarr_dandiset(new_dandiset: SampleDandiset) -> SampleDandiset:
     return new_dandiset
 
 
+@pytest.fixture()
+def create_video_files(tmp_path):
```
jwodder, line 399: Fixtures should have noun names, not verb names, so this should be just "`video_files`."

dandi/tests/fixtures.py
```diff
@@ -390,6 +395,87 @@ def zarr_dandiset(new_dandiset: SampleDandiset) -> SampleDandiset:
...
+            writer1.write(np.random.randint(0, 255, (nx, ny, 3)).astype("uint8"))
+            writer2.write(np.random.randint(0, 255, (nx, ny, 3)).astype("uint8"))
+        writer1.release()
+        writer2.release()
+        video_paths.append((movie_file1, movie_file2))
+    return video_paths
+
+
+@pytest.fixture()
+def create_video_nwbfiles(create_video_files):
```
jwodder, line 435: I don't see a reason to have this in a separate fixture from `create_video_files`.  I think the code could be decently simplified if they were combined.

dandi/tests/test_organize.py
```diff
@@ -256,3 +257,42 @@ def error_link(src: Any, dest: Any) -> NoReturn:
...
+            # get iamgeseries objects as dict(id=object_id, external_files=[])
+            ext_file_objects = _get_image_series(nwbfile)
+            for ext_file_ob in ext_file_objects:
+                for no, name in enumerate(ext_file_ob["external_files"]):
+                    video_files_organized.append(name)
+                    # check if external_file arguments are correctly named according to convention:
+                    filename = Path(
+                        f"{vid_folder.name}/{ext_file_ob['id']}_external_file_{no}"
+                    )
+                    assert str(filename) == str(Path(name).with_suffix(""))
```
jwodder, line 294: There's no need to cast both sides to a `str`.  `Path` objects can be compared for equality just fine.

jwodder (changes requested):

dandi/cli/cmd_organize.py
Saksham20: thanks, changes here: https://github.com/dandi/dandi-cli/pull/841/commits/4241edca5a9a434ab9d75a071b60c89af8f34bb5

dandi/cli/cmd_organize.py
Saksham20: thanks, changes here: https://github.com/dandi/dandi-cli/pull/841/commits/4241edca5a9a434ab9d75a071b60c89af8f34bb5

dandi/cli/cmd_organize.py
Saksham20: https://github.com/dandi/dandi-cli/pull/841/commits/4241edca5a9a434ab9d75a071b60c89af8f34bb5

dandi/tests/fixtures.py
Saksham20: yes makes sense given that they work as variables inside tests

dandi/tests/fixtures.py
Saksham20: yes, correct. Initially I had it setup to work with multiple extensions but different ones require codecs so I dropped that. fixed here https://github.com/dandi/dandi-cli/pull/841/commits/a43c81a7aa990ddb66def6b9f93633c410b6910c

dandi/tests/fixtures.py
Saksham20: https://github.com/dandi/dandi-cli/pull/841/commits/a43c81a7aa990ddb66def6b9f93633c410b6910c

dandi/cli/cmd_organize.py
jwodder: @Saksham20 You missed this comment.

dandi/pynwb_utils.py
```diff
@@ -230,9 +232,93 @@ def _get_pynwb_metadata(path: Union[str, Path]) -> Dict[str, Any]:
...
+    """
+    This method supports _get_pynwb_metadata() in retrieving all ImageSeries
+    related metadata from an open nwb file.
+    Specifically it pulls out the ImageSeries uuid, name and all the
+    externally linked files
+    named under the argument 'external_file'.
+    Parameters
+    ----------
+    nwb: pynwb.NWBFile
+    Returns
```
yarikoptic, line 251:
```suggestion
    named under the argument 'external_file'.
    
    Parameters
    ----------
    nwb: pynwb.NWBFile

    Returns
```

requires empty lines before sections

dandi/pynwb_utils.py
```diff
@@ -230,9 +232,93 @@ def _get_pynwb_metadata(path: Union[str, Path]) -> Dict[str, Any]:
...
+    metadata: List[dict]
+        list of dictionaries containing the metadata gathered from the nwbfile
+    dandiset_path: str
+        base path of dandiset
+    """
+    for meta in metadata:
+        if not all(i in meta for i in ["path", "dandi_path", "external_file_objects"]):
+            lgr.warning(
+                "could not rename external files, update metadata"
+                'with "path", "dandi_path", "external_file_objects"'
```
yarikoptic, line 295:
```suggestion
                'could not rename external files, update metadata '
                'with "path", "dandi_path", "external_file_objects"'
```
space to separate from `with` but also consistently use `'`

dandi/pynwb_utils.py
```diff
@@ -230,9 +232,93 @@ def _get_pynwb_metadata(path: Union[str, Path]) -> Dict[str, Any]:
...
+                # retrieve nwb neurodata object of the given object id:
+                container_list = [
+                    child
+                    for child in nwb.children
+                    if ext_file_dict["id"] == child.object_id
+                ]
+                if len(container_list) == 0:
+                    continue
+                else:
+                    container = container_list[0]
```
yarikoptic, line 311: is that a guarantee that there is only 1 element? ie. could we add

```suggestion
                    container = container_list[0]
                    assert len(container_list) == 1
```
or what is the expectation here on the other possible entries?

yarikoptic (changes requested):
minor comments.

dandi/pynwb_utils.py
Saksham20: yes, there is only 1 element expected in `container_list`. All containers in an nwbfile have a unique object id so at max there should only be one that matches the expected `ext_file_dict["id"]`

dandi/cli/cmd_organize.py
Saksham20: thanks for catching that: https://github.com/dandi/dandi-cli/pull/841/commits/d1a71e584e6be75c38b8e2dfbae93a5497c3a604

jwodder (approved):

PR #843, opened 2021-11-24
Error on NWB files with external links

### LINKED ISSUE #840: detect external links

jwodder:
Closes #840.

### DIFF
--- a/dandi/pynwb_utils.py
+++ b/dandi/pynwb_utils.py
@@ -362,3 +362,27 @@ def copy_nwb_file(src, dest):
     ) as iow:
         iow.write(ior.read().copy(), link_data=False)
     return dest
+
+
+@metadata_cache.memoize_path
+def nwb_has_external_links(filepath):
+    with h5py.File(filepath, "r") as fp:
+        visited = set()
+
+        # cannot use `file.visititems` because it skips external links
+        # (https://github.com/h5py/h5py/issues/671)
+        def visit(path="/"):
+            if isinstance(fp[path], h5py.Group):
+                for key in fp[path].keys():
+                    key_path = path + "/" + key
+                    if key_path not in visited:
+                        visited.add(key_path)
+                        if isinstance(
+                            fp.get(key_path, getlink=True), h5py.ExternalLink
+                        ) or visit(key_path):
+                            return True
+            elif isinstance(fp.get(path, getlink=True), h5py.ExternalLink):
+                return True
+            return False
+
+        return visit()
--- a/dandi/pynwb_utils.py
+++ b/dandi/pynwb_utils.py
@@ -356,11 +356,10 @@ def copy_nwb_file(src, dest):
         dest = op.join(dest, op.basename(src))
     else:
         os.makedirs(op.dirname(dest), exist_ok=True)
-    # The simplest way yoh could find
-    with pynwb.NWBHDF5IO(src, "r") as ior, pynwb.NWBHDF5IO(
-        dest, "w", manager=ior.manager
-    ) as iow:
-        iow.write(ior.read().copy(), link_data=False)
+    with pynwb.NWBHDF5IO(src, "r") as ior, pynwb.NWBHDF5IO(dest, "w") as iow:
+        data = ior.read()
+        data.generate_new_id()
+        iow.export(ior, nwbfile=data)
     return dest
 
 
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -49,7 +49,9 @@ def get_metadata(path):
             return None
 
     if nwb_has_external_links(path):
-        raise NotImplementedError("NWB files with external links are not supported")
+        raise NotImplementedError(
+            f"NWB files with external links are not supported: {path}"
+        )
 
     # First read out possibly available versions of specifications for NWB(:N)
     meta["nwb_version"] = get_nwb_version(path)
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -14,6 +14,7 @@ from .pynwb_utils import (
     get_nwb_version,
     ignore_benign_pynwb_warnings,
     metadata_cache,
+    nwb_has_external_links,
 )
 from .utils import ensure_datetime, get_mime_type, get_utcnow_datetime
 
@@ -46,6 +47,9 @@ def get_metadata(path):
         except ValueError as exc:
             lgr.debug("Failed to get metadata for %s: %s", path, exc)
             return None
+
+    if nwb_has_external_links(path):
+        raise NotImplementedError("NWB files with external links are not supported")
 
     # First read out possibly available versions of specifications for NWB(:N)
     meta["nwb_version"] = get_nwb_version(path)
--- a/dandi/tests/test_pynwb_utils.py
+++ b/dandi/tests/test_pynwb_utils.py
@@ -1,14 +1,16 @@
+from datetime import datetime, timezone
 import re
 
-import pynwb
+import numpy as np
+from pynwb import NWBHDF5IO, NWBFile, TimeSeries
 
-from ..pynwb_utils import _sanitize_nwb_version
+from ..pynwb_utils import _sanitize_nwb_version, nwb_has_external_links
 
 
 def test_pynwb_io(simple1_nwb):
     # To verify that our dependencies spec is sufficient to avoid
     # stepping into known pynwb/hdmf issues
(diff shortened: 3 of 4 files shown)


jwodder:
@yarikoptic The tests appear to be failing because [this call](https://github.com/dandi/dandi-cli/blob/a6767f6f2d4469a4eddbcd9aab27bd7b4e1a43d6/dandi/tests/test_organize.py#L182) to [`copy_nwb_file()`](https://github.com/dandi/dandi-cli/blob/a6767f6f2d4469a4eddbcd9aab27bd7b4e1a43d6/dandi/pynwb_utils.py#L335) produces a file with an external link.

@bendichter What's the best way to copy an NWB file while giving it a new `object_id` without creating an external link?

dandi/pynwb_utils.py
```diff
@@ -362,3 +362,27 @@ def copy_nwb_file(src, dest):
...
+                        visited.add(key_path)
+                        if isinstance(
+                            fp.get(key_path, getlink=True), h5py.ExternalLink
+                        ) or visit(key_path):
+                            return True
+            elif isinstance(fp.get(path, getlink=True), h5py.ExternalLink):
+                return True
+            return False
+
+        return visit()
```
bendichter, line 388: Ah yes, thank you for tidying up this function.

bendichter:
@jwodder

> What's the best way to copy an NWB file while giving it a new object_id without creating an external link?

I'm surprised you are running into this issue, especially since link_data is False. @rly would be the person to ask about copying NWB files using PyNWB.

yarikoptic:
pinged @rly on slack

oruebel:
> What's the best way to copy an NWB file while giving it a new `object_id` without creating an external link?

I think a combination of the [generate_new_id](https://hdmf.readthedocs.io/en/stable/hdmf.container.html#hdmf.container.AbstractContainer.generate_new_id) and [``NWBHDFIO.export`` ](https://hdmf.readthedocs.io/en/stable/export.html) functions should work for this.

If you just want to resolve external links, the [``NWBHDF5IO.copy_file`` ](https://hdmf.readthedocs.io/en/stable/hdmf.backends.hdf5.h5tools.html?highlight=HDF5IO#hdmf.backends.hdf5.h5tools.HDF5IO.copy_file) may also do the trick. This function works lower level by using the ``copy`` method from ``h5py``and has been deprecated in favor of ``export`, but I figured I'd mention is just in case the [source code](https://github.com/hdmf-dev/hdmf/blob/e5f72370c38a12ffab744613408c54cd17921c40/src/hdmf/backends/hdf5/h5tools.py#L344-L385) of that function are useful for you.

jwodder:
@oruebel

> I think a combination of the `generate_new_id` and `NWBHDFIO.export` functions should work for this.

How exactly do I combine those two functions?  If I try doing this:

```python
with pynwb.NWBHDF5IO(src, "r") as ior, pynwb.NWBHDF5IO(dest, "w") as iow:
    data = ior.read()
    data.generate_new_id()
    iow.export(container=data)
```

I get `TypeError: NWBHDF5IO.export: missing argument 'src_io', unrecognized argument: 'container'`.

If I instead do this:

```python
with pynwb.NWBHDF5IO(src, "r") as ior, pynwb.NWBHDF5IO(dest, "w") as iow:
    data = ior.read()
    data.generate_new_id()
    iow.export(src_io=ior)
```

then the copy will still have the same object ID as the original.

If I do this:

```python
with pynwb.NWBHDF5IO(src, "r") as ior, pynwb.NWBHDF5IO(dest, "w") as iow:
    ior.generate_new_id()
    iow.export(src_io=ior)
```

then I get `AttributeError: 'NWBHDF5IO' object has no attribute 'generate_new_id'`.

rly:
Please try:
```python
with pynwb.NWBHDF5IO(src, "r") as ior, pynwb.NWBHDF5IO(dest, "w") as iow:
    data = ior.read()
    data.generate_new_id()
    iow.export(ior, container=data)
```

jwodder:
@rly That fails with "`TypeError: NWBHDF5IO.export: unrecognized argument: 'container'`" (I'm using hdmf 3.1.1.)

rly:
My bad. 'container' is the argument name for `HDF5IO`. Please try:
```python
with pynwb.NWBHDF5IO(src, "r") as ior, pynwb.NWBHDF5IO(dest, "w") as iow:
    data = ior.read()
    data.generate_new_id()
    iow.export(ior, nwbfile=data)
```

jwodder:
@rly That works; thanks.

yarikoptic:
tox.ini is very popular today - keeps conflicting

dandi/pynwb_utils.py
```diff
@@ -356,9 +356,32 @@ def copy_nwb_file(src, dest):
...
+                for key in fp[path].keys():
+                    key_path = path + "/" + key
+                    if key_path not in visited:
+                        visited.add(key_path)
+                        if isinstance(
+                            fp.get(key_path, getlink=True), h5py.ExternalLink
+                        ) or visit(key_path):
+                            return True
+            elif isinstance(fp.get(path, getlink=True), h5py.ExternalLink):
+                return True
```
yarikoptic, line 384: @bendichter what use case could trigger this codepath?  would be nice to add it to the test

jwodder:
@yarikoptic Conflict resolved.

yarikoptic:
ok, let's proceed with it as is.  Thank you!

PR #845, opened 2021-12-01
Set new environment variables required by API server

mvandenburgh:
Sets new URL environment variables now required by https://github.com/dandi/dandi-api/pull/613.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,8 +19,6 @@ jobs:
       NO_ET: 1
       DANDI_ALLOW_LOCALHOST_URLS: 1
       DANDI_TESTS_PERSIST_DOCKER_COMPOSE: "1"
-      DJANGO_DANDI_WEB_APP_URL: http://localhost:8085
-      DJANGO_DANDI_API_URL: http://localhost:8000
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


mvandenburgh:
Looks like I can't request specific reviewers until I'm added to the dandi-cli team, @jwodder and/or @yarikoptic can you review this?

.github/workflows/test.yml
```diff
@@ -19,6 +19,8 @@ jobs:
       NO_ET: 1
       DANDI_ALLOW_LOCALHOST_URLS: 1
       DANDI_TESTS_PERSIST_DOCKER_COMPOSE: "1"
+      DJANGO_DANDI_WEB_APP_URL: http://localhost:8085
+      DJANGO_DANDI_API_URL: http://localhost:8000
```
jwodder, line 23: What exactly is the point of setting the environment variables here and in `tox.ini`?  Aren't these only used by the Django instance?  Why do they need to be set on the client side, especially considering that (currently) nothing in the client uses them?

jwodder (approved):

waxlamp:
In order to make the failing tests in https://github.com/dandi/dandi-api/pull/613 pass, we'll need a new release of dandi-cli. @jwodder can you help us out with that?

jwodder:
@waxlamp dandi v0.32.2 has been released.

.github/workflows/test.yml
mvandenburgh: Yeah you're right, I removed those.

(sorry forgot to submit this comment last week, doing so now just to close this thread)

PR #846, opened 2021-12-06
Remove some unused functions

jwodder:


jwodder:
The test failures are due to #847.

jwodder:
@yarikoptic We need a release with #845 in it, so I'm going to add the "release" label to this PR and merge it soon.  Is that OK?

PR #849, opened 2021-12-07
Remove dandi/core/ in favor of copy in dandischema

jwodder:
As instructed [here](https://github.com/dandi/dandischema/pull/100#discussion_r764118777).

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,11 +89,6 @@ Please see [DEVELOPMENT.md](./DEVELOPMENT.md) file.
 
 ## 3rd party components included
 
-### dandi/core/digests/dandietag.py
-
-From <https://github.com/girder/django-s3-file-field> as of v0.1.1-10-g829b9b0
-Copyright (c) 2019-2021 Kitware, Inc., Apache 2.0 license
-
 ### dandi/tests/skip.py
 
 From https://github.com/ReproNim/reproman, as of v0.2.1-40-gf4f026d


yarikoptic:
Thanks!  Let's hope that nobody used those helpers directly (checked our dandisets tools -- nope), otherwise (and ideally) we should have provided some deprecation cycle with all the warnings + shim imports.

yarikoptic:
BTW, how do you look @jwodder  at just making `dandi-cli` finally to start following semver closer and us to mark this one as `major`? ;)

jwodder:
@yarikoptic Marking a PR as "major" would mean that the next release would be v1.0.0, which we shouldn't do at this point.  Note that [the semver spec](https://semver.org/#spec-item-4) permits breaking changes at any point during 0.x.y development.

yarikoptic:
> v1.0.0, which we shouldn't do at this point

why?

> Note that the semver spec permits breaking changes at any point during 0.x.y development.

I am aware of that.  So, yes -- it just "disclaims" responsibility for breaking API.  If we release 1.0.0, we indeed might plan better for the next "major" release, or just progress through a number of them until some (e.g. 10.0.0) could be declared to be a "stable" one and we only then start paying more attention to not breaking API.

jwodder:
@yarikoptic I'd say we shouldn't release v1 until at least the Python API rewrite is done, possibly including the rewrite of `download()` etc. to make it return pyout dicts.

PR #850, opened 2021-12-13
Remove generatorify.py

jwodder:
* `generator_from_callback()` hasn't been used since we removed the Girder code.  If we need something like it again, we can use [`callback_iter()` from more-itertools](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.callback_iter).
* `callback_from_generator()` has never been used.


yarikoptic:
thanks, let's proceed!

PR #857, opened 2022-01-04
ENH: update species interpretation from NWB file

satra:
This adds ability to the CLI to resolve NCBITaxon urls and infer names of species from them. It also changes the way the name field is populated.

### DIFF
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -9,6 +9,7 @@ from xml.dom.minidom import parseString
 
 from dandischema import models
 import requests
+import tenacity
 
 from . import __version__, get_logger
 from .dandiset import Dandiset
@@ -398,6 +399,11 @@ species_map = [
 
 
 @lru_cache(maxsize=None)
+@tenacity.retry(
+    reraise=True,
+    stop=tenacity.stop_after_attempt(3),
+    wait=tenacity.wait_exponential(exp_base=1.25, multiplier=1.25),
+)
 def parse_purlobourl(url: str, lookup: ty.Optional[ty.Tuple[str, ...]] = None):
     """Parse an Ontobee URL to return properties of a Class node
 
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -1,10 +1,14 @@
 from datetime import datetime
+from functools import cache
 import os
 import os.path as op
 import re
+import typing as ty
 from uuid import uuid4
+from xml.dom.minidom import parseString
 
 from dandischema import models
+import requests
 
 from . import __version__, get_logger
 from .dandiset import Dandiset
@@ -393,10 +397,8 @@ species_map = [
 ]
 
 
-def parse_purlobourl(url, lookup=["rdfs:label", "oboInOwl:hasExactSynonym"]):
-    from xml.dom.minidom import parseString
-
-    import requests
+@cache
+def parse_purlobourl(url: str, lookup: ty.Tuple[str] = None):
 
     req = requests.get(url, allow_redirects=True)
     doc = parseString(req.text)
@@ -411,6 +413,8 @@ def parse_purlobourl(url, lookup=["rdfs:label", "oboInOwl:hasExactSynonym"]):
     if elfound is None:
         return None
     values = {}
+    if lookup is None:
+        lookup = ("rdfs:label", "oboInOwl:hasExactSynonym")
     for key in lookup:
         elchild = elfound.getElementsByTagName(key)
         if elchild:
@@ -432,7 +436,7 @@ def extract_species(metadata):
                     break
             if value_id is None:
                 value_id = value_orig
-                lookup = ["rdfs:label", "oboInOwl:hasExactSynonym"]
+                lookup = ("rdfs:label", "oboInOwl:hasExactSynonym")
                 try:
                     result = parse_purlobourl(value, lookup=lookup)
                 except ConnectionError:
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -408,15 +408,13 @@ def parse_purlobourl(url: str, lookup: ty.Optional[ty.Tuple[str, ...]] = None):
 
     req = requests.get(url, allow_redirects=True)
     doc = parseString(req.text)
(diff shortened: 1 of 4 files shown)


dandi/metadata.py
```diff
@@ -341,45 +345,128 @@ def extract_sex(metadata):
...
+        ["melanogaster", "fruit fly"],
+        None,
+        "http://purl.obolibrary.org/obo/NCBITaxon_7227",
+        "Drosophila melanogaster - Fruit fly",
+    ),
+]
+
+
+@lru_cache(maxsize=None)
+def parse_purlobourl(url: str, lookup: ty.Tuple[str] = None):
```
jwodder, line 401: The type of `lookup` needs to be `Optional[Tuple[str, ...]]`.

dandi/metadata.py
```diff
@@ -341,45 +345,128 @@ def extract_sex(metadata):
...
-            value = "Common fruit fly"
-        elif value.startswith("http"):
-            value_id = value
-            value = None
+    value_orig = metadata.get("species", None)
+    value_id = None
+    if value_orig is not None and value_orig != "":
+        value = value_orig.lower().rstrip("/")
+        if value.startswith("http://purl.obolibrary.org/obo/NCBITaxon_".lower()):
+            for entry in species_map:
```
jwodder, line 432: Please do `for names, genus, url, species in species_map` (or some other set of names) instead for better readability.

satra:
@jwodder - this is ready to merge. the failing test is unrelated to this i think.

dandi/metadata.py
```diff
@@ -341,45 +345,133 @@ def extract_sex(metadata):
...
+        ["melanogaster", "fruit fly"],
+        None,
+        "http://purl.obolibrary.org/obo/NCBITaxon_7227",
+        "Drosophila melanogaster - Fruit fly",
+    ),
+]
+
+
+@lru_cache(maxsize=None)
+def parse_purlobourl(url: str, lookup: ty.Optional[ty.Tuple[str]] = None):
```
jwodder, line 401: This declares `lookup` to be an optional tuple of one string.  For an arbitrary number of strings, you need to write `Optional[Tuple[str, ...]]`.

yarikoptic:
```suggestion
def parse_purlobourl(url: str, lookup: ty.Optional[ty.Tuple[str, ...]] = None):
```
?

dandi/metadata.py
```diff
@@ -341,45 +345,133 @@ def extract_sex(metadata):
...
-        elif value.startswith("http"):
-            value_id = value
-            value = None
+    value_orig = metadata.get("species", None)
+    value_id = None
+    if value_orig is not None and value_orig != "":
+        value = value_orig.lower().rstrip("/")
+        if value.startswith("http://purl.obolibrary.org/obo/NCBITaxon_".lower()):
+            for common_names, prefix, uri, name in species_map:
+                if value.split("//")[1] in uri:
```
yarikoptic, line 439: why is it `in` here and not exact comparison like

```suggestion
                if value.split("//")[1] == uri.split("//")[1]:
```
so we do not match `http://purl.obolibrary.org/obo/NCBITaxon_954` to `http://purl.obolibrary.org/obo/NCBITaxon_9544` etc?

dandi/metadata.py
```diff
@@ -341,45 +345,133 @@ def extract_sex(metadata):
...
+        ["melanogaster", "fruit fly"],
+        None,
+        "http://purl.obolibrary.org/obo/NCBITaxon_7227",
+        "Drosophila melanogaster - Fruit fly",
+    ),
+]
+
+
+@lru_cache(maxsize=None)
+def parse_purlobourl(url: str, lookup: ty.Optional[ty.Tuple[str, ...]] = None):
```
yarikoptic, line 401: since can raise `ConnectionError` which then causes return of None "above", I wonder if it is worth retrying it a few times at least?  Would need an import of `tenacity` which we use elsewhere and then smth like

```suggestion
@tenacity.retry(
    reraise=True,
    stop=tenacity.stop_after_attempt(3),
    wait=tenacity.wait_exponential(exp_base=1.25, multiplier=1.25)
)
def parse_purlobourl(url: str, lookup: ty.Optional[ty.Tuple[str, ...]] = None):
```
?

dandi/metadata.py
```diff
@@ -341,45 +345,131 @@ def extract_sex(metadata):
...
+                    value = None
+                else:
+                    value = None
+                    if result is not None:
+                        value = " - ".join(
+                            [result[key] for key in lookup if key in result]
+                        )
         else:
+            for common_names, prefix, uri, name in species_map:
+                if any([key in value for key in common_names]) or (
```
yarikoptic, line 456: very minor to save entire 2 bytes!
```suggestion
                if any(key in value for key in common_names) or (
```

yarikoptic:
Looks LGTM now.  Thank you @satra and sorry for trickling individual comments without aggregating them into a review.

satra:
@jwodder and @yarikoptic - feel free to merge. i think the codecov thing is out of wack a bit.

dandi/metadata.py
```diff
@@ -341,45 +346,136 @@ def extract_sex(metadata):
...
-        elif value.startswith("http"):
-            value_id = value
-            value = None
+    value_orig = metadata.get("species", None)
+    value_id = None
+    if value_orig is not None and value_orig != "":
+        value = value_orig.lower().rstrip("/")
+        if value.startswith("http://purl.obolibrary.org/obo/NCBITaxon_".lower()):
+            for common_names, prefix, uri, name in species_map:
+                if value.split("//")[1] == uri.split("//")[1]:
```
yarikoptic, line 443: Shouldn't we have
```suggestion
                if value.split("//")[1] == uri.lower().rstrip("/").split("//")[1]:
```
since we do that for value?. And since our urls do have upper case letters, I wonder now how tests passed (I can't see coverage report if next lines are covered)

satra: it's possible it never matched and went on to the next part which queried the url.

PR #859, opened 2022-01-10
Assorted code cleanups

jwodder:


yarikoptic:
Cool, thank you!  Glanced through and nothing triggered an immediate concern, and tests passed, so should be all good.

PR #860, opened 2022-01-10
Add a wrapper function for the calls to `find_files()` in tests

jwodder:
So that I don't have to keep writing `sorted(map(Path, find_files(r".*", [dirpath])))`


yarikoptic:
coolio -- I am all for reducing code duplication!

PR #861, opened 2022-01-10
Replace the use of pyfakefs with a `tmp_home` fixture

jwodder:
This eliminates the dependency on the slow and occasionally-troublesome pyfakefs in favor of just monkeypatching `$HOME` to a temporary directory when testing out keyring support.


yarikoptic:
looks LGTM and I am all for faster/more lightweight testing! Thank you!

PR #863, opened 2022-01-11
Include "dandiset" field in /uploads/initialize/ requests

### LINKED ISSUE #856: Add support for uploading embargoed files
The dandi-api changes are currently still in draft PRs, but the external API shouldn't change.

* [x] https://github.com/dandi/dandi-api/pull/643 adds an `embargo_status` field to the dandiset list and fetch endpoints. If `embargo_status == "EMBARGOED"`, then any assets should be uploaded as embargoed assets. See https://github.com/dandi/dandi-api/blob/65dc262ea565d03b089cd435c8a0d912638a95ae/doc/design/embargo-mvp.md?plain=1#L64-L74 and https://github.com/dandi/dandi-api/blob/65dc262ea565d03b089cd435c8a0d912638a95ae/doc/design/embargo-mvp.md?plain=1#L108
* When uploading an embargoed asset, `{..., "embargoed_dandiset": "123456"}` should be included in the JSON request body. This will begin an embargoed upload. No other upload endpoints need this additional argument. See https://github.com/dandi/dandi-api/blob/65dc262ea565d03b089cd435c8a0d912638a95ae/doc/design/embargo-mvp.md?plain=1#L142-L155
* Creating an asset from the blob_id will work exactly the same as normal uploads, but this has not yet been implemented.

edit my @yarikoptic, other relevant PRs
- [ ] https://github.com/dandi/dandi-api/pull/647

jwodder:
Closes #856.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -1103,6 +1103,7 @@ class RemoteDandiset:
                         "algorithm": "dandi:dandi-etag",
                         "value": filetag,
                     },
+                    "dandiset": self.identifier,
                 },
             )
         except requests.HTTPError as e:


yarikoptic:
thank you @jwodder ! additional fresh reference to support providing dandiset even for non-embargoed datasets: https://github.com/dandi/dandi-api/pull/647#discussion_r783189148  .  But I think then we should 
- schedule this PR for `release` (added)
- and declare all prior releases bad in `.et`
or should we wait to do the latter some time later for some reason?

jwodder:
@yarikoptic I know of no reason to wait to update `.et`.  Should that be done in this PR?  Also, would we want to update [the cli-minimal-version reported by the redirector](https://github.com/dandi/redirector/blob/8c3d997a422ef9661c2e0619e828e32be9e38446/serve.py#L167)?

(Side note: [etelemetry is currently broken and useless on Python 3.9+](https://github.com/sensein/etelemetry-client/issues/37); although a fix has been merged, it has yet to be released.  @satra Are you a maintainer of etelemetry?  Can you make a release or kick someone who can?)

yarikoptic:
I am curious also how et will use et to announce user that et needs to be upgraded on 3.9? Sounds like chicken and egg issue. 
May be we could just rely on 'the cli-minimal-version reported by the redirector' as for deciding if client is good enough? Sooner or later we should stop relying on et anyways to ensure consistency - it should be based on information obtained from the specific API server interacting with directly anyways.

satra:
@jwodder - new release of etelemetry is on pypi.

yarikoptic:
Let's merge, and IIRC how et works, we will announce bad versions after release is out

PR #866, opened 2022-01-17
Update bad versions list in `.et`

jwodder:
See discussion in https://github.com/dandi/dandi-cli/pull/863.


yarikoptic:
I guess let's do it and see if it would help us

PR #870, opened 2022-01-20
Make navigate() and navigate_url() auto-authenticate for embargoed Dandisets

### LINKED ISSUE #867: Call dandi_authenticate() during dandi download
Presently, `dandi_authenticate()` is only done on modification operations, i.e. `upload` or `delete`. To read embargoed dandisets, the API server needs to know who is requesting data, so `dandi_authenticate()` also needs to be called during `download`.

jwodder:
Closes #867.

Nonfunctional without https://github.com/dandi/dandi-archive/pull/831.

### DIFF
--- a/dandi/dandiarchive.py
+++ b/dandi/dandiarchive.py
@@ -29,7 +29,7 @@ from abc import ABC, abstractmethod
 from contextlib import contextmanager
 import re
 from time import sleep
-from typing import Iterator, Optional, Tuple
+from typing import Iterable, Iterator, Optional, Tuple
 from urllib.parse import unquote as urlunquote
 
 from pydantic import AnyHttpUrl, BaseModel, parse_obj_as, validator
@@ -129,9 +129,9 @@ class ParsedDandiURL(ABC, BaseModel):
 
     @contextmanager
     def navigate(
-        self, strict: bool = False
+        self, strict: bool = False, authenticate: Optional[bool] = None
     ) -> Iterator[
-        Tuple[DandiAPIClient, Optional[RemoteDandiset], Iterator[BaseRemoteAsset]]
+        Tuple[DandiAPIClient, Optional[RemoteDandiset], Iterable[BaseRemoteAsset]]
     ]:
         """
         A context manager that returns a triple of a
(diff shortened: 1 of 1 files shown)


dandi/dandiarchive.py
```diff
@@ -142,15 +142,27 @@ def navigate(
         If ``strict`` is true, then `get_dandiset()` is called with
         ``lazy=False`` and `get_assets()` is called with ``strict=True``; if
         ``strict`` is false, the opposite occurs.
+
+        If ``authenticate`` is true, then
+        `~dandi.dandiapi.DandiAPIClient.dandi_authenticate()` will be called on
+        the client before returning it.  If it is `None` (the default), the
+        method will only be called if the resource(s) the URL refers to are
+        embargoed.
```
yarikoptic, line 150:
```suggestion
        method will only be called if the request to URL requires authentication
        (e.g., if the resource(s) are embargoed).
```
just to make it a bit closer to the handling, and for future proofing (e.g., if Private/Restricted access model is added).

dandi/dandiarchive.py
```diff
@@ -203,6 +215,27 @@ def get_asset_ids(self, client: DandiAPIClient) -> Iterator[str]:
...
+            dandiset = self.get_dandiset(client, lazy=not strict)
+            try:
+                assets = list(self.get_assets(client, strict=strict))
+            except requests.HTTPError as e:
+                if e.response.status_code == 401 and authenticate is not False:
+                    lgr.info("Resource is embargoed; authenticating ...")
+                    client.dandi_authenticate()
+                    assets = list(self.get_assets(client, strict=strict))
+                else:
+                    raise
```
yarikoptic, line 236: we were in a similar position somewhere else with code duplication, but forgot how we solved it (if we did) -- feels like needing some helper/decorator but its operation would depend on the value of `authenticate` kwarg.  I guess decorator could `get` it from the invocation's `kwargs` (and better to position strict and authenticate after `*` to ensure use of kwarg instead of posarg).  Also relating to the suggestion for docstring, it is just an assumption that it is "embargoed" if `401` -- might be some new feature or a bug etc whenever API would return 401 so might be better to adjust log message to correspond situation better.

dandi/dandiarchive.py
```diff
@@ -339,12 +372,21 @@ def navigate_url(
         If true, then `get_dandiset()` is called with ``lazy=False`` and
         `get_assets()` is called with ``strict=True``; if false, the opposite
         occurs.
+    :param Optional[bool] authenticate:
+        If true, then `~dandi.dandiapi.DandiAPIClient.dandi_authenticate()`
+        will be called on the client before returning it.  If `None` (the
+        default), the method will only be called if the resource(s) the URL
+        refers to are embargoed.
```
yarikoptic, line 379: please adjust similarly to the other docstring to not limit to "embargoed"

yarikoptic (changes requested):
minor requests:
- adjust docstrings to not limit to embargoed
- ideally -- remove code duplication?

dandi/dandiarchive.py
jwodder: I don't feel the code is repeated enough to warrant the complexity of adding a decorator.

Issue #871, opened 2022-01-20
Number of "Text Dandiset" on staging server grows


yarikoptic:
https://gui-staging.dandiarchive.org/#/dandiset/search?search=Test%20Dandiset%20Wodder shows having 3 pages.

I guess they aren't cleaned up/removed properly at the end of the fixture?

jwodder:
@yarikoptic The dandi-cli tests don't touch staging.  These would have to have been created by the dandi/dandisets backups2datalad tests.

PR #872, opened 2022-01-21
Give a "unique" name to test dandiset

yarikoptic:
Prompted by confusion while tracking down reason for
https://github.com/dandi/dandi-cli/issues/871


jwodder:
@yarikoptic Is the only purpose of this PR to distinguish dandi-cli's test Dandisets from dandisets/tools'?

I would like to hold off on this PR due to the potential for conflicts with #853.  Also, I have a plan for further refactoring of the test Dandiset fixtures which I plan to do after that one is merged; the Dandisets can all easily be given actually unique names then.

yarikoptic:
indeed -- just to give easy means to identify the "sideeffect artifacts".  sounds good! let's just close then.

PR #874, opened 2022-01-21
Add `new_dandiset` fixture

jwodder:
This PR factors out the repeated code for creating a new test Dandiset and uploading files to it into a fixture.  It builds on top of/depends on #853 and thus has to be merged after it.

### DIFF
--- a/dandi/tests/fixtures.py
+++ b/dandi/tests/fixtures.py
@@ -62,7 +62,8 @@ def simple1_nwb_metadata(tmpdir_factory):
 @pytest.fixture(scope="session")
 def simple1_nwb(simple1_nwb_metadata, tmpdir_factory):
     return make_nwb_file(
-        str(tmpdir_factory.mktemp("data").join("simple1.nwb")), **simple1_nwb_metadata
+        str(tmpdir_factory.mktemp("simple1").join("simple1.nwb")),
+        **simple1_nwb_metadata,
     )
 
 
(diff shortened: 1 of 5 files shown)


dandi/tests/fixtures.py
```diff
@@ -296,75 +297,56 @@ def upload(self, paths=None, **kwargs) -> None:
...
+    (new_dandiset.dspath / "subdir1" / "apple.txt").write_text("Apple\n")
+    (new_dandiset.dspath / "subdir2").mkdir()
+    (new_dandiset.dspath / "subdir2" / "banana.txt").write_text("Banana\n")
+    (new_dandiset.dspath / "subdir2" / "coconut.txt").write_text("Coconut\n")
+    new_dandiset.upload_kwargs["allow_any_path"] = True
+    new_dandiset.upload()
+    return new_dandiset
+
+
+@pytest.fixture()
```
yarikoptic, line 343: I wonder if both `text_` and `zarr_` dandisets should become `(scope="session")` since they look like the ones to be reused for read-only purposes. And have some dedicated `new_zarr_dandiset` (and similar `new_text_dandiset`) with scope to the test for the use cases where needed to mutate?

jwodder: I think that would only be a good idea if we could somehow prevent the read-only Dandiset fixtures from being modified (both on disk and on the server); otherwise, it'd be too easy to accidentally use the wrong fixture and cause a test failure that ends up being hard to debug.  I don't think the repeated recreating of Dandisets slows the tests down all that much.

yarikoptic: yeah, inability to guarantee immutability also worried me.  I even was reminded today about https://github.com/pytest-dev/pytest/issues/5044 after I thought that such "presumably a read only use of dandist" could be tested in fixture "turndown" code -- then we could catch cases where a test unexpectedly modifies a dandiset.  I was thinking that there could be 
- code which creates a new dandiset
- test scoped fixture which creates new dandiset only if doesn't exist, yields to the test, verifies that test did not change it.  If it did -- destroys pre-created dandiset (so it would be created a new on next run) and `.fail`s the test with a message that immutability assumption was modified.

But I guess it might be all too much for this PR, so I am ok to proceed with this PR as is unless you see interest in above.

jwodder: Let's stick with the PR as it is now.

PR #877, opened 2022-01-25
Enable type-checking

jwodder:
This PR adds a tox testenv and CI action for type-checking the code with mypy.  Note that much of the code base still does not have type annotations, which mypy is fine with; the plan is to annotate everything piecemeal so as not to end up with a massive PR that would likely create merge conflicts with everything else.


yarikoptic:
Great, thank you!

PR #887, opened 2022-01-27
Add type-annotations to various files

jwodder:
Builds on top of #877.

### DIFF
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -805,7 +805,7 @@ class DownloadDirectory:
 
     def append(self, blob: bytes) -> None:
         if self.fp is None:
-            raise ValueError("DownloadDirectory.append() called outside of context")
+            raise ValueError("DownloadDirectory.append() called outside of context manager")
         self.fp.write(blob)
 
 
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -15,21 +15,43 @@ from shutil import rmtree
 import sys
 from threading import Lock
 import time
-from typing import Any, Callable, Dict, Iterator, Optional, Tuple
+from types import TracebackType
+from typing import (
+    IO,
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    Iterator,
+    List,
+    Optional,
+    Sequence,
+    Tuple,
+    Type,
+    Union,
+)
 
 from dandischema.models import DigestType
+from fasteners import InterProcessLock
 import humanize
 from interleave import FINISH_CURRENT, interleave
 import requests
 
 from . import get_logger
 from .consts import RETRY_STATUSES, dandiset_metadata_file
-from .dandiapi import AssetType, RemoteZarrAsset
-from .dandiarchive import DandisetURL, MultiAssetURL, SingleAssetURL, parse_dandi_url
+from .dandiapi import AssetType, RemoteAsset, RemoteDandiset, RemoteZarrAsset
+from .dandiarchive import (
+    DandisetURL,
+    MultiAssetURL,
+    ParsedDandiURL,
+    SingleAssetURL,
+    parse_dandi_url,
+)
 from .dandiset import Dandiset
 from .exceptions import NotFoundError
-from .files import DandisetMetadataFile, find_dandi_files
+from .files import LocalAsset, find_dandi_files
 from .support.digests import get_digest, get_zarr_checksum
+from .support.iterators import IteratorWithAggregation
 from .support.pyout import naturalsize
 from .utils import (
     abbrev_prompt,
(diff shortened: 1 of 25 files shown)


dandi/metadata.py
```diff
@@ -176,12 +194,14 @@ def _check_decimal_parts(age_parts):
             age_parts[-1],
             flags=re.I,
         )
+        if m is None:
+            raise ValueError(age_parts)
```
yarikoptic, line 198:
```suggestion
            raise ValueError(f"Failed to parse the trailing part of age {age_parts[-1]!r}")
```
or something like that
- other parts have no relation to parsing of the last one
- to provide a hint on what/why the value is wrong

may be even worth placing that regex into a variable and including it in the error message to make it even more informative

dandi/download.py
```diff
@@ -757,22 +795,24 @@ def __exit__(self, exc_type, exc_value, traceback):
...
                 rmtree(self.dirpath, ignore_errors=True)
             self.lock = None
             self.fp = None
             self.offset = None
-        return False
 
-    def append(self, blob):
+    def append(self, blob: bytes) -> None:
+        if self.fp is None:
+            raise ValueError("DownloadDirectory.append() called outside of context")
```
yarikoptic, line 808:
```suggestion
            raise ValueError("DownloadDirectory.append() called outside of context manager")
```
just to make it clearer

dandi/metadata.py
```diff
@@ -467,35 +486,40 @@ def extract_species(metadata):
...
 
 
-def extract_model(modelcls, metadata, **kwargs):
-    m = modelcls.unvalidated()
+def extract_model(modelcls: Type[M], metadata: dict, **kwargs: Any) -> M:
+    m = cast(M, modelcls.unvalidated())
     for field in m.__fields__.keys():
         value = kwargs.get(field, extract_field(field, metadata))
-        if value is not Ellipsis:
+        if value is not None:
```
yarikoptic, line 513: I wonder (didn't do archaeologic expedition) and have some vague memory that we chose `Ellipsis` for some reason, i.e. that value could really be `None` and thus assigned.  Do you remember/know the reason why Ellipsis was used and why it could be RFed safely now to become good simple `None`?

dandi/metadata.py
```diff
@@ -574,14 +600,27 @@ def extract_digest(metadata):
...
     if field in FIELD_EXTRACTORS:
         return FIELD_EXTRACTORS[field](metadata)
     else:
-        return metadata.get(field, ...)
+        return metadata.get(field)
 
 
-neurodata_typemap = {
+if TYPE_CHECKING:
+    if sys.version_info[:2] >= (3, 8):
```
yarikoptic, line 611: It should work as a simpler
```suggestion
    if sys.version_info >= (3, 8):
```
right?

dandi/tests/fixtures.py
```diff
@@ -117,13 +131,32 @@ def organized_nwb_dir2(simple1_nwb_metadata, simple2_nwb, tmp_path_factory):
     return tmp_path
 
 
-def get_gitrepo_fixture(url, committish=None, scope="session"):
+if TYPE_CHECKING:
+    if sys.version_info[:2] >= (3, 8):
```
yarikoptic, line 135: same here simpler
```suggestion
    if sys.version_info >= (3, 8):
```
?

dandi/upload.py
```diff
@@ -1,36 +1,49 @@
...
 
 from . import lgr
 from .consts import DRAFT, dandiset_identifier_regex, dandiset_metadata_file
 from .exceptions import NotFoundError
-from .files import DandisetMetadataFile, LocalAsset, find_dandi_files
+from .files import DandiFile, DandisetMetadataFile, LocalAsset, find_dandi_files
 from .utils import ensure_datetime, get_instance, pluralize
 
+if TYPE_CHECKING:
+    if sys.version_info[:2] >= (3, 8):
```
yarikoptic, line 19: same here -- simpler
```suggestion
    if sys.version_info >= (3, 8):
```
?

dandi/utils.py
```diff
@@ -19,20 +12,34 @@
...
 from semantic_version import Version
 
-#
-# Additional handlers
-#
 from . import __version__, get_logger
 from .consts import DandiInstance, known_instances, known_instances_rev
 from .exceptions import BadCliVersionError, CliVersionTooOldError
 
+if sys.version_info[:2] >= (3, 8):
```
yarikoptic, line 38:
```suggestion
if sys.version_info >= (3, 8):
```

yarikoptic (commented):
Looks good. left some minor recommendations/questions

dandi/metadata.py
jwodder: As far as I recall, `Ellipsis` was used because we wanted extracted fields that evaluated to `None` to still be assigned `None`.  However, this was apparently unnecessary, as all of the fields in question default to `None` anyway when unset, so not assigning extracted `None`s doesn't create any problems.

Also, for the record, the reason I had to drop the use of `Ellipsis` here was because it can't be used as a type in annotations prior to Python 3.10.

yarikoptic:
Thank you @jwodder , let's proceed!

PR #891, opened 2022-01-29
Fix neural data type to metadata extractor

### LINKED ISSUE #890: improper metadata extraction for neural data types
this dataset (https://dandiarchive.org/dandiset/000220/draft) does not seem to correspond to any specific type. unlike other nwb dandisets where the assets summary does have information about approach/modality etc.,. this one doesn't.

make me think that we are either not covering something in the dandi-cli parser or the info was added in the inappropriate place.

> from @bendichter: regarding 
This dataset contains the neurodata type "CurrentClampSeries", which is correctly used to infer that the data is type "icephys". I'm not sure why that neurodata type itself isn't being pulled into the metadata automatically, but I guess it must be an error with the dandi-cli.

located the bug. PR forthcoming.

satra:
closes #890


dandi/metadata.py
```diff
@@ -776,6 +776,7 @@ def process_ndtypes(asset, nd_types):
     technique = set()
     variables = set()
     for val in nd_types:
+        val = val.split()[0]
```
jwodder, line 779: You should first check that `val` is neither empty nor all-whitespace in order to avoid failures on indexing.

dandi/tests/test_metadata.py
```diff
@@ -359,3 +360,283 @@ def test_species():
...
+    "ndtypes,asset_dict",
+    [
+        (
+            ["ElectricalSeries"],
+            {
+                "approach": ["electrophysiological approach"],
+                "measurementTechnique": [
+                    "multi electrode extracellular electrophysiology recording technique"
+                ],
+                "variableMeasured": ["ElectricalSeries"],
```
jwodder, line 375: What's the point of these dict values being lists if they're all just one element long?

dandi/tests/test_metadata.py
```diff
@@ -359,3 +360,283 @@ def test_species():
...
+        encodingFormat="application/x-nwb",
+        digest={"dandi:dandi-etag": "0" * 32 + "-1"},
+        path="test.nwb",
+    )
+    asset = process_ndtypes(asset, ndtypes)
+    for key in ["approach", "measurementTechnique"]:
+        if asset_dict.get(key) is None:
+            assert getattr(asset, key) == []
+        else:
+            assert getattr(getattr(asset, key)[0], "name") == asset_dict.get(key)[0]
```
jwodder, line 640: Instead of `getattr(getattr(asset, key)[0], "name")`, just write `getattr(asset, key)[0].name`, and likewise for line 642.

dandi/metadata.py
satra: i don't think nd_types can be empty or all whitespace by definition of pynwb. we can put checks, but i don't like necessarily checking things like this.

dandi/tests/test_metadata.py
satra: they could be extended in the future. i most realistic settings they will be more than one element long.

dandi/metadata.py
yarikoptic, line 779: might be weekend/monday and the cold, but I still find it difficult to identify from the long (and seemingly not specific to the issue) added test why such splitting solves some problem.  @satra could you please digest it for me a bit?

PR #893, opened 2022-01-31
Cancel a Zarr upload if an error occurs partway through

jwodder:


dandi/files.py
```diff
@@ -827,13 +827,25 @@ def iter_upload(
...
-                        size = fut.result()
-                        bytes_uploaded += size
-                        yield {
-                            "status": "uploading",
-                            "upload": 100 * bytes_uploaded / stat.size,
-                            "current": bytes_uploaded,
-                        }
+                        try:
+                            size = fut.result()
+                        except Exception as e:
```
yarikoptic, line 832: could they be retried?  During the upload of millions of files for days some intermittent failures are probably unavoidable, so dandi-cli should retry whenever possible. Looking at `_upload_zarr_file` I don't see any retries.  Could you please check if those PUT requests could be retried?

jwodder: `_upload_zarr_file()` uses `RESTFullAPIClient`, which already automatically retries on 500, 502, 503, and 504.  Are there any other statuses you want retried?

yarikoptic: ah, cool.  not sure yet of any extra.  I guess we could add later after/if we encounter any (e.g `206 Partial Content (RFC 7233)` although that one might be relevant only for GETs).

Then let's proceed.  I think interrupting early would be beneficial ATM to not drag too long to discover a problem.

PR #896, opened 2022-02-01
Adding BIDS support

yarikoptic:
- [x] include stock and our DANDI tuned schemas
- [ ] provide needed support code for intermediate BIDS dedicated validation command (e.g. `dandi bids-validate [--bids-version VERSION] TOPOFBIDSDATASET [PATHs*]`)
   - [x] if `--bids-version` is not provided -- take the one within `dataset_description.json`, if provided -- use that version
   - [ ] parse entities for metadata
   - [x] validate filenames according to schema
       - [ ] corresponding dandisets should have "proper" BIDSVersion set to point to one or another version of the BIDS schema 
   - [ ] load extra metadata from sidecar files
- [ ] (might be separate issue/PR) together with @yarikoptic and @jwodder figure out how to integrate bids-validation seamlessly into `dandi validate`. Tricky aspects
   - [x] discovery of paths -- if bids dataset, we should probably just allow all files (besides in `.dotdirs`) not just `.nwb` since BIDS has all kinds of .json, .tsv, etc.  Hint: we can deduce all possible to be present extensions from the schema!
   - [ ] need to operate on paths with awareness of the top level . Actually should have been done also for DANDI layout so we do not have smth like `customsum/sub-01/...`


TheChymera:
As of 48c4992 this should be usable for simple applications with the dandi-extended BIDS schema.

You can get this enhancement branch with:

```console
git clone git@github.com:dandi/dandi-cli.git
cd dandi-cli
git fetch
git checkout -b enh-bids origin/enh-bids
```
and installing from the source.

You should be able to perform a basic BIDS validation of filenames with, e.g.
```console
ANDI_DEVEL=1 dandi validate-bids --schema="1.6.0+dandi001" /path/to/your/bids/dataset/like/000026/rawdata
```

This will produce a `bids-validator-report_*.log` file in your current directory, reporting which regex patterns have been used, which files could not be matched, and which *required* files were not found.

More features will become available shortly, if you experience any issues, feel free to report them here.

TheChymera:
@yarikoptic a general issue resulting from the fact that we have tried to re-use functions from `bids-specification` (i.e. `schemacode`) in as far as feasible, is that this PR will need to ship 2 files from that package. These files might diverge over time... ideally we would push for this to be included upstream and call the function from there. This might constrain how much the function should be dandified, so ideally we would set up some compatibility code in `validate.py` instead of modifying it too much.

yarikoptic:
to expedite bringing this functionality to dandi-cli, I would not worry about establishing the best/ultimate solution long term. long term -- we will continue participation in the next generation of bids-validator/validation based on the schema and strive to (re)use whatever would come from BIDS.  Meanwhile we could
- be half fancy and use `git subtree` (see some instructions in now abandoned [dandi-core](https://github.com/dandi/dandi-core#developers-information) on howto) functionality to bring the entire relevant folder from bids-specification repo here while allowing for "proper" updates later using `git` to bring in changes
- `subtree` might be more pain than gain ATM, so just
  - copy and commit needed files (as-is, no changes whatsoever) in a dedicated folder/submodule
  - initiate branch "3rd-party" (or alike) at that point
  - adjust anything needed to be adjusted in them in this `enh-bids` branch
  - this way later on we can either "cherry-pick" (manually) desired patches from here into bids-specification (while adjusting paths if needed), or easily update our "copy" with fresh one from bids-specification by doing that in "3rd-party" branch and merging it back into `enh-bids` or whatever other branch at a later point

LeeKamentsky:
Hi all.. using this - seeing it's a pull request, I'll report a problem here instead of reporting a bug.
I created a fresh Anaconda environment and ran "pip install ." to install dandi-cli and its dependencies. I ran `dandi validate-bids ...` and got:
```
  File "/mnt/beegfs/Lee/envs/dandi-cli/bin/dandi", line 33, in <module>
    sys.exit(load_entry_point('dandi', 'console_scripts', 'dandi')())
  File "/mnt/beegfs/Lee/envs/dandi-cli/lib/python3.10/site-packages/click/core.py", line 1128, in __call__
  File "/mnt/beegfs/Lee/envs/dandi-cli/lib/python3.10/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/mnt/beegfs/Lee/envs/dandi-cli/lib/python3.10/site-packages/click/core.py", line 1659, in invoke
  File "/mnt/beegfs/Lee/envs/dandi-cli/lib/python3.10/site-packages/click/core.py", line 1395, in invoke
  File "/mnt/beegfs/Lee/envs/dandi-cli/lib/python3.10/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/mnt/beegfs/Lee/envs/dandi-cli/lib/python3.10/site-packages/click/decorators.py", line 38, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/home/leek/src/dandi-cli/dandi/cli/base.py", line 102, in wrapper
    return f(*args, **kwargs)
  File "/home/leek/src/dandi-cli/dandi/cli/cmd_validate.py", line 25, in validate_bids
    validate_bids_(
  File "/home/leek/src/dandi-cli/dandi/validate.py", line 25, in validate_bids
    from .bids_validator_xs import load_all, validate_all, write_report
  File "/home/leek/src/dandi-cli/dandi/bids_validator_xs.py", line 5, in <module>
    import yaml
ModuleNotFoundError: No module named 'yaml'
```
I'm currently looking at how to get that dependency (pyyaml?) which I'm assuming should have been installed during setup.

satra:
@TheChymera - perhaps bids-validator can use the same yaml library that dandi uses, which is `ruamel.yaml`

TheChymera:
@LeeKamentsky thank you for giving this a try. `pyyaml` will not be needed as of c488a73f3109f4a07bdbe3b79384d75cb5939664 . Let us know if you were able to give it another try.

LeeKamentsky:
@TheChymera it seems to be running. Unfortunately, we are having a metadata slowdown on our storage server and going through all of the .ngff files has prevented it from completing.

TheChymera:
Hi @LeeKamentsky , during our DANDI meetup today we realized the `.ngff` files (traversable as directories) should not, in fact be traversed. If it's still running it might be best to just interrupt it and try again after we address this. Could you send me an `.ngff` file from your collection so I can test it locally before I push (ideally a small one)?

satra:
@TheChymera - you should be able to create arbitrary files for yourself with something like this and the ome-zarr-py library.

```python
import zarr
import numpy as np

from ome_zarr.writer import write_image

store = zarr.NestedDirectoryStore('test.ngff')
root = zarr.group(store=store, overwrite=True)

mean_val=10
rng = np.random.default_rng(0)
data = rng.poisson(mean_val, size=(1, 1, 100, 100, 100)).astype(np.uint8)
write_image(image=data, group=root)
```

TheChymera:
@LeeKamentsky apologies for the delay, I had tried to do this in a smarter fashion, but it will take a while before that is done. There is an interim fix available as of the last commit, which should perform just as well.

yarikoptic:
I thought that first you would implement `dandi bids-validate [--bids-version VERSION] TOPOFBIDSDATASET [PATHs*]` thus TOPOFBIDSDATASET would be known/specified.  Decided not to bother with a dedicated command?

As for integration with `dandi validate` if you would like to autodetect dataset being a BIDS dataset, then checking for REQUIRED `dataset_description.json` is IMHO Ok, since that file also would then be used to determine BIDSVersion to validate against.  Without that file, there is no BIDS version to validate against.  Making it all loose regexes (`3.`) would just make validation more "forgiving" and thus will not catch misspecifications, thus IMHO should be avoided.

satra:
> autodetect dataset being a BIDS dataset

in dandi a dandiset could have multiple bids datasets, so one should check to for this file to determine the roots of each dataset, and mark any top level folders at the same level as a bids dataset that doesn't have this file an error. i think 26 will be a good use case for this.

yarikoptic:
@satra I thought it was temporary and we would like to aim to have separate (but linked via metadata relationship) bids datasets. If we allow multiple into the same dandiset, IMHO makes validation etc less straightforward and arbitrary

satra:
i don't think they would appreciate that since the stereological data (extracted by humans from the microscopy data) would be completely separated from the source data. and bids provides a standard way of having a data collection have multiple datasets. also some of the combined images like T1maps would be separated. given the small size of the MR components, it would seem weird to do so in this case. i.e. the practical downside outweighs the cleanliness.

but navigating trees to find dataset_descriptions should not be hard. so we should try to support this internally. it may catch several issues as well.

yarikoptic:
> and bids provides a standard way of having a data collection have multiple datasets

what do you have in mind ?

satra:
> what do you have in mind ?

did you see the bids dataset referencing PR ? this doesn't mean it can't be done separately. but the bids-specification also has suggestions for mixing rawdata and derivatives, and 26 is not very different from it.

yarikoptic:
> > what do you have in mind ?
> 
> did you see the bids dataset referencing PR ?

brr... but I guess you mean if I saw https://gui.dandiarchive.org/#/dandiset/000026/draft ;) yes, I did.  It is not a BIDS compliant dataset.  It does contain possibly BIDS compliant datasets.

> but the bids-specification also has suggestions for mixing rawdata and derivatives, and 26 is not very different from it.

I guess you mean the https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#source-vs-raw-vs-derived-data section?    Note that it is really just a suggestion on how to organize multiple datasets, which doesn't really make place it into some compliance with BIDS, and it says that: *"The above example is just a convention that can be useful for organizing raw, source, and derived data while maintaining BIDS compliance of the raw data folder. "*.  As such, we can't even then validate the entire dandiset compliance to some kind of standard (e.g., BIDS) or "agreement" (our DANDI naming convention), unless we indeed formalize some DANDI way to compose multiple (BIDS or not) dandisets together (which could follow aforementioned BIDS recommendation).

NB. asking a question on "rawdata/": https://github.com/bids-standard/bids-specification/issues/1009 .

TheChymera:
NOTE
Just for personal reference, there will remain some lint failures as this is being tracked together with https://github.com/bids-standard/bids-specification/pull/969/ which has different conventions (see `Analogous to:` commit lines) — to be fixed after relevant features are in and right before pulling.

TheChymera:
@yarikoptic @jwodder @satra I think this might be ready for a review.
Getting some new and exciting test issues on ubuntu:
```
requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: http://localhost:8000/api/zarr/b07b7d51-d670-4895-a509-53f1c2cc29cd/upload/
```
but it looks like a server accessibility issue unrelated to this PR.

yarikoptic:
@TheChymera please take PR out of Draft mode if you think it is ready for review/merge ;-)

dandi/tests/expected_bids_validator_xs_write.log
```diff
@@ -0,0 +1,7 @@
+
```
jwodder, line 1: Please place this file and the other `*.log` file under `dandi/tests/data` (possibly in a dedicated subdirectory).

jwodder (changes requested):

jwodder:
@yarikoptic @TheChymera At what point will this be integrated into the rest of the asset validation code — as part of this PR or a separate PR?

yarikoptic:
Let's do as a separate PR, since I think we better also RF how we collect/report all warnings/errors while at it (#943) and also have nwbinpsector possibly also in at that point (#941) so we could right away integrate/harmonize across all such 3 validations (dandischema, BIDS, nwbinspector)

yarikoptic:
ok, let's proceed since @jwodder approved.

PR #900, opened 2022-02-04
Add zarr-checksum method to `dandi digest`

### LINKED ISSUE #881: Add support for Zarr checksums to `dandi digest`

jwodder:
Closes #881.

### DIFF
--- a/dandi/support/digests.py
+++ b/dandi/support/digests.py
@@ -86,12 +86,9 @@ def get_digest(filepath: Union[str, Path], digest: str = "sha256") -> str:
     if digest == "dandi-etag":
         return cast(str, get_dandietag(filepath).as_str())
     elif digest == "zarr-checksum":
-        path = Path(filepath)
-        if path.is_dir():
-            return get_zarr_checksum(path)
-        else:
-            digest = "md5"
-    return Digester([digest])(filepath)[digest]
+        return get_zarr_checksum(Path(filepath))
+    else:
+        return Digester([digest])(filepath)[digest]
 
 
 @checksums.memoize_path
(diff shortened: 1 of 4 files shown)


dandi/support/digests.py
```diff
@@ -85,8 +85,13 @@ def __call__(self, fpath: Union[str, Path]) -> Dict[str, str]:
...
     if digest == "dandi-etag":
         return cast(str, get_dandietag(filepath).as_str())
-    else:
-        return Digester([digest])(filepath)[digest]
+    elif digest == "zarr-checksum":
+        path = Path(filepath)
+        if path.is_dir():
+            return get_zarr_checksum(path)
+        else:
+            digest = "md5"
```
yarikoptic, line 93: shouldn't this logic just fold into `get_zarr_checksum` since IMHO it leaks implementation detail (of "zarr-checksum") into outside code.

jwodder: The only implementation detail here is the use of MD5.  `get_zarr_checksum()` is only meant to operate on directories.

yarikoptic: so it remains confusing why `get_digest(digest="zarr-checksum")` can operate on a file while `get_zarr_checksum()` can't. IMHO should be consistent one way or another. Do we need to allow `get_digest(digest="zarr-checksum")` to operate on a file or could raise an error?

jwodder: > Do we need to allow `get_digest(digest="zarr-checksum")` to operate on a file or could raise an error?

I don't really care either way.

jwodder: `get_zarr_checksum()` already raises an error if it's passed a file.  Did you mean `get_digest(path, digest = "zarr-checksum")`?

yarikoptic: yes. why should it behave differently?

jwodder: Because `get_digest()` is called by `dandi digest`, and users are likely to try to run `dandi digest -d zarr-checksum` on a file.

yarikoptic: so -- if the checksum is defined for `zarr` then it should also be reported by `get_zarr_checksum`.  If it is not defined -- neither should report it.  If we decide to go "not defined" (keep exception in `get_zarr_checksum`) then `dandi digest -d zarr-checksum` should error out

jwodder: `get_zarr_checksum()` now accepts file paths.

yarikoptic:
Thank you @jwodder, let's proceed!

PR #907, opened 2022-02-10
Mint Zarr assets at start of upload and sync Zarr contents

### LINKED ISSUE #884: Modify Zarr assets in-place when uploading to an extant Zarr path
Currently, when a Zarr is uploaded to a remote path where a Zarr asset already exists, a new Zarr asset is created from scratch, which involves uploading all of the files in the Zarr.  The client should instead optimize for the common case (a Zarr being reuploaded after some local modifications) by modifying the Zarr asset already on the server, only uploading new & changed Zarr entries and deleting any old ones.

CC @dchiquito: Just to confirm, this is a reasonable thing to implement, right?  Can a Zarr object be modified after it's made into an asset, and does the server correctly handle modification of Zarr data that multiple assets point to?

### LINKED ISSUE #906: provide "update zarr" functionality
zarr folders are heavy, and likely to fail to upload at any specific point in time and would need to be "continued" or "fixed up" (that is our C2 in #903).

Also relates to https://github.com/dandi/dandi-cli/issues/903 as possibly solution might incorporate "in band" compute of the checksum across files to be uploaded (or not).

We need to

- generate an asset immediately when we initiate a new `zarr_id` -- that would associate a given dandiset (id) and zarr path with a `zarr_id` so we could re-upload
- when we encounter already existing zarr asset for the path, we should not initiate a new zarr but modify existing
  - unless it is completely empty (all of those have the same checksum; but also size 0 and filenumber 0) we need to
    - get a list of files in the zarr
    - get a list of files locally
    - for those which are present on both, compare checksums
    - "schedule" the ones missing or different on the server to be uploaded
    - after uploads complete, remove files which are not present locally but present remotely
  - if it was completely empty, we can avoid getting a list and assume it is `[]` and proceed with the same algorithm really, so no need of special casing

jwodder:
Closes #884.
Closes #906.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -1591,9 +1591,10 @@ class RemoteZarrEntry(BasePath):
     #: The ID of the Zarr backing the asset
     zarr_id: str
     _known_dir: Optional[bool] = field(default=None, compare=False, repr=False)
+    _digest: Optional[Digest] = field(default=None, compare=False, repr=False)
 
     def _get_subpath(
-        self, name: str, isdir: Optional[bool] = None
+        self, name: str, isdir: Optional[bool] = None, checksum: Optional[str] = None
     ) -> "RemoteZarrEntry":
         if not name or "/" in name:
             raise ValueError(f"Invalid path component: {name!r}")
@@ -1602,7 +1603,17 @@ class RemoteZarrEntry(BasePath):
         elif name == "..":
             return self.parent
         else:
-            return replace(self, parts=self.parts + (name,), _known_dir=isdir)
+            if checksum is not None and isdir is not None:
+                if isdir:
+                    algorithm = models.DigestType.dandi_zarr_checksum
+                else:
+                    algorithm = models.DigestType.md5
+                digest = Digest(algorithm=algorithm, value=checksum)
+            else:
+                digest = None
+            return replace(
+                self, parts=self.parts + (name,), _known_dir=isdir, _digest=digest
+            )
 
     @property
     def parent(self) -> "RemoteZarrEntry":
@@ -1613,6 +1624,7 @@ class RemoteZarrEntry(BasePath):
                 self,
                 parts=self.parts[:-1],
                 _known_dir=True if self._known_dir is not None else None,
+                _digest=None,
             )
 
     def _isdir(self) -> bool:
@@ -1660,9 +1672,9 @@ class RemoteZarrEntry(BasePath):
         for name in listing.dirnames:
             if name == "." or name == "..":
                 continue
-            yield self._get_subpath(name, isdir=True)
+            yield self._get_subpath(name, isdir=True, checksum=listing.checksums[name])
         for name in listing.filenames:
-            yield self._get_subpath(name, isdir=False)
+            yield self._get_subpath(name, isdir=False, checksum=listing.checksums[name])
 
     def get_digest(self) -> Digest:
         """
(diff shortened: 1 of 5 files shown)


dandi/files.py
```diff
@@ -768,7 +774,8 @@ def iter_upload(
         :param int jobs: Number of threads to use for uploading; defaults to 5
         :param RemoteAsset replacing:
             If set, replace the given asset, which must have the same path as
-            the new asset
+            the new asset; if the old asset is a Zarr, the Zarr will be updated
+            & reused for the new asset
```
yarikoptic, line 778: ha. Originally I thought we would just keep modifying the zarr, and thus leaving the asset the same, not a new one upon each attempt to finish up upload or update it.  API allows for that but I think having a new asset to replace an old one if we are to change zarr, makes sense.

WDYT @satra -- may be I am missing some use case/aspect where we do want to keep the asset to be the same?

dandi/files.py
```diff
@@ -794,26 +801,77 @@ def iter_upload(
...
+        to_upload: List[dict] = []
+        for p in stat.files:
+            pdigest = p.get_digest().value
+            item = {"path": str(p), "etag": pdigest}
+            try:
+                e = old_zarr_entries.pop(str(p))
+            except KeyError:
+                to_upload.append(item)
+            else:
+                if e.is_dir():
```
yarikoptic, line 839: having a unittest to cover this codeblock (just delete some folder and/or files within a zarr and reupload?) would be nice

yarikoptic (changes requested):
looks good. seems in need of test tune up to trigger "delete" logic coverage

dandi/files.py
satra: @jwodder - that's essentially what i wanted. so seems like nothing needs to change. then in that case if the upload crashes, does the CLI check that an existing asset is there, retrieves the zarr id and uploads to that location?

regarding @yarikoptic comment, i don't think assets need to remain the same. however, my point about when a new asset is minted and who mints it still needs to be resolved with respect to updates that the server does (checksum, contentsize, timestamps).

jwodder: @satra 

> then in that case if the upload crashes, does the CLI check that an existing asset is there, retrieves the zarr id and uploads to that location?

Yes.

yarikoptic: > when a new asset is minted and who mints it still needs to be resolved with respect to updates that the server does (checksum, contentsize, timestamps).

indeed a good question, and something worth asking @dchiquito : if we e.g. mint an asset for an empty zarr, then upload into that zarr, would that asset get its record and metadata updated with new size/checksum upon zarr's upload `/complete`? (feels unlikely, but possible, didn't check)

dandi/files.py
jwodder: Done.

satra:
> I'd still be interested in if there's a better way to delete a directory than deleting all its contents individually.

the only thing i can think of is a dandi api request that launches a task that internally lists all objects with a given prefix, and then deletes all of them. this can take a while though for zarr trees and we have to be careful on what that prefix can be for a request.

dandi/dandiapi.py
```diff
@@ -1707,6 +1711,8 @@ def get_digest(self) -> Digest:
 
         :raises NotFoundError: if the path does not exist in the Zarr asset
         """
+        if self._digest is not None:
+            return self._digest
```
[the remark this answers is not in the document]
yarikoptic: so is it the `get_listing` below which would set it?  this is all makes it not obvious, and worth at least a comment here.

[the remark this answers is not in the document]
jwodder: Comment added.

yarikoptic:
@jwodder - please resolve freshly came up conflict and let's merge to carry forward

jwodder:
@yarikoptic Conflict resolved.

PR #908, opened 2022-02-14
Add dedicated HTTP404Error exception class

jwodder:
This cuts down on a lot of simple repetition by adding a dedicated `HTTP404Error` exception class subclassing `requests.HTTPError` that is used specifically for 404 errors.


yarikoptic:
LGTM, thank you!

PR #909, opened 2022-02-14
improve DandiApiClient.__init__ docstring

### LINKED ISSUE #889: DandiAPIClient not working for staging
works as expected:

```python
client = DandiAPIClient()
client.get_dandiset("000055").get_metadata()
```

however this throws an error:

```python
client = DandiAPIClient("https://gui-staging.dandiarchive.org")
client.get_dandiset("101267").get_metadata()
```

```python
---------------------------------------------------------------------------
HTTPError                                 Traceback (most recent call last)
/opt/conda/lib/python3.8/site-packages/dandi/dandiapi.py in _get_data(self)
    636             try:
--> 637                 self._data = self.client.get(f"/dandisets/{self.identifier}/")
    638             except requests.HTTPError as e:

/opt/conda/lib/python3.8/site-packages/dandi/dandiapi.py in get(self, path, **kwargs)
    293         """
--> 294         return self.request("GET", path, **kwargs)
    295 

/opt/conda/lib/python3.8/site-packages/dandi/dandiapi.py in request(self, method, path, params, data, files, json, headers, json_resp, retry_statuses, **kwargs)
    265                 lgr.error("%s: %s", msg, result.text)
--> 266             raise requests.HTTPError(msg, response=result)
    267 

HTTPError: Error 404 while sending GET request to https://gui-staging.dandiarchive.org/dandisets/101267/

During handling of the above exception, another exception occurred:

NotFoundError                             Traceback (most recent call last)
/tmp/ipykernel_316/3875352354.py in <module>
      1 client = DandiAPIClient("https://gui-staging.dandiarchive.org")
----> 2 client.get_dandiset("101267").get_metadata()

/opt/conda/lib/python3.8/site-packages/dandi/dandiapi.py in get_metadata(self)
    848         `dandischema.models.Dandiset` instance
    849         """
--> 850         return models.Dandiset.parse_obj(self.get_raw_metadata())
    851 
    852     def get_raw_metadata(self) -> Dict[str, Any]:

/opt/conda/lib/python3.8/site-packages/dandi/dandiapi.py in get_raw_metadata(self)
    856         """
    857         try:
--> 858             return cast(Dict[str, Any], self.client.get(self.version_api_path))
    859         except requests.HTTPError as e:
    860             if e.response.status_code == 404:

/opt/conda/lib/python3.8/site-packages/dandi/dandiapi.py in version_api_path(self)
    729         Dandiset are made
    730         """
--> 731         return f"/dandisets/{self.identifier}/versions/{self.version_id}/"
    732 
    733     @property

/opt/conda/lib/python3.8/site-packages/dandi/dandiapi.py in version_id(self)
    647         """The identifier for the Dandiset version"""
    648         if self._version_id is None:
--> 649             self._version_id = self.version.identifier
    650         return self._version_id
    651 

/opt/conda/lib/python3.8/site-packages/dandi/dandiapi.py in version(self)
    655         if self._version is None:
    656             if self._version_id is None:
--> 657                 self._get_data()
    658             if self._data is not None:
    659                 for vattr in ["most_recent_published_version", "draft_version"]:

/opt/conda/lib/python3.8/site-packages/dandi/dandiapi.py in _get_data(self)
    638             except requests.HTTPError as e:
    639                 if e.response.status_code == 404:
--> 640                     raise NotFoundError(f"No such Dandiset: {self.identifier}")
    641                 else:
    642                     raise

NotFoundError: No such Dandiset: 101267
```

bendichter:
It was unclear to me what value I should use to use the staging instance. I propose this improvement to the init docstring

fix #889

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -320,7 +320,7 @@ class DandiAPIClient(RESTFullAPIClient):
         """
         Construct a client instance.
 
-        :param str api_url: API URL.
+        :param str api_url: Base API URL of the server to interact with.
         - For DANDI production, use  ``"https://api.dandiarchive.org/api"``
         - For DANDI staging, use ``"https://api-staging.dandiarchive.org/api"``
         - If no URL is supplied, the value of the  :envvar:`DANDI_INSTANCE` environment
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -323,7 +323,7 @@ class DandiAPIClient(RESTFullAPIClient):
         :param str api_url: Base API URL of the server to interact with.
         - For DANDI production, use  ``"https://api.dandiarchive.org/api"``
         - For DANDI staging, use ``"https://api-staging.dandiarchive.org/api"``
-        - If no URL is supplied, the value of the  :envvar:`DANDI_INSTANCE` environment
+        - If no URL is supplied, the value of the :envvar:`DANDI_INSTANCE` environment
         variable (or default ``"dandi"``) is passed into ``for_dandi_instance()``.
         :param str token: User API Key. Note that different instance APIs have different
         keys.
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -323,8 +323,8 @@ class DandiAPIClient(RESTFullAPIClient):
         :param str api_url: Base API URL of the server to interact with.
         - For DANDI production, use  ``"https://api.dandiarchive.org/api"``
         - For DANDI staging, use ``"https://api-staging.dandiarchive.org/api"``
-        - If no URL is supplied, the value of the :envvar:`DANDI_INSTANCE` environment
-        variable (default value: ``"dandi"``) is passed into ``for_dandi_instance()``.
+        - If no URL is supplied, the URL is looked up in `known_instances` using the value of the  
+        :envvar:`DANDI_INSTANCE` environment variable (default value: ``"dandi"``).
         :param str token: User API Key. Note that different instance APIs have different
         keys.
         """
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -320,15 +320,15 @@ class DandiAPIClient(RESTFullAPIClient):
         """
         Construct a client instance.
 
-        :param api_url: API URL. For DANDI production, use ``"dandi"``. For DANDI staging,
-        use ``"https://api-staging.dandiarchive.org/api"``. If no URL is supplied, the
-        value of the  :envvar:`DANDI_INSTANCE` environment variable is used. If that
-        variable is not defined, ``"dandi"`` is used by default.
-        :type method: str
-        :param token: User API Key. By default, uses :envvar:`DANDI_API_KEY`.
-        Note that different API instances have different keys.
-        :type method: str
+        :param str api_url: API URL.
+        - For DANDI production, use  ``"https://api.dandiarchive.org/api"``
+        - For DANDI staging, use ``"https://api-staging.dandiarchive.org/api"``
+        - If no URL is supplied, the value of the  :envvar:`DANDI_INSTANCE` environment
+        variable (or default ``"dandi"``) is passed into ``for_dandi_instance()``.
+        :param str token: User API Key. Note that different instance APIs have different
+        keys.
         """
+
         check_dandi_version()
         if api_url is None:
             instance_name = os.environ.get("DANDI_INSTANCE", "dandi")
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -324,7 +324,7 @@ class DandiAPIClient(RESTFullAPIClient):
         - For DANDI production, use  ``"https://api.dandiarchive.org/api"``
         - For DANDI staging, use ``"https://api-staging.dandiarchive.org/api"``
         - If no URL is supplied, the value of the :envvar:`DANDI_INSTANCE` environment
-        variable (or default ``"dandi"``) is passed into ``for_dandi_instance()``.
+        variable (default value: ``"dandi"``) is passed into ``for_dandi_instance()``.
         :param str token: User API Key. Note that different instance APIs have different
         keys.
         """
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -318,10 +318,16 @@ class DandiAPIClient(RESTFullAPIClient):
         self, api_url: Optional[str] = None, token: Optional[str] = None
     ) -> None:
         """
-        Construct a client instance from a server's base API URL and an
-        optional authentication token/API key.  If no URL is supplied, the URL
-        of the instance named in the :envvar:`DANDI_INSTANCE` environment
-        variable (default value: ``dandi``) is used.
+        Construct a client instance.
+
+        :param api_url: API URL. For DANDI production, use ``"dandi"``. For DANDI staging,
+        use ``"https://api-staging.dandiarchive.org/api"``. If no URL is supplied, the
+        value of the  :envvar:`DANDI_INSTANCE` environment variable is used. If that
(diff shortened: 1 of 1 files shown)


dandi/dandiapi.py
```diff
@@ -318,10 +318,16 @@ def __init__(
         self, api_url: Optional[str] = None, token: Optional[str] = None
     ) -> None:
         """
-        Construct a client instance from a server's base API URL and an
-        optional authentication token/API key.  If no URL is supplied, the URL
-        of the instance named in the :envvar:`DANDI_INSTANCE` environment
-        variable (default value: ``dandi``) is used.
+        Construct a client instance.
+
+        :param api_url: API URL. For DANDI production, use ``"dandi"``. For DANDI staging,
```
jwodder, line 323: *For DANDI production, use ``"dandi"``* — This is wrong.  For production, use `"https://api.dandiarchive.org/api"` or pass `"dandi"` to the `for_dandi_instance()` classmethod.

dandi/dandiapi.py
```diff
@@ -318,10 +318,16 @@ def __init__(
...
+        use ``"https://api-staging.dandiarchive.org/api"``. If no URL is supplied, the
+        value of the  :envvar:`DANDI_INSTANCE` environment variable is used. If that
+        variable is not defined, ``"dandi"`` is used by default.
+        :type method: str
+        :param token: User API Key. By default, uses :envvar:`DANDI_API_KEY`.
```
jwodder, line 328: "By default, uses :envvar:`DANDI_API_KEY`." — Incorrect.  By default, no key is used.

dandi/dandiapi.py
```diff
@@ -318,10 +318,16 @@ def __init__(
...
         """
-        Construct a client instance from a server's base API URL and an
-        optional authentication token/API key.  If no URL is supplied, the URL
-        of the instance named in the :envvar:`DANDI_INSTANCE` environment
-        variable (default value: ``dandi``) is used.
+        Construct a client instance.
+
+        :param api_url: API URL. For DANDI production, use ``"dandi"``. For DANDI staging,
+        use ``"https://api-staging.dandiarchive.org/api"``. If no URL is supplied, the
+        value of the  :envvar:`DANDI_INSTANCE` environment variable is used. If that
```
jwodder, line 325: "the value of the :envvar:`DANDI_INSTANCE` environment variable is used." — The value is not used directly; rather, the API URL of the instance whose name is stored in the envvar is used.

jwodder (changes requested):

dandi/dandiapi.py
bendichter: @jwodder oh, I see. How would you feel about adding:

```
if api_url in known_instances:
    api_url = known_instances[api_url].api
```

then the user can provide "dandi" , dandi-api-local-docker-tests, and "dandi-staging". I could see why you might not want this, since these are not "URL"s, but I feel like it is more user friendly, particularly if the API addresses change down the road. Maybe we could change the variable name from `api_url` to `api` so that it can accept the instance names without being mislabeled. My first preference would be to do this, which would be changing the API to the way I intuitively expected it to work, but if you don't think that's a good idea I think improving the docstring would be sufficient.

jwodder: No.  There's already a way to construct clients from instance names, and it's the `for_dandi_instance()` classmethod.

dandi/dandiapi.py
jwodder: It's not that different from the command line, as the command line only authenticates if there's a need to (i.e., if downloading an embargoed Dandiset or if uploading anything).  There are circumstances in which one would want to access the API without authenticating (e.g., in order to see only what's publicly available), and I believe it's better for a low-level constructor to default to not pulling things from the environment.  (I'd also prefer not honoring `DANDI_INSTANCE` here, but Yarik asked for it.)

bendichter: ok, fair enough

bendichter:
@jwodder how's this?

dandi/dandiapi.py
```diff
@@ -318,11 +318,17 @@ def __init__(
...
-        optional authentication token/API key.  If no URL is supplied, the URL
-        of the instance named in the :envvar:`DANDI_INSTANCE` environment
-        variable (default value: ``dandi``) is used.
+        Construct a client instance.
+
+        :param str api_url: API URL.
+        - For DANDI production, use  ``"https://api.dandiarchive.org/api"``
+        - For DANDI staging, use ``"https://api-staging.dandiarchive.org/api"``
+        - If no URL is supplied, the value of the  :envvar:`DANDI_INSTANCE` environment
+        variable (or default ``"dandi"``) is passed into ``for_dandi_instance()``.
```
jwodder, line 327: "is passed into ``for_dandi_instance()``" — I don't like this, as it's technically inaccurate.  How about "is used as the name of the API instance à la `for_dandi_instance()`"?

dandi/dandiapi.py
```diff
@@ -318,11 +318,17 @@ def __init__(
...
-        Construct a client instance from a server's base API URL and an
-        optional authentication token/API key.  If no URL is supplied, the URL
-        of the instance named in the :envvar:`DANDI_INSTANCE` environment
-        variable (default value: ``dandi``) is used.
+        Construct a client instance.
+
+        :param str api_url: API URL.
+        - For DANDI production, use  ``"https://api.dandiarchive.org/api"``
+        - For DANDI staging, use ``"https://api-staging.dandiarchive.org/api"``
+        - If no URL is supplied, the value of the  :envvar:`DANDI_INSTANCE` environment
```
jwodder, line 326:
```suggestion
        - If no URL is supplied, the value of the :envvar:`DANDI_INSTANCE` environment
```

dandi/dandiapi.py
jwodder, line 327:
```suggestion
        variable (default value: ``"dandi"``) is passed into ``for_dandi_instance()``.
```

dandi/dandiapi.py
```diff
@@ -318,11 +318,17 @@ def __init__(
...
-        optional authentication token/API key.  If no URL is supplied, the URL
-        of the instance named in the :envvar:`DANDI_INSTANCE` environment
-        variable (default value: ``dandi``) is used.
+        Construct a client instance.
+
+        :param str api_url: Base API URL of the server to interact with.
+        - For DANDI production, use  ``"https://api.dandiarchive.org/api"``
+        - For DANDI staging, use ``"https://api-staging.dandiarchive.org/api"``
+        - If no URL is supplied, the value of the :envvar:`DANDI_INSTANCE` environment
+        variable (default value: ``"dandi"``) is passed into ``for_dandi_instance()``.
```
bendichter, line 327:
```suggestion
        - If no URL is supplied, the URL is looked up in `known_instances` using the value of the  
        :envvar:`DANDI_INSTANCE` environment variable (default value: ``"dandi"``).
```

dandi/dandiapi.py
```diff
@@ -318,11 +318,17 @@ def __init__(
...
-        Construct a client instance from a server's base API URL and an
-        optional authentication token/API key.  If no URL is supplied, the URL
-        of the instance named in the :envvar:`DANDI_INSTANCE` environment
-        variable (default value: ``dandi``) is used.
+        Construct a client instance.
+
+        :param str api_url: Base API URL of the server to interact with.
+        - For DANDI production, use  ``"https://api.dandiarchive.org/api"``
+        - For DANDI staging, use ``"https://api-staging.dandiarchive.org/api"``
+        - If no URL is supplied, the URL is looked up in `known_instances` using the value of the
```
jwodder, line 326:
```suggestion
        - If no URL is supplied, the URL is looked up in `known_instances` using the value of the
```

bendichter:
@jwodder thanks for your help on this! I was clearly confused in a few places

jwodder (approved):

PR #910, opened 2022-02-15
Remove unused tqdm dependency

jwodder:
The only place tqdm is currently "used" is in a commented-out line in `dandi/organize.py`.


yarikoptic:
FTR: originally added (commented out) to organize in 262543749baf2885792bd2b8b7b564c9d01cf072 .

Thank you @jwodder, no need to drag it indeed.

PR #911, opened 2022-02-15
Specify Dandiset ID when creating a Zarr

jwodder:
For https://github.com/dandi/dandi-archive/pull/897

### DIFF
--- a/dandi/files.py
+++ b/dandi/files.py
@@ -797,7 +797,9 @@ class ZarrAsset(LocalDirectoryAsset[LocalZarrEntry]):
         yield {"status": "initiating upload"}
         lgr.debug("%s: Beginning upload", asset_path)
         bytes_uploaded = 0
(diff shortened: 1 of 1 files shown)


yarikoptic:
@dchiquito since CI is all green, should I assume that it is just safe to provide `dandiset_id` even now without that https://github.com/dandi/dandi-archive/pull/897 yet merged?

dchiquito:
The API server will indeed swallow any unused arguments silently, so passing integration tests are expected.
I personally think that specifying undefined parameters is undefined behavior, and ignoring them is a perfectly valid way of handling them. If you think otherwise feel free to file a dandi-archive issue

yarikoptic:
> If you think otherwise feel free to file a dandi-archive issue

done: https://github.com/dandi/dandi-archive/issues/900  Meanwhile, let's proceed with this PR.  I will not mark it for release for now -- let's see integration testing in dandi-archive to turn green!

PR #912, opened 2022-02-16
Use full asset path as name of Zarr object

jwodder:
See https://github.com/dandi/dandi-archive/pull/897#issuecomment-1041703701


yarikoptic:
ok, let's see if that would help dandi-archive anyhow. Thank you @jwodder

PR #916, opened 2022-02-17
Make `dandi upload` look for dandiset.yaml in parents of common path

### LINKED ISSUE #892: dandi upload --dandiset-path?
`dandi upload --help` mentions a `--dandiset-path` arg, but I don't see one in the options. What is the proper way to set this without being in that directory?

```
$ dandi upload --help
Usage: dandi upload [OPTIONS] [PATHS]...

  Upload Dandiset files to DANDI Archive.

  The target Dandiset to upload to must already be registered in the archive,
  and a `dandiset.yaml` file must exist in the local `--dandiset-path`.

  Local Dandiset should pass validation.  For that, the assets should first be
  organized using the `dandi organize` command.

  By default all .nwb files in the Dandiset (excluding directories starting
  with a period) will be considered for the upload.  You can point to specific
  files you would like to validate and have uploaded.

Options:
  -e, --existing [error|skip|force|overwrite|refresh]
                                  What to do if a file found existing on the
                                  server. 'skip' would skipthe file, 'force' -
                                  force reupload, 'overwrite' - force upload
                                  if either size or modification time differs;
                                  'refresh' - upload only if local
                                  modification time is ahead of the remote.
                                  [default: refresh]
  -J, --jobs N[:M]                Number of files to upload in parallel and,
                                  optionally, number of upload threads per
                                  file
  --sync                          Delete assets on the server that do not
                                  exist locally
  --validation [require|skip|ignore]
                                  Data must pass validation before the upload.
                                  Use of this option is highly discouraged.
                                  [default: require]
  -i, --dandi-instance [dandi|dandi-api-local-docker-tests|dandi-devel|dandi-staging]
                                  DANDI instance to use  [env var:
                                  DANDI_INSTANCE;default: dandi]
  --help                          Show this message and exit.
```

jwodder:
Closes #892.

### DIFF
--- a/dandi/upload.py
+++ b/dandi/upload.py
@@ -30,7 +30,6 @@ def upload(
     paths: Optional[List[Union[str, Path]]] = None,
     existing: str = "refresh",
     validation: str = "require",
-    dandiset_path: Union[str, Path, None] = None,
     dandi_instance: str = "dandi",
     allow_any_path: bool = False,
     upload_dandiset_metadata: bool = False,
@@ -42,11 +41,15 @@ def upload(
     from .dandiapi import DandiAPIClient
     from .dandiset import APIDandiset, Dandiset
 
-    dandiset_ = Dandiset.find(dandiset_path)
+    if paths:
+        paths = [Path(p).absolute() for p in paths]
+        dandiset_ = Dandiset.find(os.path.commonpath(paths))
+    else:
+        dandiset_ = Dandiset.find(None)
     if not dandiset_:
         raise RuntimeError(
-            f"Found no {dandiset_metadata_file} anywhere.  "
-            "Use 'dandi download' or 'organize' first"
+            f"Found no {dandiset_metadata_file} anywhere in common ancestor of"
+            " paths.  Use 'dandi download' or 'organize' first."
         )
 
     instance = get_instance(dandi_instance)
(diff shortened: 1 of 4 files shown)


dandi/upload.py
```diff
@@ -335,7 +332,7 @@ def upload_agg(*ignored: Any) -> str:
 
     if sync:
         relpaths: List[str] = []
-        for p in original_paths:
+        for p in paths:
```
yarikoptic, line 335: I wonder if there was some good reason to use `original_paths` here instead of `.absolute()`ified ones... can't think of anything and (to be removed) comment on expanding and "validating" does not help (was there some which was removed? or did we use normpath not abspath at some point?)

yarikoptic (approved):
I am a little confused on either there was a good reason before to iterate through `original_paths` which now would be RFed away... but can't see why it was done that way

dandi/upload.py
yarikoptic: I see, so they were overloaded with the ones from `find*files` which is no longer the case iirc, so not needed indeed.  I don't think `absolute` should have negative effect there.. let's proceed. thanks!

yarikoptic:
marking for a release.  not sure already if it wouldn't be too much for `auto` to swallow.  If it would be, @jwodder -- please take care about manually releasing.

jwodder:
@yarikoptic The release workflow failed (because of auto) while trying to build & upload to PyPI.  Fortunately, that was the only thing left to do manually.

PR #919, opened 2022-02-21
Don't repeat API requests when finding subdir files to delete before Zarr upload

jwodder:
Before starting to upload a Zarr, the client compares the on-disk tree to the contents of the remote Zarr object being updated (if any).  Among other checks, if it finds a remote directory at the same path as a local file, it deletes the directory, which consists of deleting all of the files under it recursively.  Currently, the files are retrieved by making numerous queries to the API, but this PR changes the code to instead go through the already-fetched collection of remote files.


yarikoptic:
LGTM, thank you @jwodder !

PR #922, opened 2022-02-22
Recognize video files as non-generic assets

### LINKED ISSUE #920: Add VIDEO_FILE_EXTENSIONS into the "default" set
part of https://github.com/dandi/dandi-cli/issues/918 by @Saksham20 .  to avoid needing DANDI_DEVEL modes of operation on video files, we should treat them as 1st class citizens for `organize` and `upload` operations

jwodder:
Closes #920.

### DIFF
--- a/dandi/files.py
+++ b/dandi/files.py
@@ -45,6 +45,7 @@ import zarr
 from . import get_logger
 from .consts import (
     MAX_ZARR_DEPTH,
+    VIDEO_FILE_EXTENSIONS,
     ZARR_MIME_TYPE,
     ZARR_UPLOAD_BATCH_SIZE,
     EmbargoStatus,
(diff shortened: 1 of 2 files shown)


Saksham20:
@jwodder, @yarikoptic  is this good to go?

yarikoptic:
@jwodder -- is that all that is needed and PR just needs to be taken out of draft (although having some unittest tuned or added might be good) or there is more to do?

jwodder:
@yarikoptic You still need to answer my question in the issue: #920.

dandi/files.py
```diff
@@ -518,6 +519,19 @@ def get_validation_errors(
...
+    EXTENSIONS: ClassVar[List[str]] = VIDEO_FILE_EXTENSIONS
+
+    def get_metadata(
+        self,
+        digest: Optional[Digest] = None,
+        ignore_errors: bool = True,
+    ) -> BareAsset:
+        metadata = get_default_metadata(self.filepath, digest=digest)
+        metadata.path = self.path
+        return metadata
```
yarikoptic, line 532: why not to move this implementation of `get_metadata` into `LocalFileAsset` to provide such default implementation and thus to avoid duplication across subclasses which just need to use the default?

yarikoptic: could there be a test added which would ensure that `upload` does recognize video files as "first class citizen"?

jwodder: Done.

jwodder: For testing, what about just checking [here](https://github.com/dandi/dandi-cli/blob/a1a7fd553cf300967adc060302c7e6f6c1b921c8/dandi/tests/test_files.py#L23) that files with one of more of the video extensions are recognized as Dandi files?

yarikoptic: ok

jwodder: Done.

yarikoptic:
Thank you @jwodder, let's see where the changes would lead us ;)

PR #923, opened 2022-02-22
Minimize/optimize Zarr digestion when uploading

### LINKED ISSUE #913: multithread digest'ing of zarr folders
ATM, if I run `dandi digest` on a hot (was done before, so IO is fast) folder, `dandi digest` gets just 30% CPU busy and takes > 20 sec, whenever a parallelized example for checksumming takes about x20 times less, and goes above 100% CPU.

```
(dandi-devel) jovyan@jupyter-yarikoptic:/shared/ngffdata$ time DANDI_CACHE=ignore dandi digest -d zarr-checksum test64.ngff/0/0/0/0
test64.ngff/0/0/0/0: 53ad9d9fe8b4f34882fdaea76599da22
2022-02-16 18:51:44,202 [    INFO] Logs saved in /home/jovyan/.cache/dandi-cli/log/20220216185120Z-4990.log

real    0m24.034s
user    0m8.524s
sys     0m5.120s
(dandi-devel) jovyan@jupyter-yarikoptic:/shared/ngffdata$ time DANDI_CACHE=ignore dandi digest -d zarr-checksum test64.ngff/0/0/0/0
test64.ngff/0/0/0/0: 53ad9d9fe8b4f34882fdaea76599da22
2022-02-16 18:52:24,406 [    INFO] Logs saved in /home/jovyan/.cache/dandi-cli/log/20220216185202Z-5127.log

real    0m22.499s
user    0m8.369s
sys     0m6.190s

(dandi-devel) jovyan@jupyter-yarikoptic:/shared/ngffdata$ time python /shared/io-utils/fastio_md5.py test64.ngff/0/0/0/0
Total: 7200

real    0m1.358s
user    0m1.545s
sys     0m2.164s
```

related PR introducing multithreaded walk in fscacher (benefit is not yet 100% clear): https://github.com/con/fscacher/pull/67

### LINKED ISSUE #914: do not fscache individual files digests for zarr-checksum
A follow up to #913 which might have more timing information.  For that issue timings I disabled (after first physically moving aside the entire fsccher's cache for dandi-digests) fscacher, and the run took about 23 seconds.  Whenever I stopped disabling cache, the run took over 5 minutes (so fscacher gave 1500% overhead if I got it right):
```
(dandi-devel) jovyan@jupyter-yarikoptic:/shared/ngffdata$ time dandi digest -d zarr-checksum test64.ngff/0/0/0/0
test64.ngff/0/0/0/0: 53ad9d9fe8b4f34882fdaea76599da22
2022-02-16 19:02:54,861 [    INFO] Logs saved in /home/jovyan/.cache/dandi-cli/log/20220216185712Z-5383.log

real    5m43.311s
user    0m24.188s
sys     0m13.128s
```
and rerunning,  took 4 sec (which is better than original 24sec, but slower than just full recompute could be, see #913):

```
(dandi-devel) jovyan@jupyter-yarikoptic:/shared/ngffdata$ time dandi digest -d zarr-checksum test64.ngff/0/0/0/0
test64.ngff/0/0/0/0: 53ad9d9fe8b4f34882fdaea76599da22
2022-02-16 19:05:03,854 [    INFO] Logs saved in /home/jovyan/.cache/dandi-cli/log/20220216190459Z-5515.log

real    0m4.576s
user    0m2.878s
sys     0m4.593s
```

and that is using fscacher of https://github.com/con/fscacher/pull/67 .... may be that one needs really needs to become more efficient .

FWIW

<details>
<summary>with this patch I disabled caching individual files digests but added one for zarr folder</summary> 

```patch
(dandi-devel) jovyan@jupyter-yarikoptic:~/dandi-cli$ git diff
diff --git a/dandi/support/digests.py b/dandi/support/digests.py
index 2226ea8..74e5199 100644
--- a/dandi/support/digests.py
+++ b/dandi/support/digests.py
@@ -81,7 +81,7 @@ class Digester:
 checksums = PersistentCache(name="dandi-checksums", envvar="DANDI_CACHE")
 
 
-@checksums.memoize_path
+#@checksums.memoize_path
 def get_digest(filepath: Union[str, Path], digest: str = "sha256") -> str:
     if digest == "dandi-etag":
         return cast(str, get_dandietag(filepath).as_str())
@@ -96,6 +96,7 @@ def get_dandietag(filepath: Union[str, Path]) -> DandiETag:
     return DandiETag.from_file(filepath)
 
 
+@checksums.memoize_path
 def get_zarr_checksum(
     path: Path,
     basepath: Optional[Path] = None,
```
</details>

<details>
<summary>and it ran "fast" in those original 22 sec, and reloaded result in fscacher in its 3-4 sec</summary> 

```shell
(dandi-devel) jovyan@jupyter-yarikoptic:/shared/ngffdata$ mv  ~/.cache/fscacher/dandi-checksums{,.aside2}
(dandi-devel) jovyan@jupyter-yarikoptic:/shared/ngffdata$ time dandi digest -d zarr-checksum test64.ngff/0/0/0/0
test64.ngff/0/0/0/0: 53ad9d9fe8b4f34882fdaea76599da22
2022-02-16 19:22:10,832 [    INFO] Logs saved in /home/jovyan/.cache/dandi-cli/log/20220216192149Z-6188.log

real    0m22.338s
user    0m8.445s
sys     0m6.815s
(dandi-devel) jovyan@jupyter-yarikoptic:/shared/ngffdata$ time dandi digest -d zarr-checksum test64.ngff/0/0/0/0
test64.ngff/0/0/0/0: 53ad9d9fe8b4f34882fdaea76599da22
2022-02-16 19:23:02,610 [    INFO] Logs saved in /home/jovyan/.cache/dandi-cli/log/20220216192259Z-8240.log

real    0m3.401s
user    0m2.926s
sys     0m5.002s
```
</details>

Meanwhile, I think it would be worth to just disable the fscaching of both individual file digests in the zarr archive (overhead from storing that many digests on initial run seems too great to ignore) and zarr folder altogether (until we make fscaching of folder more efficient).  

Not sure if alternatively we should/could come up with some smarter policy/specification for what to cache, e.g. we could parametrize `memoize_path` to cache only if a file is larger than some specified size (e.g. 500KB or 1MB in the case of digests), but it would still need some `io.stat` to make the decision etc thus altogether might still have some overhead... but I think it might make it more flexible/generic.

### LINKED ISSUE #915: zarr upload: do not pre-digest anything
A distilled variant of #903, which if decided to proceed with could take precedence over improving any related code (#913 for more efficient digesting of zarr folders; and #914 - disabling fscaching of individual files in zarr):

- disable digesting of zarr folders in `upload` -- always assume that they differ
- do not digest individual files unless really needed -- remote file is present, the same size, so its ETag is the only way to say for sure if it is the same
- for now keep digesting for providing digests to mint upload urls for the batch.  But I am afraid we better disable fscaching of those digests (thus in part #914) since seems to provide too notable impact
  - ideally there should be a separate process/thread which would take care about pre-digesting files which might be missing digests for the next batch.  This way we could minimize delay between batches
  - md5 digest should be computed in-band while uploading and compared against ETag for the file whenever done
- overall zarr archive checksum'ing on the server will be disabled, so we should then stop checking if our upload of the zarr is consistent with the remote upon completion.

Is that about right @satra, are we disabling checksumming of zarrs on dandi-api during upload @dchiquito ?

### LINKED ISSUE #926: --digest zarr-checksum incorrectly identifies directory as "empty"
```shell
$> dandi digest --digest zarr-checksum dandi
2022-02-25 13:57:48,587 [    INFO] Logs saved in /home/yoh/.cache/dandi-cli/log/20220225185747Z-698970.log
Error: Cannot compute a Zarr checksum for an empty directory

$> ls dandi
__init__.py   conftest.py  dandiarchive.py  due.py         metadata.py     support/   validate/
__pycache__/  consts.py    dandiset.py      exceptions.py  misctypes.py    tests/     validate.py
_version.py   core/        delete.py        files.py       organize.py     upload.py
cli/          dandiapi.py  download.py      keyring.py     pynwb_utils.py  utils.py

$> dandi --version
0.36.0+4.gf467fe8
```

jwodder:
Closes #913.
Closes #914.
Closes #915.
Closes #926.


dandi/support/digests.py
```diff
@@ -103,22 +107,93 @@ def get_zarr_checksum(
...
     if path.is_file():
         return cast(str, get_digest(path, "md5"))
+    root: Tuple[str, ...]
     if basepath is None:
         basepath = path
-    dirs = {}
-    files = {}
+        root = ()
+    else:
+        root = path.relative_to(basepath).parts
```
yarikoptic, line 115: this code path (and block below depending on `root` being iterable) seems to not covered by a test.

dandi/support/digests.py
```diff
@@ -103,22 +107,93 @@ def get_zarr_checksum(
 ) -> str:
```
yarikoptic, line 107: please add a docstring describing the purpose of `basepath` and `known` since hard to impossible to guess

dandi/files.py
```diff
@@ -838,64 +836,98 @@ def iter_upload(
...
-                    pps = str(pp)
-                    if pps in old_zarr_entries:
-                        if old_zarr_entries[pps].is_file():
+
+        to_upload = EntryUploadTracker()
+        if old_zarr_entries:
+            to_delete: List[RemoteZarrEntry] = []
+            digesting: List[Future[Optional[Tuple[LocalZarrEntry, str]]]] = []
+            yield {"status": "comparing against remote Zarr"}
+            with ThreadPoolExecutor(max_workers=jobs or 5) as executor:
```
yarikoptic, line 845: did you try this PR on some size-able zarr upload from e.g. hub to our staging S3?  I wonder about how long it took e.g. to start with uploading for a fresh new zarr (so nothing on server yet) and for e.g. some interrupted upload after some files were already transmitted; and what speed did you observe?

yarikoptic (commented):
some few initial questions/comments

dandi/support/digests.py
jwodder: Done.

dandi/files.py
jwodder: Not yet.  Do you have a Zarr or script for creating one to recommend using, or should I just jump straight to trying to upload `test128.ngff`?  (CC: @satra)

yarikoptic: I have also downloaded that sample large-ish  (~37GB) zarr @satra uploaded to staging , into the hub: `/shared/zarr/b35681cd-936a-477c-ac51-06661612d4df` . Would be interesting to gauge upload speed for that one into staging (as a new zarr of cause), and then interrupt and continue.  @satra also placed a helper tool (like dstat) under `/shared/io-utils/dool` to monitor "wire bandwidth" (in addition to what client would be reporting)

jwodder: @yarikoptic While testing out the upload, I found an inefficiency: the first thing `upload()` does when processing an asset is get its local size, which for Zarrs means iterating over the entire directory tree, which is later iterated over again inside `iter_upload()`.  Should Zarr uploads delay this calculation until the latter point?

jwodder: @yarikoptic I just finished a complete upload of the given Zarr in a single upload session, which took 63 minutes, the first couple minutes of which were spent in pre-upload directory traversals.  The upload itself alternated between uploading files in fast bursts of about 10 seconds each and waiting about 10-20 seconds for `/zarr/{zarr_id}/upload/complete/` requests to return.  Next I'll try interrupting and resuming.

yarikoptic: re size compute iteration: I guess iteration through zarr to determine the size might just need to be improved (that threaded walk?) since probably otherwise would be too tricky for "compute out of band and update progress reporting etc accordingly later". Or you see how to make it delayed "nicely"?

So -- we still have some dandi-api activities (I guess checksumming) to return for `/complete` which should be either optimized or somehow shifted to be "out of band" besides may be the final batch upload?  (attn @satra @dchiquito @waxlamp )

jwodder: @yarikoptic By "delay", I meant leaving the display of the "SIZE" column for a Zarr blank until all the local entries have been traversed while determining what to upload, at which point (just before starting upload proper) `iter_upload()` can just emit a `{"size": total_size}` dict, the size having been calculated while iterating.

satra: if this is simply for the metadata, i would say the client could provide it but there is no reason for server to trust it. and since the server will compute size and checksum out of band, the server will populate those fields in the metadata. this is similar to the PR that @AlmightyYakob is working on that will also support out of band ingestion.

jwodder: @satra This is not for the metadata; this is for the pyout display.

satra: ah in that case, i would say don't worry about total size. just provide a summary at the end.

yarikoptic: > @yarikoptic By "delay", I meant leaving the display of the "SIZE" column for a Zarr blank until all the local entries have been traversed while determining what to upload, at which point (just before starting upload proper) `iter_upload()` can just emit a `{"size": total_size}` dict, the size having been calculated while iterating.

that sounds great to me.

yarikoptic: @satra - don't worry about size.  But we need to figure out what to do with `/complete` for finishing a batch upload for zarr which seems to add ~200% overhead on top of transfer time in @jwodder's experiment.

dandi/support/digests.py
```diff
@@ -101,24 +105,105 @@ def get_zarr_checksum(
     basepath: Optional[Path] = None,
     known: Optional[Dict[str, str]] = None,
 ) -> str:
+    """
+    Compute the Zarr checksum for a file or directory tree.  The checksum for a
+    subdirectory of a Zarr can be computed by setting ``path`` to the path to
+    the subdirectory and setting ``basepath`` to the path to the root of the
+    Zarr.
```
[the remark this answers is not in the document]
yarikoptic: may be only for "known"?  but even then, could just have abs (or consistently relative) paths and it would all still work, wouldn't it?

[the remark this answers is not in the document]
jwodder: When calculating the checksum for a directory, the paths used in the intermediate JSON document are relative to the root of the Zarr, not to the directory in question, so we need to know the path to the root when working on a subdirectory of a Zarr.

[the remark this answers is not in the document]
yarikoptic: >  the paths used in the intermediate JSON document are relative to the root of the Zarr, not to the directory in question,

yikes -- I guess I didn't pay close attention.  Do you know why is that so, i.e. is there a reason for "relative to the top" relative paths?  that means that renaming of a subdirectory changes its (portion) of the checksum for no content change within it -- odd/inefficient.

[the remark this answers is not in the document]
jwodder: I don't know.  That's just how Daniel designed it.

jwodder:
After using this branch to upload a 38GB Zarr from the Hub in just over an hour, I renamed the Zarr, started a new upload, let it run a little bit, interrupted it, and then started uploading again.  That worked out fine (after a fix).  I then tried another upload, without having changed anything, and it took the client almost exactly 30 minutes to compare the local & remote Zarrs and determine that nothing needed to be uploaded.

One possible change that would likely make this faster would be if the Zarr directory listings returned from `/zarr/{zarr_id}.zarr/{dirpath}/` requests included file sizes; they currently do not, which means the client has to make a HEAD request for every remote entry it wants to get the size of.

dandi/support/digests.py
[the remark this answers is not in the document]
yarikoptic: ok, filed https://github.com/dandi/dandischema/issues/119 , nothing to "fix" here, so resolving

yarikoptic:
> One possible change that would likely make this faster would be if the Zarr directory listings returned from `/zarr/{zarr_id}.zarr/{dirpath}/` requests included file sizes; they currently do not, which means the client has to make a HEAD request for every remote entry it wants to get the size of.

Filed https://github.com/dandi/dandi-archive/issues/925 . As a file size change is likely to be a "rare" indicator for having a change within file, I wonder if we should stick to it, or just immediately proceed to checksumming, which is to happen anyways for overall check if there is anything to upload whenever it is exactly the same (at took you 30 minutes), right?

Meanwhile, I will just proceed with the merge of this one as is to ease testing etc.

PR #925, opened 2022-02-24
Let `dandi delete` take paths to local Zarrs

jwodder:


yarikoptic:
not fully grasping it, but that is ok -- must be doing the right thing! thank you @jwodder !

PR #928, opened 2022-03-01
checking for duplicate video files in external_files

Saksham20:
If there happen to be multiple nwb files linked to one video file via the `ImageSeries.external_files`, the `move` option of `video_mode` will fail. This PR enforces the user to change the option to one of `copy/symlink/hardlink' instead by raising an Exception.

### DIFF
--- a/dandi/organize.py
+++ b/dandi/organize.py
@@ -886,7 +886,8 @@ def organize(
     if media_files_mode == "move":
         videos_list = []
         for meta in metadata:
-            videos_list.extend(meta["external_file_objects"].get("external_files", []))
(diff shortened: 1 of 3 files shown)


yarikoptic:
interesting! in principle, the same video file could indeed be used (e.g. the same stimuli) across multiple files.  It would be a bit wasteful at the level of filesystem to copy them, but for DANDI archive which deduplicates assets (or at datalad dandisets level, unless on a crippled filesystem) -- it doesn't really matter. Cool.  So, let's allow for silently copying etc. Thank you @Saksham20 !

PR #930, opened 2022-03-01
Wait for zarr upload cancels

### LINKED ISSUE #929: Update zarr upload cancelling
As of https://github.com/dandi/dandi-archive/pull/911 (currently still unreleased), zarr uploads are cancelled asynchronously instead of in the cancel request. This means that the `DELETE /zarr/{zarr_id}/upload/` is dispatched as normal, but then the CLI needs to idle until `GET /zarr/{zarr_id}` has `upload_in_progress == False`, where `upload_in_progress` is a newly added field.

dchiquito:
Fixes #929 

I'm not sure at how best to go about testing this.

### DIFF
--- a/dandi/files.py
+++ b/dandi/files.py
@@ -961,6 +961,10 @@ class ZarrAsset(LocalDirectoryAsset[LocalZarrEntry]):
                                 f.cancel()
                             executor.shutdown()
                             client.delete(f"/zarr/{zarr_id}/upload/")
+                            upload_in_progress = True
+                            while upload_in_progress:
+                                r = client.get(f"/zarr/{zarr_id}/")
+                                upload_in_progress = r["upload_in_progress"]
                             raise
                         else:
                             bytes_uploaded += size


jwodder:
@dchiquito Why exactly do we need to wait for the cancel to complete?  The only time the client cancels an upload is when an error occurs, at which point we're not going to be uploading anything else to the Zarr for the rest of the process.

yarikoptic:
> @dchiquito Why exactly do we need to wait for the cancel to complete? The only time the client cancels an upload is when an error occurs, at which point we're not going to be uploading anything else to the Zarr for the rest of the process.

but can't we resume upload to that zarr in subsequent process i.e. wouldn't there still be an asset associated with that zarr?  IIRC now we first create asset for a new zarr, and then start uploading.  Then if we cancel upload, it is ok -- asset and possibly in bad state zarr already exist.  We can try to reupload/tune up.  I don't see though the point of waiting for  "canceling upload" process to finish since what if a next upload is getting requested -- will it succeed @dchiquito ?  is such waiting just to provide some good UX to "automatically wait" until being able to try to (re)upload again?

dchiquito:
I wasn't sure of how exactly the CLI was using the API, and it seemed plausible to me that it might want to retry an upload batch after canceling. If it doesn't make sense to do that immediately after cancelling, then do feel free to close this PR.

The change in dandi-api also introduced a new field `upload_in_progress` to `GET /zarr/{zarr_id}/`, so you can now tell in advance if an upload is already in progress without simply attempting it and checking if the request errors out. Not sure if that's relevant here, but thought I'd mention it.

yarikoptic:
> I wasn't sure of how exactly the CLI was using the API, and it seemed plausible to me that it might want to retry an upload batch after canceling. If it doesn't make sense to do that immediately after cancelling, then do feel free to close this PR.

It might make sense to reupload immediately after canceling. My point was that we shouldn't hold waiting upon canceling of the upload (as this PR does) but might want to do that for the upload to start.  If I am canceling, there might be a need such as "get me back into the terminal", instead of waiting more or needing 2nd Ctrl-C to cause exception while handling an exception.  Filed #942 with my brief description on how I think such waiting could be implemented. 

> The change in dandi-api also introduced a new field `upload_in_progress` to `GET /zarr/{zarr_id}/`, so you can now tell in advance if an upload is already in progress without simply attempting it and checking if the request errors out.

errorring out doesn't provide a similar response field?

yarikoptic:
I would leave it for @jwodder to decide to close this in favor of #942

PR #931, opened 2022-03-02
check for existance of video path else assume relative path

Saksham20:
if the user uses the relative path of video file wrt nwbfile then we must find the absolute path of video before performing copy/move/symlink/hardlink

### DIFF
--- a/dandi/organize.py
+++ b/dandi/organize.py
@@ -256,6 +256,10 @@ def organize_external_files(
             ):
                 new_path = op.join(dandiset_path, op.dirname(e["dandi_path"]), name_new)
                 name_old_str = str(name_old)
+                if not op.isabs(name_old_str):
+                    name_old_str = op.join(op.dirname(e["path"]), name_old_str)
+                if not op.exists(name_old_str):
+                    lgr.error("%s does not exist", name_old_str)
                 os.makedirs(op.dirname(new_path), exist_ok=True)
                 if files_mode == "symlink":
                     os.symlink(name_old_str, new_path)


yarikoptic:
looks LGTM to me, but would be nice to have a unittest for that code path/case.  WDYT @jwodder ?

dandi/organize.py
```diff
@@ -256,6 +256,10 @@ def organize_external_files(
             ):
                 new_path = op.join(dandiset_path, op.dirname(e["dandi_path"]), name_new)
                 name_old_str = str(name_old)
+                if not op.isabs(name_old_str):
+                    name_old_str = op.join(op.dirname(e["path"]), name_old_str)
+                if not op.exists(name_old_str):
```
jwodder, line 261: Right now, if the file doesn't exist, you're just logging a message and then continuing as it if does.  Shouldn't you at least add a `continue`?

Saksham20: I am just logging the error, but would want the`shutil.copy()` and `shutil.move()` or the symlink/hardlink to throw a `FileNotFoundError`

yarikoptic: why to hope/rely on some other code/error whenever you know that it must be coming?

Saksham20: so we could explicitly throw a `FileNotFoundError` like https://github.com/dandi/dandi-cli/pull/931/commits/f23ca3677cc9b3b2259894ecf6440bad1f2fcc7e

yarikoptic:
ok, let's proceed. Thank you @Saksham20 !

PR #932, opened 2022-03-02
skip handling URLs in external files references as we do for local paths

Saksham20:
if the external_files contain weblinks to media files, skip re-writing that attribute and do not do any copy/move operation.

### DIFF
--- a/dandi/organize.py
+++ b/dandi/organize.py
@@ -254,6 +254,8 @@ def organize_external_files(
                     ext_file_dict["external_files_renamed"],
                 )
             ):
+                if any([str(name_old).startswith(i) for i in ["http", "ftp"]]):
+                    continue
                 new_path = op.join(dandiset_path, op.dirname(e["dandi_path"]), name_new)
                 name_old_str = str(name_old)
                 os.makedirs(op.dirname(new_path), exist_ok=True)
--- a/dandi/organize.py
+++ b/dandi/organize.py
@@ -255,7 +255,7 @@ def organize_external_files(
                     ext_file_dict["external_files_renamed"],
                 )
             ):
-                if is_url(name_old):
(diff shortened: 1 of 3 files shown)


dandi/organize.py
```diff
@@ -254,6 +254,8 @@ def organize_external_files(
                     ext_file_dict["external_files_renamed"],
                 )
             ):
+                if any([str(name_old).startswith(i) for i in ["http", "ftp"]]):
```
yarikoptic, line 257: can't have `ftpdownloaded.mp3`? check at least (and no need for `[]` AFAIK) for 

```suggestion
                if any(str(name_old).startswith(i) for i in ["http://", "https://", "ftp://"]):
```

dandi/pynwb_utils.py
```diff
@@ -315,7 +315,8 @@ def rename_nwb_external_files(metadata: List[dict], dandiset_path: str) -> None:
                         ext_file_dict["external_files_renamed"],
                     )
                 ):
-                    container.external_file[no] = str(name_new)
+                    if not any([str(name_old).startswith(i) for i in ["http", "ftp"]]):
```
yarikoptic, line 318: 2nd encounter -- thus better add a short function `is_url` or alike implementing that , and reuse in both spots

yarikoptic: actually -- we already have `dandi.utils.is_url` which you might just want to extend with `ftp://`. Note that it also takes `dandi:` ones as urls, imho legit

Saksham20: thanks for pointing that out, I have now used that function instead by adding the "ftp" check to that

yarikoptic:
ok, let's proceed. Thank you @Saksham20 !

PR #933, opened 2022-03-02
Update for change in Zarr checksum format

jwodder:
To accompany https://github.com/dandi/dandischema/pull/120

Cannot be tested until after dandischema 0.6.0 is released (and then further edits will be needed to update some test cases that hardcode checksums).


yarikoptic:
0.6.0 was released, so time to "test"/merge/release asap @jwodder ?

jwodder:
@yarikoptic I already updated the tests.  The release should include #938 as well.

PR #940, opened 2022-03-16
dandi upload: Make `--jobs` default to `5:5` and document it

### LINKED ISSUE #927: dandi upload parallel jobs
When you do `dandi upload`, it's not clear how many parallel jobs are used by default. I would have expected the program to automatically detect how many jobs are available. Upon some digging, I found this `iter_upload` method, which claims the number of jobs defaults to 5 but does not set it to 5 in the method signature. 

https://github.com/dandi/dandi-cli/blob/a1a7fd553cf300967adc060302c7e6f6c1b921c8/dandi/files.py#L765-L792

Then the default to 5 is set down here:

https://github.com/dandi/dandi-cli/blob/a1a7fd553cf300967adc060302c7e6f6c1b921c8/dandi/files.py#L845

If you really want to hard-code 5, wouldn't it be easier to just do it in the function signature like:

```python
jobs: Optional[int] = 5
```
And then it would be nice to indicate this in the cli help.

A better option might be to use what is [now default for Python >=3.8](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor): `max_workers=min(32, os.cpu_count() + 4)`

jwodder:
Closes #927.

### DIFF
--- a/dandi/cli/cmd_upload.py
+++ b/dandi/cli/cmd_upload.py
@@ -25,7 +25,10 @@ from .base import (
     "-J",
     "--jobs",
     type=IntColonInt(),
-    help="Number of files to upload in parallel and, optionally, number of upload threads per file",
+    help=(
+        "Number of assets to upload in parallel and, optionally, number of"
+        " upload threads per asset  [default: 5:5]"
+    ),
 )
 @click.option(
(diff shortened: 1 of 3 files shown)


yarikoptic:
Thank you @jwodder -- let's try it out and see either it improves or not UX.

PR #941, opened 2022-03-20
Integrate NWBInspector with DANDI validation

CodyCBakerPhD:
Hello everyone!

This is a first attempt at integrating the [NWBInspector](https://github.com/neurodatawithoutborders/nwbinspector) in place of existing metadata validation applied to single NWBAsset instances. This approach will make including other types of metadata checking on NWBAssets much easier to extend; all you would have to do is add the specific check to the `list` of `_required_nwb_checks` at the [top of dandi/files](https://github.com/catalystneuro/dandi-cli/blob/4e0252dece8b2bb97cd2da0c6cf15a9cfca76374/dandi/files.py#L80).

No significant overall change to your CLI behavior is expected here, up to some changes we may have to make on our end to support PyNWB validation on older NWB schemas.

For 'next steps' after this PR that _would_ change behavior by adding extra metadata conditions, this would be very easy to do and could immediately include checking that Subject fields follow certain patterns, such as age/duration in ISO 8601, species being in Latin binomial form, etc...

I just wanted to check first off that this is the desired injection point for utilizing the generalized Inspector during explicit validation (see [PR #40](https://github.com/dandi/handbook/pull/40) for documentation changes suggesting users to run the **full** NWBInspector prior to DANDI validation, but this is purely optional).

### DIFF
--- a/dandi/files.py
+++ b/dandi/files.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from collections import deque
+from collections.abc import Iterable
 from concurrent.futures import Future, ThreadPoolExecutor, as_completed
 from contextlib import closing
 from dataclasses import dataclass, field, replace
@@ -44,6 +45,7 @@ from dandischema.models import DigestType, get_schema_version
 from pydantic import ValidationError
 import requests
 import zarr
+import nwbinspector
 
 from . import get_logger
 from .consts import (
@@ -64,7 +66,6 @@ from .dandiapi import (
 from .exceptions import UnknownAssetError
 from .metadata import get_default_metadata, get_metadata, nwb2asset
 from .misctypes import DUMMY_DIGEST, BasePath, Digest, P
-from .pynwb_utils import validate as pynwb_validate
 from .support.digests import (
     get_dandietag,
     get_digest,
@@ -77,7 +78,10 @@ lgr = get_logger()
 
 # TODO -- should come from schema.  This is just a simplistic example for now
 _required_dandiset_metadata_fields = ["identifier", "name", "description"]
-_required_nwb_metadata_fields = ["subject_id"]
+_required_nwb_metadata_fields = [
+    nwbinspector.check_subject_exists,
+    nwbinspector.check_subject_id_exists,
+]
 
 
 @dataclass  # type: ignore[misc]  # <https://github.com/python/mypy/issues/5374>
@@ -507,8 +511,8 @@ class NWBAsset(LocalFileAsset):
         self,
         schema_version: Optional[str] = None,
         devel_debug: bool = False,
-    ) -> List[str]:
-        errors: List[str] = pynwb_validate(self.filepath, devel_debug=devel_debug)
+    ) -> Iterable:
+        errors = []
         if schema_version is not None:
             errors.extend(
                 super().get_validation_errors(
@@ -518,21 +522,25 @@ class NWBAsset(LocalFileAsset):
         else:
             # make sure that we have some basic metadata fields we require
             try:
-                meta = get_metadata(self.filepath)
+                errors.extend(
+                    [
+                        error.message
+                        for error in nwbinspector.inspect_nwb(
+                            nwbfile_path=self.filepath,
+                            checks=_required_nwb_metadata_fields,
+                        )
+                    ]
+                )
             except Exception as e:
                 if devel_debug:
                     raise
                 lgr.warning(
-                    "Failed to read metadata in %s: %s",
+                    "Failed to inspect NWBFile in %s: %s",
                     self.filepath,
                     e,
                     extra={"validating": True},
                 )
-                errors.append(f"Failed to read metadata: {e}")
-            else:
-                errors.extend(
-                    _check_required_fields(meta, _required_nwb_metadata_fields)
-                )
+                errors.append(f"Failed to inspect NWBFile: {e}!")
         return errors
 
 
@@ -926,7 +934,6 @@ class ZarrAsset(LocalDirectoryAsset[LocalZarrEntry]):
             for local_entry in self.iterfiles():
                 total_size += local_entry.size
                 to_upload.register(local_entry)
-
         yield {"status": "initiating upload", "size": total_size}
         lgr.debug("%s: Beginning upload", asset_path)
         bytes_uploaded = 0
--- a/dandi/files.py
+++ b/dandi/files.py
@@ -13,7 +13,6 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from collections import deque
-from collections.abc import Iterable
 from concurrent.futures import Future, ThreadPoolExecutor, as_completed
 from contextlib import closing
 from dataclasses import dataclass, field, replace
@@ -511,8 +510,8 @@ class NWBAsset(LocalFileAsset):
         self,
         schema_version: Optional[str] = None,
         devel_debug: bool = False,
-    ) -> Iterable:
-        errors = []
+    ) -> List[str]:
+        errors: List[str] = []
         if schema_version is not None:
             errors.extend(
                 super().get_validation_errors(
@@ -540,7 +539,7 @@ class NWBAsset(LocalFileAsset):
                     e,
                     extra={"validating": True},
                 )
-                errors.append(f"Failed to inspect NWBFile: {e}!")
+                errors.append(f"Failed to inspect NWBFile: {e}")
         return errors
 
 
--- a/dandi/files.py
+++ b/dandi/files.py
@@ -930,6 +930,7 @@ class ZarrAsset(LocalDirectoryAsset[LocalZarrEntry]):
             for local_entry in self.iterfiles():
                 total_size += local_entry.size
                 to_upload.register(local_entry)
+
         yield {"status": "initiating upload", "size": total_size}
         lgr.debug("%s: Beginning upload", asset_path)
         bytes_uploaded = 0
--- a/dandi/files.py
+++ b/dandi/files.py
@@ -42,9 +42,9 @@ from dandischema.models import BareAsset, CommonModel
 from dandischema.models import Dandiset as DandisetMeta
 from dandischema.models import DigestType, get_schema_version
 from pydantic import ValidationError
+from nwbinspector import inspect_nwb, check_subject_exists, check_subject_id_exists
 import requests
 import zarr
-import nwbinspector
 
 from . import get_logger
 from .consts import (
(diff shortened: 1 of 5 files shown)


dandi/files.py
```diff
@@ -508,7 +508,7 @@ def get_validation_errors(
         schema_version: Optional[str] = None,
         devel_debug: bool = False,
     ) -> List[str]:
-        errors: List[str] = pynwb_validate(self.filepath, devel_debug=devel_debug)
```
CodyCBakerPhD, line 511: Note that `pynwb.validate` comes for free with the NWBInspector: https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/nwbinspector/nwbinspector.py#L324-L333

However, I noticed your `pynwb_validate` helper does some additional filtering on top of this, so we may have to update our end to ensure it behaves 100% the same way.

dandi/files.py
```diff
@@ -518,21 +518,25 @@ def get_validation_errors(
...
             try:
-                meta = get_metadata(self.filepath)
+                errors.extend(
+                    [
+                        error.message
+                        for error in inspect_nwb(
+                            nwbfile_path=self.filepath,
+                            checks=_required_nwb_checks,
+                        )
+                    ]
```
CodyCBakerPhD, line 528: Note two items here:

(a) I notice in your validation API you expect the output to be a list of strings; our native output from the Inspector is a sophisticated [data class](https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/nwbinspector/register_checks.py#L35-L75) that also tracks a bunch of other contextual information, the most immediately important of which here is likely the `message`, which is why I'm only pulling out that item.

If desired, we could figure out some sort of 'nice' f-string that incorporates other relevant attributes. But for now, the form of the output can be seen in the modified and added tests for this PR

(b) All the highest-level `inspect` functionalities such as `inspect_nwb` used here are generators that yield data streams of messages; this iteration is being collapsed in this stage similar to the existing functionality, but if desired we could look into yielding here as well.

dandi/files.py
yarikoptic: I wonder if instead of replacing "our" invocation of nwb's stock `.validate` we could just instruct inspector to skip nwb stock validation?

CodyCBakerPhD: Would be better for everyone else to user the superior version here, no? The inspector validation is only the vanilla type right now, but we should improve it by porting over your modifications.

dandi/files.py
yarikoptic: We definitely should refactor validation output to support more informative structured records. For the purpose of this PR let's KISS **but** it relates to #896 for adding BIDS validation.

But I thought that inspector just provides "hints" or "warnings" not "error" class of notifications. 
In refactoring we might provide generalization over types of records known to bids-validator and nwb inspector.  Filed https://github.com/dandi/dandi-cli/issues/943 to start dialog on that topic. Haven't looked yet into nwbinspector records

dandi/files.py
CodyCBakerPhD: Never mind, Ben walked me through what this issue is about. Looks like a PR on PyNWB is in progress: https://github.com/NeurodataWithoutBorders/pynwb/pull/1432

When it is, maybe we should update the functionality so the ignoring of 2.10 is no longer needed within `dandi`, and the `validate` can be one of the burdens of the Inspector.

For now, we will add a flag to the Inspector to skip validation so we can run yours instead.

yarikoptic: ;-) superior is good.  I guess we would just need to encode (if we haven't yet) desired assumptions on nwb validation in our tests to guarantee that nwbinspector refactorings do not make them go away etc.

setup.cfg
```diff
@@ -50,7 +50,7 @@ install_requires =
     pycryptodomex  # for EncryptedKeyring backend in keyrings.alt
     pydantic >= 1.9.0
     pynwb >= 1.0.3,!=1.1.0
-    nwbinspector >= 0.3.1
+    nwbinspector @ git+https://github.com/NeurodataWithoutBorders/nwbinspector@dev
```
yarikoptic, line 53: note that for PR to get merged we better have a released version.

Moreover: filed https://github.com/NeurodataWithoutBorders/nwbinspector/issues/116 , attn @jwodder .

dandi/files.py
yarikoptic: please digest it for me a bit more @CodyCBakerPhD -- will we need to wait for new pynwb release?

dandi/files.py
```diff
@@ -77,7 +77,7 @@
 
 # TODO -- should come from schema.  This is just a simplistic example for now
 _required_dandiset_metadata_fields = ["identifier", "name", "description"]
-_required_nwb_metadata_fields = ["subject_id"]
+_required_nwb_checks = [check_subject_exists, check_subject_id_exists]
```
yarikoptic, line 80: so is the idea here that we will need to adjust this list of checks to cover only the checks we care about?

could there instead be some kind of annotation of checks in nwbinspector, e.g. with `dandi` tag of some kind, so we could just ask here for all checks with `"dandi"` tag instead of asking for specifically these checks?

asking because hardcoding a list of checks here would require not only a release of nwbinspector whenever we add a new test but also changes to dandi-cli and its release.  Best to have a workflow which would allow for needing only a single library release, e.g. `nwbinspector` and asking people to update it instead of updating both.

setup.cfg
CodyCBakerPhD: Yep, that's no problem, just wanted to make sure it works for now.

CodyCBakerPhD:
> please digest it for me a bit more @CodyCBakerPhD -- will we need to wait for new pynwb release?

@yarikoptic No no, I'll restore your validation method and we'll include the changes to the inspector with a new release on that.

The PyNWB changes are just for the long-term plan, once they support validation on cached namespaces themselves then we can run the validation through NWBInspector appropriately on pre-2.10 files, instead of skipping them like we do here.

dandi/files.py
CodyCBakerPhD: > so is the idea here that we will need to adjust this list of checks to cover only the checks we care about?

For hard-enforcing this level of validation, yes. Basically what you were originally doing with exact metadata assertions (`subject_id`), but now each assertion is a 

> could there instead be some kind of annotation of checks in nwbinspector, e.g. with dandi tag of some kind, so we could just ask here for all checks with "dandi" tag instead of asking for specifically these checks?

So, we have a [config functionality](https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/nwbinspector/internal_configs/dandi.inspector_config.yaml) that alters Importance levels of checks, but still includes all original checks from the base registry. You can tell it to SKIP various checks at whim, though, but for this it might be better to instead have a SELECT operation.

> asking because hardcoding a list of checks here would require not only a release of nwbinspector whenever we add a new test but also changes to dandi-cli and its release. Best to have a workflow which would allow for needing only a single library release, e.g. nwbinspector and asking people to update it instead of updating both.

I see what you mean, here. We will have to think about how best to allow this on our end, then. Will get back soon

dandi/files.py
CodyCBakerPhD: > But I thought that inspector just provides "hints" or "warnings" not "error" class of notifications.

Precisely; the Inspector only returns messages, we don't/can't distinguish errors vs. warnings most of the time. The behavior occurring here is to basically consider ANY message returned as an error to prevent validation+upload.

CodyCBakerPhD: Well, technically, there are two types of explicit errors that we can and do catch, (i) actual Python errors that occur during the process, these get returned as traceback messages but inspection proceeds, and (ii) pynwb validation, but we're planning to disable that here according the other discussion.

You could *technically* consider anything else from the inspector a 'warning', but the intended use here is for things like critical missing metadata to be considered as errors in the context of validation/upload.

yarikoptic: the question here -- should list comprehension be adjusted to only consider errors and skipping any hints etc, or those wouldn't be included ATM any ways? Or should we specify that config should be "dandi"? (I guess it relates to discussion on "tag"ing which checks to run for dandi [below](https://github.com/dandi/dandi-cli/pull/941#discussion_r831507352))

bendichter: @yarikoptic we are using the nwb inspector in two different modes.
1. provide a report, tailored to the needs of dandi, recommending changes to the file
2. validating the file to ensure all essential requirements are met.

this PR is for mode 2. Mode 1 is taken care of by this PR: https://github.com/dandi/handbook/pull/40 and requires no changes to the dandi cli

setup.cfg
```diff
@@ -50,6 +50,7 @@ install_requires =
     pycryptodomex  # for EncryptedKeyring backend in keyrings.alt
     pydantic >= 1.9.0
     pynwb >= 1.0.3,!=1.1.0
+    nwbinspector @ git+https://github.com/NeurodataWithoutBorders/nwbinspector@validate_optional
```
CodyCBakerPhD, line 53: I'll update this once we have a new release with all the desired fixes in place for this PR

dandi/files.py
```diff
@@ -518,21 +518,27 @@ def get_validation_errors(
         else:
             # make sure that we have some basic metadata fields we require
             try:
-                meta = get_metadata(self.filepath)
+                errors.extend(
+                    [
+                        error.message
+                        for error in inspect_nwb(
+                            nwbfile_path=self.filepath,
+                            skip_validate=True,
```
CodyCBakerPhD, line 526: This allows your local v2.10-skipping method to run instead of ours.

dandi/files.py
```diff
@@ -518,21 +518,27 @@ def get_validation_errors(
...
+                            config=load_config(filepath_or_keyword="dandi"),
+                            importance_threshold=Importance.CRITICAL,
```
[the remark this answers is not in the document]
CodyCBakerPhD: Now, looking towards the future you envision for easily extending the behavior of `dandi validate` to include the list of extra metadata checks that you want (Icephys cell things, ontologies, specific forms of subject fields, etc): these would be very easy to include simply by following the comments I left in the config file. That is, when you are all ready for that change, you can 
Now, we do _strongly_ recommend however that, for further customization and control for any `dandi validate`-exclusive purposes, you should move that config file (or make your own) over to live on the DANDI side of things. I believe this is the best option for both of us, since it is more general than you having to hard-code a list of explicit checks or metadata fields (you automatically inherit any additional checks/debugs we ever add to the NWBInspector) but it would also soften _your own_ dependency on _our_ release cycle if you want to make any on-the-fly adjustments to  metadata considerations.

Either way, this integration is perhaps beginning to err on the side of meriting a joint meeting - please schedule with Ben if you'd like to discuss all this in person.

setup.cfg
```diff
@@ -50,6 +50,7 @@ install_requires =
     pycryptodomex  # for EncryptedKeyring backend in keyrings.alt
     pydantic >= 1.9.0
     pynwb >= 1.0.3,!=1.1.0
+    nwbinspector @ git+https://github.com/NeurodataWithoutBorders/nwbinspector@dev
```
jwodder, line 53: Please release a new version of nwbinspector so that this PR can be merged (and so that a conda-forge feedstock can be created).

jwodder (changes requested):

setup.cfg
CodyCBakerPhD: @jwodder Hey there - I think we might need 1 or 2 bug fixes plus a new feature. I'm on vaca right now, but I'll be back on Thursday to try to push this out before the end of the week

CodyCBakerPhD: @jwodder Pushed new version

PR #953, opened 2022-04-07
Specify all queues in test celery worker

dchiquito:
The celery worker in the docker-compose test configuration needs to
explicitly specify all queues that it monitors. Several queues have been
added recently so the worker definition needs to include those new
queues.

Fixes https://github.com/dandi/dandi-archive/issues/1011


jwodder (approved):

yarikoptic:
@jwodder approved, so let's proceed

Issue #948, opened 2022-04-06
BIDS validator tests create `bids-validator-report_*.log` files in repository root


jwodder:
Any transient files created as part of tests should be created in a temporary directory, ideally one provisioned using pytest's `tmp_path` fixture.

PR #954, opened 2022-04-07
Fixed test_write_report() log path issue

TheChymera:
Addressing: https://github.com/dandi/dandi-cli/issues/948 

Fix for another function is ongoing, hopefully done within the hour.


yarikoptic:
Please rebase on current master, there was a fix to stalling.

dandi/tests/test_bids_validator_xs.py
```diff
@@ -273,8 +273,8 @@ def test_write_report():
     ]
 
     report_path = os.path.join(
-        os.path.abspath(os.path.dirname(__file__)),
-        "data/output_bids_validator_xs_write.log",
+        tmp_path,
+        "output_bids_validator_xs_write.log",
```
yarikoptic, line 277: Also attend to subsequent use of `__file__` below

TheChymera: That points to the example output file in `tests/data`.

dandi/bids_validator_xs.py
```diff
@@ -644,6 +650,11 @@ def validate_bids(
         regex_schema,
         debug=debug,
     )
-    write_report(validation_result)
+
+    if report_path:
+        if isinstance(report_path, str):
+            write_report(validation_result, report_path=report_path)
+        else:
+            write_report(validation_result)
```
yarikoptic, line 658: that is why codecov complains since this line is not covered AFAIK.
FWIW code logic is a bit confusing due to support of the `bool` value here. I would have not bothered with "default location" (via `bool`). But if it is supported -- add a test which would test this code path to make codecov happy.

TheChymera: The problem is that if the logs as per the default path are to be dropped in the current working directory, there's no way to not drop them in the current directory :o)

I think it's an important feature since typing in the datetime pattern is annoying, and it makes sense to be able to rapidly generate unique logs when running the command in batch.

My current approach was to move the default path from the current working directory to `/var/tmp/` — let's see how well that goes down with windows tests.

TheChymera:
Hi @jwodder — I understand this is somewhat urgent and @yarikoptic might not be with us next week. Can you have a look at this and let me know if it's ok to merge even with the slight test coverage decrease? The log files get dropped either under pytest's `tmp_path` or under `/var/tmp/` which is now the default location if logging is desired but a path not specified by the user.

dandi/bids_validator_xs.py
TheChymera: well, the tests pass on windows, which is cool, I guess.

jwodder:
@TheChymera Running the tests locally still produces 33 `bids-validator-report_*.log` files in the root of the repository.

Issue #950, opened 2022-04-06
validate-bids should use lowest supported BIDS version for earlier ones


yarikoptic:
per our face-to-face discussion, if BIDSVersion has no present schemadata -- use `1.7.0+012` we carry after issuing a warning about that. As for version comparison -- seems to NOT work with `semantic_version` we already use

```shell
(git)lena:~/proj/dandi/dandi-cli-master[master]git
$> python -c 'from semantic_version import Version; print(Version("1.7.0")<Version("1.7.0+012"))' 
False

$> python -c 'from semantic_version import Version; print(Version("1.7.0+011")<Version("1.7.0+012"))'
False

$> python -c 'from semantic_version import Version; print(Version("1.7.0+013")<Version("1.7.0+012"))'
False
```
so let's not sweat about that for now with all the `+`s we have ATM.

TheChymera:
Solved in https://github.com/dandi/dandi-cli/pull/955/commits/5fc7b0652e8eb9a998e12e84f7273e12e062f21e other than not presenting a warning. Not sure what sort of warning would be appropriate...

As for version comparison Python referencing of the ASCII table has us covered:

```python
>>> "1.7.0"<"1.7.0+012"
True
>>> "1.7.0+011"<"1.7.0+012"
True
>>> "1.7.0+013"<"1.7.0+012"
False
```

AFAICT the only surprises that this could cause are due to the fact that ASCII lists uppercase before lowercase, so:

```python
>>> "1.7.0+012+DANDI002"<"1.7.0+012+dandi001"
True
```

yarikoptic:
I would prefer to avoid relying on plain string comparison of semantic versions composed from numbers, could just bite somehow unexpectedly e.g
```
>>> "1.7.0+012"< "1.7.0+99"
True
```
it is just a matter of semver to have "Identifiers MUST NOT be empty. Build metadata MUST be ignored when determining version precedence. "

<details>
<summary>although difference between <= and == and < confuses me -- may be a bug in semver?</summary> 

```python
$> python3 -c 'from semantic_version import Version; print(Version("1.7.0")==Version("1.7.0+012"))'
False

$> python3 -c 'from semantic_version import Version; print(Version("1.7.0")<Version("1.7.0+012"))'
False

$> python3 -c 'from semantic_version import Version; print(Version("1.7.0")<=Version("1.7.0+012"))'
True

```
</details>

[...]

PR #955, opened 2022-04-11
Validator fixes: fallback to our version, skip dotdirs on windows, etc

TheChymera:
Miscellaneous but important improvements developed in parallel.
I will ping for review later, there are a few more.

### DIFF
--- a/dandi/bids_validator_xs.py
+++ b/dandi/bids_validator_xs.py
@@ -35,6 +35,9 @@ def _get_paths(bids_paths):
         "/.dandi",
         "/.datalad",
         "/.git",
+        "\\.dandi",
+        "\\.datalad",
+        "\\.git",
     ]
     # `.bidsignore` is not, in fact, a BIDS file, as per:
     # https://github.com/bids-standard/bids-specification/issues/980
--- a/dandi/bids_validator_xs.py
+++ b/dandi/bids_validator_xs.py
@@ -128,8 +128,10 @@ def _extension_safety(extension):
     * Period safety: https://github.com/bids-standard/bids-specification/issues/1055
     * Hopefully this function will be deprecated soon, but it will not break safe entries.
     """
-    if extension[0] == ".":
-        extension = extension[1:]
+    if extension == "None":
+        return ""
+    if "*" in extension:
+        extension = extension.replace(".", "\\.")
     if "*" in extension:
         extension = extension.replace("*", ".*?")
 
@@ -146,7 +148,7 @@ def _add_extensions(regex_string, variant):
         regex_extensions = "({})".format("|".join(fixed_variant_extensions))
     else:
         regex_extensions = fixed_variant_extensions[0]
-    regex_string = f"{regex_string}\\.{regex_extensions}"
+    regex_string = f"{regex_string}{regex_extensions}"
 
     return regex_string
 
@@ -217,7 +219,7 @@ def load_top_level(
                 extension = _extension_safety(extension)
                 safe_extensions.append(extension)
                 extensions_regex = "|".join(safe_extensions)
-            regex = f".*?/{top_level_filename}\\.({extensions_regex})$"
+            regex = f".*?/{top_level_filename}({extensions_regex})$"
         else:
             regex = f".*?/{top_level_filename}$"
         regex_entry = {
--- a/dandi/bids_validator_xs.py
+++ b/dandi/bids_validator_xs.py
@@ -635,6 +635,14 @@ def validate_bids(
         If string, the string will be used as the output path.
         If the variable evaluates as False, no log will be written.
 
+    Returns
+    -------
+    results : dict
+        A dictionary reporting the target files for validation, the unmatched files and unmatched
+        regexes, and optionally the itemwise comparison results.
+        Keys include "schema_tracking", "path_tracking", "path_listing", "match_listing", and
+        optionally "itemwise"
+
     Examples
     --------
     >>> from schemacode import validator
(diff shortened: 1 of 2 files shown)


jwodder:
@TheChymera Is this PR finished/ready for review?  If not, please mark it as a draft until it is done.

TheChymera:
@jwodder should be good to go now. Sorry for the delay :(

dandi/bids_validator_xs.py
```diff
@@ -195,14 +217,12 @@ def load_top_level(
...
             for extension in extensions:
-                if extension[0] == ".":
-                    periodsafe_extensions.append(extension[1:])
-                else:
-                    periodsafe_extensions.append(extension)
-                extensions_regex = "|".join(periodsafe_extensions)
-                regex = f".*?/{top_level_filename}\\.({extensions_regex})$"
+                extension = _extension_safety(extension)
+                safe_extensions.append(extension)
+                extensions_regex = "|".join(safe_extensions)
```
jwodder, line 224: Shouldn't this line be unindented one level?

jwodder (changes requested):

dandi/bids_validator_xs.py
TheChymera: @jwodder you're right. Thank you.

jwodder (approved):

dandi/bids_validator_xs.py
```diff
@@ -195,14 +217,12 @@ def load_top_level(
...
             for extension in extensions:
-                if extension[0] == ".":
-                    periodsafe_extensions.append(extension[1:])
-                else:
-                    periodsafe_extensions.append(extension)
-                extensions_regex = "|".join(periodsafe_extensions)
-                regex = f".*?/{top_level_filename}\\.({extensions_regex})$"
+                extension = _extension_safety(extension)
+                safe_extensions.append(extension)
+            extensions_regex = "|".join(safe_extensions)
```
yarikoptic, line 224: eh,can't do suggestion since covers a deleted line but could be just following for a more concise etc code

```python
extensions_regex = "|".join(map(_extension_safety, extensions))
```
right?

dandi/bids_validator_xs.py
```diff
@@ -35,6 +35,9 @@ def _get_paths(bids_paths):
         "/.dandi",
         "/.datalad",
         "/.git",
+        "\\.dandi",
+        "\\.datalad",
+        "\\.git",
```
yarikoptic, line 40: why or what for, would ever `os.walk` return names with escaped `.`s?

jwodder: The dots aren't escaped here.  Two backslashes in a non-raw string equals a backslash, which is the Windows path separator.

yarikoptic: Doh, right.. is there a function to normalize one way or another on windows? You know my allergy to duplication ;-)

jwodder: If you mean always getting the same path separator regardless of OS, no.  I think the better way would be to use `pathlib` instead of string manipulations, both here and throughout the BIDS code, but that might be big enough for a separate PR.

dandi/bids_validator_xs.py
```diff
@@ -581,31 +604,28 @@ def select_schema_dir(
...
-                        except KeyError:
-                            return utils.get_schema_path()
-                    else:
+                    try:
                         schema_version = dataset_info["BIDSVersion"]
+                    except KeyError:
+                        schema_version = schema_min_version
+        if schema_min_version:
+            if schema_version < schema_min_version:
+                schema_version = schema_min_version
```
yarikoptic, line 613: re using semver (needs import) and warning discussed in https://github.com/dandi/dandi-cli/issues/950#issuecomment-1101174851

```suggestion
            if Version(schema_version) <= Version(schema_min_version):
                if schema_version != schema_min_version:
                    lgr.warning("BIDSVersion %s is less than the smallest known %s, so we will use the known one",
                        schema_version, schema_min_version)
                schema_version = schema_min_version
```

dandi/bids_validator_xs.py
```diff
@@ -581,31 +604,28 @@ def select_schema_dir(
...
-                    if force_select:
-                        try:
-                            schema_version = dataset_info["BIDSVersion"]
-                        except KeyError:
-                            return utils.get_schema_path()
-                    else:
+                    try:
                         schema_version = dataset_info["BIDSVersion"]
+                    except KeyError:
+                        schema_version = schema_min_version
```
yarikoptic, line 610: above can be oneliner
```
schema_version = dataset_info.get("BIDSVersion", schema_min_version)
```
unless you may be want to warn the user that dataset at that path lacks BIDSVersion which is REQUIRED.  So -- addd here a `lgr.warning` stating that this dataset_description lacks required field and you are choosing the schema_min_version for that

dandi/bids_validator_xs.py
yarikoptic: ok, let's make them then
```
fr"{os.sep}.dandi",
```
 or just add that `os.sep` while using these in the loop below.

dandi/bids_validator_xs.py
TheChymera: yep!

dandi/bids_validator_xs.py
TheChymera: Makes sense, the actual complaint about REQUIRED might best come out in the report — i.e. report the file is broken in the same place where we would report all broken files, rather than report here as a special case.

yarikoptic:
I adjusted title to be descriptive (will go into changelog), added labels including the one for release -- let's proceed. Thank you @TheChymera

PR #958, opened 2022-04-13
Add RemoteDandiset.has_data_standard() convenience function

bendichter:
@jwodder, what do you think of something like this?


dandi/dandiapi.py
```diff
@@ -1071,6 +1071,17 @@ def iter_upload_raw_asset(
             self, metadata=asset_metadata, jobs=jobs, replacing=replace_asset
         )
 
+    def is_nwb(self):
```
jwodder, line 1074: This needs a docstring and a type annotation.

dandi/dandiapi.py
```diff
@@ -1071,6 +1071,17 @@ def iter_upload_raw_asset(
...
+        assetsSummary = self.get_raw_metadata()["assetsSummary"]
+        if "dataStandard" not in assetsSummary:
+            return False
+        return any(
+            [
```
jwodder, line 1079: There's no reason to create a list here.  Get rid of the square brackets.

jwodder (changes requested):

dandi/dandiapi.py
```diff
@@ -1071,6 +1071,20 @@ def iter_upload_raw_asset(
             self, metadata=asset_metadata, jobs=jobs, replacing=replace_asset
         )
 
+    def is_nwb(self) -> bool:
+        """
+        Returns True if the dandiset contains NWB file assets.
```
jwodder, line 1076:
```suggestion
        Returns True if the Dandiset contains one or more NWB file assets.
```

jwodder (changes requested):

dandi/dandiapi.py
```diff
@@ -1071,6 +1071,20 @@ def iter_upload_raw_asset(
             self, metadata=asset_metadata, jobs=jobs, replacing=replace_asset
         )
 
+    def is_nwb(self) -> bool:
```
yarikoptic, line 1074: @jwodder I feel that we had a similar case somewhere where I have pushed for replacing `is_nwb` and `is_zarr` with something like `format` property .  Do you remember what was it? 
@bendichter I think that situation is similar here and we should not overfit for NWB here and needing to grow API for any type of file formats which might be in the dandiset.  Moreover, `is_nwb` is IMHO a suboptimal name. if anything should be `has_nwb` since dandiset is not of NWB format.

What if there is a more general `has_any(format: str)` where it would understand `format` *as-is* if `.startswith("RRID")` and otherwise knows a few mappings, such as the used RRIR for `NWB`, and may be right away one for zarr if we know of any (@satra)?

This was we would not need to expand API and duplicate code to support more "formats"

jwodder: @yarikoptic I think you're thinking of the discussion that lead to [this commit](https://github.com/dandi/dandi-cli/commit/c0a32efff54c34e2b79c2d551def3a396a2e1fb6).

bendichter: what would be my new syntax for determining if a dandiset contains NWB files?

yarikoptic: @jwodder -- thank you! Indeed it is `BaseRemoteAssset.asset_type`  https://github.com/dandi/dandi-cli/blob/master/dandi/dandiapi.py#L1353 I had in mind.  After starting to craft a patch for @bendichter I realized that it allowed to differentiate at different "level" -- only `blob` vs `zarr`, not at the "higher" (if not orthogonal) level of the asset format or "mime" type. `blob` could be a nifti or nwb or any other of hundreds of mimetypes or in other words `dataStandard`s as we call that in our assets summary which @bendichter is to use here.

So I think it could be something like (prototyping to expedite since not sure what part of "syntax" you had in mind @bendichter  ;) ):

```
    def has_data_standard(data_standard: str) -> bool:
        """
        Returns True if the Dandiset contains one or more assets conformant to a given type.
    
       `data_standard` should be an `RRID:` identifier, or a string for one of the known data standards
       (ATM known only `nwb`). Code then checks for the corresponding RRID to be listed in the
       "dataStandard" field of the assetsSummary of the dandiset.
        """
        known = {
            'nwb': "RRID:SCR_015242",
        }
        if not data_standard.startswith('RRID:'):
                if data_standard.lower() in known_types:
                    data_standard = known_types[data_standard.lower()]
                else:
                    raise ValueError(f"Do not know about data standard {data_standard}. Provide an RRID or one of known: {', '.join(known)}")
        assetsSummary = self.get_raw_metadata()["assetsSummary"]
        if "dataStandard" not in assetsSummary:
            return False
        return any(
            x["identifier"] == data_standard for x in assetsSummary["dataStandard"]
        )
```
    
I also wonder if we should discriminate the use case of `dataStandard` being not populated and return None, so code outside could tell apart the case where such is not populated. Code outside would still be able to use it as a `bool`.  But not sure if worth it.

bendichter: I meant how would I use this function as a user. So it would be `remote_dandiset.has_data_standard("nwb")`? That works for me.

I'm fine with returning `None` if no data standard is defined, though I probably wouldn't use that feature.

yarikoptic: > `remote_dandiset.has_data_standard("nwb")`? That works for me.

yes, that is what I thought about. @jwodder -- do you see a better naming/signature?  What do you think about `None` for return?

jwodder: @yarikoptic I have no opinion on the name.  I think a lack of `dataStandard` should result in an exception, as it's an exceptional condition.

bendichter: @jwodder I think that would require me to use this method within a try/except which I'd rather not do.

yarikoptic: ok, looking at the schema and code

```shell
$> git grep -p dataStand
dandischema/metadata.py=def _add_asset_to_stats(assetmeta: Dict[str, Any], stats: _stats_type) -> None:
dandischema/metadata.py:    stats["dataStandard"] = stats.get("dataStandard", [])
dandischema/metadata.py:        if models.nwb_standard not in stats["dataStandard"]:
dandischema/metadata.py:            stats["dataStandard"].append(models.nwb_standard)
dandischema/metadata.py:        if models.bids_standard not in stats["dataStandard"]:
dandischema/metadata.py:            stats["dataStandard"].append(models.bids_standard)
dandischema/models.py=class AssetsSummary(DandiBaseModel):
dandischema/models.py:    dataStandard: Optional[List[StandardsType]] = Field(readOnly=True)
```
dataStandard is optional and populated only if there is any asset, so empty dandisets do not have it , as verified on `curl -X GET "https://api.dandiarchive.org/api/dandisets/000233/versions/draft/" -H  "accept: application/json" `, so I think it would not be an exceptional condition. Ok, let's keep it just to `False` in case of its absence.

yarikoptic: @bendichter what are your plans on this PR - should we generalize it the way I proposed or some other way, or forget/close for now?

yarikoptic: `has_any` is too wide semantically for me, and it should not default to `NWB` IMHO, so what about suggested `has_data_standard(data_standard: str)` ? ;-)

bendichter: @yarikoptic ok I can adjust this PR to that syntax

bendichter:
@yarikoptic before I write tests for this, how does this syntax look?

dandi/dandiapi.py
```diff
@@ -1071,19 +1073,24 @@ def iter_upload_raw_asset(
...
-        assetsSummary = self.get_raw_metadata()["assetsSummary"]
-        if "dataStandard" not in assetsSummary:
+        if data_standard in DATA_STANDARD_MAP:
+            rrid = DATA_STANDARD_MAP[data_standard]
+        elif data_standard.startswith("RRID:"):
+            rrid = data_standard
+        else:
+            raise ValueError(
+                f"'data_standard' must be an RRID (of form 'RRID:XXX_NNNNNNN`) or one "
+                f"of the following values: {DATA_STANDARD_MAP.keys()}"
```
jwodder, line 1088: `dict.keys()` doesn't stringify in a human-friendly way.  Use `", ".join(DATA_STANDARD_MAP.keys())` instead.

dandi/dandiapi.py
```diff
@@ -1071,19 +1073,24 @@ def iter_upload_raw_asset(
...
 
-    def is_nwb(self) -> bool:
+    def has_data_standard(self, data_standard: str) -> bool:
         """
-        Returns True if the Dandiset contains one or more NWB file assets.
-
-        This is determined by checking for "RRID:SCR_015242" in the "dataStandard" field
-        of the assetsSummary of the dandiset.
+        Returns True if the Dandiset contains one or more files of the indicated
+        standard. Otherwise, returns False.
```
jwodder, line 1079: The docstring should document the accepted data standards (or, at the very least, the RRID input format).

jwodder (changes requested):

dandi/dandiapi.py
```diff
@@ -1071,6 +1073,25 @@ def iter_upload_raw_asset(
...
+        """
+        if data_standard in DATA_STANDARD_MAP:
+            rrid = DATA_STANDARD_MAP[data_standard]
+        elif data_standard.startswith("RRID:"):
+            rrid = data_standard
+        else:
+            raise ValueError(
+                f"'data_standard' must be an RRID (of form 'RRID:XXX_NNNNNNN`) or one "
+                f"of the following values: {DATA_STANDARD_MAP.keys()}"
```
bendichter, line 1088:
```suggestion
                f"of the following values: {", ".join(DATA_STANDARD_MAP.keys())}"
```

dandi/dandiapi.py
```diff
@@ -1071,6 +1076,31 @@ def iter_upload_raw_asset(
...
+        if data_standard in DATA_STANDARD_MAP:
+            rrid = DATA_STANDARD_MAP[data_standard]
+        elif data_standard.startswith("RRID:"):
+            rrid = data_standard
+        else:
+            raise ValueError(
+                f"'data_standard' must be an RRID (of form 'RRID:XXX_NNNNNNN`) or one "
+                f"of the following values: {", ".join(DATA_STANDARD_MAP.keys())}"
+            )
+        assets_summary = self.get_raw_metadata()["assetsSummary"]
```
yarikoptic, line 1099: well, `assetsSummary` seems to be non optional, so must be present in a valid metadata record, but if not we would get `KeyError` here... which could be good as to alert user that smth is wrong with metadata in that dandisets.  But should it be an exception really or just a warning?  I would lean to a be a warning and just return `False`. WDYT @jwodder and @satra ?

yarikoptic:
> @yarikoptic before I write tests for this, how does this syntax look?

looks ok to me ;) left few insignificant comments

yarikoptic:
@bendichter I have converted PR back to draft since needs more work.  some comments are left above, and also would be nice to add some rudimentary unittest. Overall I think it would be a nice addition, so would be great to see if finished.

bendichter:
I could use some guidance in how to write a test for this

jwodder:
@bendichter Could you rebase or make an empty commit or something in order for the CI to run again?

jwodder:
@bendichter 

> Can I just use published dandisets on the archive?

That's one option.  Normally, I'd suggesting just using the `nwb_dandiset` and `text_dandiset` fixtures, but that would require the Dockerized server to produce the `assetsSummary` immediately, and I don't know how long that takes.

bendichter:
ok, I added some tests. Let me know if these will work

jwodder:
@bendichter Please run pre-commit on your code.

bendichter:
@jwodder  oh, I see. Is there a reason to not set the pre-commit hooks as GitHub Actions? We do this for NeuroConv and it works great. All PRs are automatically Black formatted, so there is no need for developers to worry about installing and running pre-commit.

dandi/dandiapi.py
yarikoptic: @jwodder and @bendichter WDYT - better be strict or robust?

bendichter: I'd be fine with changing this to throw a warning if assetsSummary is missing

dandi/dandiapi.py
```diff
@@ -1223,6 +1228,30 @@ def iter_upload_raw_asset(
...
+        if data_standard in DATA_STANDARD_MAP:
+            rrid = DATA_STANDARD_MAP[data_standard]
+        elif data_standard.startswith("RRID:"):
+            rrid = data_standard
+        else:
+            raise ValueError(
+                "'data_standard' must be an RRID (of form 'RRID:XXX_NNNNNNN`) or one "
+                f"of the following values: {', '.join(DATA_STANDARD_MAP.keys())}"
+            )
+        assets_summary = self.get_raw_metadata()["assetsSummary"]
```
bendichter, line 1250:
```suggestion
        assets_summary = self.get_raw_metadata().get("assetsSummary")
        if assets_summary is None:
            warnings.warn(f"The raw metadata of RemoteDandiset {self.identifier} does not contain 'assetsSummary'.")
```

yarikoptic: I guess we would need more like
```suggestion
        assets_summary = self.get_raw_metadata().get("assetsSummary")
        if assets_summary is None:
            warnings.warn(
                f"The raw metadata of RemoteDandiset {self.identifier} does not contain 'assetsSummary'. "
                f"Assuming that it does not contain {data_standard}.")
            return False
```
or alike or otherwise kaboom would happen on the following line.

yarikoptic:
I think it is still a legit and potentially useful interface. @bendichter - what do you think?  it might benefit from adhering to the same design principle of using Enum's instead of free strings, as RFed into by #1357 yesterday

bendichter:
@yarikoptic  what do you mean by Enum? Do you mean using Literal in the typehints?

bendichter:
OK, that looks like a fine approach. I'm not able to work on this right now. Feel free to take this over and implement Enums if you want to push this through now, otherwise, I can get to this next week

dandi/dandiapi.py
```diff
@@ -1223,6 +1228,35 @@ def iter_upload_raw_asset(
...
+        else:
+            raise ValueError(
+                "'data_standard' must be an RRID (of form 'RRID:XXX_NNNNNNN`) or one "
+                f"of the following values: {', '.join(DATA_STANDARD_MAP.keys())}"
+            )
+        assets_summary = self.get_raw_metadata().get("assetsSummary")
+        if assets_summary is None:
+            warnings.warn(
+                f"The raw metadata of RemoteDandiset {self.identifier} does not contain 'assetsSummary'. "
+                f"Assuming that it does not contain {data_standard}.")
```
jwodder, line 1254: @bendichter You do not appear to be using pre-commit; if you were, I'm quite certain black would have moved that closing parenthesis to the next line.  Please install pre-commit for the repository and run it.

bendichter: @jwodder in this particular case, I accepted @yarikoptic 's proposed change in GitHub. I'll make sure pre-commit is installed and running properly for future PRs

yarikoptic:
> OK, that looks like a fine approach. I'm not able to work on this right now. Feel free to take this over and implement Enums if you want to push this through now, otherwise, I can get to this next week

:letsdoit:! @bendichter -- would you have time/effort to finish it up?

yarikoptic:
I will close this for now -- we know where to find it @bendichter  ;-)

PR #961, opened 2022-04-15
Better detection of empty directories when download-syncing Zarrs

jwodder:


yarikoptic:
not clear from the description on how it is better, but seems to come with a test to demonstrate its superiority, so must be right :) Thank you @jwodder

PR #962, opened 2022-04-18
BIDS metadata read-in pilot.

TheChymera:
Will return minimal BIDS metadata via DANDI commands:

```console
chymera@decohost ~/.data2/dandi $ dandi ls /home/chymera/.data2/dandi/000026/derivatives/EPIC/sub-I46/ses-MRI/anat/sub-I46_ses-MRI_flip-3_VFA.nii.gz
2022-04-18 06:05:44,701 [ WARNING] A newer version (0.39.1) of dandi/dandi-cli is available. You are using 0.38.0+43.g648cb22
2022-04-18 06:05:44,875 [    INFO] Note: NumExpr detected 12 cores but "NUMEXPR_MAX_THREADS" not set, so enforcing safe limit of 8.
2022-04-18 06:05:44,875 [    INFO] NumExpr defaulting to 8 threads.
- bids_version: 1.7.0+012+dandi001
  path: /home/chymera/.data2/dandi/000026/derivatives/EPIC/sub-I46/ses-MRI/anat/sub-I46_ses-MRI_flip-3_VFA.nii.gz
  session_id: MRI
  size: 0
  subject_id: I46
2022-04-18 06:05:45,999 [    INFO] Logs saved in /home/chymera/.cache/dandi-cli/log/20220418100544Z-27525.log
```

- [ ] update BIDS version reporting once support is merged
   - [ ] will also require including the used BIDS version in the validator output directory
- [ ] nicely crawl up to the `dataset_description.json` to populate additional fields

### DIFF
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -25,6 +25,7 @@ from uuid import uuid4
 from xml.dom.minidom import parseString
 
 from dandischema import models
+import h5py
 import requests
 import tenacity
 
@@ -43,7 +44,27 @@ from .utils import ensure_datetime, get_mime_type, get_utcnow_datetime
 
 lgr = get_logger()
 
+# Remove hard-coding when current version fallback is merged.
+BIDS_VERSION = "1.7.0+012+dandi001"
 
+BIDS_TO_DANDI = {
+    "subject": "subject_id",
+    "session": "session_id",
+}
+
+
+def _rename_bids_keys(bids_metadata, conversion=BIDS_TO_DANDI):
+    """Standardize BIDS metadata field naming to match DANDI."""
+    converted_bids_metadata = {}
+    for key in bids_metadata.keys():
+        if key in conversion:
+            converted_bids_metadata[conversion[key]] = bids_metadata[key]
+        else:
+            converted_bids_metadata[key] = bids_metadata[key]
+    return converted_bids_metadata
+
+
+# Disable this for clean hacking
 @metadata_cache.memoize_path
 def get_metadata(path: Union[str, Path]) -> Optional[dict]:
     """Get selected metadata from a .nwb file or a dandiset directory
@@ -70,6 +91,19 @@ def get_metadata(path: Union[str, Path]) -> Optional[dict]:
         except ValueError as exc:
             lgr.debug("Failed to get metadata for %s: %s", path, exc)
             return None
+
+    # Pretty fragile way to determine that it's not nwb.
+    # Ideally can find something better.
+    try:
+        h5py.File(path)
+    except OSError:
+        from .bids_validator_xs import validate_bids
+
+        meta = validate_bids(path, schema_version=BIDS_VERSION)
+        meta = meta["match_listing"][0]
+        meta["bids_version"] = BIDS_VERSION
+        meta = _rename_bids_keys(meta)
+        return meta
 
     if nwb_has_external_links(path):
         raise NotImplementedError(
@@ -195,7 +229,9 @@ def _check_decimal_parts(age_parts: List[str]) -> bool:
             flags=re.I,
         )
         if m is None:
-            raise ValueError(f"Failed to parse the trailing part of age {age_parts[-1]!r}")
+            raise ValueError(
+                f"Failed to parse the trailing part of age {age_parts[-1]!r}"
+            )
         age_parts = age_parts[:-1] + [m[i] for i in range(1, 3) if m[i]]
     decim_part = ["." in el for el in age_parts]
     return not (any(decim_part) and any(decim_part[:-1]))
--- a/dandi/cli/cmd_ls.py
+++ b/dandi/cli/cmd_ls.py
@@ -368,6 +368,7 @@ def get_metadata_ls(
         if (
             not op.isdir(path)
             and "nwb_version" not in rec
+            and "bids_version" not in rec
             and (keys and "nwb_version" in keys)
         ):
             # Let's at least get that one
--- a/dandi/consts.py
+++ b/dandi/consts.py
@@ -30,6 +30,7 @@ metadata_nwb_subject_fields = (
 metadata_nwb_dandi_fields = ("cell_id", "slice_id", "tissue_sample_id", "probe_ids")
 
 metadata_nwb_computed_fields = (
+    "bids_version",
     "number_of_electrodes",
     "number_of_units",
     "nwb_version",
@@ -75,7 +76,7 @@ PUBLISHED_VERSION_REGEX = r"[0-9]+\.[0-9]+\.[0-9]+"
 
 #: Regular expression for a valid Dandiset version identifier.  This regex is
 #: not anchored.
-VERSION_REGEX = fr"(?:{PUBLISHED_VERSION_REGEX}|draft)"
+VERSION_REGEX = rf"(?:{PUBLISHED_VERSION_REGEX}|draft)"
 
 
 class EmbargoStatus(Enum):


TheChymera:
This now correctly reports the schema version as well as the fallback warning:

```console
chymera@decohost ~/.data2/dandi $ dandi ls /home/chymera/.data2/dandi/000026/derivatives/EPIC/sub-I46/ses-MRI/anat/sub-I46_ses-MRI_flip-3_VFA.nii.gz
2022-04-21 18:01:27,253 [    INFO] Note: NumExpr detected 12 cores but "NUMEXPR_MAX_THREADS" not set, so enforcing safe limit of 8.
2022-04-21 18:01:27,253 [    INFO] NumExpr defaulting to 8 threads.
/home/chymera/.data2/dandi/000026/derivatives/EPIC/sub-I46/ses-MRI/anat/sub-I46_ses-MRI_flip-3_VFA.nii.gz
2022-04-21 18:01:28,372 [ WARNING] BIDSVersion 1.5.0 is less than the minimal working 1.7.0+012+dandi001. Falling back to 1.7.0+012+dandi001. To force the usage of earlier versions specify them explicitly when calling the validator.
- bids_schema_version: 1.7.0+012+dandi001
  path: /home/chymera/.data2/dandi/000026/derivatives/EPIC/sub-I46/ses-MRI/anat/sub-I46_ses-MRI_flip-3_VFA.nii.gz
  session_id: MRI
  size: 0
  subject_id: I46
2022-04-21 18:01:29,905 [    INFO] Logs saved in /home/chymera/.cache/dandi-cli/log/20220421220126Z-13527.log
```

@satra could you let me know if this corresponds to what you mentioned during week's meeting? Am I missing something (other than a lot of metadata still not being read in)?

satra:
@TheChymera - thank you. this is a good starting point. 

we also need to add information about modality, technique, variablesmeasured. 
presumably this can be mapped on from the bids schema. also it would be nice to augment encodingformat to suggest nifti or other types of files, rather than just gzip.

eventually we should add some metadata about instruments as well into the asset metadata, when available in the corresponding json file.

TheChymera:
I think this would be good for a preliminary merge, again to see how well it integrates with workflows. Dealing with the open task, which requires extending the validation capabilities, is more of a validator issue than integrating what the validator currently does with `dandi ls`.

@jwodder @yarikoptic would you mind reviewing this? The only part I am a bit concerned by is [how we “detect” BIDS](https://github.com/dandi/dandi-cli/pull/962/files#diff-efe33fb24af3f1d5572d20580fa315cc3b2a8d6a69992c260fd900f643f2e924R96). Happy to do it nicer if you have any suggestions.

dandi/bids_validator_xs.py
```diff
@@ -711,6 +711,16 @@ def validate_bids(
         regex_schema,
         debug=debug,
     )
+    # Record schema version.
+    # Not sure whether to incorporate in validation_result.
+    if bids_schema_dir == os.path.join(
+        os.path.abspath(os.path.dirname(__file__)),
+        "data/schema",
```
jwodder, line 718: Will this work on Windows?  It would be better to instead do `"data", "schema"` — or, even better, use `pathlib` instead.

dandi/metadata.py
```diff
@@ -71,6 +91,19 @@ def get_metadata(path: Union[str, Path]) -> Optional[dict]:
             lgr.debug("Failed to get metadata for %s: %s", path, exc)
             return None
 
+    # Pretty fragile way to determine that it's not nwb.
+    # Ideally can find something better, like `is_nwb` or `is_bids`.
+    try:
+        h5py.File(path)
```
jwodder, line 97: I do not like this one bit.  How would a human looking at a dataset know whether something is BIDS or not?

jwodder, line 97: I do not like this one bit.  How would a human looking at a dataset know whether something is BIDS or not?

jwodder (changes requested):

dandi/metadata.py
satra: also bids datasets themselves can have nwb files. 

i would say bids datasets in dandi would still have dandiset.yaml. and that the check would be that each bids dataset in a dandiset will have a dataset_description.json.

algorithm for is_bids:
- walk up from file to find dataset_description.json and a dandiset.yaml 

if dandiset.yaml occurs before a dataset_description then the dataset is not bids. 

also for this PR, which is for extracting metadata, it seems the simplest check would be, is there a same named sidecar json file?

this also means when dandi cli downloads a bids dandiset, it has to download the relevant dataset_description.json (there are multiple). this is to support partial uploads for large dandisets that are bids. the alternative would be to break up all bids datasets whether "raw" or "derived" into separate dandisets, and at least for the ones we have now, it would make that ugly, and we need a way to link better through provenance.

@TheChymera - would be happy to chat briefly about this. i think in the first pass this can be simplified quite a bit, and all it may require is simply parsing the file name into sub, session, sample, stain, modality.

TheChymera: @jwodder improved detection as per @satra 's suggestion. Also made the path creation more windows-robust (though I think it *might* have worked anyway, we noticed in some other PRs that `os.path` automagically windows-ized things on windows).

dandi/metadata.py
```diff
@@ -43,7 +44,63 @@
...
+BIDS_TO_DANDI = {
+    "subject": "subject_id",
+    "session": "session_id",
+}
+
+
+def _rename_bids_keys(bids_metadata, conversion=BIDS_TO_DANDI):
+    """Standardize BIDS metadata field naming to match DANDI."""
+    converted_bids_metadata = {}
+    for key in bids_metadata.keys():
```
yarikoptic, line 58:
```suggestion
    for key in bids_metadata:
```

yarikoptic, line 58:
```suggestion
    for key, value in bids_metadata.values():
```

dandi/metadata.py
```diff
@@ -43,7 +44,63 @@
...
+        search.
+
+    Returns
+    -------
+    bool
+    """
+    from .utils import find_files
+
+    path = Path(path)
+    if [i for i in find_files(root_marker, path)]:
```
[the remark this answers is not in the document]
yarikoptic: but I overall I don't think you need this special clause -- it should all work in consistent loop below,  I will suggest...

dandi/metadata.py
```diff
@@ -43,7 +44,63 @@
...
+        return True
+    search_path = path
+    while True:
+        search_path = search_path.parent
+        bids_marker_candidate = search_path / root_marker
+        end_marker_candidate = search_path / end_marker
+        if bids_marker_candidate.is_file():
```
yarikoptic, line 95: could be a symlink, 
```suggestion
        if bids_marker_candidate.is_file() or bids_marker_candidate.is_link():
```
just check if the right name, and iirc there is no `.lexists` analog in pathlib (since then we could just use that one).
The same for end_marker condition below

dandi/metadata.py
```diff
@@ -43,7 +44,63 @@
...
+    """
+    from .utils import find_files
+
+    path = Path(path)
+    if [i for i in find_files(root_marker, path)]:
+        return True
+    search_path = path
+    while True:
+        search_path = search_path.parent
+        bids_marker_candidate = search_path / root_marker
```
yarikoptic, line 93: align with variable names:
```suggestion
        root_marker_candidate = search_path / root_marker
```
(adjust usage below).

dandi/metadata.py
```diff
@@ -71,6 +128,20 @@ def get_metadata(path: Union[str, Path]) -> Optional[dict]:
...
             return None
 
+    # Somewhat less fragile search than previous proposals,
+    # could still be augmented with `_is_nwb` to disambiguate both cases
+    # at the detection level.
+    if _path_in_bids(path):
+        from .bids_validator_xs import validate_bids
+
+        _meta = validate_bids(path)
+        meta = _meta["match_listing"][0]
```
yarikoptic, line 138: note: referencings like `[0]` 
- make it not clear what / why we are getting the first element only
- source of trouble since would crash if list/tuple is empty
- sooner than later we should overhaul that `validate_bids` to provide a result which makes returned value "more obvious"

dandi/metadata.py
```diff
@@ -71,6 +128,20 @@ def get_metadata(path: Union[str, Path]) -> Optional[dict]:
...
+        meta["bids_schema_version"] = _meta["bids_schema_version"]
+        meta = _rename_bids_keys(meta)
+        return meta
+    else:
+        h5py.File(path)
```
yarikoptic, line 143: what is this added `else` for?

yarikoptic (changes requested):
Code needs some added tests and RF

dandi/metadata.py
TheChymera: agree, in principle that `[0]` is not nice, but (a) possibly a fix for another PR, (b) it actually makes sense here. A file name could be matched by 2 patterns. That in turn could be an issue with the schema (in which case QA should be done in upstream BIDS, not here), or it could be having `participants.json` *and* `participants.tsv` — not entirely sure whether BIDS agrees that this is broken, and even if it is probably not a reason to invalidate any of the files specifically. Either or both could contain valid metadata.

dandi/bids_validator_xs.py
```diff
@@ -711,6 +711,19 @@ def validate_bids(
...
+    # Record schema version.
+    # Not sure whether to incorporate in validation_result.
+    if bids_schema_dir == os.path.join(
+        os.path.abspath(os.path.dirname(__file__)),
+        "data",
+        "schema",
+    ):
+        # Declare we are using live version,
+        # string will evaluate as larger than numbered versions.
+        schema_version = "live"
```
yarikoptic, line 723: oh well, semver would puke on it quite loud happen someone decide to use that:

```
$> python3 -c 'from semantic_version import Version; print(Version("live.0.0")>Version("1.7.0+012"))'
  File "/usr/lib/python3/dist-packages/semantic_version/base.py", line 105, in __init__
    major, minor, patch, prerelease, build = self.parse(version_string, partial)
  File "/usr/lib/python3/dist-packages/semantic_version/base.py", line 311, in parse
    raise ValueError('Invalid version string: %r' % version_string)
ValueError: Invalid version string: 'live.0.0'
1 24249 ->1.....................................:Wed 11 May 2022 04:25:08 PM EDT:.
lena:~
$> python3 -c 'from semantic_version import Version; print(Version("live")>Version("1.7.0+012"))' 
  File "/usr/lib/python3/dist-packages/semantic_version/base.py", line 105, in __init__
    major, minor, patch, prerelease, build = self.parse(version_string, partial)
  File "/usr/lib/python3/dist-packages/semantic_version/base.py", line 311, in parse
    raise ValueError('Invalid version string: %r' % version_string)
ValueError: Invalid version string: 'live'
```

since I still hope that eventually we arrive at proper semver comparison, let's do indeed some rudiculous large number as you had it but semver friendly:

```suggestion
        schema_version = "99999.0.0"
```

and semver would be happy too and it would work until we retire.  I will accept my invaluable contribution right away to see if tests survive

TheChymera:
oh, I was trying to commit this, but you already did. Ok `"9999.0.0"` sounds good, it's what I was going for with 9999 initially. It's what we use for Getnoo packages to mark them as “live”. In any case, good for merge now? @yarikoptic

yarikoptic:
yeap, let's proceed and see where it takes us ;)

yarikoptic:
I will add it a `minor` level of importance although it is not breaking (hopefully) anything as a notable semantical change in operation ;)

dandi/metadata.py
```diff
@@ -71,6 +116,19 @@ def get_metadata(path: Union[str, Path]) -> Optional[dict]:
...
+    # at the detection level.
+    if _path_in_bids(path):
+        from .bids_validator_xs import validate_bids
+
+        _meta = validate_bids(path)
+        meta = _meta["match_listing"][0]
+        meta["bids_schema_version"] = _meta["bids_schema_version"]
+        meta = _rename_bids_keys(meta)
+        return meta
+    h5py.File(path)
```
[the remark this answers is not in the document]
yarikoptic: I feel like I had/asked the same question -- missed in the re-review so still interested to hear ;)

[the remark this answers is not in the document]
yarikoptic: or may be then I thought "it is an undocumented intended way to check if it is an .nwb (hdf5) file and if not -- raise exception so outside code would then fall into extracting metadata from bids"... is that it @TheChymera ?

Issue #946, opened 2022-04-05
Analyze and possibly speed up testing


yarikoptic:
@satra raised concern in https://github.com/dandi/dandischema/pull/128#issuecomment-1086380260  and since we (or well, I) enforce our goodness of downstream testing against dandi-cli upon other components, it would be great to indeed ensure that we do not impair DX.  Need to 
- analyze what tests/fixtures we spend most time on?
- analyze and implement what could be speed up to speed up testing?

there was effort/attempt to parallelize testing (#194) which IIRC encountered a "show stopper" at that point, may be worth resurrecting since most of the tests are probably bottlenecking in IO and parallelization should have positive effect.

jwodder:
From running the tests locally, the slowest 10 durations as reported by pytest are:

```
============================= slowest 10 durations =============================
43.32s call     dandi/tests/test_bids_validator_xs.py::test_bids_datasets
42.35s call     dandi/tests/test_validate.py::test_validate_bids
31.97s setup    dandi/tests/test_dandiapi.py::test_upload
14.49s call     dandi/tests/test_dandiapi.py::test_publish_and_manipulate
11.70s teardown dandi/tests/test_validate.py::test_validate_bids
8.49s call     dandi/tests/test_upload.py::test_upload_different_zarr_entry_conflicts
8.14s call     dandi/tests/test_dandiapi.py::test_get_dandiset_published_other_version[True]
8.05s call     dandi/tests/test_upload.py::test_upload_different_zarr_file_to_parent_dir
7.14s call     dandi/tests/test_download.py::test_download_newest_version
6.65s call     dandi/tests/test_dandiapi.py::test_get_dandiset_published_other_version[False]
```

Note that the "setup" and "teardown" entries are actually for the setup & teardown of the Dockerized dandi-archive environment.

yarikoptic:
@TheChymera we might need to look into making bids testing leaner

TheChymera:
one thing I can do is strip [the current testdata repo](https://github.com/dandi/bids-examples) down to a smaller amount of datasets. Other than that, they can probably also be parallelized, but validation is pretty quick at this stage so it's mainly download.

TheChymera:
https://github.com/dandi/bids-examples/commit/0553aef7d9e7e1d5f042f54178a3a0b31e92bf16 took it down from 34s locally to 24s — another thing that's taking a bit is the full debugging output to stdout. I could:

[...]

PR #964, opened 2022-04-19
Combine subtests of `test_publish_and_manipulate`

jwodder:
Part of #946.

This reduces the runtime of the test from 14.49s to 9.82s.


yarikoptic:
I haven't done careful analysis if all testing is preserved but looks kosher.

coolio, thanks @jwodder .

PR #971, opened 2022-04-21
keeping logs cleaner in dedicated directory

TheChymera:
Very minor fix, prevents default logs from unceremoniously cluttering `/var/tmp/`.

### DIFF
--- a/dandi/bids_validator_xs.py
+++ b/dandi/bids_validator_xs.py
@@ -452,8 +452,8 @@ def validate_all(
 
 def write_report(
     validation_result,
-    report_path="{logdir}/bids-validator-report_{date}.log",
-    datetime_format="%Y%m%d-%H%M%S",
+    report_path="{logdir}/bids-validator-report_{datetime}-{pid}.log",
+    datetime_format="%Y%m%d%H%M%SZ",
 ):
     """Write a human-readable report based on the validation result.
 
@@ -465,10 +465,10 @@ def write_report(
         The "itemwise" value, if present, should be a list of dictionaries, with keys including
         "path", "regex", and "match".
     report_path : str, optional
-        A path under which the report is to be saved, `logdir` and `date` are available
-        as variables for string formatting, and will be expanded to the application log
-        directory and current datetime (as per the `datetime_format` parameter),
-        respectively.
+        A path under which the report is to be saved, `datetime`, `logdir`, and `pid`
+        are available as variables for string formatting, and will be expanded to the
+        current datetime (as per the `datetime_format` parameter), application log
+        directory, and process ID, respectively.
     datetime_format : str, optional
         A datetime format, optionally used for the report path.
 
@@ -481,7 +481,8 @@ def write_report(
 
     report_path = report_path.format(
         logdir=logdir,
-        date=datetime.datetime.now().strftime(datetime_format),
+        datetime=datetime.datetime.utcnow().strftime(datetime_format),
+        pid=os.getpid(),
     )
     report_path = os.path.abspath(os.path.expanduser(report_path))
     try:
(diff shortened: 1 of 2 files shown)


dandi/bids_validator_xs.py
```diff
@@ -450,7 +450,7 @@ def validate_all(
 
 def write_report(
     validation_result,
-    report_path="/var/tmp/bids-validator-report_{}.log",
+    report_path="/var/tmp/bids-validator/report_{}.log",
```
satra, line 453: why not write this in the same directory where dandi-cli puts its logs?

yarikoptic: +1 on that -- there should be no hardcoded OS-specific paths. ATM we (still, might switch to platformdirs later) use appdirs:
```
$> git grep appdi
dandi/cli/command.py:import appdirs
dandi/cli/command.py:    logdir = appdirs.user_log_dir("dandi-cli", "dandi")
```

yarikoptic (changes requested):
replace OS specific paths

dandi/bids_validator_xs.py
TheChymera: fixed and added better logger support as well.

dandi/bids_validator_xs.py
```diff
@@ -519,6 +528,7 @@ def write_report(
         else:
             f.write("All mandatory BIDS files were found.\n")
         f.close()
+    lgr.info(f"BIDS validation log written to {report_path}")
```
yarikoptic, line 531: Please don't use f-strings for logging, use %s and place value as argument. Rationale could be found in some prior PRs

dandi/bids_validator_xs.py
```diff
@@ -450,7 +452,7 @@ def validate_all(
 
 def write_report(
     validation_result,
-    report_path="/var/tmp/bids-validator-report_{}.log",
+    report_path="{logdir}/bids-validator-report_{date}.log",
```
yarikoptic, line 455: do not use `/` but `os.sep` to provide OS agnostic one?

To say the truth -- I do not see really a reason for all this automated saving into some generic directory which is disconnected from the dataset where validation is done.  Imagine that `ls` by default would have `ls`ed content into some global location.

If you were to be fancy, I would have also defined some `dataset` path variable (but it isn't known, would need to be passed).  But then again -- why bother since outside code could easily provide it.

Also even with a detailed datetime to a second, collision could happen and thus default one better has some process ID . Altogether then may be make it by default the same as we have for logs

```
dandi/cli/command.py:        logdir, "{:%Y%m%d%H%M%SZ}-{}.log".format(datetime.utcnow(), os.getpid())
```
which includes that PID and explicit time zone (or lack of such).

And since it is `datetime`, call it such instead of `date`  to be more factual.

dandi/bids_validator_xs.py
```diff
@@ -450,7 +452,7 @@ def validate_all(
...
     datetime_format="%Y%m%d-%H%M%S",
```
yarikoptic, line 456:
```suggestion
    datetime_format="%Y%m%d%H%M%SZ",
```
per above comment, and use utcnow

yarikoptic (changes requested):
if to be fancy let's be fancy consistently with the logs logging ;)

TheChymera:
@yarikoptic watiting for tests. All fixes suggested seem to be in, except `/`.

As per https://ppb.chymera.eu/d69110.txt (the `report_path` variable value is returned in between repeated prints of the strings “lulu” and “lala”) I think os.path somehow already fixes this up, though I am not sure where since I don't have windows locally.

Given that this still auto-works, I think we might leave it as `/` so as to not clutter the string with additional variables for formatting. Alternatively, if you want to make it extra-safe, I can explicitly replace `/` with `os.sep`... tough again, that won't actually be doing anything.

yarikoptic:
didn't test on windows if would be storing ok given a mix of `\` and `/` in the path, but let's assume so and proceed

PR #972, opened 2022-04-21
Strip trailing slash from URLs in `/server-info`

jwodder:
See https://github.com/dandi/dandi-archive/issues/1052


yarikoptic:
I think there is no side effects but didn't check -- let's see in master

PR #974, opened 2022-04-21
Update `test_parse_api_url[DANDI:000027]`

### LINKED ISSUE #968: `test_parse_api_url[DANDI:000027]` broken
Due to changes in the redirector etc., `parse_dandi_url("DANDI:000027")` is currently returning a `DandisetURL` in which the `version_id` is `None`, whereas it was previously `"draft"`, leading to a failing test.  I'm uncertain exactly what the critical behavior change on the archive side was.

CC: @yarikoptic

jwodder:
Closes #968.


yarikoptic:
thanks!

PR #982, opened 2022-04-25
performance: do not duplicate schema loading

TheChymera:
New speedscope profile: https://ppb.chymera.eu/51d5f7.svg


yarikoptic:
still green and no drop in coverage so must have had no undesired side-effects. Assuming this API is not used in other packages (yet), let's just proceed

PR #984, opened 2022-04-25
Increase Zarr depth limit to 7 directories

### LINKED ISSUE #983: ngff upload running into validation errors
i don't think there is a directory more the 5 levels into the ngff file. also, if there is, it would be good to know an example of one in the error.

```
$ dandi validate .
2022-04-25 09:34:22,018 [   ERROR] micr/sub-MITU01_ses-20210521h17m17s06_sample-178_stain-NPY_run-1_chunk-1_SPIM.ngff: 1 error(s)
2022-04-25 09:34:22,019 [   ERROR]   Zarr directory tree more than 5 directories deep
```

this error repeats for all ngff files in the directory

jwodder:
Closes #983.

### DIFF
--- a/dandi/consts.py
+++ b/dandi/consts.py
@@ -75,7 +75,7 @@ PUBLISHED_VERSION_REGEX = r"[0-9]+\.[0-9]+\.[0-9]+"
 
 #: Regular expression for a valid Dandiset version identifier.  This regex is
 #: not anchored.
-VERSION_REGEX = fr"(?:{PUBLISHED_VERSION_REGEX}|draft)"
+VERSION_REGEX = rf"(?:{PUBLISHED_VERSION_REGEX}|draft)"
 
 
 class EmbargoStatus(Enum):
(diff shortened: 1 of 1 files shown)


dandi/consts.py
```diff
@@ -75,7 +75,7 @@
 
 #: Regular expression for a valid Dandiset version identifier.  This regex is
 #: not anchored.
-VERSION_REGEX = fr"(?:{PUBLISHED_VERSION_REGEX}|draft)"
+VERSION_REGEX = rf"(?:{PUBLISHED_VERSION_REGEX}|draft)"
```
yarikoptic, line 78: curious: is there a difference and/or standardization of the order of `r` and `f`?

jwodder: There's no practical difference, but the latest version of black prefers `rf` for some reason.

yarikoptic: gotcha -- thank you!

PR #987, opened 2022-04-26
Upgrade dandischema to 0.7.x series

yarikoptic:
per @satra request

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,7 @@ install_requires =
     appdirs
     click
     click-didyoumean
-    dandischema ~= 0.6.0
+    dandischema ~= 0.7.0
     etelemetry >= 0.2.2
     fasteners
     fscacher


satra:
the errors seem to be related to the url field. since the schema was changed to support a couple of external environment variables, perhaps some additional change is needed in the CLI tests.

jwodder:
cf. https://github.com/dandi/dandi-schema/pull/132

yarikoptic:
@jwodder if you see what to do please proceed

jwodder:
Even after updating the metadata test cases, the publication tests are still failing with "Metadata version 0.6.3 is not allowed. Allowed are: 0.6.2, 0.4.4, 0.5.1, 0.5.2, 0.6.0, 0.6.1.", likely due to the fact that dandi-archive is still using dandischema 0.6.0.

jwodder:
@yarikoptic The tests are passing now.

yarikoptic:
Cool, but I guess we must not  merge/release until we release dandi archive first

yarikoptic:
ok, archive is deployed. This upgrade is not strictly needed AFAIK since dandischema would just upgrade from 0.6.2 to 0.6.3 on server. But should not hurt. so let's proceed

Issue #990, opened 2022-04-28
Should error out with some meaningful error


yarikoptic:
most likely because it nowhere finds a dataset_description.json (since there is none). Should report an error stating that instead of just letting it crash with an Error

```shell
lena:/tmp/test
$> dandi validate-bids sub-MITU01_ses-MR_SSSPDmap.nii.gz
2022-04-28 18:05:16,509 [    INFO] Logs saved in /home/yoh/.cache/dandi-cli/log/20220428220515Z-1782325.log
Error: '<' not supported between instances of 'NoneType' and 'str'
```

PR #994, opened 2022-05-02
Setting version to min version if not available

TheChymera:
Closes: https://github.com/dandi/dandi-cli/issues/990


yarikoptic:
Looks lgtm albeit not accompanied with a unit test, so mileage can vary. Let's proceed, I will add patch level label

PR #1001, opened 2022-05-05
ENH: delete - allow to specify --force in CLI

### LINKED ISSUE #993: non-interactive delete
a command like this:

> find . -maxdepth 1 -name "ses-*" -type d | cut -c 3- | sort | xargs -I sessionname bash -c "dandi delete https://dandiarchive.org/dandiset/000108/draft/files?location=sub-MITU01%2Fsessionname%2Fmicroscopy%2F"

does not work since delete expects a human acknowledgment to the delete process. it would be nice to have a `--yes` option so that it can be used like above.

yarikoptic:
internal implementation already had that option, so we just needed
to pass it through

Closes #993

### DIFF
--- a/dandi/cli/cmd_delete.py
+++ b/dandi/cli/cmd_delete.py
@@ -14,7 +14,7 @@ from .base import devel_debug_option, instance_option, map_to_click_exceptions
 @instance_option()
 @devel_debug_option()
 @map_to_click_exceptions
-def delete(paths, skip_missing, dandi_instance, devel_debug=False, force=False):
+def delete(paths, skip_missing, dandi_instance, force, devel_debug=False):
     """Delete dandisets and assets from the server.
 
     PATH could be a local path or a URL to an asset, directory, or an entire
--- a/dandi/cli/cmd_delete.py
+++ b/dandi/cli/cmd_delete.py
@@ -5,11 +5,16 @@ from .base import devel_debug_option, instance_option, map_to_click_exceptions
 
 @click.command()
 @click.option("--skip-missing", is_flag=True)
+@click.option(
+    "--force",
+    is_flag=True,
+    help="Force deletion without requesting interactive confirmation",
+)
 @click.argument("paths", nargs=-1, type=click.Path(exists=False, dir_okay=True))
 @instance_option()
 @devel_debug_option()
 @map_to_click_exceptions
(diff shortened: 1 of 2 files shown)


dandi/cli/cmd_delete.py
```diff
@@ -5,11 +5,16 @@
 
 @click.command()
 @click.option("--skip-missing", is_flag=True)
+@click.option(
+    "--force",
```
jwodder, line 9: This option should also be listed in `docs/source/cmdline/delete.rst`.

dandi/cli/cmd_delete.py
```diff
@@ -5,11 +5,16 @@
...
+    is_flag=True,
+    help="Force deletion without requesting interactive confirmation",
+)
 @click.argument("paths", nargs=-1, type=click.Path(exists=False, dir_okay=True))
 @instance_option()
 @devel_debug_option()
 @map_to_click_exceptions
-def delete(paths, skip_missing, dandi_instance, devel_debug=False):
+def delete(paths, skip_missing, dandi_instance, devel_debug=False, force=False):
```
jwodder, line 17:
```suggestion
def delete(paths, skip_missing, dandi_instance, force, devel_debug=False):
```

jwodder (changes requested):

dandi/cli/cmd_delete.py
jwodder: Because the `--devel-debug` option is only defined when `DANDI_DEVEL` is set, `devel_debug` is not always passed to the command function when it's called, so it needs a default value.  `force`, on the other hand, is always passed to the command function, so giving it a default value is pointless and potentially misleading.

yarikoptic: Got it - makes sense. Let's do it, thanks!

dandi/cli/cmd_delete.py
yarikoptic: Previously in https://github.com/dandi/dandi-cli/pull/831 sphinx-cli ... formalized that in a new  https://github.com/dandi/dandi-cli/issues/1002

jwodder (changes requested):

jwodder (approved):

PR #1004, opened 2022-05-09
TEMP: just to see how CI is doing

yarikoptic:


yarikoptic:
not quite sure about codecov complaining about that line since I think it should be deterministic (but I guess there is still some room for different order of execution leading to different code paths with those threads) in https://app.codecov.io/gh/dandi/dandi-cli/compare/1004/changes#D1L32  but otherwise all good.

PR #1009, opened 2022-05-11
organize - sanitize ; as well within the entity value

yarikoptic:
User reported filenames with ; in them.  Although I do not think
it caused any trouble, seeing such character in the filename is
quite "surprising" to say the least, and since ; has meaning in
POSIX as separator between different commands I think best to be avoided.
Also it is not part of alphanumerics of BIDS.

But with that I wonder if we should add other punctuation to be
subject for sanitization -- e.g.  ,[](){}?  WDYT @dandi/dandi-cli ???

### DIFF
--- a/dandi/organize.py
+++ b/dandi/organize.py
@@ -400,7 +400,7 @@ def _sanitize_value(value, field):
     Of particular importance is _ which we use, as in BIDS, to separate
     _key-value entries
     """
-    value = re.sub("[_*:%@]", "-", value)
+    value = re.sub("[_*:%@;]", "-", value)
     if field != "extension":
         value = value.replace(".", "-")
     return value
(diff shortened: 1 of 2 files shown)


jwodder:
@yarikoptic Regarding possible other characters to sanitize, I believe the characters that are forbidden in Windows filenames are `\/<>:|"?*%`.

dandi/organize.py
```diff
@@ -400,7 +400,7 @@ def _sanitize_value(value, field):
     Of particular importance is _ which we use, as in BIDS, to separate
     _key-value entries
     """
-    value = re.sub("[_*:%@]", "-", value)
+    value = re.sub("[_*:%@;]", "-", value)
```
yarikoptic, line 403:
```suggestion
    value = re.sub("[_*\/<>:|\"?*%@;]", "-", value)
```
expanding with suggested by @jwodder as not allowed on windows

PR #1010, opened 2022-05-11
Removing trace from previous commit

TheChymera:
@yarikoptic the issue @jwodder picked up on. I'm pretty sure this was a trace from the previous “autodetection” approach. Best see if purging it has unintended test consequences first. Thanks :)


TheChymera:
should be good, coverage difference is artifactual.

yarikoptic:
Thanks! Let's proceed!

PR #1011, opened 2022-05-12
Built-in BIDS support for `dandi upload`

TheChymera:
@yarikoptic @satra still as draft, some things might still be weird, and need to come up with tests.

Eventually it will look like this (contains both normal and validation ignore usage), and if the number of invalid files is not 18k, the output also does not need to contain 18k lines :)
https://ppb.chymera.eu/1c32cb

### DIFF
--- a/dandi/bids_utils.py
+++ b/dandi/bids_utils.py
@@ -1,15 +1,13 @@
-from typing import Optional
-
-import click
-
 from .utils import pluralize
 
 
 def evaluate_validation(
     validation_result: dict,
-    allow_errors: Optional[bool] = False,
-    cli_output: Optional[bool] = False,
+    allow_errors: bool = False,
+    cli_output: bool = False,
 ) -> bool:
+    if allow_errors:
+        return True
     missing_files = [
         pattern["regex"]
         for pattern in validation_result["schema_tracking"]
@@ -29,6 +27,8 @@ def evaluate_validation(
         )
         error_list.append(error_substring)
     if error_list:
+        import click
+
         if cli_output:
             error_string = " and ".join(error_list)
             error_string = f"Summary: {error_string}."
@@ -37,6 +37,5 @@ def evaluate_validation(
                 bold=True,
                 fg="red",
             )
-        if not allow_errors:
-            return False
+        return False
     return True
--- a/dandi/bids_utils.py
+++ b/dandi/bids_utils.py
@@ -3,11 +3,46 @@ from .utils import pluralize
 
 def evaluate_validation(
     validation_result: dict,
-    allow_errors: bool = False,
-    cli_output: bool = False,
+    allow_invalid_filenames: bool = False,
+    allow_missing_files: bool = False,
 ) -> bool:
-    if allow_errors:
+    """Determine whether a dataset validation result marks it as valid.
+
+    Parameters
+    ----------
+    validation_result: dict
+        Dictionary as returned by `dandi.bids_validator_xs.validate_bids()`.
+    allow_missing_files: bool, optional
+        Whether to consider the dataset invalid if any mandatory files are not present.
+    allow_invalid_filenames: bool, optional
+        Whether to consider the dataset invalid if any filenames inside are invalid.
+
+    Returns
+    -------
+    bool: whether the dataset validation result marks it as valid.
+
+    """
+
+    if allow_invalid_filenames and allow_missing_files:
         return True
+    missing_files = [
+        i["regex"] for i in validation_result["schema_tracking"] if i["mandatory"]
+    ]
+    invalid_filenames = validation_result["path_tracking"]
+
+    if missing_files and not allow_missing_files:
+        return False
+    if invalid_filenames and not allow_invalid_filenames:
+        return False
+    else:
+        return True
+
+
+def report_errors(
+    validation_result: dict,
+):
+    import click
+
     missing_files = [
         pattern["regex"]
         for pattern in validation_result["schema_tracking"]
@@ -27,15 +62,16 @@ def evaluate_validation(
         )
         error_list.append(error_substring)
     if error_list:
-        import click
-
-        if cli_output:
-            error_string = " and ".join(error_list)
-            error_string = f"Summary: {error_string}."
-            click.secho(
-                error_string,
-                bold=True,
-                fg="red",
-            )
-        return False
-    return True
+        error_string = " and ".join(error_list)
+        error_string = f"Summary: {error_string}."
+        click.secho(
+            error_string,
+            bold=True,
+            fg="red",
+        )
+    else:
+        click.secho(
+            "All filenames are BIDS-valid and no mandatory files are missing.",
+            bold=True,
+            fg="green",
+        )
--- a/dandi/bids_utils.py
+++ b/dandi/bids_utils.py
@@ -1,0 +1,42 @@
+from typing import Optional
+
+import click
+
+from .utils import pluralize
+
+
+def evaluate_validation(
+    validation_result: dict,
+    allow_errors: Optional[bool] = False,
+    cli_output: Optional[bool] = False,
+) -> bool:
+    missing_files = [
+        pattern["regex"]
+        for pattern in validation_result["schema_tracking"]
+        if pattern["mandatory"]
+    ]
+    error_list = []
+    if missing_files:
+        error_substring = (
+            f"{pluralize(len(missing_files), 'filename pattern')} required "
+            "by BIDS could not be found"
+        )
+        error_list.append(error_substring)
+    if validation_result["path_tracking"]:
+        error_substring = (
+            f"{pluralize(len(validation_result['path_tracking']), 'filename')} "
+            "did not match any pattern known to BIDS"
+        )
+        error_list.append(error_substring)
+    if error_list:
+        if cli_output:
+            error_string = " and ".join(error_list)
+            error_string = f"Summary: {error_string}."
+            click.secho(
+                error_string,
+                bold=True,
+                fg="red",
+            )
+        if not allow_errors:
+            return False
+    return True
--- a/dandi/bids_utils.py
+++ b/dandi/bids_utils.py
@@ -40,7 +40,7 @@ def evaluate_validation(
 
 def report_errors(
     validation_result: dict,
-):
+) -> None:
     import click
 
     missing_files = [
--- a/dandi/bids_utils.py
+++ b/dandi/bids_utils.py
@@ -1,7 +1,7 @@
 from .utils import pluralize
 
 
-def evaluate_validation(
+def is_valid(
     validation_result: dict,
     allow_invalid_filenames: bool = False,
     allow_missing_files: bool = False,
--- a/dandi/bids_validator_xs.py
+++ b/dandi/bids_validator_xs.py
@@ -611,17 +611,28 @@ def select_schema_dir(
                 )
             else:
                 with open(dataset_description) as f:
-                    dataset_info = json.load(f)
                     try:
-                        schema_version = dataset_info["BIDSVersion"]
-                    except KeyError:
-                        lgr.warning(
-                            "BIDSVersion is not specified in "
-                            "`dataset_description.json`. "
-                            "Falling back to %s.",
+                        dataset_info = json.load(f)
+                    except json.decoder.JSONDecodeError:
+                        lgr.error(
+                            "The `%s` file could not be loaded. "
+                            "Please check whether the file is valid JSON. "
+                            "Falling back to the %s BIDS version.",
+                            dataset_description,
                             schema_min_version,
                         )
                         schema_version = schema_min_version
+                    else:
+                        try:
+                            schema_version = dataset_info["BIDSVersion"]
+                        except KeyError:
+                            lgr.warning(
+                                "BIDSVersion is not specified in "
+                                "`dataset_description.json`. "
+                                "Falling back to %s.",
+                                schema_min_version,
+                            )
+                            schema_version = schema_min_version
         if not schema_version:
             lgr.warning(
                 "No BIDSVersion could be found for the dataset. Falling back to %s.",
--- a/dandi/tests/test_upload.py
+++ b/dandi/tests/test_upload.py
@@ -181,10 +181,10 @@ def test_upload_sync_folder(
         text_dandiset.dandiset.get_asset_by_path("subdir2/banana.txt")
 
 
-def test_upload_sync_bids(mocker, bids_dandiset):
-    confirm_mock = mocker.patch("click.confirm", return_value=True)
-    bids_dandiset.upload()
-    confirm_mock.assert_called_with("Delete 1 asset on server?")
+def test_upload_bids(mocker: MockerFixture, bids_dandiset: SampleDandiset) -> None:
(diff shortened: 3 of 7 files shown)


dandi/upload.py
```diff
@@ -383,3 +402,46 @@ def check_replace_asset(
...
+            bids_datasets_to_validate = set()
+            for p in bids_lookup_paths:
+                for bd in bids_datasets:
+                    if p.is_relative_to(bd):
+                        bids_datasets_to_validate.add(bd)
+                        break
+        else:
+            bids_datasets_to_validate = bids_datasets
+        bids_datasets_to_validate = sorted(bids_datasets_to_validate)
+        for bd in sorted(bids_datasets_to_validate):
```
jwodder, line 441: You're sorting twice.

dandi/upload.py
```diff
@@ -383,3 +402,46 @@ def check_replace_asset(
...
+    if bids_datasets:
+        lgr.debug(
+            "Detected %d BIDS datasets at following paths: %s",
+            len(bids_datasets),
+            ", ".join([str(i) for i in bids_datasets]),
+        )
+
+    if validation != "skip":
+        if bids_lookup_paths:
+            bids_datasets_to_validate = set()
```
jwodder, line 432: Change this to a list so that the type is the same regardless of `bids_lookup_paths`.

dandi/validate.py
```diff
@@ -33,9 +39,37 @@ def validate_bids(
...
+    if error_list:
+        error_string = " and ".join(error_list)
+        error_string = f"Summary: {error_string}."
+        click.secho(
+            error_string,
+            bold=True,
+            fg="red",
+        )
+        if not allow_errors:
+            raise SystemExit(1)
```
jwodder, line 72: Library functions should not raise `SystemExit`.  Instead, return a value that is used by `cmd_validate.py` to decide whether to raise `SystemExit` (or, even better, use `ctx.exit(1)` in `cmd_validate.py`).

dandi/upload.py
```diff
@@ -61,6 +66,20 @@ def upload(
...
 
+    # pre-validate BIDS datasets before going for individual
+    # files etc
+    bids_datasets = _bids_discover_and_validate(dandiset_.path, paths, validation)
+
+    if bids_datasets:
+        bids_datasets = [str(i) for i in bids_datasets]
+        if not allow_any_path:
+            lgr.info(
+                "Setting allow_any_path to True since we detected %s under: %s",
```
jwodder, line 77: Under what?

jwodder (changes requested):

dandi/upload.py
```diff
@@ -383,3 +398,54 @@ def check_replace_asset(
...
+                    try:
+                        p.relative_to(bd)
+                    except ValueError:
+                        pass
+                    else:
+                        bids_datasets_to_validate.append(bd)
+                        break
+        else:
+            bids_datasets_to_validate = bids_datasets
+        bids_datasets_to_validate = sorted(bids_datasets_to_validate)
```
jwodder, line 446:
```suggestion
        bids_datasets_to_validate.sort()
```

dandi/upload.py
```diff
@@ -383,3 +398,54 @@ def check_replace_asset(
 
 def skip_file(msg: Any) -> Dict[str, str]:
     return {"status": "skipped", "message": str(msg)}
+
+
+def _bids_discover_and_validate(
+    dandiset_path: str,
+    paths: Optional[List[Union[str, Path]]] = None,
+    validation: str = "require",
+):
```
jwodder, line 407:
```suggestion
) -> List[str]:
```

dandi/upload.py
```diff
@@ -61,6 +61,21 @@ def upload(
             " paths.  Use 'dandi download' or 'organize' first."
         )
 
+    # pre-validate BIDS datasets before going for individual
+    # files etc
+    bids_datasets = _bids_discover_and_validate(dandiset_.path, paths, validation)
+
+    if bids_datasets:
+        bids_datasets = [str(i) for i in bids_datasets]
```
jwodder, line 69: Aren't the elements of `bids_datasets` already strings?

dandi/upload.py
```diff
@@ -383,3 +398,54 @@ def check_replace_asset(
...
+                    except ValueError:
+                        pass
+                    else:
+                        bids_datasets_to_validate.append(bd)
+                        break
+        else:
+            bids_datasets_to_validate = bids_datasets
+        bids_datasets_to_validate = sorted(bids_datasets_to_validate)
+        for bd in bids_datasets_to_validate:
+            _ = validate_bids(bd, allow_errors=validation == "ignore")
```
jwodder, line 448:
```suggestion
            validate_bids(bd, allow_errors=validation == "ignore")
```

dandi/validate.py
jwodder: @TheChymera You did not address this comment.

dandi/upload.py
```diff
@@ -383,3 +398,54 @@ def check_replace_asset(
...
+    """
+    from .utils import find_files
+    from .validate import validate_bids
+
+    if paths:
+        bids_lookup_paths = set(paths)
+    else:
+        bids_lookup_paths = None
+    bids_descriptions = map(
+        Path, find_files(r"dataset_description\.json$", dandiset_path)
```
jwodder, line 422:
```suggestion
        Path, find_files(r"(^|[/\x5C])dataset_description\.json$", dandiset_path)
```

jwodder (changes requested):

dandi/upload.py
TheChymera: No, they are `PosixPath` objects.

TheChymera:
@jwodder putting it on draft for now to not waste your time. Need to test some things re typing first.

TheChymera:
@jwodder ok, figured out the typing issues :) Everything else you mentioned should already be in.

dandi/upload.py
```diff
@@ -61,6 +61,21 @@ def upload(
...
 
+    # pre-validate BIDS datasets before going for individual
+    # files etc
+    bids_datasets = _bids_discover_and_validate(dandiset_.path, paths, validation)
+
+    if bids_datasets:
+        _bids_datasets = [str(i) for i in bids_datasets]
+        if not allow_any_path:
+            lgr.info(
+                "Setting allow_any_path to True since we detected %s under the "
```
jwodder, line 72:
```suggestion
                "Enabling --allow-any-path since we detected %s under the "
```

(Because this function is most often invoked via the command line.)

dandi/upload.py
```diff
@@ -383,3 +398,59 @@ def check_replace_asset(
...
+
+    References:
+    - unification of validation records: https://github.com/dandi/dandi-cli/issues/943
+    - validation "design doc": https://github.com/dandi/dandi-cli/pull/663
+    """
+    from .utils import find_files
+    from .validate import validate_bids
+
+    if paths:
+        bids_lookup_paths = set(paths)
```
jwodder, line 418: What's the point of converting this to a set?

dandi/validate.py
```diff
@@ -1,15 +1,20 @@
...
 
 
 def validate_bids(
-    *paths: str,
+    *paths: Union[str, Path],
     schema_version: Optional[str] = None,
     devel_debug: bool = False,
     report: Optional[str] = None,
+    allow_errors: Optional[bool] = False,
 ) -> Any:
```
jwodder, line 18:
```suggestion
) -> bool:
```

jwodder (changes requested):

dandi/upload.py
```diff
@@ -383,3 +398,59 @@ def check_replace_asset(
...
+
+    if validation != "skip":
+        if bids_lookup_paths:
+            bids_datasets_to_validate = list()
+            for p in bids_lookup_paths:
+                if isinstance(p, str):
+                    p = Path(p)
+                for bd in bids_datasets:
+                    try:
+                        p.relative_to(bd)
```
jwodder, line 440: This adds `bd` to `bids_datasets_to_validate` if `p` is a child path of `bd`.  Don't we also want to add if `bd` is a child path of `p`?

dandi/validate.py
```diff
@@ -33,9 +45,38 @@ def validate_bids(
...
+    if validation_result["path_tracking"]:
+        error_substring = (
+            f"{pluralize(len(validation_result['path_tracking']), 'filename')} "
+            "did not match any pattern known to BIDS"
+        )
+        error_list.append(error_substring)
+    if error_list:
+        error_string = " and ".join(error_list)
+        error_string = f"Summary: {error_string}."
+        click.secho(
```
jwodder, line 72: I don't like this.  Errors in a library function should be reported via a return value or exception for printing by the caller.

@yarikoptic Thoughts on how BIDS dataset validation errors should be displayed when uploading?

jwodder (changes requested):

dandi/upload.py
TheChymera: you're right.

dandi/upload.py
```diff
@@ -383,3 +398,67 @@ def check_replace_asset(
...
+            "Detected %s under following paths: %s",
+            pluralize(len(bids_datasets), "BIDS dataset"),
+            ", ".join(str(i) for i in bids_datasets),
+        )
+
+    if validation != "skip":
+        if bids_lookup_paths:
+            bids_datasets_to_validate = list()
+            for p in bids_lookup_paths:
+                print(f"p:{p}")
```
jwodder, line 436: Are these `print`s just for develomental debugging?  If they're really needed, replace them with `lgr.debug()` calls with more informative messages.  Otherwise, get rid of them.

dandi/validate.py
TheChymera: This I am unsure about, as it stands the main “library function” for validation (`bids_validator_xs.validate_bids()`) just returns a nested dictionary which we could build our reporting on. If we don't want to have this `click` call happen in `validate.py` we would need to reproduce this in both `upload.py` and `cli/cmd_validate.py` — because both modules use validation and (I woud assume?) need identical reporting to not confuse people when they get one sort of output for validation, and another sort when they try to upload.

dandi/upload.py
TheChymera: there were these and some more traces of my debugging, purged them all and force-pushed.

dandi/validate.py
yarikoptic: Agree with @jwodder that there should be no `print` or `click.anything` in the library, non-CLI code. Details should be logged (.error or .warning), and display/reaction during upload should be exactly the same  as for nwb files. Check details, but IIRC we would just error out that file upload unless --validation is `skip` or `ignore`.  Overall: just mimic behavior on what is happening for NWB files ATM.  Let me know if I should analyze and provide more specific "desires".

TheChymera:
@jwodder is this better like this design-wise? If so I can just nice-ify it a bit so all the strings don't get constructed if we don't need the output.

dandi/bids_utils.py
```diff
@@ -0,0 +1,42 @@
+from typing import Optional
+
+import click
+
+from .utils import pluralize
+
+
+def evaluate_validation(
+    validation_result: dict,
+    allow_errors: Optional[bool] = False,
```
jwodder, line 10: `Optional[...]` means "can be `None`", not "is an optional argument."  It does not seem to apply here.

jwodder (changes requested):

jwodder:
@TheChymera Yes, this is better.

dandi/upload.py
```diff
@@ -383,3 +399,68 @@ def check_replace_asset(
...
+            pluralize(len(bids_datasets), "BIDS dataset"),
+            ", ".join(str(i) for i in bids_datasets),
+        )
+
+    if validation != "skip":
+        if bids_lookup_paths:
+            bids_datasets_to_validate = list()
+            for p in bids_lookup_paths:
+                if isinstance(p, str):
+                    p = Path(p)
```
yarikoptic, line 438: here and below lots of code/logic not covered by tests ... need more tests ??

dandi/validate.py
```diff
@@ -26,16 +27,22 @@ def validate_bids(
         If string, the string will be used as the output path.
         If the variable evaluates as False, no log will be written.
 
+    Returns
+    -------
+    bool
```
yarikoptic, line 32: diff above says it returns `-> dict`

needs fixing

yarikoptic (changes requested):

dandi/bids_validator_xs.py
```diff
@@ -611,17 +611,28 @@ def select_schema_dir(
...
-                            "BIDSVersion is not specified in "
-                            "`dataset_description.json`. "
-                            "Falling back to %s.",
+                        dataset_info = json.load(f)
+                    except json.decoder.JSONDecodeError:
+                        lgr.error(
+                            "The `%s` file could not be loaded. "
+                            "Please check whether the file is valid JSON. "
+                            "Falling back to the %s BIDS version.",
+                            dataset_description,
```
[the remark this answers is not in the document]
TheChymera: There are fixes for this in the upstream code. I was hoping we would be using that ASAP, but will pull in the mean time.

dandi/bids_utils.py
```diff
@@ -0,0 +1,41 @@
+from .utils import pluralize
+
+
+def evaluate_validation(
+    validation_result: dict,
+    allow_errors: bool = False,
+    cli_output: bool = False,
+) -> bool:
+    if allow_errors:
+        return True
```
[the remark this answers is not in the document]
TheChymera: working on test data

dandi/bids_utils.py
```diff
@@ -0,0 +1,41 @@
+from .utils import pluralize
+
+
+def evaluate_validation(
+    validation_result: dict,
+    allow_errors: bool = False,
+    cli_output: bool = False,
+) -> bool:
```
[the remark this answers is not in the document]
TheChymera: Think this was resolved after our conversation.

dandi/tests/test_upload.py
```diff
@@ -181,6 +181,12 @@ def test_upload_sync_folder(
         text_dandiset.dandiset.get_asset_by_path("subdir2/banana.txt")
 
 
+def test_upload_bids(mocker: MockerFixture, bids_dandiset: SampleDandiset) -> None:
+    iter_upload_spy = mocker.spy(LocalFileAsset, "iter_upload")
+    bids_dandiset.upload(existing="forced")
+    iter_upload_spy.assert_called()
```
yarikoptic, line 187: @jwodder what is the best way to extend this unittest to also query the server after upload to ensure that the specific assets were uploaded?  I have tried to add `local_dandi_api` fixture and use it (while in pdb) but failed so far.

TheChymera:
@yarikoptic @jwodder should be done, unless you notice some additional issues. Not sure what's going on with the doctests, but probably unrelated.

yarikoptic:
I will check it out tomorrow morning. Thanks!

jwodder (approved):

yarikoptic:
I have reservations since testing is quite "limited" as for number of use cases code tries to cover with the discovery of bids datasets... also tried on some dummy ds210 from bids-examples, it still required explicit paths and --allow-any-path... will file separate issues for that. For now -- let's just proceed.

PR #1021, opened 2022-05-27
Fetch paginated results concurrently

### LINKED ISSUE #1019: multithread get_assets (or may be other endpoints) as well pagination
eyeballing the output of snippet in  https://github.com/dandi/dandi-cli/issues/1018#issuecomment-1137227203 shows wait times after each page.  That could be avoided.  Pagination output returns `count` in the initial response. our pagination is simple as in specify the page number (no "next" token).  So it should be trivial to 
- estimate how many pages we would need `ceil(count/page_size)`
- send multiple requests in parallel e.g. for the next 5 pages
- assemble them back in original order

pagination is similar for other end points (e.g. `/dandisets/{versions__dandiset__pk}/versions/{versions__version}/assets/paths/`) and I think overall generic. So I think we can implement it generally for any pagination requests to submit multiple pages requests in parallel and speed up interactions with the archive this way.

jwodder:
Closes #1019.

See https://github.com/dandi/dandi-cli/issues/1019#issuecomment-1139666971 for concerns about the runtime of the test.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -113,6 +113,11 @@ class RESTFullAPIClient:
         if headers is not None:
             session.headers.update(headers)
         self.session = session
+        #: Default number of items to request per page when paginating (`None`
+        #: means to use the server's default)
+        self.page_size: Optional[int] = None
+        #: How many pages to fetch at once when parallelizing pagination
+        self.page_lookahead: int = 5
 
     def __enter__(self: T) -> T:
         return self
@@ -305,7 +310,6 @@ class RESTFullAPIClient:
         path: str,
         page_size: Optional[int] = None,
         params: Optional[dict] = None,
-        lookahead: int = 5,
     ) -> Iterator:
         """
         Paginate through the resources at the given path: GET the path, yield
@@ -314,10 +318,16 @@ class RESTFullAPIClient:
 
         If the first ``"next"`` key is the same as the initially-requested URL
         but with the ``page`` query parameter set to ``2``, then the remaining
-        pages are fetched concurrently in separate threads, ``lookahead``
+        pages are fetched concurrently in separate threads, `page_lookahead`
         (default 5) at a time.  This behavior requires the initial response to
         contain a ``"count"`` key giving the number of items across all pages.
+
+        :param Optional[int] page_size:
+            If non-`None`, overrides the client's `page_size` attribute for
+            this sequence of pages
         """
+        if page_size is None:
+            page_size = self.page_size
         if page_size is not None:
             if params is None:
                 params = {}
@@ -354,7 +364,7 @@ class RESTFullAPIClient:
         pages = (r["count"] + page_size - 1) // page_size
         pageno_iter = iter(range(2, pages + 1))
 
-        with ThreadPoolExecutor(max_workers=lookahead) as pool:
+        with ThreadPoolExecutor(max_workers=self.page_lookahead) as pool:
             futures: deque[Future[list]] = deque()
 
             def get_page(pageno: int) -> list:
@@ -371,7 +381,7 @@ class RESTFullAPIClient:
                     return
                 futures.append(pool.submit(get_page, i))
 
-            for _ in range(lookahead):
+            for _ in range(self.page_lookahead):
                 submit_job()
             while futures:
                 res = futures.popleft().result()
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -381,12 +381,16 @@ class RESTFullAPIClient:
                     return
                 futures.append(pool.submit(get_page, i))
 
-            for _ in range(self.page_lookahead):
-                submit_job()
-            while futures:
-                res = futures.popleft().result()
-                submit_job()
-                yield from res
+            try:
+                for _ in range(self.page_lookahead):
+                    submit_job()
+                while futures:
+                    res = futures.popleft().result()
+                    submit_job()
+                    yield from res
+            finally:
+                for f in futures:
+                    f.cancel()
 
 
 class DandiAPIClient(RESTFullAPIClient):
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -1,5 +1,8 @@
+from __future__ import annotations
+
 from abc import ABC, abstractmethod
 from collections import deque
+from concurrent.futures import Future, ThreadPoolExecutor
 from dataclasses import dataclass, field, replace
 from datetime import datetime
 from enum import Enum
@@ -54,6 +57,7 @@ from .utils import (
     chunked,
     ensure_datetime,
     is_interactive,
+    is_page2_url,
 )
 
 lgr = get_logger()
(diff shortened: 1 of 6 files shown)


dandi/dandiapi.py
```diff
@@ -301,25 +305,78 @@ def paginate(
...
+                except StopIteration:
+                    return
+                futures.append(pool.submit(get_page, i))
+
+            for _ in range(lookahead):
+                submit_job()
+            while futures:
+                res = futures.popleft().result()
+                submit_job()
+                yield from res
```
yarikoptic, line 379: why not just prepare futures for all pages ahead of time? i.e. smth like
```suggestion
            for i in range(2, pages + 1):        
                futures.append(pool.submit(get_page, i))

            while futures:
                yield futures.popleft().result()
```

yarikoptic:
FTR: this version (as of 0.39.6-15-g9ddba4c5) takes 11 seconds to list 000108 whenever master (no parallelization) takes 37 sec on my laptop from home. Cool! Thank you @jwodder

dandi/dandiapi.py
jwodder: Because, if the user stops iterating partway through the results, your way would lead to every page still being fetched, whereas the current code would only fetch five more pages.

yarikoptic: hm, why would it? (self-educational question) I would have expected that leaving ThreadPoolExecutor disables execution of anything, so no other one would be operating. But if needed really -- should just cancel all futures I guess. I did similar in https://github.com/datalad/datalad/blob/master/datalad/support/parallel.py#L226 and that should then not leave any running... so I guess smth like

```python
            try:
                for i in range(2, pages + 1):        
                    futures.append(pool.submit(get_page, i))
                while futures:
                    yield futures.popleft().result()
            finally:
                while futures:
                    futures.popleft().cancel()
```
so even those five do not survive or in other words -- worth doing for your version as well... but I still wonder why/if that is needed... will try now

yarikoptic: FWIW -- seems similar pattern to e.g. (besides that there it is a "simpler" `.as_completed`) (also with cancels) as https://github.com/dandi/dandi-cli/blob/HEAD/dandi/files.py#L1218 ...

jwodder: The pool's `__exit__` method is only called when control reaches the end of the `with ThreadPoolExecutor:` block.  If the caller abandons the pagination partway through, execution of the `pagination()` generator will still be in the `while futures:` loop, and the jobs in the pool won't be stopped, and they'll continue to use up resources until garbage collection (or reference counting or whatever) does its thing. 

Similarly, with your suggested code, if the caller abandons the pagination partway through, the `while futures: yield futures.popleft().result()` will never finish, and so the `finally:` won't be executed.  The only way to ensure that the thread pool is shutdown when iteration is over is to call the paginator like:

```
from contextlib import closing

with closing(client.paginate(...)) as it:
    for obj in it:
         ...
```

and then every iterator that "yields from" the paginator (like `get_assets()`) would also have to be wrapped in `closing()` as well, along with any iterators that wrapped those, etc.

yarikoptic: <details>
<summary>confirmed with this script</summary> 

```python
from collections import deque
from concurrent.futures import Future, ThreadPoolExecutor

def get_page(page: int):
    print(f"getting page {page}")
    import time
    time.sleep(0.5)
    print(f"got page {page}")
    return f"returning page {page}"

lookahead = 5
pages = 100

with ThreadPoolExecutor(max_workers=lookahead) as pool:
    futures: deque[Future[list]] = deque()
    try:
        for i in range(2, pages + 1):
            futures.append(pool.submit(get_page, i))
        print(f"Created all {len(futures)}")
        while futures:
            print(f"Got {futures.popleft().result()}")
    finally:
        if True:
            print(f"CANCELING remaining {len(futures)}")
            while futures:
                futures.popleft().cancel()

print("DONE")


```
</details>
that canceling is indeed needed and that otherwise (`if False` in finally) it keeps plowing... that is odd since makes little sense to me when would I want such behavior while using the `ThreadPoolExecutor`

jwodder: OK, I just tried running this as a demonstration:

```python
import sys

def gen():
    try:
        for i in range(5):
            yield i
    finally:
        print(sys.exc_info())
        print("This will never be printed.")

for x in gen():
    print(x)
    if x == 3:
        break

"""
# Alternative:
it = gen()
print(next(it))
print(next(it))
print(next(it))
print("After")
"""
```

And Python ended up automatically closing the generator (specifically, raising `GeneratorExit` within it) after the `break` (and, for the "alternative" block, after "After" was printed).  I'm going to have to look into exactly under what circumstances that happens.

jwodder: @yarikoptic So then are you fine with the code as it is now?

yarikoptic: I am not yet catching exactly how it (investigation of "incompletely" explored generator) would relate besides showing that `finally` would always be executed (desired).

FWIW: in `finally` there could simply be `pool.shutdown(cancel_futures=True)` which would cancel the futures.

yarikoptic: > @yarikoptic So then are you fine with the code as it is now?

I would prefer a more consistent with other similar codes pattern of `try`/`finally: pool.shutdown(cancel_futures=True)`.

jwodder: Done (using `f.cancel()`, since `cancel_futures` was only added in Python 3.9).

dandi/dandiapi.py
```diff
@@ -301,25 +310,87 @@ def paginate(
...
+                    return
+                futures.append(pool.submit(get_page, i))
+
+            try:
+                for _ in range(self.page_lookahead):
+                    submit_job()
+                while futures:
+                    res = futures.popleft().result()
+                    submit_job()
+                    yield from res
```
yarikoptic, line 390: I guess I was not specific enough in https://github.com/dandi/dandi-cli/pull/1021#discussion_r883959010 . What I meant: If I got your investigation right in https://github.com/dandi/dandi-cli/pull/1021#discussion_r883936703 - we will reach `finally` in any case, even though may be slightly delayed waiting for `gc`  to kick in, so I think that simpler/analogous to other places 

```suggestion
            try:
                for i in range(2, pages + 1):        
                    futures.append(pool.submit(get_page, i))
                while futures:
                    yield futures.popleft().result()
```
would be just fine (would also eliminate that definition of `pageno_iter` above) or what am I still missing?

jwodder: Submitting everything to the pool at once could still result in far more pages being fetched than are needed, e.g., if the user iterates through the results slowly (so that there's enough time for dozens or hundreds of pages to be fetched in the background) and still ends up breaking out of the loop early on.

yarikoptic: Well, I would then use your argument of "slow iteration" that what if user iterates slowly for awhile (laptop/bandwidth/server is busy with something else) and then really fast: Your approach would cause the the slow down of fetching during "slow" phase, instead of fetching as fast as it can to provide results during "fast" phase.

Overall, if we had to deal with such effects, other code paths of similar kind should have got similar code refactoring but I truly do not think it is necessary.  The goal for this function is to fetch most efficiently, so let's stick to that and a uniform code pattern.

yarikoptic: Thank you @jwodder !

yarikoptic:
Thank you @jwodder -- let's proceed ! not sure when/if would get to merge other PRs so marking for release as well -- let's see it in action

PR #1024, opened 2022-05-31
Add `dandi move` command

### LINKED ISSUE #1006: dandi move command
Prior mentioning/need for such (`dandi mv`) command is at https://github.com/dandi/dandi-cli/issues/848#issuecomment-988055068 and it came up again in context of zarr archives. Since they are mutable ATM, asset ids are changing upon metadata change and not stored locally, there is no sense of identity of a zarr folder.  That forbids any efficient renaming of a zarr archive locally. 

With `dandi move`, similar one to `dandi delete` we have, we could rename locally and on the server really fast, thus making subsequent `dandi upload` if so needed update an existing zarr archive instead of uploading a new one and deleting an old one. 

Extra points to keep in mind
- IIRC `dandi organize` can operate in-place (within existing dandiset).  Then in such scenario "by default" it should use `dandi move` internally to move files. But it should be some kind of a switch which would do server side change as well or not.
- Originally I thought to file it as `dandi mv` but decided to name it `dandi move` instead -- https://github.com/dandi/dandi-cli/issues/1005

jwodder:
Closes #1006.

To do:
- [ ] Write tests to drive development
- [ ] Implement `move()` function
- [ ] Implement `dandi move` command
- [ ] Document `dandi move` in Sphinx docs

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -1576,6 +1576,18 @@ class RemoteAsset(BaseRemoteAsset):
         """
         ...
 
+    def rename(self, dest: str) -> None:
+        """
+        .. versionadded:: 0.41.0
+
+        Change the path of the asset on the server to the given value and
+        update the `RemoteAsset` in place.  If another asset already exists at
+        the given path, a `requests.HTTPError` is raised.
+        """
+        md = self.get_raw_metadata().copy()
+        md["path"] = dest
+        self.set_raw_metadata(md)
+
     def delete(self) -> None:
         """Delete the asset"""
         self.client.delete(self.api_path)
--- a/dandi/tests/test_dandiapi.py
+++ b/dandi/tests/test_dandiapi.py
@@ -13,6 +13,7 @@ import click
 from dandischema.models import UUID_PATTERN, DigestType, get_schema_version
 import pytest
 from pytest_mock import MockerFixture
+import requests
 import responses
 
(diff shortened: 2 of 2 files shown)


docs/source/cmdline/delete.rst
```diff
@@ -19,7 +19,7 @@ Options
 
     Force deletion without requesting interactive confirmation
 
-.. option:: -i, --instance <instance-name>
+.. option:: -i, --dandi-instance <instance-name>
```
yarikoptic, line 22: just to make sure -- you are not changing API here, and rather just fixing up documentation right? (i.e. that it was `dandi_instance` already)

jwodder: Correct.

docs/source/cmdline/move.rst
```diff
@@ -0,0 +1,91 @@
...
+
+If there is more than one source path, or if the destination path either names
+an existing directory or ends in a trailing forward slash (``/``), then the
+source assets are placed within the destination directory.  Otherwise, the
+single source path is renamed to the given destination path.
+
+Alternatively, if the ``--regex`` option is given, then there must be exactly
+two arguments on the command line: a `Python regular expression`_ and a
+replacement string, possibly containing regex backreferences.  :program:`dandi
+move`: will then apply the regular expression to the path of every asset in the
```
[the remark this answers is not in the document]
jwodder: Fixed.

PR #1026, opened 2022-06-01
Fix calculation of download paths for asset path prefix URLs

### LINKED ISSUE #1025: downloading individual file works but throws an error
I am using the following to use the dandi cli to download an individual file

```bash
dandi download https://dandiarchive.org/dandisets/000053/versions/"draft"/assets/?path="sub-npJ3/sub-npJ3_ses-20190504_behavior.nwb"
```

which works (and is very fast). The correct file is downloaded to my cwd, and is intact. However, the command throws an error:

```pytb
Producing value for row ('sub-npJ3_ses-20190504_behavior.nwb/.',) failed:
  File "/Users/bendichter/opt/miniconda3/lib/python3.9/site-packages/pyout/interface.py", line 279, in _print_async_exceptions
    future.result()
  File "/Users/bendichter/opt/miniconda3/lib/python3.9/concurrent/futures/_base.py", line 438, in result
    return self.__get_result()
  File "/Users/bendichter/opt/miniconda3/lib/python3.9/concurrent/futures/_base.py", line 390, in __get_result
    raise self._exception
  File "/Users/bendichter/opt/miniconda3/lib/python3.9/concurrent/futures/thread.py", line 52, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/bendichter/opt/miniconda3/lib/python3.9/site-packages/pyout/interface.py", line 482, in async_fn
    for i in gen:
  File "/Users/bendichter/opt/miniconda3/lib/python3.9/site-packages/dandi/download.py", line 711, in _download_file
    os.utime(path, (time.time(), ensure_datetime(mtime).timestamp()))
NotADirectoryError: [Errno 20] Not a directory
```

Am I using this command correctly?

jwodder:
Closes #1025.

### DIFF
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -247,10 +247,7 @@ def download_generator(
             path = op.normpath(path)
             if not isinstance(parsed_url, DandisetURL):
                 if isinstance(parsed_url, MultiAssetURL):
-                    folder_path = op.normpath(parsed_url.path)
-                    path = op.join(
-                        op.basename(folder_path), op.relpath(path, folder_path)
-                    )
+                    path = multiasset_target(parsed_url.path, path)
                 elif isinstance(parsed_url, SingleAssetURL):
                     path = op.basename(path)
                 else:
@@ -1035,3 +1032,15 @@ class ProgressCombiner:
             lgr.warning(
                 "Unexpected download status dict for %r received: %r", path, status
             )
+
+
+def multiasset_target(url_path: str, asset_path: str) -> str:
+    """
+    When downloading assets for a `MultiAssetURL` with `~MultiAssetURL.path`
+    equal to ``url_path``, calculate the path (relative to the output path) at
+    which to save the asset with path ``asset_path``.
+    """
+    prefix = op.dirname(url_path.rstrip("/"))
+    if prefix:
+        prefix += os.sep
+    return asset_path[len(prefix) :]
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -1043,4 +1043,5 @@ def multiasset_target(url_path: str, asset_path: str) -> str:
     prefix = op.dirname(url_path.rstrip("/"))
     if prefix:
         prefix += os.sep
+    assert(asset_path.startswith(prefix))
     return asset_path[len(prefix) :]
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -1042,6 +1042,6 @@ def multiasset_target(url_path: str, asset_path: str) -> str:
     """
     prefix = op.dirname(url_path.rstrip("/"))
     if prefix:
-        prefix += os.sep
-    assert(asset_path.startswith(prefix))
+        prefix += "/"
+    assert asset_path.startswith(prefix)
     return asset_path[len(prefix) :]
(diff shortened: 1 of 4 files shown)


dandi/tests/test_download.py
```diff
@@ -703,3 +703,20 @@ def test_progress_combiner(
...
     assert outputs == expected
+
+
+@pytest.mark.parametrize(
+    "url_path,asset_path,target",
+    [
+        ("", "foo/bar", "foo/bar"),
+        ("fo", "foo/bar", "foo/bar"),
+        ("foo", "foo/bar", "foo/bar"),
+        ("foo/", "foo/bar", "foo/bar"),
```
yarikoptic, line 714: note: so this one shows that it is a bit different from how I thought it in https://github.com/dandi/dandi-cli/issues/1025#issuecomment-1143964516 where it would have been
```suggestion
        ("foo/", "foo/bar", "bar"),
```
because of  trailing / in the `url_path`, but code rstrips it thus making it "consistent" between the two forms

dandi/download.py
```diff
@@ -1035,3 +1032,15 @@ def feed(self, path: str, status: dict) -> Iterator[dict]:
...
             )
+
+
+def multiasset_target(url_path: str, asset_path: str) -> str:
+    """
+    When downloading assets for a `MultiAssetURL` with `~MultiAssetURL.path`
+    equal to ``url_path``, calculate the path (relative to the output path) at
+    which to save the asset with path ``asset_path``.
+    """
+    prefix = op.dirname(url_path.rstrip("/"))
```
yarikoptic, line 1043: I think I might have `lstrip('/')`-ed both url_path and asset_path for consistency here since upstairs asset_path is lstripped but not sure if url_path is, which could then bring odd results below in `asset_path[len(prefix) :]`... and then have one test case which relies on ensuring that they are lstripped

dandi/download.py
```diff
@@ -1035,3 +1032,15 @@ def feed(self, path: str, status: dict) -> Iterator[dict]:
...
+    if prefix:
+        prefix += os.sep
+    return asset_path[len(prefix) :]
```
yarikoptic, line 1046:
```suggestion
    assert(asset_path.startswith(prefix))
    return asset_path[len(prefix) :]
```
correct? possibly relates to the lstripping above.

yarikoptic (commented):
It is a bit different from what I thought about in https://github.com/dandi/dandi-cli/issues/1025#issuecomment-1143964516 which would have introduced "rsync" like semantic for handling trailing `/` in the path, but that is Ok.  Might even be better that would be consistent like that.  But how would it behave for `path` component in the rival `.../files/?location={path}/` -- would it be the same (might be worth checking while at it)?

I would also like to add guards before `[prefix:]`slicing, left comments

dandi/download.py
jwodder: It's my understanding that Dandi Archive doesn't permit asset paths with leading slashes, and thus a `path` argument in a URL with a leading slash would match nothing.  Is that not the case?

jwodder: It turns out the archive does allow paths with leading slashes; filed https://github.com/dandi/dandi-archive/issues/1111.  In the meantime, slashes are now stripped from both ends of `url_path`.

yarikoptic:
ok, let's proceed - thank you @jwodder !

Issue #1037, opened 2022-06-30
dandi validate-bids crashes with IndexError: list index out of range


yarikoptic:
trying on datalad dandiset of 000108 (note -- is being updated)

```
(git-annex)lena:~/proj/dandi/dandisets/000108[draft]git
$> dandi --pdb validate-bids sub-MITU01/ses-20210521h17m17s06/micr/sub-MITU01_ses-20210521h17m17s06_sample-178_stain-NN_run-1_chunk-1_SPIM.ome.zarr
2022-06-30 16:37:17,222 [ WARNING] BIDSVersion 1.4.0 is less than the minimal working 1.7.0+012+dandi001. Falling back to 1.7.0+012+dandi001. To force the usage of earlier versions specify them explicitly when calling the validator.
2022-06-30 16:37:18,059 [    INFO] Logs saved in /home/yoh/.cache/dandi-cli/log/20220630203713Z-3420148.log
Traceback (most recent call last):
  File "/home/yoh/proj/dandi/dandi-cli-master/venvs/dev3/bin/dandi", line 33, in <module>
    sys.exit(load_entry_point('dandi', 'console_scripts', 'dandi')())
  File "/home/yoh/proj/dandi/dandi-cli-master/venvs/dev3/lib/python3.9/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/home/yoh/proj/dandi/dandi-cli-master/venvs/dev3/lib/python3.9/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/home/yoh/proj/dandi/dandi-cli-master/venvs/dev3/lib/python3.9/site-packages/click/core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/yoh/proj/dandi/dandi-cli-master/venvs/dev3/lib/python3.9/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/yoh/proj/dandi/dandi-cli-master/venvs/dev3/lib/python3.9/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/home/yoh/proj/dandi/dandi-cli-master/venvs/dev3/lib/python3.9/site-packages/click/decorators.py", line 38, in new_func
    return f(get_current_context().obj, *args, **kwargs)
[...]

PR #1050, opened 2022-07-08
Updating BIDS validator and schema to contemporary upstream equivalent

TheChymera:
Brought our code back in sync with upstream BIDS, as [we finally have upstream `ome.zarr` support](https://github.com/bids-standard/bids-specification/pull/1104) and there is no longer any reason to maintain a fork of the specification.
This also fixes: https://github.com/dandi/dandi-cli/issues/1037

Ideally this will be the last update of bundled code as we are working on importing the built-in Python validator as a library.
Should the need for specification forking arise in the future, this should best be handled via the parameterized schema directory.

~~Draft for now as there might be inconsistencies I haven't spotted.~~


TheChymera:
Additionally, we might want to start using the reference bids-examples repository for validation (rather than our own fork).
To be able to perform all relevant tests, this would require having at least one broken dataset in the examples, so that we can verify errors as well: https://github.com/bids-standard/bids-examples/issues/327

In any case, using vanilla upstream would add another ~10s to the BIDS validator tests; as our fork had stripped a lot of the datasets. While we will prospectively no longer run the validator tests themselves (these would be run by the bids-schemacode package), we still need BIDS example data to construct our derivative BIDS datasets for higher-level function testing...

Any ideas if we can somehow check out only some of the files to reduce download time? @yarikoptic

dandi/bids_validator_xs.py
```diff
@@ -453,7 +467,7 @@ def validate_all(
 
 def write_report(
     validation_result,
-    report_path="{logdir}/bids-validator-report_{datetime}-{pid}.log",
+    report_path="/var/tmp/bids-validator/report_{datetime}-{pid}.log",
```
yarikoptic, line 470: looks wrong.
So this is the code which is borrowed from 3rd party (upstream `bids-specification`)?   could such change be introduced "upstream"?

A possible workflow for carrying 3rd party components to make it clear(er) on what changes we introduce here and carry them through the upgrades of the thirdparty code (we used it in PyMVPA)

- keep all thirtd-party code in a consistent descriptive location, e.g. `dandi.support.third_party` (IIRC submodule cannot start with `3` so we will use `third`)
- establish "third-party" branch at some point where pristine versions (no modifications) of thirdparty code are introduced. (that branch could even have no common base with main tree, just carry that `dandi/support/third_party` folder)
- introduce new versions of third-party **only** in that branch. merge that branch into `master` when that is done. Then any changes to upstream code would merge with our code

as we hope that all of this would be a legit pypi package soon, may be "too late" for us here, but please first inspect `git log` on 3rd party content to see if any other change was "dropped" (has some commit what was not migrated upstream)

TheChymera: set it up in a separate submodule for now. A couple of changes are still pending upstream in bids-specification.

yarikoptic: > set it up in a separate submodule for now

git-rotten mind of me immediately thought about `git submodules`. I guess we could have used that one as well (instead of branching mechanism I described above), but I guess could be avoided.



> A couple of changes are still pending upstream in bids-specification.

I guess 
- https://github.com/bids-standard/bids-specification/pull/1120
- https://github.com/bids-standard/bids-specification/pull/1122
- https://github.com/bids-standard/bids-specification/pull/1138
?  seems all could be "adopted" merged in here later , we should not wait for them needing to amass even more changes, let's finish/merge this PR to the current state of things in BIDS validation land.

dandi/cli/cmd_validate.py
```diff
@@ -10,13 +10,11 @@
...
-@click.option("--report", help="Specify path to write a report under.")
 @click.option(
-    "--report-flag",
+    "--report",
     "-r",
     is_flag=True,
-    help="Whether to write a report under a unique path in the current directory. "
-    "Only usable if `--report` is not already used.",
+    help="Whether to write a report under a unique path in the current directory. ",
 )
```
yarikoptic, line 18: so we are breaking CLI ... why not just to keep `--report` to be an option to provide the path and if not specified, assume that no report writing was requested - print to the screen?  
Note: I might not be even able to write to current directory (dataset might be owned by someone else).

TheChymera: functionality wasn't broken, help text was just incorrect.

yarikoptic (changes requested):
`--report` handling still needs to be fixed and tested better

dandi/validate.py
```diff
@@ -22,10 +24,8 @@ def validate_bids(
         BIDS schema version to use, this setting will override the version specified in the dataset.
     devel_debug : bool, optional
         Whether to trigger debugging in the BIDS validator.
-    report_path : bool or str, optional
-        If `True` a log will be written using the standard output path of `.write_report()`.
-        If string, the string will be used as the output path.
-        If the variable evaluates as False, no log will be written.
+    report : bool or str, optional
```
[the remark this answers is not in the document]
TheChymera: Why would it explicitly check if it's a string? Any true value is true.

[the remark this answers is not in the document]
yarikoptic: do you have a test where report is saved into specified here `str`?

I see that below if is assigned to `report_path = "{log_dir}/bids-validator-report_{{datetime}}-{{pid}}.log"` indeed for any `if report` thus if I say `report="here.log"` it would still save into that path, or am I wrong?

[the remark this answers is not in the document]
TheChymera: the docstring is outdated.

dandi/validate.py
```diff
@@ -36,12 +36,24 @@ def validate_bids(
     Notes
     -----
     Can be used from bash, as:
-        DANDI_DEVEL=1 dandi validate-bids --schema="1.7.0+012+dandi001" --report="my.log" /my/path
+        DANDI_DEVEL=1 dandi validate-bids --report="my.log" /my/path
     """
-    from .bids_validator_xs import validate_bids as validate_bids_
+    from .support.bids.validator import validate_bids as validate_bids_
 
+    if report:
```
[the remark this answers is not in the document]
TheChymera: done.

dandi/validate.py
```diff
@@ -36,12 +36,24 @@ def validate_bids(
     Notes
     -----
     Can be used from bash, as:
-        DANDI_DEVEL=1 dandi validate-bids --schema="1.7.0+012+dandi001" --report="my.log" /my/path
+        DANDI_DEVEL=1 dandi validate-bids --report="my.log" /my/path
```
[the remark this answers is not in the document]
TheChymera: done. As said before, given that DANDI has its own logging control, we are no longer supporting custom `report` values.

TheChymera:
@yarikoptic 

> so, it got broken that user no longer would be able to say `--report myvalidation.log`

It didn't get broken, the feature was simply removed. I added it initially because I thought it would be helpful, and I decided that in fact it is not. The `validation()` function which I understand we want to merge BIDS into at some point does not have such a parameter, and since DANDI doesn't manage logging this way I see no reason to do it for this function specifically...

dandi/support/bids/validator.py
```diff
@@ -545,12 +545,14 @@ def write_report(
 
 def _find_dataset_description(my_path):
     candidate = os.path.join(my_path, "dataset_description.json")
-    if my_path == "/":
+    # Windows support... otherwise we could do `if my_path == "/"`.
+    if my_path == "/" or "\\" and "/" not in my_path:
```
jwodder, line 549: This needs to be `if my_path == "/" or my_path == "\\" ...`.  I also suggest you write either `if (my_path == "/" or my_path == "\\") and "/" not in my_path` or `if my_path == "/" or (my_path == "\\" and "/" not in my_path)` to make it clear which one you mean.

TheChymera: @jwodder thanks for the feedback, but better disregard whatever I'm doing until I remove draft status. I seem to have run into some windows compatibility issue bigger than just this line, and I wouldn't want to waste your time on my troubleshooting :)

Also, on the general comment, `if my_path == "/" or my_path == "\\" ...` — this won't work, as the top level on windows ends up being `C:`... but again, pls don't worry about this.

dandi/cli/cmd_validate.py
```diff
@@ -10,27 +10,29 @@
...
     is_flag=True,
-    help="Whether to write a report under a unique path in the current directory. "
-    "Only usable if `--report` is not already used.",
+    help="Whether to write a report under a unique path in the DANDI log directory.",
 )
 @click.argument("paths", nargs=-1, type=click.Path(exists=True, dir_okay=True))
 @devel_debug_option()
 @map_to_click_exceptions
 def validate_bids(
     paths, schema=None, devel_debug=False, report=False, report_flag=False
```
yarikoptic, line 23: forgotten `report_flag` here

dandi/validate.py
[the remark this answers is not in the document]
TheChymera: in any case this is an outdated diff.

dandi/cli/cmd_validate.py
```diff
@@ -10,30 +10,41 @@
...
 @click.argument("paths", nargs=-1, type=click.Path(exists=True, dir_okay=True))
 @devel_debug_option()
 @map_to_click_exceptions
 def validate_bids(
-    paths, schema=None, devel_debug=False, report=False, report_flag=False
+    paths,
+    schema=None,
+    devel_debug=False,
+    report=False,
+    report_path="",
```
jwodder, line 31: Only `devel_debug` should have a default argument, as it's the only option that's not always present.  Giving the other arguments defaults is misleading, as those are never used.

dandi/support/bids/validator.py
```diff
@@ -204,7 +208,7 @@ def load_top_level(
     Parameters
     ----------
     my_schema : dict
-        A nested dictionary, as returned by `dandi.support.bids.schema.load_schema()`.
+        A nested dictionary, as returned by `schemacode.schema.load_schema()`.
```
jwodder, line 211: What is "schemacode" here?

dandi/support/bids/validator.py
```diff
@@ -453,7 +467,7 @@ def validate_all(
 
 def write_report(
     validation_result,
-    report_path="{logdir}/bids-validator-report_{datetime}-{pid}.log",
+    report_path="/var/tmp/bids-validator/report_{datetime}-{pid}.log",
```
jwodder, line 470: I suggest still using a variable for the log base directory, but set it to `$TMPDIR` (falling back to `/tmp`) by default.

dandi/support/bids/validator.py
```diff
@@ -689,12 +701,42 @@ def log_errors(validation_result):
...
+        if i_value.endswith("/"):
+            if i_value != "/":
+                pseudofile_suffixes.append(i_value[:-1])
+    return pseudofile_suffixes
+
+
 def validate_bids(
     bids_paths,
-    schema_reference_root="{module_path}/support/bids/schemadata/",
+    schema_reference_root="/usr/share/bids-schema/",
```
jwodder, line 735: Under what conditions is this default used, and why would we expect anything to be in this directory?

dandi/tests/test_validate.py
```diff
@@ -0,0 +1,33 @@
...
+    report_expression = os.path.join(log_dir, f"bids-validator-report_*-{pid}.log")
+    assert len(glob(report_expression)) == 1
+
+    # TEST CISTOM REPORT PATH:
+    # Replace explicit dataset with `os.listdir(bids_examples)[1]` whenever we
+    # implement light data cloning analogous to (`[1]` because `[0]` is .git):
+    # https://github.com/bids-standard/bids-specification/pull/1143
+    selected_dataset = os.path.join(bids_examples, "asl003")
+    _ = validate_bids(
+        selected_dataset, report_path="/tmp/inplace_bids-validator-report.log"
```
jwodder, line 29: I suggest using pytest's `tmp_path` instead of `/tmp` here.

jwodder (changes requested):

dandi/support/bids/validator.py
TheChymera: Again, this is for upstream usage, we wrap this command and pass the DANDI log directory via appdirs. `/tmp/` is tricky as it's ephemeral on some systems, and users might not know how their system handles it. `/var/tmp` is the safer bet.

Compare:
https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s15.html
to
https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s18.html
Particularly:
> Programs must not assume that any files or directories in /tmp are preserved between invocations of the program.

is pretty much a non-starter for logging.

dandi/cli/cmd_validate.py
TheChymera: Apologies, but I don't understand what you mean by this. `dandi.cli.cmd_validate.validate()` also has default values for the others.

jwodder: @TheChymera When an option isn't given on the command line, click takes care of filling in the default.  This means that a value will always be passed in to the function, so the default values for `schema`, `report`, and `report_path` given in the signature will never be used.  Specifying default values for the function arguments is misleading, especially in the case of `report_path`, as click's default there is `None`, not an empty string.

`devel_debug` is an exception, as that option is only added to the command if an environment variable is present, and when the envvar isn't present, no `devel_debug` argument is passed to `validate_bids()`, so it needs a default.

I am aware that Yarik has misunderstood click's default handling in the past, leading to some command functions having superfluous argument defaults, but that's no reason to continue the practice.

TheChymera: @yarikoptic better? :)

dandi/support/bids/validator.py
```diff
@@ -50,23 +54,19 @@ def _get_paths(bids_paths):
...
-            if any(f"{i}/" in root for i in treat_as_file_suffix):
-                continue
-            if any(f"{i}\\" in root for i in treat_as_file_suffix):
-                continue
+        for root, dirs, file_names in os.walk(bids_path, topdown=True):
+            if root.endswith(tuple(pseudofile_suffixes)):
+                # Add the directory name to the validation paths list.
+                path_list.append(Path(root).as_posix() + "/")
+                # Do not index the contents of the directory.
+                dirs[:] = []
```
jwodder, line 69: Shouldn't you add a `continue` here so that no file paths in the directory get appended to the path list?

dandi/support/bids/validator.py
```diff
@@ -50,23 +54,19 @@ def _get_paths(bids_paths):
...
             # will break if BIDS ever puts meaningful data under `/.{dandi,datalad,git}*/`
             if any(exclude_subdir in root for exclude_subdir in exclude_subdirs):
```
jwodder, line 71: I suggest removing the `os.sep`s from `exclude_subdirs` and changing this line to:

```suggestion
            if os.path.basename(root) in exclude_subdirs:
```

Unlike the current code, this correctly handles things like `foo/.dandibar`.

Better yet, apply `exclude_subdirs` to `dirs` rather than `root` in order to actually avoid recursing into excluded subdirs.

jwodder (changes requested):

dandi/support/bids/validator.py
TheChymera: it's not needed, when topdown is True, the directory list can be modified in-place (fourth paragraph here: https://docs.python.org/3/library/os.html#os.walk )

jwodder: I meant that the code falls through at line 69 to line 73, so any files at the first level of a directory with a pseudofile suffix will still be added to `path_list`, won't they?

TheChymera: ah, yes, you are right. Thanks for picking up on this.

TheChymera:
@jwodder thanks for your input :3 good to go now?

yarikoptic:
ok, I think that main discussions/questions were addressed/resolved. Let's proceed and see where it takes us. Thank you @TheChymera !

PR #1053, opened 2022-07-08
DOC: describe the purpose for --instance in download

yarikoptic:
attn @waxlamp -- is it better?

### DIFF
--- a/dandi/cli/cmd_download.py
+++ b/dandi/cli/cmd_download.py
@@ -88,9 +88,12 @@ Download a file or entire folder from DANDI.
 )
 @instance_option(
     default=None,
-    help="Instance of DANDI archive to use. If URLs provided, we would ensure that "
-    "they point to the same instance as specified. If no URLs provided, that "
-    "instance will be used for identifier found in local dandiset.yaml.",
+    help=(
+        "DANDI Archive instance to download from. If any URLs are provided,"
+        " they must point to the given instance. If no URL is provided, and"
+        " there is a local dandiset.yaml file, the Dandiset with the identifier"
+        " given in the file will be downloaded from the given instance."
+    ),
 )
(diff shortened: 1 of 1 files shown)


dandi/cli/cmd_download.py
```diff
@@ -86,7 +86,12 @@ def get_metavar(self, param):
 @click.option(
     "--sync", is_flag=True, help="Delete local assets that do not exist on the server"
 )
-@instance_option(default=None)
+@instance_option(
+    default=None,
+    help="Instance of DANDI archive to use. If URLs provided, we would ensure that "
+    "they point to the same instance as specified. If no URLs provided, that "
+    "instance will be used for identifier found in local dandiset.yaml.",
```
jwodder, line 93: Suggested rewording:

> DANDI Archive instance to download from.  If any URLs are provided, they must point to the given instance.  If no URL is provided, and there is a local dandiset.yaml file, the Dandiset with the identifier given in the file will be downloaded from the given instance.

yarikoptic: could you please send it as suggestion (select multiple lines before clicking on suggestion icon)

jwodder, line 93:
```suggestion
    help=(
        "DANDI Archive instance to download from. If any URLs are provided,"
        " they must point to the given instance. If no URL is provided, and"
        " there is a local dandiset.yaml file, the Dandiset with the identifier"
        " given in the file will be downloaded from the given instance."
    ),
```

jwodder (changes requested):

dandi/cli/cmd_download.py
yarikoptic: indeed is better than mine, accepted

PR #1056, opened 2022-07-13
Replace numpy.asscalar() with numpy.ndarray.item()

jwodder:
[`asscalar()`](https://numpy.org/doc/1.21/reference/generated/numpy.asscalar.html) has been deprecated since numpy 1.16, with `numpy.ndarray.item()` recommended instead.  Moreover, it was [completely removed in numpy 1.23](https://numpy.org/doc/stable/release/1.23.0-notes.html#expired-deprecations), released June 22.  Note that none of our tests seem to cover the case where an NWB has any probe_ids.

### DIFF
--- a/dandi/pynwb_utils.py
+++ b/dandi/pynwb_utils.py
@@ -10,7 +10,6 @@ import dandischema
 from fscacher import PersistentCache
 import h5py
 import hdmf
-import numpy as np
 import pynwb
 from pynwb import NWBHDF5IO
 import semantic_version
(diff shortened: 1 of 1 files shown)


yarikoptic:
> Note that none of our tests seem to cover the case where an NWB has any probe_ids.

and due to `if` within list comprehension we did not even know about that through coverage. Oh well, let's just proceed for now. Thank you @jwodder !  I will mark for release so people with new numpy do not experience crashes etc.

PR #1062, opened 2022-07-15
Mention in help text that 'draft' is not supported in DANDI:<dandiset>/<version>

### LINKED ISSUE #1058: download doesn't handle DANDI:<dandiset id>[/<version id>] although says that it does
might be related to some prior issue with identifiers.org and/or our redirector (@satra might remember) which is not yet fixed. Then we should adjust documentation to not say that we support it.  If that is some internal bug -- should be fixed.

```
$> dandi download DANDI:000029/draft
2022-07-14 13:51:01,804 [    INFO] Logs saved in /home/yoh/.cache/dandi-cli/log/20220714175100Z-591196.log
Error: We do not know how to map URL DANDI:000029/draft to our servers.
Accepted resource identifier patterns:
 - DANDI:<dandiset id>[/<version id>]
 - https://dandiarchive.org/...
...
```

jwodder:
Closes #1058.

### DIFF
--- a/dandi/dandiarchive.py
+++ b/dandi/dandiarchive.py
@@ -424,11 +424,11 @@ class _dandi_url_parser:
         #       for not only "dandiarchive.org" URLs
         (
             re.compile(
-                fr"DANDI:{DANDISET_ID_REGEX}(?:/{PUBLISHED_VERSION_REGEX})?",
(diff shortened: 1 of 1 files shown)


yarikoptic:
thanks!

PR #1063, opened 2022-07-15
dandi move: Fix resolution of nonexistent dests when in a subdirectory

### LINKED ISSUE #1057: move command should take destination as a relative path too for local operation
running the following command moves the file from the current directory into the root directory of the dandiset. this seems unintuitive for a unix user.

> /mnt/beegfs/Lee/dandi/sub-MITU01/ses-20211016h18m33s42/micr$ dandi move sub-MITU01_sample-100_photo.jpg sub-MITU01_ses-20211016h18m33s42_sample-100_photo.jpg

jwodder:
Fixes #1057.

### DIFF
--- a/dandi/move.py
+++ b/dandi/move.py
@@ -216,16 +216,17 @@ class LocalizedMover(Mover):
         Given a sequence of input source paths and a destination path, return a
         sorted list of all assets that will be moved/renamed
         """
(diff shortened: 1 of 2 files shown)


yarikoptic:
Didn't test but the added unittest seems to test exactly what is needed, so let's proceed. Thank you @jwodder . Let's release with this

PR #1064, opened 2022-07-15
dandi move: Include all mismatches in error messages

### LINKED ISSUE #1060: move: report all problematic paths at once (not just error out on the first)
Compliment to fresh https://github.com/dandi/dandi-cli/issues/1059  : if there are more than one file not present on the remote server, `move` would error out informing only about one of them, thus making it difficult to e.g. first use `dandi upload` to upload those missing files. 
Since such listing of files could be quite long, I wonder even if we should may be store them into a file for user's convenience and print that filename, while displaying only first few to the screen.

```shell
lena:/tmp/000029
$> echo data > sub-anm369964/newfile.dat

$> dandi move sub-anm369964 sub-anm369964-moved 
2022-07-14 13:56:45,356 [    INFO] Fetching list of assets for Dandiset 000029
2022-07-14 13:56:45,495 [    INFO] Logs saved in /home/yoh/.cache/dandi-cli/log/20220714175643Z-592564.log
Error: Mismatch between local and remote servers: asset 'sub-anm369964/newfile.dat' only exists locally

$> echo data2 > sub-anm369964/newfile2.dat            

> dandi move sub-anm369964 sub-anm369964-moved
2022-07-14 14:01:01,147 [    INFO] Fetching list of assets for Dandiset 000029
2022-07-14 14:01:01,257 [    INFO] Logs saved in /home/yoh/.cache/dandi-cli/log/20220714180059Z-593263.log
Error: Mismatch between local and remote servers: asset 'sub-anm369964/newfile.dat' only exists locally
```

so it complained only about newfile.dat, nothing about newfile2.dat

jwodder:
Closes #1060.

### DIFF
--- a/dandi/move.py
+++ b/dandi/move.py
@@ -714,30 +714,33 @@ class LocalRemoteMover(Mover):
         any differences.
         """
         # Recall that the Movements are sorted by src path
+        mismatches = []
         for lm, rm in zip_longest(local_moves, remote_moves):
             if rm is None:
-                raise AssetMismatchError(f"asset {lm.src!r} only exists locally")
+                mismatches.append(f"Asset {lm.src!r} only exists locally")
             elif lm is None:
-                raise AssetMismatchError(f"asset {rm.src!r} only exists remotely")
+                mismatches.append(f"Asset {rm.src!r} only exists remotely")
             elif lm.src < rm.src:
-                raise AssetMismatchError(f"asset {lm.src!r} only exists locally")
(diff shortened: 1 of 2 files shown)


yarikoptic:
oh well, it wouldn't be very easy to copy / paste all those paths to be fed into `dandi upload` (which was the original motivation) but at least possible to dump into a file and post-process if so really desired, so let's proceed with this for now. Thank you @jwodder

PR #1066, opened 2022-07-15
Support DANDI:<ID>/draft and stop using identifiers.org for any of the DANDI: URIs

### LINKED ISSUE #1065: download etc: support DANDI:<dandiset id>/draft
following discussion and argumentation by @satra in https://github.com/dandi/dandi-cli/issues/1058#issuecomment-1184949961 .

yarikoptic:
Closes #1065 

If we are to support `DANDI:<ID>/draft` which cannot go through the identifiers.org, I saw no reason to go through identifiers.org for **any** `DANDI:<ID>` since that, if anything, could only introduce inconsistency.

I have left also initial version which creates a dedicated rule for `/draft` version in a separate commit "just in case".

@satra -- please review to agree that there would be no reason to go through identifiers.org

@jwodder -- please review for code changes etc


dandi/dandiarchive.py
```diff
@@ -442,7 +444,7 @@ class _dandi_url_parser:
                 flags=re.I,
             ),
             {"handle_redirect": "pass"},
-            "https://identifiers.org/DANDI:<dandiset id>[/<version id>] (<version id> cannot be 'draft')",
+            "https://identifiers.org/DANDI:<dandiset id>[/<version id>]",
```
satra, line 447: i would revert this change, since on identifiers.org version id indeed cannot be draft.

jwodder:
@yarikoptic This PR should also include an update to [`docs/source/ref/urls.rst`](https://github.com/dandi/dandi-cli/blob/master/docs/source/ref/urls.rst).

dandi/dandiarchive.py
yarikoptic: indeed, I have `sed`ed too much... will rewrite and force-push

yarikoptic:
> @yarikoptic This PR should also include an update to [`docs/source/ref/urls.rst`](https://github.com/dandi/dandi-cli/blob/master/docs/source/ref/urls.rst).

thanks! adjusted in 49e1d132795721b24f3deeda15b793e49c6e123b

jwodder (approved):

PR #1071, opened 2022-07-18
Do not traverse .ome.zarr and .ngff folders while looking for BIDS datasets

### LINKED ISSUE #1069: Hanging dandi upload call
Attempt to run dandi upload leading to indefinite stalling. At first I thought it may be due to output not piping correctly, but logs revealed no activity.

On version 0.45.1

Command
`DANDI_DEVEL=1 dandi upload --validation skip --allow-any-path ./`

Log in it's entirety
```
2022-07-18T14:09:44-0400 [INFO    ] dandi 15159:139926543738688 dandi v0.45.1, hdmf v1.5.4, pynwb v1.2.1, h5py v2.10.0
2022-07-18T14:09:44-0400 [INFO    ] dandi 15159:139926543738688 sys.argv = ['/mnt/beegfs/satra/miniconda3/envs/dandi/bin/dandi', 'upload', '--validation', 'skip', '--allow-any-path', './']
2022-07-18T14:09:44-0400 [INFO    ] dandi 15159:139926543738688 os.getcwd() = /mnt/beegfs/Lee/dandi/sub-MITU01/ses-20211111h11m08s18/micr
2022-07-18T14:09:45-0400 [DEBUG   ] urllib3.connectionpool 15159:139926543738688 Starting new HTTPS connection (1): rig.mit.edu:443
2022-07-18T14:09:45-0400 [DEBUG   ] urllib3.connectionpool 15159:139926543738688 https://rig.mit.edu:443 "GET /et/projects/dandi/dandi-cli HTTP/1.1" 200 579
2022-07-18T14:09:45-0400 [DEBUG   ] dandi 15159:139926543738688 No newer (than 0.45.1) version of dandi/dandi-cli found available
2022-07-18T14:10:21-0400 [INFO    ] dandi 15159:139926543738688 Logs saved in /home/smarx/.cache/dandi-cli/log/20220718180944Z-15159.log
```

yarikoptic:
Fixes #1069

TODO
- [ ] check on real case if that helps enough

### DIFF
--- a/dandi/upload.py
+++ b/dandi/upload.py
@@ -419,6 +419,7 @@ def _bids_discover_and_validate(
         find_files(
             r"(^|[/\x5C])dataset_description\.json$",
             dandiset_path,
+            # for cases like sub-MITU01h3_..._chunk-4_SPIM.ngff.source
             dirs_avoid=r"\.(ngff|zarr)(\..*)?$",
         ),
     )


dandi/upload.py
```diff
@@ -413,8 +413,14 @@ def _bids_discover_and_validate(
...
     from .validate import validate_bids
 
+    lgr.debug("Discovering root directories of BIDS datasets")
     bids_descriptions = map(
-        Path, find_files(r"(^|[/\x5C])dataset_description\.json$", dandiset_path)
+        Path,
+        find_files(
+            r"(^|[/\x5C])dataset_description\.json$",
+            dandiset_path,
+            dirs_avoid=r"\.(ngff|zarr)$",
```
yarikoptic, line 422: ha -- the case at hands has `micr/sub-MITU01h3_ses-20211202h20m54s11_sample-20_stain-NN_run-1_chunk-4_SPIM.ngff.source` so I would need to add allowing for more extensions for such cases.

yarikoptic: Let's allow for any other suffix as well:

```suggestion
            dirs_avoid=r"\.(ngff|zarr)(\..*)?$",
```

jwodder: Note that the main upload code currently only regards directories ending in `.ngff` and `.zarr` as Zarrs; `*.ngff.source` and the like are just treated as regular directories.

yarikoptic:
ok -- with this I fail (incorrect token) on our target use case within 3 seconds, so we manage to traverse enough fast enough, so effectively addresses the case. but may be we would tune bids detection more - see #1073

satra:
@yarikoptic - does the check still run when `--allow-any-path` is used? as in is there a necessity of any bids-specific traversal other than `dandiset.yaml` detection?

and secondarily, if we don't use that flag and do `--validation skip` then does it care for anything else other than the specific directory contents and the presence of a `dataset_description.json`. i.e. it should not care about files in the other subject's directory.

(and the one ngff file in MITU01 is being removed).

dandi/upload.py
```diff
@@ -413,8 +413,14 @@ def _bids_discover_and_validate(
...
     from .validate import validate_bids
 
+    lgr.debug("Discovering root directories of BIDS datasets")
     bids_descriptions = map(
-        Path, find_files(r"(^|[/\x5C])dataset_description\.json$", dandiset_path)
+        Path,
+        find_files(
+            r"(^|[/\x5C])dataset_description\.json$",
+            dandiset_path,
+            dirs_avoid=r"\.(ngff|zarr)(\..*)?$",
```
TheChymera, line 422: is this bit `(\..*)?` required? I know there is `ome.zarr` — but is there `zarr.ome` or `.ngff.ome`?

yarikoptic: there was a resolved [comment](https://github.com/dandi/dandi-cli/pull/1071#discussion_r923893998):

ha -- the case at hands has micr/sub-MITU01h3_ses-20211202h20m54s11_sample-20_stain-NN_run-1_chunk-4_SPIM.ngff.source so I would need to add allowing for more extensions for such cases.

Let's add a comment so we know *why* later on

```suggestion
            # for cases like sub-MITU01h3_..._chunk-4_SPIM.ngff.source
            dirs_avoid=r"\.(ngff|zarr)(\..*)?$",
```

dandi/upload.py
yarikoptic: oh.  Thanks! prepared separate PR for that: https://github.com/dandi/dandi-cli/pull/1077 .

yarikoptic:
filed https://github.com/dandi/dandi-cli/issues/1078 for a failed test, filed https://github.com/dandi/dandi-cli/pull/1077 for a separate aspect identified by @jwodder , got blessing from @jwodder. So Let's proceed with this but I will delay release -- may be more PRs would get merged in short time.

PR #1077, opened 2022-07-19
Treat .ngff.source as .ngff/.zarr in upload

yarikoptic:
Alerted to by @jwodder in https://github.com/dandi/dandi-cli/pull/1071#discussion_r923895609 . Added a helper `has_suffixes`.

### DIFF
--- a/dandi/files.py
+++ b/dandi/files.py
@@ -75,7 +75,7 @@ from .support.digests import (
     get_zarr_checksum,
     md5file_nocache,
 )
-from .utils import chunked, pluralize, yaml_load
+from .utils import chunked, has_suffixes, pluralize, yaml_load
 
 lgr = get_logger()
 
@@ -1138,7 +1138,7 @@ def dandi_file(
         if not any(filepath.iterdir()):
             raise UnknownAssetError("Empty directories cannot be assets")
         for dirclass in LocalDirectoryAsset.__subclasses__():
-            if filepath.suffix in dirclass.EXTENSIONS:
+            if has_suffixes(filepath.suffix, dirclass.EXTENSIONS):
                 return dirclass(filepath=filepath, path=path)  # type: ignore[abstract]
         raise UnknownAssetError(
             f"Directory has unrecognized suffix {filepath.suffix!r}"
@@ -1147,7 +1147,7 @@ def dandi_file(
         return DandisetMetadataFile(filepath=filepath)
     else:
         for fileclass in LocalFileAsset.__subclasses__():
-            if filepath.suffix in fileclass.EXTENSIONS:
+            if has_suffixes(filepath.suffix, fileclass.EXTENSIONS):
                 return fileclass(filepath=filepath, path=path)  # type: ignore[abstract]
         return GenericAsset(filepath=filepath, path=path)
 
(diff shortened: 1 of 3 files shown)


dandi/utils.py
```diff
@@ -403,6 +403,24 @@ def find_parent_directory_containing(
         path = path.parent  # go up
 
 
+def has_suffixes(filename: Union[str, Path], suffixes: Union[set, list, tuple]) -> set:
+    """For a filename, return the set of matching suffixes among provided.
+
+    Could be used as a boolean indicator on either filename has any of the suffixes
```
jwodder, line 412:
```suggestion
    Could be used as a boolean indicator as to whether the filename has any suffixes
```

dandi/utils.py
```diff
@@ -403,6 +403,24 @@ def find_parent_directory_containing(
         path = path.parent  # go up
 
 
+def has_suffixes(filename: Union[str, Path], suffixes: Union[set, list, tuple]) -> set:
```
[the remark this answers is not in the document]
yarikoptic: cool, will rewrite commit locally with needed import. Thanks!

dandi/tests/test_utils.py
```diff
@@ -459,3 +460,42 @@ def test_get_mime_type(filename: str, mtype: str) -> None:
...
+        ("blah", {".ngff", ".dat"}, set()),
+        ("blah", set(), set()),
+        ("blah.ome", set(), set()),
+    ],
+)
+def test_has_suffixes(filename: str, suffixes: set, target: set) -> None:
+    # as is
+    assert has_suffixes(filename, suffixes) == target
+
+    # poor Yarik doesn't know nice way for combining multiple .parametrize
```
jwodder, line 485: Try:

```
@pytest.mark.parametrize("filename,suffixes,target", [ ... ])
@pytest.mark.parametrize("suffix_t", [list, set, tuple])
def test_has_suffixes(filename: str, suffixes: set[str], target: set[str], suffix_t: type) -> None:
```

dandi/files.py
```diff
@@ -1147,7 +1147,7 @@ def dandi_file(
         return DandisetMetadataFile(filepath=filepath)
     else:
         for fileclass in LocalFileAsset.__subclasses__():
-            if filepath.suffix in fileclass.EXTENSIONS:
+            if has_suffixes(filepath.suffix, fileclass.EXTENSIONS):
```
jwodder, line 1150: Are you sure we want, say, `foo.nwb.bar` to be treated as an NWB?

jwodder (changes requested):

dandi/tests/test_utils.py
yarikoptic: I didn't dare to try -- works wonderfully ;)  will push this change even though I think we might close this PR ;)

satra:
those are indeed mistakes in upload the source folders shouldn't have gone out. i've alerted folks.

yarikoptic:
> those are indeed mistakes in upload the source folders shouldn't have gone out. i've alerted folks.

well , the question is what should we do to prevent "hammer effect" of such user mistakes? because now it would have lead to upload of each and one of those files as a separate asset, thus populating 100k assets in the DB.  It should survive but clearly not a pleasant thing to do if could be avoided...  I think this PR we can just close, but let's continue on that question in https://github.com/dandi/dandi-cli/issues/1079

Issue #1055, opened 2022-07-08
BIDS: does not upload anything BIDS if not explicitly specifying and --allow-any-path


yarikoptic:
I thought that #1011 should have made it magically happened to use our `dandi.utils.find_files` (while excluding VCS) if bids dataset is detected, or may be just "valid bids" paths... 

<details>
<summary>but nothing like that happens -- nothing gets uploaded unless I explicitly use --allow-any-paths in DANDI_DEVEL=1 mode:</summary> 

```shell
(git)lena:~/proj/bids/bids-examples[master]ds210
$> dandi upload -i dandi-staging       
2022-07-08 19:52:39,208 [ WARNING] BIDSVersion 1.0.2 is less than the minimal working 1.7.0+012+dandi001. Falling back to 1.7.0+012+dandi001. To force the usage of earlier versions specify them explicitly when calling the validator.
2022-07-08 19:52:39,924 [   ERROR] The `.*?/README$` regex pattern file required by BIDS was not found.
2022-07-08 19:52:39,925 [   ERROR] The `.*?/CHANGES$` regex pattern file required by BIDS was not found.
2022-07-08 19:52:39,925 [ WARNING] The `/home/yoh/proj/bids/bids-examples/ds210/task-cuedSGT_echo-1_bold.json` file was not matched by any regex schema entry.
2022-07-08 19:52:39,925 [ WARNING] The `/home/yoh/proj/bids/bids-examples/ds210/task-cuedSGT_echo-2_bold.json` file was not matched by any regex schema entry.
2022-07-08 19:52:39,925 [ WARNING] The `/home/yoh/proj/bids/bids-examples/ds210/task-cuedSGT_echo-3_bold.json` file was not matched by any regex schema entry.
2022-07-08 19:52:39,925 [ WARNING] The `/home/yoh/proj/bids/bids-examples/ds210/task-rest_echo-1_bold.json` file was not matched by any regex schema entry.
2022-07-08 19:52:39,925 [ WARNING] The `/home/yoh/proj/bids/bids-examples/ds210/task-rest_echo-2_bold.json` file was not matched by any regex schema entry.
2022-07-08 19:52:39,925 [ WARNING] The `/home/yoh/proj/bids/bids-examples/ds210/task-rest_echo-3_bold.json` file was not matched by any regex schema entry.
2022-07-08 19:52:40,494 [    INFO] Found 1 files to consider
[...]

PR #1080, opened 2022-07-19
User notification if datasets are invalid.

TheChymera:
Addressing: https://github.com/dandi/dandi-cli/issues/1055#issuecomment-1189149375

Example:
```console
chymera@decohost ~/src/bids-examples $ dandi upload -i dandi-staging eeg_cbm/
2022-07-19 14:17:00,460 [    INFO] Note: NumExpr detected 12 cores but "NUMEXPR_MAX_THREADS" not set, so enforcing safe limit of 8.
2022-07-19 14:17:00,460 [    INFO] NumExpr defaulting to 8 threads.
2022-07-19 14:17:01,688 [ WARNING] BIDSVersion 1.0.2 is less than the minimal working 1.7.0+012+dandi001. Falling back to 1.7.0+012+dandi001. To force the usage of earlier versions specify them explicitly when calling the validator.
2022-07-19 14:17:02,435 [   ERROR] The `.*?/CHANGES$` regex pattern file required by BIDS was not found.
2022-07-19 14:17:02,435 [    INFO] Logs saved in /home/chymera/.cache/dandi-cli/log/20220719181659Z-26640.log
Error: Found 1 BIDS dataset, which did not pass validation:
 * /home/chymera/src/bids-examples/eeg_cbm
To resolve this, perform the required changes or set the validation parameter to "skip" or "ignore".
```

### DIFF
--- a/dandi/tests/test_upload.py
+++ b/dandi/tests/test_upload.py
@@ -185,6 +185,7 @@ def test_upload_bids_invalid(
     mocker: MockerFixture, bids_dandiset_invalid: SampleDandiset
 ) -> None:
     iter_upload_spy = mocker.spy(LocalFileAsset, "iter_upload")
+    # Does it fail when it should fail?
     with pytest.raises(RuntimeError):
         bids_dandiset_invalid.upload(existing="forced")
     iter_upload_spy.assert_not_called()
--- a/dandi/tests/test_upload.py
+++ b/dandi/tests/test_upload.py
@@ -185,8 +185,8 @@ def test_upload_bids_invalid(
     mocker: MockerFixture, bids_dandiset_invalid: SampleDandiset
 ) -> None:
     iter_upload_spy = mocker.spy(LocalFileAsset, "iter_upload")
-    bids_dandiset_invalid.upload(existing="forced")
-    # Check whether upload failed
+    with pytest.raises(RuntimeError):
+        bids_dandiset_invalid.upload(existing="forced")
     iter_upload_spy.assert_not_called()
 
 
--- a/dandi/tests/test_upload.py
+++ b/dandi/tests/test_upload.py
@@ -181,9 +181,20 @@ def test_upload_sync_folder(
         text_dandiset.dandiset.get_asset_by_path("subdir2/banana.txt")
 
 
-def test_upload_bids(mocker: MockerFixture, bids_dandiset: SampleDandiset) -> None:
+def test_upload_bids_invalid(
+    mocker: MockerFixture, bids_dandiset_invalid: SampleDandiset
+) -> None:
     iter_upload_spy = mocker.spy(LocalFileAsset, "iter_upload")
-    bids_dandiset.upload(existing="forced")
+    bids_dandiset_invalid.upload(existing="forced")
+    # Check whether upload failed
+    iter_upload_spy.assert_not_called()
+
+
+def test_upload_bids_validation_ignore(
+    mocker: MockerFixture, bids_dandiset: SampleDandiset
+) -> None:
+    iter_upload_spy = mocker.spy(LocalFileAsset, "iter_upload")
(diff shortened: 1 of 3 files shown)


dandi/upload.py
```diff
@@ -453,6 +453,19 @@ def _bids_discover_and_validate(
             )
             if valid:
                 validated_datasets.append(bd)
+        invalid_datasets = []
+        for i in bids_datasets_to_validate:
+            if i not in validated_datasets:
+                invalid_datasets.append(str(i))
+        if invalid_datasets:
+            raise RuntimeError(
```
yarikoptic, line 461: should not raise exception if validation = 'ignore' (as exception suggests).

and if `ignore` it is worthwhile to lgr.warning (instead of exception) that information.  So just format the message, then then 

```python
if validation == 'ignore':
    lgr.warning(msg)
else:
    raise RuntimeError(msg) 
```

dandi/upload.py
```diff
@@ -453,6 +453,19 @@ def _bids_discover_and_validate(
             )
             if valid:
                 validated_datasets.append(bd)
+        invalid_datasets = []
+        for i in bids_datasets_to_validate:
+            if i not in validated_datasets:
+                invalid_datasets.append(str(i))
```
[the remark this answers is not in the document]
TheChymera: done

dandi/upload.py
TheChymera: not really, the logic is encoded here:

```python
valid = is_valid(
                validator_result,
                allow_missing_files=validation == "ignore",
                allow_invalid_filenames=validation == "ignore",
            )
```

The function starts off with this:

```python
if allow_invalid_filenames and allow_missing_files:
        return True
```

so there's no case for which behaviour would not be equivalent.

yarikoptic: I don't think any of that is related to my concern. Am I correct that having this exception here, it would be raised if any found BIDS dataset is invalid, correct?  If yes -- I will not be able to upload **any** file from any dataset since dandi cli would just crash before due to this exception, right? and that is what I am trying to avoid

TheChymera: Right, but if `validation == "ignore"` all datasets will always be evaluated as valid. The only case where you could get an exception is if the validation code breaks and one is raised, to bypass running validation altogether we have `validation == "skip"`.

yarikoptic: ok, as too me it is too much of assumptions on the code behavior somewhere not nearby. but ok, let's just proceed for now and see how it goes.

Meanwhile please address typing issues which are still there!

TheChymera:
@yarikoptic should be done

dandi/tests/test_upload.py
```diff
@@ -181,9 +181,20 @@ def test_upload_sync_folder(
...
 
-def test_upload_bids(mocker: MockerFixture, bids_dandiset: SampleDandiset) -> None:
+def test_upload_bids_invalid(
+    mocker: MockerFixture, bids_dandiset_invalid: SampleDandiset
+) -> None:
     iter_upload_spy = mocker.spy(LocalFileAsset, "iter_upload")
-    bids_dandiset.upload(existing="forced")
+    with pytest.raises(RuntimeError):
+        bids_dandiset_invalid.upload(existing="forced")
+    iter_upload_spy.assert_not_called()
```
yarikoptic, line 190: Per our [discussion](https://github.com/dandi/dandi-cli/pull/1080#discussion_r924875265) about exception, should this be working:

```suggestion
    iter_upload_spy.assert_not_called()
    bids_dandiset_invalid.upload(existing="forced", validation="ignore")
    iter_upload_spy.assert_called()
```
? if so let's add it in?

TheChymera: sounds good, thansk

TheChymera:
@yarikoptic :3

yarikoptic:
Thanks, let's proceed!

PR #1084, opened 2022-07-26
ENH: list localhost:8085  as the web UI for dandi-api-local-docker-tests

yarikoptic:
It is not really for the tests but if local dandi-archive instance is
established following instructions. This follows @jwodder idea instead of
https://github.com/dandi/dandi-cli/pull/1003 with more explicit dedicated instance

Fixes #1003


jwodder (approved):

yarikoptic:
oh right, duplication ...

yarikoptic:
and dang -- I already did folding into that instance within #1003 -- I will finalize that one instead

PR #1088, opened 2022-08-01
Factor out common fields in nwb2asset() and get_default_metadata()

### LINKED ISSUE #456: Rewrite nwb2asset() to use get_default_metadata() as a base
As mentioned [here](https://github.com/dandi/dandi-cli/pull/452#discussion_r589607072).

jwodder:
Closes #456.

### DIFF
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -636,9 +636,7 @@ def extract_session(metadata: dict) -> Optional[List[models.Session]]:
     ]
 
 
-def extract_digest(
-    metadata: dict,
-) -> Optional[Dict[models.DigestType, str]]:
+def extract_digest(metadata: dict) -> Optional[Dict[models.DigestType, str]]:
     if "digest" in metadata:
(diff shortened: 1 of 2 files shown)


dandi/metadata.py
```diff
@@ -636,9 +636,7 @@ def extract_session(metadata: dict) -> Optional[List[models.Session]]:
     ]
 
 
-def extract_digest(
-    metadata: dict,
-) -> Optional[Dict[models.DigestType, str]]:
+def extract_digest(metadata: dict) -> Optional[Dict[models.DigestType, str]]:
```
yarikoptic, line 639: I see no content difference -- why is it jumping like this then -- did we change blacking settings since last time this file was modified?

jwodder: No, I just noticed that the signature could fit on one line and felt compelled to fix it.

yarikoptic:
Thanks!  I dislike functions with undocumented side-effects, but that is ok for now. Let's proceed.

PR #1100, opened 2022-08-11
Update video NWB fixtures for next pynwb version

### LINKED ISSUE #1099: Troubleshoot failing/stalling tests on master
see https://github.com/dandi/dandi-cli/actions/runs/2837734398 which has

[test (ubuntu-18.04, 3.7, dev-deps)](https://github.com/dandi/dandi-cli/runs/7781214399?check_suite_focus=true#step:8:28741)
Process completed with exit code 1.
[test (ubuntu-18.04, 3.10, normal)](https://github.com/dandi/dandi-cli/runs/7781213938?check_suite_focus=true)
The job running on runner GitHub Actions 4 has exceeded the maximum execution time of 360 minutes.
[test (ubuntu-18.04, 3.10, normal)](https://github.com/dandi/dandi-cli/runs/7781213938?check_suite_focus=true#step:8:26430)
The operation was canceled.

jwodder:
Closes #1099.

### DIFF
--- a/dandi/tests/fixtures.py
+++ b/dandi/tests/fixtures.py
@@ -521,7 +521,7 @@ def _create_nwb_files(video_list):
             name=f"MouseWhiskers{no}",
             format="external",
             external_file=[str(vid_1), str(vid_2)],
-            starting_frame=[0, 0],
+            starting_frame=[0, 2],
             starting_time=0.0,
             rate=150.0,
         )
--- a/dandi/tests/fixtures.py
+++ b/dandi/tests/fixtures.py
@@ -521,7 +521,7 @@ def _create_nwb_files(video_list):
             name=f"MouseWhiskers{no}",
             format="external",
             external_file=[str(vid_1), str(vid_2)],
-            starting_frame=[0],
+            starting_frame=[0, 0],
             starting_time=0.0,
             rate=150.0,
         )


dandi/tests/fixtures.py
```diff
@@ -521,7 +521,7 @@ def _create_nwb_files(video_list):
             name=f"MouseWhiskers{no}",
             format="external",
             external_file=[str(vid_1), str(vid_2)],
-            starting_frame=[0],
+            starting_frame=[0, 0],
             starting_time=0.0,
```
yarikoptic, line 525: @bendichter so is it some general convention that if it is a scalar (not a list), like here for `starting_time` - it would apply to all external files, but if it is a sequence, then it must correspond in number of items? (kinda makes sense to me, but want to make sure, or we better right away convert starting_time into `[0.0, 0.0]`? 

I also wonder if that lead to unloadable nwb files -- here we generate them on the fly, but what if we stored that one produced with prior nwb with a list with a single value -- would it load with current version?

bendichter: When you have multiple vidoe files in a single `ImageSeries`, you are saying that you want to concatenate these videos in time to create a single ImageSeries.

`starting_time` indicates the time of the first frame of the (first) video with respect to the `NWBFile.session_start_time`, so it would not make sense to convert this value to a vector and the schema would not (or at least should not) allow that.

Here is the definition of the `starting_frame` attribute in the [NWB schema](https://nwb-schema.readthedocs.io/en/latest/format.html?highlight=starting_frame#imageseries):

> Each external image may contain one or more consecutive frames of the full ImageSeries. This attribute serves as an index to indicate which frames each file contains, to faciliate random access. The ‘starting_frame’ attribute, hence, contains a list of frame numbers within the full ImageSeries of the first frame of each file listed in the parent ‘external_file’ dataset. Zero-based indexing is used (hence, the first element will always be zero). For example, if the ‘external_file’ dataset has three paths to files and the first file has 5 frames, the second file has 10 frames, and the third file has 20 frames, then this attribute will have values [0, 5, 15]. If there is a single external file that holds all of the frames of the ImageSeries (and so there is a single element in the ‘external_file’ dataset), then this attribute should have value [0].
> 
> Data Type: int32
> Dimensions: [‘num_files’]
> Shape: [None]
> Name: starting_frame

So for example of there are 3 videos that have 100, 200, and 300 frames, the `starting_frame` should be `[0, 101, 301]`.These values need to be integers and are not allowed to be floats.

Does this answer your question, @yarikoptic?

yarikoptic: Thank you @bendichter - it clarifies alot!  I should have RTFM I guess but it is so much easier to just ask you ;)  But here are follow up questions
- is `starting_frame=[0, 0],` *legit* ? 
- or is it legit to have mismatch in number of frames in each video file and their indexing within `starting_frame` (e.g. `[0, 1]` whenever first file has >1 frames)?
-   if nwb would decide later that it is illegit I guess this test would start failing again since we should have it `[0, get_nframes(vid_1)]` right?

bendichter: Well the second number should be the number of frames in the first video, so I would say that [0,0] is not 100% legit. [0,1] would mean that there is one single frame, which is not likely but I suppose it is possible.

dandi/tests/fixtures.py
```diff
@@ -521,7 +521,7 @@ def _create_nwb_files(video_list):
             name=f"MouseWhiskers{no}",
             format="external",
             external_file=[str(vid_1), str(vid_2)],
-            starting_frame=[0],
+            starting_frame=[0, 0],
```
bendichter, line 524:
```suggestion
            starting_frame=[0, 10],
```

jwodder: Where did you get 10?  To me, [it looks like](https://github.com/dandi/dandi-cli/blob/7c42ed91c084e978f9904fd7495f8fcf97239b1a/dandi/tests/fixtures.py#L467-L486) there are two frames per video.

bendichter:
@jwodder oh I didn't see that, I was just changing the value to be more plausible. OK, let's change it to 2 then

yarikoptic:
Thank you @jwodder and @bendichter , let's proceed!

PR #1101, opened 2022-08-17
Add support for age range 

### LINKED ISSUE #1090: age metadata extraction preventing upload
I was just on a call with a user trying to upload data and she was blocked by a few things.

One of the problems was that she was expressing age as "24 to 48 hrs". This is valid in NWB. Our recommended best practice is to use ISO 8601 duration format, but that is not a requirement. DANDI failed with a message "failed to...". After some digging it became clear the message was "failed to extract metadata" and that age was the culprit. This was really bad user experience. She has been trying to upload this data for weeks, spending hours of her time on debugging and was clearly exasperated at this point. Even with me remote controlling into her computer it took 40 minutes to sort all of this out. In the end, she'll need to remove the age field and run her conversions again, which creates the same result as if dandi had simply ignored age values it cannot parse to begin with. That would be my preferred behavior: to permit upload and simply not extract age metadata if it cannot be parsed.

We have had several conversations with the NWB team about how best to handle cases like this in the standard when the exact age is not known. I think ultimately it might be useful to have lower_bound and upper_bound as optional attributes of the age dataset, both of which, if included, must be ISO 8601, so that age ranges can be parsed by DANDI. Until then, would it be possible to be a little more relaxed about age metadata?

djarecka:
Adding range to the option for age. Using an ISO8601 extension for intervals, i.e:
`<ISOduration>/<ISOduration>` - both limits known
`/<ISOduration>` or `<ISOduration>/` - one limit is  unknown
`/` - both limits are unknown

Please check the tests for the new examples that will be covered, closes #1090 

I've realized that there is some redundant code for the iso formatting (some in utils.py), will try to clean it a bit later

### DIFF
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -246,8 +246,8 @@ def _check_range_limits(limits: List[List[str], List[str]]):
         ok = False
     if not ok:
         raise ValueError(
-            "The upper limit has to be larger than the lower limit "
-            "and they should have a consistent units."
+            "The upper limit has to be larger than the lower limit, "
+            "and they should have consistent units."
         )
 
 
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -287,7 +287,7 @@ def parse_age(age: Optional[str]) -> Tuple[str, str]:
                 limits.append([""])
             else:
                 raise ValueError(
-                    f"Age that use / for range need to use ISO8601 format, "
+                    f"Ages that use / for range need to use ISO8601 format, "
                     f"but {el!r} found."
                 )
         age_f = limits[0] + ["/"] + limits[1]
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -201,7 +201,7 @@ def _parse_hours_format(age: str) -> Tuple[str, List[str]]:
         return age, []
 
 
-def _check_decimal_parts(age_parts: List[str]):
+def _check_decimal_parts(age_parts: List[str]) -> None:
     """checking if decimal parts are only in the lowest order component"""
     # if the last part is the T component I have to separate the parts
     decim_part = ["." in el for el in age_parts]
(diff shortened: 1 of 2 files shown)


dandi/metadata.py
```diff
@@ -230,7 +277,24 @@ def parse_age(age: Optional[str]) -> Tuple[str, str]:
...
+        age = age.replace(" ", "")
+        limits = []
+        for el in age.split("/"):
+            if el.startswith("P"):
+                limits.append(_parse_iso8601(el))
+            elif el == "":  # start or end of range is unknown
+                limits.append([""])
+            else:
+                raise ValueError(
+                    f"Age that use / for range need to use ISO8601 format, "
```
jwodder, line 290:
```suggestion
                    f"Ages that use / for range need to use ISO8601 format, "
```

dandi/metadata.py
```diff
@@ -180,28 +195,60 @@ def _parse_hours_format(age: str) -> Tuple[str, List[str]]:
...
+                break
+            else:  # lower limit has higher unit
+                ok = False
+                break
+    if len(lower) > len(upper):  # lower has still more elements
+        ok = False
+    if not ok:
+        raise ValueError(
+            "The upper limit has to be larger than the lower limit "
+            "and they should have a consistent units."
```
jwodder, line 250:
```suggestion
            "The upper limit has to be larger than the lower limit, "
            "and they should have consistent units."
```

jwodder (changes requested):

djarecka:
@jwodder - i have some unrelated to my changes issues. I can fix spelling to please the linter, but there are also tests in `test_video_organize` that gives error under dev-deps. Do you know what is wrong?

jwodder:
@djarecka The dev-deps failures are known and are being dealt with in #1100; ignore them for now.  As for the linter, see if the error is still there after you address codespell's concerns.

djarecka:
codespell complains about name of the file and regex... 
https://github.com/dandi/dandi-cli/blob/master/dandi/tests/test_pynwb_utils.py#L43
will think about it

yarikoptic:
Thank you @djarecka and @jwodder ! Let's proceed!

djarecka:
@bendichter - it was merged sooner than I thought, so not sure if you had a chance to check the examples :)

PR #1104, opened 2022-08-18
Structured validation results

### LINKED ISSUE #943: validate: RF to collect/output more informative structured records
Currently `validate` is just collecting a list of string error representations.  As @CodyCBakerPhD mentions in https://github.com/dandi/dandi-cli/pull/941/files#r830671373 , nwbinspector provides more informative result records with contextual information about the error etc.

Similarly for BIDS validation being added in #896 error messages are also serialized/simplified, although stock `bids-validator` has quite an extensive registry of warnings/errors: [google spreadsheet](https://docs.google.com/spreadsheets/d/181vtw2bt2nHcUgw__G_gFSkKGyuxOxUNr5wHrZXY_uI/edit?usp=sharing)

Looking forward we should establish a ValidationResult data type/class/record which would be expressive to cover validation results from various validators.  I foresee it to also reflect the "source" of the validation result, so overall should have (more to be filled out later on; yet to look at what nwbinspector provides):

- `origin`: what produced this validation result
  - `name: str` , `{"pynwb.load", "pynwb.validate", "nwbinspector", "dandischema", "dandi-bids", "ome-zarr"}`
  - `version: str` -- version of the tool producing the result, e.g. `"0.3.1"`
- `severity: {'ERROR', 'WARNING', 'HINT'}`  -- for now just reflecting 
- `id: str` (or `id: list`?) -- identifier of that particular type of result. For BIDS could be its Key or numeric ID. Providing a unique string ID like BIDS does quite nice since gives immediate summary in its name 
- `scope: {'file', 'dandiset'}` -- either it is
- `path: str` - path to file or folder/dataset (e.g. BIDS of dandiset) which triggered the error
- `asset_paths: Optional[list[str]]` - defaults to `[.path]` if not defined. Otherwise - could point to an actual file (or collection of files) which are affected by a warning/error in the `.path`. E.g. `{"path": "task-broken_bold.json", "asset_paths": ["sub-01/func/sub-01_task-broken_bold.json", "sub-02/func/sub-02_task-broken_bold.json"]}`.
- `dataset_path: Optional[str]` -- path to the dataset based on validation of which message is based (e.g. could be a nested BIDS dataset, or not present for `pynwb` validations since they do not care)
- `message: str` - it would be the "Reason" of BIDS or a string interpolation of nwbinspector record?

### Use

- `dandi validate` should gain `--severity=HINT,WARNING,ERROR` switch to provide minimal level for what to include in the report
- `dandi upload` should use validate with `severity='ERROR'` thus only allowing for uploads without errors but possibly which would trigger WARNINGs and any other lower level problems 

edits:
- here is what @nellh has for data structures to encapsulate in schema-based bids-validator: https://github.com/bids-standard/bids-validator/pull/1413/files#diff-3e29891088bc511ec972c5b36dcb579d59db3552fa69b3c882e065931d790664R6
- @effigies also expressed some utopian idea -- if schema-based errors could point to specific BIDS schema file which instructed that particular structure, that would be useful; but to me it sounds like tricky to implement to make checks aware of original location of definitions in schema 
- nwbinspector's `@dataclass` describing InspectorMessage: https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/nwbinspector/register_checks.py#L38

jwodder:
Closes #943.

### DIFF
--- a/dandi/cli/cmd_validate.py
+++ b/dandi/cli/cmd_validate.py
@@ -11,7 +11,7 @@ from ..validate_types import Severity
 
 @click.command()
 @click.option(
-    "--schema", help="Validate against new BIDS schema version", metavar="VERSION"
+    "--schema", help="Validate against new BIDS schema version.", metavar="VERSION"
 )
 @click.option(
     "--report-path",
--- a/dandi/cli/cmd_validate.py
+++ b/dandi/cli/cmd_validate.py
@@ -21,6 +21,11 @@ from ..utils import pluralize
     is_flag=True,
     help="Whether to write a report under a unique path in the DANDI log directory.",
 )
+@click.option(
+    "--grouping",
+    "-g",
+    help="Write report under path, this option implies `--report/-r`.",
+)
 @click.argument("paths", nargs=-1, type=click.Path(exists=True, dir_okay=True))
 @map_to_click_exceptions
 def validate_bids(
@@ -28,8 +33,15 @@ def validate_bids(
     schema,
     report,
     report_path,
+    grouping=None,
 ):
     """Validate BIDS paths.
+
+    Parameters
+    ----------
+
+    grouping : str, optional
+        A string which is either "", "error", or "path" — "error" implemented.
 
     Notes
     -----
@@ -47,18 +59,40 @@ def validate_bids(
         schema_version=schema,
     )
 
+    issues = []
     for i in validator_result:
         if not i.severity:
             continue
-        if i.path:
-            scope = i.path
-        elif i.path_regex:
-            scope = i.path_regex
         else:
-            scope = i.dataset_path
-        display_errors(scope, [i.id], [i.severity], [i.message])
+            issues.append(i)
 
-    validation_errors = [e for e in validator_result if e.severity == Severity.ERROR]
+    purviews = [
+        list(filter(bool, [i.path, i.path_regex, i.dataset_path]))[0] for i in issues
+    ]
+    if not grouping:
+        display_errors(
+            purviews,
+            [i.id for i in issues],
+            [i.severity for i in issues],
+            [i.message for i in issues],
+        )
+    elif grouping == "path":
+        for purview in purviews:
+            applies_to = [
+                i for i in issues if purview in [i.path, i.path_regex, i.dataset_path]
+            ]
+            display_errors(
+                [purview],
+                [i.id for i in applies_to],
+                [i.severity for i in applies_to],
+                [i.message for i in applies_to],
+            )
+    else:
+        raise NotImplementedError(
+            "The `grouping` parameter values currently supported are " "path or None."
+        )
+
+    validation_errors = [i for i in issues if i.severity == Severity.ERROR]
 
     if validation_errors:
         raise SystemExit(1)
@@ -75,10 +109,18 @@ def validate_bids(
 @click.argument("paths", nargs=-1, type=click.Path(exists=True, dir_okay=True))
 @devel_debug_option()
 @map_to_click_exceptions
-def validate(paths, schema=None, devel_debug=False, allow_any_path=False):
+def validate(
+    paths, schema=None, devel_debug=False, allow_any_path=False, grouping=None
+):
     """Validate files for NWB and DANDI compliance.
 
     Exits with non-0 exit code if any file is not compliant.
+
+    Parameters
+    ----------
+
+    grouping : str, optional
+        A string which is either "", "error", or "path" — "error" not yet implemented.
     """
     from ..pynwb_utils import ignore_benign_pynwb_warnings
     from ..validate import validate as validate_
@@ -96,42 +138,35 @@ def validate(paths, schema=None, devel_debug=False, allow_any_path=False):
     # at this point, so we ignore it although ideally there should be a formal
     # way to get relevant warnings (not errors) from PyNWB
     ignore_benign_pynwb_warnings()
-    view = "one-at-a-time"  # TODO: rename, add groupped
 
-    all_files_errors = {}
+    all_files_error_messages = {}
     nfiles = 0
-    for path, errors in validate_(
+    for path, messages in validate_(
         *paths,
         schema_version=schema,
         devel_debug=devel_debug,
         allow_any_path=allow_any_path,
     ):
         nfiles += 1
-        if view == "one-at-a-time":
-            if errors:
-                display_errors(
-                    path, ["NWBError"] * len(errors), [""] * len(errors), errors
-                )
-        all_files_errors[path] = errors
+        if not grouping:
+            error_paths = [path] * len(messages)
+            errors = ["NWBError"] * len(messages)
+            severities = [""] * len(messages)
+            display_errors(error_paths, errors, severities, messages)
+        elif grouping == "path":
+            error_paths = [path]
+            errors = ["NWBError"] * len(messages)
+            severities = [""] * len(messages)
+            display_errors(error_paths, errors, severities, messages)
+        else:
+            raise NotImplementedError(
+                "The `grouping` parameter values currently supported are "
+                "path or None."
+            )
 
-    if view == "groupped":
-        # TODO: Most likely we want to summarize errors across files since they
-        # are likely to be similar
-        # TODO: add our own criteria for validation (i.e. having needed metadata)
+        all_files_error_messages[path] = messages
 
-        # # can't be done since fails to compare different types of errors
-        # all_errors = sum(errors.values(), [])
-        # all_error_types = []
-        # errors_unique = sorted(set(all_errors))
-        # from collections import Counter
-        # # Let's make it
-        # print(
-        #     "{} unique errors in {} files".format(
-        #     len(errors_unique), len(errors))
-        # )
-        raise NotImplementedError("TODO")
-
-    files_with_errors = [f for f, errors in all_files_errors.items() if errors]
+    files_with_errors = [f for f, errors in all_files_error_messages.items() if errors]
 
     if files_with_errors:
         click.secho(
@@ -150,23 +185,71 @@ def validate(paths, schema=None, devel_debug=False, allow_any_path=False):
         )
 
 
-def display_errors(scope, errors, severities=[], messages=[]):
+def _get_severity_color(severities):
     from ..validate import Severity
 
     if Severity.ERROR in severities:
-        fg = "red"
+        return "red"
     elif Severity.WARNING in severities:
-        fg = "orange"
+        return "green"
     else:
-        fg = "blue"
-    summary = f"{scope}: {pluralize(len(errors), 'issue')} detected."
-    click.secho(summary, fg=fg)
-    for error, severity, message in zip(errors, severities, messages):
-        if severity == Severity.ERROR:
-            fg = "red"
-        elif severity == Severity.WARNING:
-            fg = "green"
-        else:
-            fg = "blue"
-        error_message = f"  [{error}] {message}"
+        return "blue"
+
+
+def display_errors(purviews, errors, severities, messages):
+    """
+    Unified error display for validation CLI, which auto-resolves grouping logic based on
+    the length of input lists.
+
+    Parameters
+    ----------
+    purviews: list of str
+    errors: list of str
+    severities: list of dandi.validate.Severity
+    messages: list of str
+
+
+    Notes
+    -----
+    * There is a bit of roundabout and currently untestable logic to deal with potential cases
+    where the same error has multiple error message, could be removed in the future and removed
+    by assert if this won't ever be the case.
+    """
+
+    if all([len(i) == 1 for i in [purviews, errors, severities, messages]]):
+        fg = _get_severity_color(severities)
+        error_message = f"[{errors[0]}] {purviews[0]} — {messages[0]}"
         click.secho(error_message, fg=fg)
+    elif len(purviews) == 1:
+        group_message = f"{purviews[0]}: {pluralize(len(errors), 'issue')} detected."
+        fg = _get_severity_color(severities)
+        click.secho(group_message, fg=fg)
+        for error, severity, message in zip(errors, severities, messages):
+            error_message = f"  [{error}] {message}"
+            fg = _get_severity_color([severity])
+            click.secho(error_message, fg=fg)
+    elif len(errors) == 1:
+        fg = _get_severity_color(severities)
+        group_message = (
+            f"{errors[0]}: detected in {pluralize(len(purviews), 'purviews')}"
+        )
+        if len(set(messages)) == 1:
+            error_message += f" — {messages[0]}."
+            click.secho(group_message, fg=fg)
+            for purview, severity in zip(purviews, severities):
+                error_message = f"  {purview}"
+                fg = _get_severity_color([severity])
+                click.secho(error_message, fg=fg)
+        else:
+            error_message += "."
+            for purview, severity, message in zip(purviews, severities, messages):
+                error_message = f"  {purview} — {message}"
+                fg = _get_severity_color([severity])
+                click.secho(error_message, fg=fg)
+    else:
+        for purview, error, severity, message in zip(
+            purviews, errors, severities, messages
+        ):
+            fg = _get_severity_color([severity])
+            error_message = f"[{error}] {purview} — {message}"
+            click.secho(error_message, fg=fg)
--- a/dandi/cli/cmd_validate.py
+++ b/dandi/cli/cmd_validate.py
@@ -3,7 +3,9 @@ import os
 
 import click
 
-from .base import devel_debug_option, devel_option, lgr, map_to_click_exceptions
+from .base import devel_debug_option, devel_option, map_to_click_exceptions
+from ..utils import pluralize
+from ..validate import Severity
 
 
 @click.command()
@@ -36,7 +38,6 @@ def validate_bids(
         dandi validate-bids /my/path
     """
 
-    from ..validate import Severity
     from ..validate import validate_bids as validate_bids_
 
     validator_result = validate_bids_(  # Controller
@@ -47,24 +48,15 @@ def validate_bids(
     )
 
     for i in validator_result:
+        if not i.severity:
+            continue
         if i.path:
             scope = i.path
         elif i.path_regex:
             scope = i.path_regex
         else:
             scope = i.dataset_path
-        echo_string = f"[{i.id}] {scope}: {i.message}"
-        i_fg = ""
-        if i.severity == Severity.ERROR:
-            i_fg = "red"
-        elif i.severity == Severity.WARNING:
-            i_fg = "orange"
-        if i_fg:
-            click.secho(
-                echo_string,
-                bold=True,
-                fg=i_fg,
-            )
+        display_errors(scope, [i.id], [i.severity], [i.message])
 
     validation_errors = [e for e in validator_result if e.severity == Severity.ERROR]
 
@@ -155,10 +147,21 @@ def validate(paths, schema=None, devel_debug=False, allow_any_path=False):
         )
 
 
-def display_errors(path, errors):
-    if not errors:
-        lgr.info("%s: ok", path)
+def display_errors(scope, errors, severities=[], messages=[]):
+    if Severity.ERROR in severities:
+        fg = "red"
+    elif Severity.WARNING in severities:
+        fg = "orange"
     else:
-        lgr.error("%s: %d error(s)", path, len(errors))
-        for error in errors:
-            lgr.error("  %s", error)
+        fg = "blue"
+    summary = f"{scope}: {pluralize(len(errors), 'issue')} detected."
+    click.secho(summary, bold=True, fg="red")
+    for error, severity, message in zip(errors, severities, messages):
+        if severity == Severity.ERROR:
+            fg = "red"
+        elif severity == Severity.WARNING:
+            fg = "green"
+        else:
+            fg = "blue"
+        error_message = f"  [{error}] {message}"
+        click.secho(error_message, fg=fg)
--- a/dandi/cli/cmd_validate.py
+++ b/dandi/cli/cmd_validate.py
@@ -75,12 +75,14 @@ def validate_bids(
     "--grouping",
     "-g",
     help="How to group error/warning reporting.",
+    type=click.Choice(['none', 'path'], case_sensitive=False),
+    default="none",
 )
 @click.argument("paths", nargs=-1, type=click.Path(exists=True, dir_okay=True))
 @devel_debug_option()
 @map_to_click_exceptions
 def validate(
-    paths, schema=None, devel_debug=False, allow_any_path=False, grouping=None
+    paths, schema=None, devel_debug=False, allow_any_path=False, grouping="none",
 ):
     """Validate files for NWB and DANDI compliance.
 
--- a/dandi/cli/cmd_validate.py
+++ b/dandi/cli/cmd_validate.py
@@ -191,7 +191,7 @@ def _get_severity_color(severities):
     if Severity.ERROR in severities:
         return "red"
     elif Severity.WARNING in severities:
-        return "green"
+        return "yellow"
     else:
         return "blue"
 
--- a/dandi/cli/cmd_validate.py
+++ b/dandi/cli/cmd_validate.py
@@ -108,7 +108,10 @@ def validate(paths, schema=None, devel_debug=False, allow_any_path=False):
     ):
         nfiles += 1
         if view == "one-at-a-time":
-            display_errors(path, errors)
+            if errors:
+                display_errors(
+                    path, ["NWBError"] * len(errors), [""] * len(errors), errors
+                )
         all_files_errors[path] = errors
 
     if view == "groupped":
@@ -155,7 +158,7 @@ def display_errors(scope, errors, severities=[], messages=[]):
     else:
         fg = "blue"
     summary = f"{scope}: {pluralize(len(errors), 'issue')} detected."
-    click.secho(summary, bold=True, fg="red")
+    click.secho(summary, fg=fg)
     for error, severity, message in zip(errors, severities, messages):
         if severity == Severity.ERROR:
             fg = "red"
--- a/dandi/cli/cmd_validate.py
+++ b/dandi/cli/cmd_validate.py
@@ -129,6 +129,7 @@ def validate(
     """
     from ..pynwb_utils import ignore_benign_pynwb_warnings
     from ..validate import validate as validate_
+    from ..validate_types import Severity
 
     # Don't log validation warnings, as this command reports them to the user
     # anyway:
@@ -144,50 +145,49 @@ def validate(
     # way to get relevant warnings (not errors) from PyNWB
     ignore_benign_pynwb_warnings()
 
-    all_files_error_messages = {}
-    nfiles = 0
-    for path, messages in validate_(
+    validator_result = validate_(
         *paths,
         schema_version=schema,
         devel_debug=devel_debug,
         allow_any_path=allow_any_path,
-    ):
-        nfiles += 1
-        if not grouping:
-            error_paths = [path] * len(messages)
-            errors = ["NWBError"] * len(messages)
-            severities = [""] * len(messages)
-            display_errors(error_paths, errors, severities, messages)
-        elif grouping == "path":
-            error_paths = [path]
-            errors = ["NWBError"] * len(messages)
-            severities = [""] * len(messages)
-            display_errors(error_paths, errors, severities, messages)
+    )
+    issues = []
+    for i in validator_result:
+        if not i.severity:
+            continue
         else:
-            raise NotImplementedError(
-                "The `grouping` parameter values currently supported are "
-                "path or None."
+            issues.append(i)
+
+    purviews = [
+        list(filter(bool, [i.path, i.path_regex, i.dataset_path]))[0] for i in issues
+    ]
+    if not grouping:
+        display_errors(
+            purviews,
+            [i.id for i in issues],
+            [i.severity for i in issues],
+            [i.message for i in issues],
+        )
+    elif grouping == "path":
+        for purview in purviews:
+            applies_to = [
+                i for i in issues if purview in [i.path, i.path_regex, i.dataset_path]
+            ]
+            display_errors(
+                [purview],
+                [i.id for i in applies_to],
+                [i.severity for i in applies_to],
+                [i.message for i in applies_to],
             )
-
-        all_files_error_messages[path] = messages
-
-    files_with_errors = [f for f, errors in all_files_error_messages.items() if errors]
-
-    if files_with_errors:
-        click.secho(
-            "Summary: Validation errors in {} out of {} files".format(
-                len(files_with_errors), nfiles
-            ),
-            bold=True,
-            fg="red",
-        )
-        raise SystemExit(1)
     else:
-        click.secho(
-            "Summary: No validation errors among {} file(s)".format(nfiles),
-            bold=True,
-            fg="green",
+        raise NotImplementedError(
+            "The `grouping` parameter values currently supported are " "path or None."
         )
+
+    validation_errors = [i for i in issues if i.severity == Severity.ERROR]
+
+    if validation_errors:
+        raise SystemExit(1)
 
 
 def _get_severity_color(severities):
--- a/dandi/cli/cmd_validate.py
+++ b/dandi/cli/cmd_validate.py
@@ -96,6 +96,8 @@ def validate_bids(
 
     if validation_errors:
         raise SystemExit(1)
+    else:
+        click.secho("No errors found.", fg="green")
 
 
 @click.command()
@@ -188,6 +190,8 @@ def validate(
 
     if validation_errors:
         raise SystemExit(1)
+    else:
+        click.secho("No errors found.", fg="green")
 
 
 def _get_severity_color(severities):
--- a/dandi/cli/cmd_validate.py
+++ b/dandi/cli/cmd_validate.py
@@ -52,7 +52,6 @@ def validate_bids(
     """
 
     from ..validate import validate_bids as validate_bids_
-    from ..validate_types import Severity
 
     validator_result = validate_bids_(  # Controller
         *paths,
@@ -61,40 +60,7 @@ def validate_bids(
         schema_version=schema,
     )
 
-    issues = [i for i in validator_result if i.severity]
-
-    purviews = [
-        list(filter(bool, [i.path, i.path_regex, i.dataset_path]))[0] for i in issues
-    ]
-    if grouping == "none":
-        display_errors(
-            purviews,
-            [i.id for i in issues],
-            [i.severity for i in issues],
-            [i.message for i in issues],
-        )
-    elif grouping == "path":
-        for purview in purviews:
-            applies_to = [
-                i for i in issues if purview in [i.path, i.path_regex, i.dataset_path]
-            ]
-            display_errors(
-                [purview],
-                [i.id for i in applies_to],
-                [i.severity for i in applies_to],
-                [i.message for i in applies_to],
-            )
-    else:
-        raise NotImplementedError(
-            "The `grouping` parameter values currently supported are " "path or None."
-        )
-
-    validation_errors = [i for i in issues if i.severity == Severity.ERROR]
-
-    if validation_errors:
-        raise SystemExit(1)
-    else:
-        click.secho("No errors found.", fg="green")
+    _process_issues(validator_result, grouping)
 
 
 @click.command()
@@ -128,7 +94,6 @@ def validate(
     """
     from ..pynwb_utils import ignore_benign_pynwb_warnings
     from ..validate import validate as validate_
-    from ..validate_types import Severity
 
     # Don't log validation warnings, as this command reports them to the user
     # anyway:
@@ -150,12 +115,18 @@ def validate(
         devel_debug=devel_debug,
         allow_any_path=allow_any_path,
     )
+
+    _process_issues(validator_result, grouping)
+
+
+def _process_issues(validator_result, grouping):
+    from ..validate_types import Severity
     issues = [i for i in validator_result if i.severity]
 
     purviews = [
         list(filter(bool, [i.path, i.path_regex, i.dataset_path]))[0] for i in issues
     ]
-    if not grouping:
+    if grouping == "none":
         display_errors(
             purviews,
             [i.id for i in issues],
--- a/dandi/cli/cmd_validate.py
+++ b/dandi/cli/cmd_validate.py
@@ -49,8 +49,8 @@ def validate_bids(
         dandi validate-bids /my/path
     """
 
-    from ..validate import Severity
     from ..validate import validate_bids as validate_bids_
+    from ..validate_types import Severity
 
     validator_result = validate_bids_(  # Controller
         *paths,
@@ -186,7 +186,7 @@ def validate(
 
 
 def _get_severity_color(severities):
-    from ..validate import Severity
+    from ..validate_types import Severity
 
     if Severity.ERROR in severities:
         return "red"
@@ -205,7 +205,7 @@ def display_errors(purviews, errors, severities, messages):
     ----------
     purviews: list of str
     errors: list of str
-    severities: list of dandi.validate.Severity
+    severities: list of dandi.validate_types.Severity
     messages: list of str
 
 
--- a/dandi/cli/cmd_validate.py
+++ b/dandi/cli/cmd_validate.py
@@ -1,11 +1,12 @@
 import logging
 import os
-from typing import List
+from typing import List, cast
 
 import click
 
 from .base import devel_debug_option, devel_option, map_to_click_exceptions
 from ..utils import pluralize
+from ..validate_types import Severity
 
 
 @click.command()
@@ -39,12 +40,6 @@ def validate_bids(
     grouping="none",
 ):
     """Validate BIDS paths.
-
-    Parameters
-    ----------
-
-    grouping : str, optional
-        A string which is either "", "error", or "path" — "error" implemented.
 
     Notes
     -----
@@ -127,7 +122,6 @@ def validate(
 
 
 def _process_issues(validator_result, grouping):
-    from ..validate_types import Severity
 
     issues = [i for i in validator_result if i.severity]
 
@@ -166,7 +160,6 @@ def _process_issues(validator_result, grouping):
 
 
 def _get_severity_color(severities):
-    from ..validate_types import Severity
 
     if Severity.ERROR in severities:
         return "red"
@@ -179,19 +172,12 @@ def _get_severity_color(severities):
 def display_errors(
     purviews: List[str],
     errors: List[str],
-    severities: List[str],
+    severities: List[Severity],
     messages: List[str],
 ) -> None:
     """
     Unified error display for validation CLI, which auto-resolves grouping logic based on
     the length of input lists.
-
-    Parameters
-    ----------
-    purviews: list of str
-    errors: list of str
-    severities: list of dandi.validate_types.Severity
-    messages: list of str
 
 
     Notes
@@ -201,7 +187,7 @@ def display_errors(
     by assert if this won't ever be the case.
     """
 
-    if all(len(i) == 1 for i in [purviews, errors, severities, messages]):
+    if all(len(cast(list, i)) == 1 for i in [purviews, errors, severities, messages]):
         fg = _get_severity_color(severities)
         error_message = f"[{errors[0]}] {purviews[0]} — {messages[0]}"
         click.secho(error_message, fg=fg)
--- a/dandi/cli/cmd_validate.py
+++ b/dandi/cli/cmd_validate.py
@@ -24,7 +24,7 @@ from ..utils import pluralize
 @click.option(
     "--grouping",
     "-g",
-    help="Write report under path, this option implies `--report/-r`.",
+    help="How to group error/warning reporting.",
 )
 @click.argument("paths", nargs=-1, type=click.Path(exists=True, dir_okay=True))
 @map_to_click_exceptions
--- a/dandi/cli/cmd_validate.py
+++ b/dandi/cli/cmd_validate.py
@@ -25,6 +25,8 @@ from ..utils import pluralize
     "--grouping",
     "-g",
     help="How to group error/warning reporting.",
+    type=click.Choice(['none', 'path'], case_sensitive=False),
+    default="none",
 )
 @click.argument("paths", nargs=-1, type=click.Path(exists=True, dir_okay=True))
 @map_to_click_exceptions
@@ -33,7 +35,7 @@ def validate_bids(
     schema,
     report,
     report_path,
-    grouping=None,
+    grouping="none",
 ):
     """Validate BIDS paths.
 
@@ -69,7 +71,7 @@ def validate_bids(
     purviews = [
         list(filter(bool, [i.path, i.path_regex, i.dataset_path]))[0] for i in issues
     ]
-    if not grouping:
+    if grouping == "none":
         display_errors(
             purviews,
             [i.id for i in issues],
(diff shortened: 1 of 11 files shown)


dandi/validate.py
```diff
@@ -1,11 +1,42 @@
...
+@dataclass
+class ValidationOrigin:
+    name: str
+    version: str
+
+
+class Severity(Enum):
+    ERROR = "ERROR"
+    WARNING = "WARNING"
+    HINT = "HINT"
```
CodyCBakerPhD, line 34: Interesting, what are some examples you can think of for `HINT`s? Are these asset-level, like minor style suggestions for NWBFile metadata?

yarikoptic: For nwb within BIDS dataset might potentially be something like "place different modalities into separate .nwb files". I could come up with others later ;)

dandi/files/bids.py
```diff
@@ -66,25 +66,29 @@ def _validate(self) -> None:
...
                 ]
+                # TODO gh-943: use RFed data structures, avoid duplicating logic
                 results = validate_bids(*bids_paths)
                 self._dataset_errors: list[str] = []
-                if len(results["path_listing"]) == len(results["path_tracking"]):
-                    self._dataset_errors.append("No valid BIDS files were found")
-                for entry in results["schema_tracking"]:
-                    if entry["mandatory"]:
+                self._asset_errors = defaultdict(list)
+                for i in results:
```
yarikoptic, line 73: please avoid such overly too generic `i` since you inadvertingly would run into overloading it. Make it at least `r` for (`r`esults) or `result`

TheChymera:
As of we no longer need our fork of bids-examples https://github.com/dandi/bids-examples

N.B. me delete after this is merged.

TheChymera:
@jwodder could you review this?

dandi/validate.py
```diff
@@ -1,19 +1,59 @@
...
+    id: str
+    origin: ValidationOrigin
+    scope: Scope
+    asset_paths: Optional[list[str]] = None
+    dandiset_path: Optional[Path] = None
+    dataset_path: Optional[Path] = None
+    message: str = ""
+    metadata: Optional[dict] = None
+    path: Optional[Path] = None
+    path_regex: Optional[str] = ""
```
jwodder, line 29:
```suggestion
    path_regex: Optional[str] = None
```

dandi/validate.py
```diff
@@ -1,19 +1,59 @@
...
+
+@dataclass
+class ValidationResult:
+    id: str
+    origin: ValidationOrigin
+    scope: Scope
+    asset_paths: Optional[list[str]] = None
+    dandiset_path: Optional[Path] = None
+    dataset_path: Optional[Path] = None
+    message: str = ""
```
jwodder, line 26: Remove the default value and move the field above `asset_paths`.

dandi/validate.py
```diff
@@ -44,12 +90,103 @@ def validate_bids(
...
+                dataset_path=dataset_path,
+                dandiset_path=dandiset_path,
+            )
+        )
+
+    for pattern in validation_result["schema_tracking"]:
+        # Future proofing for standard-compliant name.
+        if ("mandatory" in pattern and pattern["mandatory"]) or (
+            "required" in pattern and pattern["required"]
+        ):
```
jwodder, line 129:
```suggestion
        if pattern.get("mandatory") or pattern.get("required"):
```

dandi/validate.py
```diff
@@ -44,12 +90,103 @@ def validate_bids(
...
+    in_path : str, optional
+        File for which to determine set path.
+    marker : str, optional
+        Filename marker which identifies the set root.
+    """
+    import os
+
+    candidate = os.path.join(in_path, marker)
+    # Windows support... otherwise we could do `if in_path == "/"`.
+    if in_path == "/" or not any(i in in_path for i in ["/", "\\"]):
```
jwodder, line 183: Assuming the goal of this line is to test whether `in_path` is the root of a drive/filesystem/whatever, I suspect a more accurate test (that won't fail on relative paths) can be constructed using `pathlib` to test whether the path equals [`Path.root`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.root), [`Path.anchor`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.anchor), or some combination thereof.

dandi/files/bids.py
```diff
@@ -15,10 +15,12 @@
 from ..metadata import add_common_metadata, prepare_metadata
 from ..misctypes import Digest
 
-BIDS_TO_DANDI = {
-    "subject": "subject_id",
-    "session": "session_id",
-}
+BIDS_ASSET_ERRORS = [
```
jwodder, line 18: If `BIDS_ASSET_ERRORS` and `BIDS_DATASET_ERRORS` are only going to be used on the right side of `in`, they would be better as sets.

dandi/tests/fixtures.py
```diff
@@ -221,8 +233,56 @@ def fixture() -> Iterator[str]:
...
+    @pytest.fixture(scope="session")
+    def fixture() -> Iterator[str]:
+        skipif.no_network()
+        skipif.no_git()
+        path = tempfile.mktemp()  # not using pytest's tmpdir fixture to not
+        # collide in different scopes etc. But we
+        # would need to remove it ourselves
+        try:
+            lgr.debug("Cloning %r into %r", url, path)
+            runout = run(
```
jwodder, line 249: Add `check=True` to the `run()` call and skip the check of `runout.returncode`.  I also don't see a reason to capture the output.

dandi/tests/fixtures.py
```diff
@@ -221,8 +233,56 @@ def fixture() -> Iterator[str]:
...
+                ],
+                capture_output=True,
+            )
+            if runout.returncode:
+                raise RuntimeError(f"Failed to clone {url} into {path}")
+            # cwd specification is VERY important, not only to achieve the correct
+            # effects, but also to avoid dropping files from your repository if you
+            # were to run `git sparse-checkout` inside the software repo.
+            _ = run(["git", "sparse-checkout", "init", "--cone"], cwd=path)
+            _ = run(["git", "sparse-checkout", "set"] + whitelist, cwd=path)
```
jwodder, line 267:
```suggestion
            run(["git", "sparse-checkout", "init", "--cone"], cwd=path, check=True)
            run(["git", "sparse-checkout", "set"] + whitelist, cwd=path, check=True)
```

dandi/tests/fixtures.py
```diff
@@ -221,8 +233,56 @@ def fixture() -> Iterator[str]:
...
+            yield path
+        finally:
+            try:
+                shutil.rmtree(path)
+            except BaseException as exc:
```
jwodder, line 272: DO NOT CATCH `BaseException`!  That catches keyboard interrupts and other things that should only be caught under very special circumstances.  Catch just `Exception` instead.

dandi/tests/fixtures.py
```diff
@@ -221,8 +233,56 @@ def fixture() -> Iterator[str]:
...
+def get_filtered_gitrepo_fixture(
+    url: str,
+    whitelist: List[str],
+) -> Callable[[], Iterator[str]]:
+    @pytest.fixture(scope="session")
+    def fixture() -> Iterator[str]:
+        skipif.no_network()
+        skipif.no_git()
+        path = tempfile.mktemp()  # not using pytest's tmpdir fixture to not
+        # collide in different scopes etc. But we
```
jwodder, line 245: Please explain this "to not collide in different scopes" thing.  Is the issue that `tmp_path` is not available when `scope="session"`?  That's when you should use [`tmp_path_factory`](https://docs.pytest.org/en/7.1.x/reference/reference.html#std-fixture-tmp_path_factory).  Is the issue that you think calling `tmp_path_factory.mktemp()` with the same arguments will return the same path?  It doesn't; each call by default creates a new directory suffixed with an incrementing counter.

dandi/tests/test_validate.py
```diff
@@ -1,35 +1,62 @@
 from glob import glob
 import os
 
-import appdirs
+# import appdirs
```
jwodder, line 4: Just delete this line.

dandi/tests/test_validate.py
```diff
@@ -1,35 +1,62 @@
...
+@pytest.mark.parametrize(
+    "dataset", ["invalid_asl003", "invalid_eeg_cbm", "invalid_pet001"]
+)
+def test_validate_bids_errors(bids_error_examples, dataset):
+    # This only checks that the error we found is correct, not that we found all errors.
+    # ideally make a list and erode etc.
+    import json
+    import pathlib
+
+    from ..validate import validate_bids
```
jwodder, line 43: I see no reason to place these imports inside the function instead of at file scope.

dandi/validate.py
TheChymera: Message is only defined for errors, not for matches, so it needs to have a default value, or am I missing something?
We could put some generic placeholder message but “File identified” over and over is just wasted memory.

jwodder: Then it would probably be better to change the type to `Optional[str]` and set the default to `None` for representing message-less results.

dandi/validate.py
TheChymera: That is indeed the intent, but paths get expanded in any case. Also, not sure what I'd set the root to other than, “when there's nothing left except `"/"`”.

dandi/tests/fixtures.py
TheChymera: Hm, ok, I tried to use tmp_path_factory, but it seems to have trouble dealing with windows:

```
2022-09-20T18:21:58.4183372Z ____________ ERROR at setup of test_upload_bids_validation_ignore _____________
2022-09-20T18:21:58.4183687Z dandi\tests\fixtures.py:259: in fixture
2022-09-20T18:21:58.4183940Z     capture_output=True,
2022-09-20T18:21:58.4184272Z C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\subprocess.py:488: in run
2022-09-20T18:21:58.4184625Z     with Popen(*popenargs, **kwargs) as process:
2022-09-20T18:21:58.4184993Z C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\subprocess.py:800: in __init__
2022-09-20T18:21:58.4185328Z     restore_signals, start_new_session)
2022-09-20T18:21:58.4185690Z C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\subprocess.py:1148: in _execute_child
2022-09-20T18:21:58.4186216Z     args = list2cmdline(args)
2022-09-20T18:21:58.4186549Z C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\subprocess.py:555: in list2cmdline
2022-09-20T18:21:58.4186915Z     needquote = (" " in arg) or ("\t" in arg) or not arg
2022-09-20T18:21:58.4187284Z E   TypeError: argument of type 'WindowsPath' is not iterable
```

As for the original code and comment, I used it to make the function analogous to `get_gitrepo_fixture()` — it's basically the same thing other than doing some git magic to really only ever download the whitelist. I think the original code might be from @yarikoptic

jwodder: I believe the problem is that `tmp_path_factory` returns a `pathlib.Path` object, yet `subprocess.run()` does not accept `Path`s in command arguments prior to Python 3.8.  You need to change `path` in the "clone" call to `str(path)`.  (Also, get rid of the `try` ... `finally` and its cleanup code, as pytest manages its temp dirs itself.)

TheChymera:
@jwodder other than that one comment I'm not sure how/whether to address, should be done now. Thank you for your help! 💛

dandi/tests/fixtures.py
jwodder: @TheChymera You missed this comment.

dandi/validate.py
yarikoptic: as I have provided an example, I will consider this one resolved ;)

dandi/cli/cmd_validate.py
```diff
@@ -36,19 +36,40 @@ def validate_bids(
...
+        else:
+            scope = i.dataset_path
+        echo_string = f"[{i.id}] {scope}: {i.message}"
+        if i.severity == Severity.ERROR:
+            click.secho(
+                echo_string,
+                bold=True,
+                fg="red",
+            )
+        if i.severity == Severity.WARNING:
```
yarikoptic, line 63: at least
```suggestion
        elif i.severity == Severity.WARNING:
```
for clarity.  In general, I would have even folded this into a single `.secho` with just changing color based on severity value to avoid duplication, but since such a minor -- not worth RFing for that

dandi/cli/cmd_validate.py
```diff
@@ -36,19 +36,40 @@ def validate_bids(
...
         *paths,
         report=report,
         report_path=report_path,
         schema_version=schema,
     )
-    valid = is_valid(validator_result)
-    report_errors(validator_result)
 
-    if not valid:
+    for i in validator_result:
```
yarikoptic, line 49: note: such "view" seems to be getting closer to our `validate` (each message separately) instead of what was in https://github.com/dandi/dandi-cli/pull/1104/files#diff-79d7645e39caaf3d1ca18c9ed87f99a1696696fa67da3464983fc07d6c974325L41 or how bids-validator presents groupped by type of error.  I guess it is ok, and we will provide alternative renderings later. But best to right away move this into some function e.g. `echo_individual_errors` or alike.

dandi/validate.py
```diff
@@ -44,12 +90,101 @@ def validate_bids(
...
+    """Get the path to the root of a file set (e.g. BIDS dataset or DANDIset).
+
+    Parameters
+    ----------
+    in_path : str, optional
+        File for which to determine set path.
+    marker : str, optional
+        Filename marker which identifies the set root.
+    """
+    import os
```
yarikoptic, line 177: in general please define such generic import to top of the file

dandi/validate.py
```diff
@@ -44,12 +90,101 @@ def validate_bids(
...
+                metadata=meta,
+                dataset_path=dataset_path,
+                dandiset_path=dandiset_path,
+            )
+        )
+
+    return our_validation_result
+
+
+def _get_set_path(in_path, marker):
```
yarikoptic, line 167: I had said before that we have already a function for such purposes and it should be used instead of reinventing the wheel!  Just use `find_parent_directory_containing` which is currently at https://github.com/dandi/dandi-cli/blob/HEAD/dandi/utils.py#L392

dandi/validate.py
```diff
@@ -44,12 +90,101 @@ def validate_bids(
...
+        version=bidsschematools.__version__,
+    )
+
+    for path in validation_result["path_tracking"]:
+        # Hard-coding exclusion here pending feature + release in:
+        # https://github.com/bids-standard/bids-specification/issues/1272
+        if path.endswith((".ERRORS", ".ERRORS.json")):
+            continue
+        dataset_path = _get_set_path(path, "dataset_description.json")
+        dandiset_path = _get_set_path(path, "dandiset.yaml")
```
yarikoptic, line 111: note that doing that for **every** path could be avoidably expensive and the same folder cannot belong to different datasets or dandisets.  Get path's folder (if not a folder), and cache results.  but see above to re-use existing `find_parent_directory_containing`

dandi/validate.py
```diff
@@ -44,12 +90,101 @@ def validate_bids(
...
+        file_path = meta.pop("path")
+        meta = {BIDS_TO_DANDI[k]: v for k, v in meta.items() if k in BIDS_TO_DANDI}
+        dataset_path = _get_set_path(file_path, "dataset_description.json")
+        dandiset_path = _get_set_path(file_path, "dandiset.yaml")
+        # Top level files do not have any other metadata other than path,
+        # which we pop and put in the object...
+        if not meta:
+            meta = {
+                "subject": None,
+            }
```
yarikoptic, line 151: is `subject` mandatory? why?  why it cannot stay an empty `{}`?

yarikoptic (changes requested):
- avoid duplicating existing code
- remove logging warnings as still there 
```
2022-09-26 13:48:28,920 [ WARNING] The `/home/yoh/datalad/openneuro/ds000001/.datalad/config` file was not matched by any regex schema entry.
[BIDS.NON_BIDS_PATH_PLACEHOLDER] /home/yoh/datalad/openneuro/ds000001/.datalad/config: File does not match any pattern known to BIDS.
```

yarikoptic:
also - you did it only for BIDS, right? we need to make it also for `validate` and thus provide adapter for nwbinspector call at https://github.com/dandi/dandi-cli/blob/HEAD/dandi/files/bases.py#L505

CodyCBakerPhD:
Hey all, looks interesting! Thought I'd give my two cents on the question @yarikoptic just raised...

> also - you did it only for BIDS, right? we need to make it also for validate and thus provide adapter for nwbinspector call at https://github.com/dandi/dandi-cli/blob/HEAD/dandi/files/bases.py#L505

For reconciling between https://github.com/dandi/dandi-cli/pull/1104/files#diff-5625da161c31ad0a0f74a12c0183cf444f1b945bfa10a7b7487d92859eaca720R18-R30 <-> and <-> https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/nwbinspector/register_checks.py#L38-L78, here's my take...

```
ValidationResult <-> InspectorMessage

id = ?  # not sure what this is
origin = ?  # DANDI inspection runs iteratively over a list of nwbfile_paths, so IDK how you want to classify that
scope = Scope.FILE  # There's technically one folder-level thing but it gets registered at file level

asset_paths = [file_path]  # only have one per InspectorMessage
dandiset_path = file_path.parent.parent  # Assuming multiple sessions per multiple subjects, otherwise nesting level might differ
dataset_path = None  # Not clear how this differs from dandiset_path

message = message
metadata = None  # Not sure what you intend for this

path = location  # assuming this is HDF5 Group information within the file
path_regex = None

importance_to_dandi_severity_map = {
    ERROR="ERROR",
    PYNWB_VALIDATION="ERROR",
    CRITICAL="ERROR",  # when using --config dandi
    BEST_PRACTICE_VIOLATION="WARNING",
    BEST_PRACTICE_SUGGESTION="HINT",
}
severity = importance_to_dandi_severity_map[importance]  # unfortunate name collision; the InspectorMessage.severity is used only for minor ordering of final report - will have to remove the `importance_treshold` currently used to filter DANDI-level CRITICAL evaluations as errors vs. validation results
```

Note: I'd strongly recommend some nice, robust docstrings describing the new data classes

Feel free to schedule a time on my [calendly](https://calendly.com/cody-c-baker) if you'd like to discuss in more detail

yarikoptic:
Thank you @CodyCBakerPhD !  Re some items you were not certain
- `id` - I think it would be smth like `nwbi.{name of the check/function which gave rise to the message}` so smth like `nwbi.check_experimenter_form` (where I would use `nwbi` to shorten `nwbinspector` and not to confuse with `nwb` itself).  It is the one for which later , if decide to group by id across files, we could see where that particular problem/check fails.
- `origin` - is for provenance of what tool produced that message, so `ValidationOrigin('nwbinspector, version=nwbinspector.__version__)` or alike.

dandi/cli/cmd_validate.py
TheChymera: [48de132](https://github.com/dandi/dandi-cli/pull/1104/commits/48de132d361d3da53b1b1775b3721a1f1d61e92b)

dandi/validate.py
TheChymera: 3acc2c93cda314f6326985999b2db6f8786de47e

CodyCBakerPhD:
> Thank you @CodyCBakerPhD ! Re some items you were not certain
> 
> id - I think it would be smth like nwbi.{name of the check/function which gave rise to the message} so smth like nwbi.check_experimenter_form (where I would use nwbi to shorten nwbinspector and not to confuse with nwb itself). It is the one for which later , if decide to group by id across files, we could see where that particular problem/check fails.
origin - is for provenance of what tool produced that message, so ValidationOrigin('nwbinspector, version=nwbinspector.__version__) or alike.

Perfect, thanks - I updated the post with that info!

dandi/validate.py
TheChymera: 290d1ed5df72f9de393f4c7390dee195a395a7fa

dandi/validate.py
TheChymera: I'm pretty sure this used to fail somehow, but I can't reproduce the failure. Maybe something that got sorted out by the `assert`s. In any case, dealt with: b76b8bbbd94dfa3ebfc969ed08482369bf55d28b

dandi/validate.py
TheChymera: ultimately not needed, actually.

b76b8bbbd94dfa3ebfc969ed08482369bf55d28b

dandi/validate.py
TheChymera: Ok, no, it is required, it's explained in the above comment. Some files only have path in `meta` and that gets popped since we store the path separately, meaning that `assert` evaluates it as false making it look like an invalid entry.

```
2022-09-26T21:25:58.7861309Z =================================== FAILURES ===================================
2022-09-26T21:25:58.7861775Z ___________________________ test_upload_bids_invalid ___________________________
2022-09-26T21:25:58.7862256Z dandi/tests/test_upload.py:188: in test_upload_bids_invalid
2022-09-26T21:25:58.7862710Z     bids_dandiset_invalid.upload(existing="force")
2022-09-26T21:25:58.7863110Z dandi/tests/fixtures.py:429: in upload
2022-09-26T21:25:58.7863499Z     **{**self.upload_kwargs, **kwargs},
2022-09-26T21:25:58.7863869Z dandi/upload.py:301: in upload
2022-09-26T21:25:58.7864249Z     for v in process_path(dfile):
2022-09-26T21:25:58.7864611Z dandi/upload.py:144: in process_path
2022-09-26T21:25:58.7865035Z     validation_errors = dfile.get_validation_errors()
2022-09-26T21:25:58.7865471Z dandi/files/bids.py:233: in get_validation_errors
2022-09-26T21:25:58.7865889Z     ) + BIDSAsset.get_validation_errors(self)
2022-09-26T21:25:58.7866304Z dandi/files/bids.py:158: in get_validation_errors
2022-09-26T21:25:58.7866738Z     return self.bids_dataset_description.get_asset_errors(self)
2022-09-26T21:25:58.7867175Z dandi/files/bids.py:89: in get_asset_errors
2022-09-26T21:25:58.7867537Z     self._validate()
2022-09-26T21:25:58.7867890Z dandi/files/bids.py:82: in _validate
2022-09-26T21:25:58.7868255Z     assert result.metadata
2022-09-26T21:25:58.7868580Z E   AssertionError
```

dandi/validate.py
TheChymera: 354a99255cc527e48606a1ad41f7ce4a8bb2cbed

dandi/validate.py
```diff
@@ -44,12 +92,89 @@ def validate_bids(
...
+
+    # Storing variable to not re-compute set paths for each individual file.
+    parent_path = None
+    for meta in validation_result["match_listing"]:
+        file_path = meta.pop("path")
+        meta = {BIDS_TO_DANDI[k]: v for k, v in meta.items() if k in BIDS_TO_DANDI}
+        if parent_path != os.path.dirname(file_path):
+            parent_path = os.path.dirname(file_path)
+            dataset_path = find_parent_directory_containing("dataset_description.json", parent_path)
+            dandiset_path = find_parent_directory_containing("dandiset.yaml", parent_path)
```
yarikoptic, line 156: pay attention to LGTM (although soon it will die). You overload these two variables in the next two lines below.  So which should it be -- should there be `else:` or just these two removed?

dandi/validate.py
yarikoptic: well, without deeper analysis I cannot really judge on why we have that `assert` i.e. why we have assumption that metadata should have something!  may be it was supposed to be `assert result.metadata is not None` or smth like that?

TheChymera: you're right aeb6d23ebbf1580e1c6ab8435431b8c92dfa932e

dandi/cli/cmd_validate.py
TheChymera: should I extend and reuse e.g. `display_errors()` (in the same file)? It uses the logger instead of click.

yarikoptic: yeah - RF to reuse `display_errors()` for consistency with `validate`.

Next step after this PR would be to merry `validate` and `validate_bids` - we should have a single `validate` and there we will expose some option to switch between different "views" over errors.

TheChymera:
Not sure if this addresses the actual issue that `dandi/cli/tests/test_command.py::test_no_heavy_imports` tries to prevent or just works around the way in which the test is implemented https://github.com/dandi/dandi-cli/pull/1104/commits/8785deb59aba004d8fdabbc9fdc21e0bbefda5fd but otherwise should be fine.

yarikoptic:
FTR -- that test was failing with

```
dandi/cli/tests/test_command.py:61: in test_no_heavy_imports
    assert not loaded_heavy
E   AssertionError: assert not {'h5py', 'numpy', 'pynwb'}
```

and the purpose is that CLI (without invoking a command) should not cause import of heavy imports which then would delay trivial `--help` and other operations.  To quickly see how we get the problem (on commit before the fix for it):

```
> mkdir /tmp/badimports; echo 'raise ValueError' > /tmp/badimports/h5py.py
> PYTHONPATH=/tmp/badimports dandi organize --help 
  File "/home/yoh/proj/dandi/dandi-cli-master/venvs/dev3/bin/dandi", line 5, in <module>
    from dandi.cli.command import main
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/cli/command.py", line 151, in <module>
    from .cmd_validate import validate, validate_bids  # noqa: E402
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/cli/cmd_validate.py", line 8, in <module>
    from ..validate import Severity
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/validate.py", line 11, in <module>
    from .files import find_dandi_files
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/files/__init__.py", line 23, in <module>
    from ._private import BIDSFileFactory, DandiFileFactory
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/files/_private.py", line 15, in <module>
    from .bases import DandiFile, GenericAsset, LocalAsset, NWBAsset, VideoAsset
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/files/bases.py", line 21, in <module>
    from nwbinspector import Importance, inspect_nwb, load_config
  File "/home/yoh/proj/dandi/dandi-cli-master/venvs/dev3/lib/python3.10/site-packages/nwbinspector/__init__.py", line 1, in <module>
    from .register_checks import available_checks, Importance
  File "/home/yoh/proj/dandi/dandi-cli-master/venvs/dev3/lib/python3.10/site-packages/nwbinspector/register_checks.py", line 8, in <module>
    import h5py
  File "/tmp/badimports/h5py.py", line 1, in <module>
    raise ValueError
ValueError
```
so the change you did sounds good to mitigate it.  I thought to check an

<details>
<summary>alternative of delaying nwbinspector imports</summary> 

```shell
diff --git a/dandi/files/bases.py b/dandi/files/bases.py
index d43bf41..338e48e 100644
--- a/dandi/files/bases.py
+++ b/dandi/files/bases.py
@@ -18,8 +18,6 @@ from dandischema.models import BareAsset, CommonModel
 from dandischema.models import Dandiset as DandisetMeta
 from dandischema.models import get_schema_version
 from etelemetry import get_project
-from nwbinspector import Importance, inspect_nwb, load_config
-from nwbinspector.utils import get_package_version
 from packaging.version import Version
 from pydantic import ValidationError
 import requests
@@ -462,6 +460,9 @@ class NWBAsset(LocalFileAsset):
         devel_debug: bool = False,
     ) -> list[str]:
+        # delay import since it is heavy
+        from nwbinspector import Importance, inspect_nwb, load_config
+        from nwbinspector.utils import get_package_version
         errors: list[str] = pynwb_validate(self.filepath, devel_debug=devel_debug)
         if schema_version is not None:
             errors.extend(

```
</details>

<details>
<summary>but it is not sufficient</summary> 

```shell
$> PYTHONPATH=/tmp/badimports dandi organize --help                        
  File "/home/yoh/proj/dandi/dandi-cli-master/venvs/dev3/bin/dandi", line 5, in <module>
    from dandi.cli.command import main
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/cli/command.py", line 151, in <module>
    from .cmd_validate import validate, validate_bids  # noqa: E402
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/cli/cmd_validate.py", line 8, in <module>
    from ..validate import Severity
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/validate.py", line 11, in <module>
    from .files import find_dandi_files
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/files/__init__.py", line 23, in <module>
    from ._private import BIDSFileFactory, DandiFileFactory
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/files/_private.py", line 15, in <module>
    from .bases import DandiFile, GenericAsset, LocalAsset, NWBAsset, VideoAsset
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/files/bases.py", line 27, in <module>
    from dandi.metadata import get_default_metadata, nwb2asset
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/metadata.py", line 33, in <module>
    from .pynwb_utils import (
  File "/home/yoh/proj/dandi/dandi-cli-master/dandi/pynwb_utils.py", line 11, in <module>
    import h5py
  File "/tmp/badimports/h5py.py", line 1, in <module>
    raise ValueError
ValueError

```
</details>

TheChymera:
@yarikoptic yep, that's what I was also trying to do, but realized this might be just a question of *delaying* imports and wholesale preventing them for the CLI functions — because in the end nwbinspector is used by the validator for nwb files.... so unless the upstream library can validate without heavy libs (like BST does without pandas), probably we can't remove this entirely...

Anyway, is this PR complete now?

yarikoptic:
> Anyway, is this PR complete now?

have you added adapting nwbinspector outputs into our records as well per [discussion above](https://github.com/dandi/dandi-cli/pull/1104#issuecomment-1258415009)?

dandi/files/bases.py
```diff
@@ -185,7 +199,23 @@ def get_validation_errors(
                     e,
                     extra={"validating": True},
                 )
-                return [str(e)]
+                # TODO: how do we get **all** errors from validation - there must be a way
+                return [
+                    ValidationResult(
```
yarikoptic, line 204: create a helper `pydantic_errors_to_validation_results(errors: list[dict]) -> list[ValidationResult]` which you would pass `e.errors` here and in the other location above on validation of the dandiset level metadata.

Each error seems to be a dict which looks like

```
*(Pdb) pprint(e.errors()[-1])
{'loc': ('version',), 'msg': 'field required', 'type': 'value_error.missing'}
*(Pdb) pprint(e.errors())
[{'ctx': {'pattern': '^(dandi|DANDI):\\d{6}(/(draft|\\d+\\.\\d+\\.\\d+))$'},
  'loc': ('id',),
  'msg': 'string does not match regex '
         '"^(dandi|DANDI):\\d{6}(/(draft|\\d+\\.\\d+\\.\\d+))$"',
  'type': 'value_error.str.regex'},
 {'loc': ('name',), 'msg': 'field required', 'type': 'value_error.missing'},
 {'loc': ('description',),
  'msg': 'field required',
  'type': 'value_error.missing'},
 {'loc': ('contributor',),
  'msg': 'field required',
  'type': 'value_error.missing'},
 {'loc': ('license',), 'msg': 'field required', 'type': 'value_error.missing'},
 {'loc': ('identifier',),
  'msg': 'field required',
  'type': 'value_error.missing'},
 {'loc': ('citation',), 'msg': 'field required', 'type': 'value_error.missing'},
 {'loc': ('assetsSummary',),
  'msg': 'field required',
  'type': 'value_error.missing'},
 {'loc': ('manifestLocation',),
  'msg': 'field required',
  'type': 'value_error.missing'},
 {'loc': ('version',), 'msg': 'field required', 'type': 'value_error.missing'}]

```

could look like following for the `id`

```
(Pdb) dict(id=':'.join(filter(bool, ("dandischema", e0.get("type", "UNKNOWN"), '+'.join(e0.get("loc", []))))))
{'id': 'dandischema:value_error.str.regex:id'}
(Pdb) 
{'id': 'dandischema:value_error.str.regex:id'}
```

for the message I guess just take `msg` if it is there.

jwodder:
@TheChymera You still have not addressed https://github.com/dandi/dandi-cli/pull/1104#discussion_r974239796

dandi/tests/fixtures.py
TheChymera: sry, missed that. Should be solved now.

TheChymera:
@jwodder how about now?

dandi/pynwb_utils.py
```diff
@@ -362,14 +387,24 @@ def validate(path: Union[str, Path], devel_debug: bool = False) -> List[str]:
...
-                errors = [e for e in errors if not re_ok_prior_210.search(str(e))]
+                errors = []
+                for e in errors:
+                    try:
+                        assert e.message
+                        if not re_ok_prior_210.search(e.message):
+                            errors.append(e)
+                    except AssertionError:
+                        errors.append(e)
+                # errors = [e for e in errors if not re_ok_prior_210.search(e.message)]
```
yarikoptic, line 406: seems to be copy of above -- should be gone
```suggestion
```

dandi/pynwb_utils.py
```diff
@@ -362,14 +387,24 @@ def validate(path: Union[str, Path], devel_debug: bool = False) -> List[str]:
...
                 errors_ = errors[:]
-                errors = [e for e in errors if not re_ok_prior_210.search(str(e))]
+                errors = []
+                for e in errors:
+                    try:
+                        assert e.message
+                        if not re_ok_prior_210.search(e.message):
+                            errors.append(e)
+                    except AssertionError:
+                        errors.append(e)
```
yarikoptic, line 405: too verbose, I really think there should be some construct like 
```suggestion
                errors = [e for e in errors if not re_ok_prior_210.search(cast(str, getattr(e, "message", ""))]
```
which would keep it concise and please type checking

dandi/pynwb_utils.py
```diff
@@ -362,14 +387,24 @@ def validate(path: Union[str, Path], devel_debug: bool = False) -> List[str]:
         if version is not None:
             # Explicitly sanitize so we collect warnings.
             # TODO: later cast into proper ERRORs
+            # Do we really need to do these acrobatics? string comparison works fine...
```
yarikoptic, line 390: what "acrobatics", could you be more specific or this comment is no longer pertinent?

yarikoptic (changes requested):
Initial pass over - left various recommendations, will do another pass later

jwodder:
@TheChymera

> how about now?

How about now what?  Are you asking for a full review?  You haven't implemented all the `--grouping` options, so it looks to me like this PR isn't done yet.

dandi/cli/cmd_validate.py
```diff
@@ -20,36 +21,83 @@
     is_flag=True,
     help="Whether to write a report under a unique path in the DANDI log directory.",
 )
+@click.option(
+    "--grouping",
+    "-g",
```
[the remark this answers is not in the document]
yarikoptic: referring to discussion with @jwodder (I think that is what his concern was) -- please list only those which are implemented ATM, so if `"error"` is not implemented -- do not list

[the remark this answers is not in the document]
TheChymera: should I then remove the explicit `NotImplementedError` (L93)?

dandi/cli/cmd_validate.py
```diff
@@ -20,36 +21,83 @@
...
-    valid = is_valid(validator_result)
-    report_errors(validator_result)
 
-    if not valid:
+    issues = []
+    for i in validator_result:
+        if not i.severity:
+            continue
+        else:
+            issues.append(i)
```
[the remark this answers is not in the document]
TheChymera: done

dandi/pynwb_utils.py
TheChymera: done 37bfe870fb4e021cec5b92eaa528e54df25d8871

dandi/pynwb_utils.py
TheChymera: done ea129678ef1aebaf556eb82737051707f7944935

dandi/cli/cmd_validate.py
```diff
@@ -84,61 +146,119 @@ def validate(paths, schema=None, devel_debug=False, allow_any_path=False):
...
+
+
+    Notes
+    -----
+    * There is a bit of roundabout and currently untestable logic to deal with potential cases
+    where the same error has multiple error message, could be removed in the future and removed
+    by assert if this won't ever be the case.
+    """
+
+    if all([len(i) == 1 for i in [purviews, errors, severities, messages]]):
```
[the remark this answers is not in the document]
TheChymera: done 743680df1df54c33beca3364b95b2bf6aaf85d9e

dandi/cli/cmd_validate.py
```diff
@@ -20,36 +21,83 @@
...
+
+    purviews = [
+        list(filter(bool, [i.path, i.path_regex, i.dataset_path]))[0] for i in issues
+    ]
+    if not grouping:
+        display_errors(
+            purviews,
+            [i.id for i in issues],
+            [i.severity for i in issues],
+            [i.message for i in issues],
```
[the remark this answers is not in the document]
TheChymera: done fbaa3d7d79fa16bb922727545a7d96f094960b4e

dandi/cli/tests/test_cmd_validate.py
```diff
@@ -1,19 +1,77 @@
...
+    assert simple3_nwb in r.output
+    assert "NWBI.check_subject_id_exists" in r.output
+
+
+def test_validate_bids_error_grouping_notification(
+    bids_error_examples, dataset="invalid_asl003"
+):
+    """Test user notification for unimplemented parameter value."""
+
+    from ..cmd_validate import validate_bids
```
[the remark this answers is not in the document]
TheChymera: d77711734cd9aca7659a80c9f1b3475e93f3b648

dandi/cli/cmd_validate.py
```diff
@@ -84,61 +146,119 @@ def validate(paths, schema=None, devel_debug=False, allow_any_path=False):
...
```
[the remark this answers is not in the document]
TheChymera: > why not to make display_errors just to get a list of them instead of breaking them apart like this just to `zip` them back together inside that function?

Well, the idea was to have a separate function which does as little as possible other than `click.secho()`. The lists need to be populated differently depending on the grouping anyway, and additional edge cases will pop up as we have e.g. multi-path errors, so we do all that in the parent function and just echo depending on the list lengths. If you want I can roll it all up in a single function.

dandi/pynwb_utils.py
TheChymera: done

dandi/cli/cmd_validate.py
```diff
@@ -20,36 +21,83 @@
...
+        )
+    elif grouping == "path":
+        for purview in purviews:
+            applies_to = [
+                i for i in issues if purview in [i.path, i.path_regex, i.dataset_path]
```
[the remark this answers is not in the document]
TheChymera: Right, here we just select the entries based on the value of the purview. i.e. if multiple issues have the same `i.path` we group them together.

dandi/cli/cmd_validate.py
```diff
@@ -84,61 +146,119 @@ def validate(paths, schema=None, devel_debug=False, allow_any_path=False):
...
+    """
+    Unified error display for validation CLI, which auto-resolves grouping logic based on
+    the length of input lists.
+
+    Parameters
+    ----------
+    purviews: list of str
+    errors: list of str
+    severities: list of dandi.validate_types.Severity
+    messages: list of str
```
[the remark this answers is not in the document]
TheChymera: done ad27bfaaead0c3bd817fe6413f55636e0149cad1

dandi/cli/cmd_validate.py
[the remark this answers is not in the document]
yarikoptic: nah -- keep it -- always good to have some final `else` for such cases .

dandi/cli/cmd_validate.py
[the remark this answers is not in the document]
yarikoptic: so severity strings or `Severity` (as Parameters says)? What I meant is that there is no point in Parameters at all if you aren't going to provide some documentation for them and only types which should be in type hints.  So can just kill this Parameters   IMHO here.

dandi/cli/cmd_validate.py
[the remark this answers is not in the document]
yarikoptic: I still feel that it is odd design here, not ready ATM to give a complete solution, could at least be `display_errors(purviews, issues)` or alike, with all needed unzipping/zipping done inside if needed.  but ok, we could RF later

dandi/cli/cmd_validate.py
```diff
@@ -20,36 +22,46 @@
...
+    grouping="none",
 ):
     """Validate BIDS paths.
 
+    Parameters
+    ----------
+
+    grouping : str, optional
+        A string which is either "", "error", or "path" — "error" implemented.
+
```
yarikoptic, line 48:
```suggestion
```

absolutely no point to have it here since we rely on `click` to provide all docs etc, why duplicate?

dandi/cli/cmd_validate.py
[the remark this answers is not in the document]
TheChymera: I'd really suggest leaving it like this for now, and see if moving that code bit to another function is still needed when we merge validation functions.

dandi/cli/cmd_validate.py
TheChymera: done 677ef33e582c8b7225c29e567e206eeacb1335c3

dandi/cli/cmd_validate.py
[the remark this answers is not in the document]
TheChymera: done 677ef33e582c8b7225c29e567e206eeacb1335c3

dandi/files/bases.py
```diff
@@ -113,23 +126,11 @@ def get_validation_errors(
...
                 if devel_debug:
                     raise
-                lgr.warning(
-                    "Unexpected validation error for %s: %s",
-                    self.filepath,
-                    e,
-                    extra={"validating": True},
-                )
-                return [f"Failed to initialize Dandiset meta: {e}"]
+                return _pydantic_errors_to_validation_results(e, str(self.filepath))
```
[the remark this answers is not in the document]
TheChymera: done 5394ebc3fbac945447f9f8271593b3b2d8bbce9f

yarikoptic:
@TheChymera type checking is red. I would advise to run it before pushing. @jwodder - don't we have type checking as part of the pre-commit routine?

jwodder:
@yarikoptic No, type-checking should not be run by pre-commit, as it can take multiple seconds to run.

TheChymera:
@yarikoptic for some reason I can't reproduce this locally:

```
chymera@decohost ~/src/dandi-cli $ tox -e typing
GLOB sdist-make: /home/chymera/src/dandi-cli/setup.py
typing inst-nodeps: /home/chymera/src/dandi-cli/.tox/.tmp/package/1/dandi-0.46.2+128.g93b7aa3.zip
typing installed: anys==0.2.1,appdirs==1.4.4,arrow==1.2.3,asciitree==0.3.3,attrs==22.1.0,bidsschematools==0.4.0,blessings==1.7,certifi==2022.9.24,cffi==1.15.1,charset-normalizer==2.1.1,ci-info==0.3.0,click==8.1.3,click-didyoumean==0.3.0,coverage==6.5.0,cryptography==38.0.1,dandi @ file:///home/chymera/src/dandi-cli/.tox/.tmp/package/1/dandi-0.46.2%2B128.g93b7aa3.zip,dandischema==0.7.1,Deprecated==1.2.13,dnspython==2.2.1,email-validator==1.3.0,entrypoints==0.4,etelemetry==0.3.0,fasteners==0.18,fqdn==1.5.1,fscacher==0.2.0,h5py==3.7.0,hdmf==3.4.6,humanize==4.4.0,idna==3.4,iniconfig==1.1.1,interleave==0.2.1,isoduration==20.11.0,jaraco.classes==3.2.3,jeepney==0.8.0,joblib==1.2.0,jsonpointer==2.3,jsonschema==4.16.0,keyring==23.9.3,keyrings.alt==4.2.0,more-itertools==9.0.0,mypy==0.982,mypy-extensions==0.4.3,natsort==8.2.0,numcodecs==0.10.2,numpy==1.22.4,nwbinspector==0.4.17,opencv-python==4.6.0.66,packaging==21.3,pandas==1.5.1,pluggy==1.0.0,py==1.11.0,pycparser==2.21,pycryptodomex==3.15.0,pydantic==1.10.2,pynwb==2.1.0,pyout==0.7.2,pyparsing==3.0.9,pyrsistent==0.18.1,pytest==7.1.3,pytest-cov==4.0.0,pytest-mock==3.10.0,python-dateutil==2.8.2,pytz==2022.5,PyYAML==6.0,requests==2.28.1,responses==0.22.0,rfc3339-validator==0.1.4,rfc3987==1.3.8,ruamel.yaml==0.17.21,ruamel.yaml.clib==0.2.6,scipy==1.9.2,SecretStorage==3.3.3,semantic-version==2.10.0,six==1.16.0,tenacity==8.1.0,toml==0.10.2,tomli==2.0.1,tqdm==4.64.1,types-appdirs==1.4.3,types-python-dateutil==2.8.19.2,types-requests==2.28.11.2,types-toml==0.10.8,types-urllib3==1.26.25.1,typing_extensions==4.4.0,uri-template==1.2.0,urllib3==1.26.12,webcolors==1.12,wrapt==1.14.1,zarr==2.13.3
typing run-test-pre: PYTHONHASHSEED='653402275'
typing run-test: commands[0] | mypy dandi
Success: no issues found in 78 source files
_________________________________________________________________ summary __________________________________________________________________
  typing: commands succeeded
  congratulations :)
chymera@decohost ~/src/dandi-cli $ git rev-parse HEAD
93b7aa39110f1e1b5654493fa767032ffe3fb72a
  ```
  
I'll fix it by testing with the GH test suite, but what's also curious is that this is part of the code I haven't touched 🤔

**Comment by jwodder** [comment]
@yarikoptic @TheChymera I fixed the typing issue in #1144.  I believe the error only shows up when using the latest version of pytest.

**Comment by TheChymera** [comment]
@jwodder thank you, yes, the error no longer happens. Apparently there was another one, but I fixed that as well. @yarikoptic ready for merge?

**Comment by yarikoptic** [comment]
ok, let's merge and then tune up from there. Thanks @TheChymera and @jwodder for the reviews.

File: dandi/tests/test_validate.py [hunk 35]
```diff
@@ -1,35 +1,58 @@
...
-    # https://github.com/bids-standard/bids-specification/pull/1143
-    selected_dataset = os.path.join(bids_examples, "asl003")
-    _ = validate_bids(selected_dataset, report=True)
+@pytest.mark.parametrize("dataset", BIDS_TESTDATA_SELECTION)
+def test_validate_bids(bids_examples, tmp_path, dataset):
+
+    selected_dataset = os.path.join(bids_examples, dataset)
+    validation_result = validate_bids(selected_dataset, report=True)
+    for i in validation_result:
+        assert not hasattr(i, "severtiy")
```
jwodder, line 18: @TheChymera Is this typo deliberate?  If it is, what's the point?  If it's not, shouldn't you instead be testing that the attribute is `None`?

TheChymera: @jwodder thank you for spotting this, indeed it was an error, and further, it was giving a false positive arising from an upstream oversight. Will sumbit a fix PR once the [reference data is fixed](https://github.com/bids-standard/bids-examples/pull/338), since currently fixing this would fail due to the data being broken.

TheChymera: @jwodder fix is here → https://github.com/dandi/dandi-cli/pull/1148

Issue #1105, opened 2022-08-18
0.46.0 release to pypi errorred out due to 2BIG


yarikoptic:
It did release fine on github but then

https://github.com/dandi/dandi-cli/runs/7903361664?check_suite_focus=true

```
ℹ  info      Created GitHub release.
ℹ  info      https://github.com/dandi/dandi-cli/releases/tag/0.46.0
ℹ  info      Running command: python -m build && twine upload dist/*
ℹ  info      Supplied Environment (name and char size):
...
	SELENIUM_JAR_PATH=35
	GOROOT_1_17_X64=35
	ARG_0=154015
✖  error     Received E2BIG from execSync.

This usually occurs when the argument list is too large for the command you are trying to run. 

Please consider disabling your 'auto-exec' usage and following this issue for updates: https://github.com/intuit/auto/issues/1294
Error: Process completed with exit code 1.
```

I will merge some PR to trigger a new release and see if it works or we just finally hit the wall with this `intuit auto` and 2 year old bug https://github.com/intuit/auto/issues/1294

yarikoptic:
0.46.1 got released fine https://pypi.org/project/dandi/

PR #1106, opened 2022-08-18
DOC: minor - boost copyright years in the docs

yarikoptic:


yarikoptic:
I don't think there is need to wait for CI. This PR is really to just trigger a release (see #1105)

PR #1108, opened 2022-08-24
Check for usage of latest NWB Inspector release

CodyCBakerPhD:
Hey all, 

as a companion to the upcoming enforcement of more rigorous subject metadata requirements for DANDI upload (all taken care of by https://github.com/NeurodataWithoutBorders/nwbinspector/pull/247) it occurred to me that in order to both (i) prevent users from bypassing these requirements by downgrading their NWB Inspector version or (ii) prevent fewer future lower bound version bumps of the `nwbinspector` package on the DANDI requirements side; it would be best to simply check that the person is using the latest PyPI/conda-forge release of the package at time of validation.

I'm not really sure how to add a test for this though, let me know if you have ideas for that.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -75,9 +75,10 @@ jobs:
         python -m pip install --upgrade pip wheel
         pip install ".[test]"
 
-    - name: Install dev version of pynwb
+    - name: Install dev version of pynwb and nwbinspector
       if: matrix.mode == 'dev-deps'
       run: |
+        pip install git+https://github.com/NeurodataWithoutBorders/nwbinspector
         pip install git+https://github.com/NeurodataWithoutBorders/pynwb
 
     - name: Run all tests
--- a/dandi/files/bases.py
+++ b/dandi/files/bases.py
@@ -17,7 +17,10 @@ from dandischema.digests.dandietag import DandiETag
 from dandischema.models import BareAsset, CommonModel
 from dandischema.models import Dandiset as DandisetMeta
 from dandischema.models import get_schema_version
+from etelemetry import get_project
 from nwbinspector import Importance, inspect_nwb, load_config
+from nwbinspector.utils import get_package_version
+from packaging.version import Version
 from pydantic import ValidationError
 import requests
 
@@ -469,6 +472,33 @@ class NWBAsset(LocalFileAsset):
         else:
             # make sure that we have some basic metadata fields we require
             try:
+                # Ensure latest version of NWB Inspector is installed and used client-side
+                try:
+                    max_version = Version(
+                        get_project(repo="NeurodataWithoutBorders/nwbinspector")[
+                            "version"
+                        ]
+                    )
+                    current_version = get_package_version(name="nwbinspector")
+
+                    if current_version < max_version:
+                        errors.append(
+                            f"NWB Inspector version {current_version} is installed - please "
+                            f"use the latest release of the NWB Inspector ({max_version}) "
+                            "when performing `dandi validate`. To update, run "
+                            "`pip install -U nwbinspector` if you installed it with `pip`."
+                        )
+                        return errors
+
+                except Exception as e:  # In case of no internet connection or other error
+                    lgr.warning(
+                        "Failed to retrieve NWB Inspector version due to %s: %s",
+                        type(e).__name__,
+                        str(e),
+                    )
+
+                # Run NWB Inspector with 'dandi' config
+                # CRITICAL errors and above are equivalent to validation errors
                 errors.extend(
                     [
                         error.message
--- a/setup.cfg
+++ b/setup.cfg
@@ -51,7 +51,7 @@ install_requires =
     pycryptodomex  # for EncryptedKeyring backend in keyrings.alt
     pydantic >= 1.9.0
     pynwb >= 1.0.3,!=1.1.0
-    nwbinspector >= 0.3.8
+    nwbinspector >= 0.4.12
     pyout >=0.5, !=0.6.0
     python-dateutil
     requests ~= 2.20
--- a/dandi/tests/fixtures.py
+++ b/dandi/tests/fixtures.py
@@ -113,8 +113,8 @@ def simple2_nwb(
         subject=pynwb.file.Subject(
             subject_id="mouse001",
             date_of_birth=datetime(2016, 12, 1, tzinfo=tzutc()),
-            sex="M",
-            species="mouse",
+            sex="U",
+            species="Mus musculus",
         ),
         **simple1_nwb_metadata,
     )
@@ -128,9 +128,9 @@ def simple3_nwb(
     return make_nwb_file(
         str(tmpdir_factory.mktemp("simple2").join("simple2.nwb")),
         subject=pynwb.file.Subject(
-            date_of_birth=datetime(2016, 12, 1, tzinfo=tzutc()),
-            sex="M",
-            species="mouse",
+            age="P1D/",
+            sex="O",
+            species="Mus musculus",
         ),
         **simple1_nwb_metadata,
     )


dandi/files/bases.py
```diff
@@ -469,6 +480,21 @@ def get_validation_errors(
         else:
             # make sure that we have some basic metadata fields we require
             try:
+                # Ensure latest version of NWB Inspector is installed and used client-side
+                current_version, max_version = self._get_inspector_versions()
+                if current_version != max_version:
```
CodyCBakerPhD, line 485: Although, perhaps using `<` instead of `!=` here could allow easier testing using dev branches of the NWB Inspector? What do you guys think?

setup.cfg
```diff
@@ -50,7 +50,7 @@ install_requires =
     pycryptodomex  # for EncryptedKeyring backend in keyrings.alt
     pydantic >= 1.9.0
     pynwb >= 1.0.3,!=1.1.0
-    nwbinspector >= 0.3.8
+    nwbinspector >= 0.4.12
```
CodyCBakerPhD, line 53: Figured I might as well bump this while I'm at it - the `get_package_version` convenience helper is only available in `>=0.4.10`.

Point still being this isn't something that would need to be done hereafter since the version check in the validation will require the latest update.

.github/workflows/test.yml
```diff
@@ -80,6 +80,11 @@ jobs:
       run: |
         pip install git+https://github.com/NeurodataWithoutBorders/pynwb
 
+    - name: Install dev version of nwbinspector
+      if: matrix.mode == 'dev-deps'
+      run: |
+        pip install git+https://github.com/NeurodataWithoutBorders/nwbinspector
```
CodyCBakerPhD, line 86: Went ahead and added the NWB Inspector dev branch to the `dev-deps` testing mode alongside PyNWB.

Let me know if you'd prefer me to revert this, though.

CodyCBakerPhD: Now, what is a bit odd here, when I ran the testing suite on a local clone of `dandi-cli` with the dev branch of `nwbinspector` also installed in the same virtualenv, all the tests passed.

Reason I find that odd, is the fixtures define metadata that ought to be flagged if those fixtures are being used to test dandi validation, specifically this line: https://github.com/dandi/dandi-cli/blob/master/dandi/tests/fixtures.py#L117, which should be "mus musculus" to pass the NWB Inspector.

I could not tell at a glance, but where in the testing suite are those fixtures used to test validation of NWB assets?

CodyCBakerPhD:
@yarikoptic @jwodder Ready for review

setup.cfg
yarikoptic: ok

dandi/files/bases.py
```diff
@@ -453,6 +456,14 @@ def get_metadata(
         metadata.path = self.path
         return metadata
 
+    def _get_inspector_versions(self):
+        url = "https://pypi.org/pypi/nwbinspector/json"
+        data = json.loads(requests.get(url=url).text)
```
yarikoptic, line 461: consider using etelemetry from @satra et al: https://pypi.org/project/etelemetry/ which we already use for dandi-cli itself. Doesn't need to be in the same PR

dandi/files/bases.py
```diff
@@ -469,6 +480,21 @@ def get_validation_errors(
         else:
             # make sure that we have some basic metadata fields we require
             try:
+                # Ensure latest version of NWB Inspector is installed and used client-side
+                current_version, max_version = self._get_inspector_versions()
+                if current_version < max_version:
+                    errors.extend(
```
yarikoptic, line 486: I would not make demands such strong as to make it The only error and overall error out.

Please 
- add a dedicated try/except around the point of your interaction with the Internet. I might be offline and I still want to be able to run validation!
- make it into a `lgr.warning` and proceed "as usual" without `return`.

dandi/files/bases.py
```diff
@@ -469,6 +480,21 @@ def get_validation_errors(
...
+                        [
+                            f"NWB Inspector version {current_version} is installed - please "
+                            f"use the latest release of the NWB Inspector ({max_version}) when "
+                            "performing `dandi validate`. "
+                            "To update, please run `pip install -U nwbinspector`."
```
yarikoptic, line 491: Note that `pip install -U` is useful but might be misleading if dandi-cli and/or nwbinspector installed through a different distribution (e.g. conda or who knows whatnot later, e.g. debian), so
```suggestion
                            "To update, run `pip install -U nwbinspector` if you installed it with `pip`."
```

yarikoptic (changes requested):
Must not error out if outdated or forbid validation if offline

.github/workflows/test.yml
yarikoptic: I would fold this `pip install` into previous step and just adjust description that it is dev version of both pynwb and nwbinspector.

Rationale
- do not duplicate conditioning on dev-deps
- shorter

yarikoptic: > I could not tell at a glance, but where in the testing suite are those fixtures used to test validation of NWB assets?

```
$> git grep simple2_nwb
dandi/cli/tests/test_command.py:def test_smoke(simple2_nwb, command):
dandi/cli/tests/test_command.py:    r = runner.invoke(command, [simple2_nwb])
dandi/tests/fixtures.py:def simple2_nwb(
dandi/tests/fixtures.py:    simple2_nwb: str, tmp_path_factory: pytest.TempPathFactory
dandi/tests/fixtures.py:        organize, ["-f", "copy", "--dandiset-path", str(tmp_path), str(simple2_nwb)]
dandi/tests/fixtures.py:    simple2_nwb: str,
dandi/tests/fixtures.py:    shutil.copy(simple2_nwb, tmp_path)
dandi/tests/test_files.py:def test_validate_simple2(simple2_nwb):
dandi/tests/test_files.py:    errors = dandi_file(simple2_nwb).get_validation_errors()
dandi/tests/test_files.py:def test_validate_simple2_new(simple2_nwb):
dandi/tests/test_files.py:    errors = dandi_file(simple2_nwb).get_validation_errors(
dandi/tests/test_organize.py:def test_ambiguous(simple2_nwb: str, tmp_path: Path) -> None:
dandi/tests/test_organize.py:    copy2 = copy_nwb_file(simple2_nwb, tmp_path)
dandi/tests/test_organize.py:    args = ["--files-mode", "copy", "-d", outdir, simple2_nwb, copy2]
dandi/tests/test_organize.py:        for f in [simple2_nwb, copy2]
```
and since used in those other fixtures:

```shell
$> git grep organized_nwb_dir | grep test_ 
dandi/tests/test_fixtures.py:def test_organized_nwb_dir(organized_nwb_dir: Path) -> None:
dandi/tests/test_fixtures.py:def test_organized_nwb_dir2(organized_nwb_dir2: Path) -> None:
dandi/tests/test_upload.py:    new_dandiset: SampleDandiset, organized_nwb_dir: str, tmp_path: Path
dandi/tests/test_upload.py:        p for p in list_paths(organized_nwb_dir) if p.name != dandiset_metadata_file
dandi/tests/test_upload.py:    parent, name = nwb_file.relative_to(organized_nwb_dir).parts
```

dandi/files/bases.py
CodyCBakerPhD: Our of curiosity, how does `dandi upload` handle internet connectivity? Would it error out here or somewhere below:

https://github.com/dandi/dandi-cli/blob/ef097cc124629556d94544b4ca40f2f1ee567425/dandi/upload.py#L67-L69

.github/workflows/test.yml
CodyCBakerPhD: > I would fold this pip install into previous step and just adjust description that it is dev version of both pynwb and nwbinspector.
> 
> Rationale
> 
> do not duplicate conditioning on dev-deps
> shorter

Sure thing

dandi/files/bases.py
CodyCBakerPhD: Tried,

```python
etelemetry.get_project(repo="NeurodataWithoutBorders/nwbinspector")
```

results in

```python
RuntimeError: Connection to server could not be made
```

dandi/files/bases.py
CodyCBakerPhD: Fixed as requested, though.

dandi/files/bases.py
```diff
@@ -12,12 +12,15 @@
 from threading import Lock
 from typing import Any, BinaryIO, Generic, Optional
 from xml.etree.ElementTree import fromstring
+import json
+from packaging import version
```
jwodder, line 16: I can tell that you didn't run pre-commit on this code.  Please do `pip install pre-commit`, then run `pre-commit install` in your clone of dandi, and then `pre-commit run -a` to ensure all your contributions so far have been formatted, linted, & isorted.

dandi/files/bases.py
jwodder, line 461:
```suggestion
        data = requests.get(url).raise_for_status().json()
```

Assuming you can't get etelemetry to work.

dandi/files/bases.py
```diff
@@ -453,6 +456,14 @@ def get_metadata(
         metadata.path = self.path
         return metadata
 
+    def _get_inspector_versions(self):
```
jwodder, line 459: You don't appear to be using `self`, so this should be a static method.

dandi/files/bases.py
```diff
@@ -469,6 +480,31 @@ def get_validation_errors(
...
+                got_inspector_version = True
+                try:
+                    current_version, max_version = self._get_inspector_versions()
+                except Exception as e:  # In case of no internet connection or other error
+                    got_inspector_version = False
+                    lgr.warning(
+                        "Failed to retrieve NWB Inspector version.",
+                        self.filepath,
+                        type(e).__name__,
+                        str(e),
```
jwodder, line 493: These arguments are just going to be ignored, as there are no placeholders in the log message to display them.  Did you mean to include format placeholders?

.github/workflows/test.yml
CodyCBakerPhD: > and since used in those other fixtures:

Ah, thanks so much! I've updated the fixtures to the new requirements as well.

dandi/files/bases.py
```diff
@@ -469,6 +480,31 @@ def get_validation_errors(
...
+                    )
+                if got_inspector_version and current_version < max_version:
+                    errors.extend(
```
jwodder, line 496: You're only adding one item, so use `errors.append()` and get rid of the brackets around the item.

jwodder (changes requested):

dandi/files/bases.py
yarikoptic: > out here or somewhere below:

I bet it should fail somewhere "here"  -- just try? ;)

dandi/files/bases.py
yarikoptic: indeed still down, eh https://github.com/sensein/etelemetry-client/issues/35  attn @satra

dandi/files/bases.py
CodyCBakerPhD: Oh, sorry - copied/pasted from https://github.com/catalystneuro/dandi-cli/blob/force_usage_of_latest_nwb_inspector_release/dandi/files/bases.py#L446-L451 and didn't look very closely - I had assumed they were extra arguments to some kind of logger warning class that captured type/message as separate properties rather than just a single string.

Updated it to be a proper format-string now.

dandi/files/bases.py
CodyCBakerPhD: Behavior is not similar - calling `raise_for_status` when response is a success returns `None`. Updated to use the direct `.json()` call, though.

dandi/files/bases.py
CodyCBakerPhD: Done

Have y'all thought about using https://pre-commit.ci/ to automatically resolve such things? Since we've started using it it's _so nice_ to never having to think about

dandi/files/bases.py
jwodder: You should use `%s` placeholders instead of f-strings for log messages, in order to avoid an error if a stringified exception contains a `%`.

CodyCBakerPhD: Can you give an example of that?

CodyCBakerPhD: > The code you linked to in your previous comment does that; note the %s's in the first argument to log.warning().

I mean, can you provide an example of the justification for

> instead of f-strings for log messages, in order to avoid an error if a stringified exception contains a %.

jwodder: If `str(e)` equals, say, `"Bad format string: '%s %q'"`, then using it in an f-string will cause it to be a substring of the first argument of `log.warning()`, and so `log.warning()` will try to fill in the `%s` with an argument that isn't there and/or choke on the `%q`.  Yes, this is an unlikely scenario in this particular instance, but it costs you nothing to do it the correct, foolproof way from the start.

dandi/files/bases.py
satra: should be back up now.

CodyCBakerPhD: `etelemetry` working now anyhow

CodyCBakerPhD: Using `etelemetry` now

It simplified enough lines so that there's really no point to having it be a separate function, so I moved all the logic into the `try` of the logger-wrapping section

CodyCBakerPhD:
Unable to view what's off with the CI right now, but I'm assuming it's the keyring issue. #1111

dandi/support/bids/utils.py
```diff
@@ -0,0 +1,200 @@
+"""
```
yarikoptic, line 1: why these files are added back in this PR? (likely some rebase went a bit "dirty")

CodyCBakerPhD: Hmmm... Yesterday I

-> implemented `etelemetry`
-> synched `catalystneuro:master` to `dandi-cli:master`
-> merged `catalystneuro:master` into this branch

will look into this to see what went wrong

dandi/support/pyout.py
```diff
@@ -36,7 +36,7 @@ def naturalsize(v):
     return humanize.naturalsize(v)
 
 
-def datefmt(v, fmt=u"%Y-%m-%d/%H:%M:%S"):
+def datefmt(v, fmt="%Y-%m-%d/%H:%M:%S"):
```
CodyCBakerPhD, line 39: pre-commit will not let me restore these changes - I tried manually putting them back to the way they were but every time I try to commit it, pre-commit triggers and undoes them.

dandi/support/bids/utils.py
CodyCBakerPhD: OK whatever caused it it must have happened before the `etelemetry` change. It's been manually removed from this branch now

dandi/support/pyout.py
CodyCBakerPhD: I can certainly try it, but you can always squash it on merge to the same effect, though, right?

*E.g.*,

image

A bit less work that way IMO

yarikoptic: this PR still has unrelated changes we do not want to merge/squash in, right?

CodyCBakerPhD: Not since I forcibly restored the changes to `supper/pyout.py`. Now the changelog represents only the related changes to the NWB Inspector and NWB asset validation.

Just wanted you to be aware that following [jwodder's instructions](https://github.com/dandi/dandi-cli/pull/1108#discussion_r959937152) to the letter resulted in pre-commit automatically doing all of the changes to that file.

yarikoptic: dandi/tests/fixtures.py is still in the diff and to be gone if you like to continue your fight!

CodyCBakerPhD: The changes to `dandi/tests/fixtures.py` are intentional updates to keep them compatible with the new validation requirements

CodyCBakerPhD: The git stash step didn't work out for me, but

```
git reset $(git merge-base master $(git branch --show-current))
git add -A
git commit --no-verify -m "manual squash"  # Need --no-verify to skip the pyout.py changes
git push -f
```

did

yarikoptic:
Thank you! Let's see how it goes

PR #1112, opened 2022-09-01
BF(workaround): blacklist 23.9.0 of keyring since introduced regression

### LINKED ISSUE #1111: [Bug]: cannot import name 'properties' from 'keyring.util'
Just FYI

Ran into this earlier today: https://github.com/catalystneuro/neuroconv/runs/8136194375?check_suite_focus=true

Already been raised upstream on https://github.com/jaraco/keyring/issues/593

For now I'm pinning to <=23.8.2 downstream, IDK what y'all would prefer to do for now here though.

yarikoptic:
Closes #1111

Ref upstream: https://github.com/jaraco/keyring/issues/593

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -45,7 +45,7 @@ install_requires =
     importlib-metadata;  python_version < "3.8"
     interleave ~= 0.1
     joblib
-    keyring
+    keyring != 23.9.0
     keyrings.alt
     packaging
     pycryptodomex  # for EncryptedKeyring backend in keyrings.alt


jwodder (approved):
Don't forget to update the conda-forge recipe after releasing!

jwodder:
On closer inspection, are you sure 23.8.2 is right version to exclude?  That thread seems to say that 23.8.2 is the last good version and 23.9.0 is the problem.

yarikoptic:
> On closer inspection, are you sure 23.8.2 is right version to exclude? That thread seems to say that 23.8.2 is the last good version and 23.9.0 is the problem.

d'oh, right, I rushed. I will fix to exclude 23.9.0

PR #1115, opened 2022-09-15
Detect link type for `organize` by linking from first src dir to dest dir

### LINKED ISSUE #1110: organize - cannot move files on Windows across drives
IIRC we had similar issue in datalad-installer recently. IIRC boils down to using shutil's move instead.

e.g.

```
(dandi) C:\Users\Pluta lab\000301>dandi organize "D:\MATLAB Analysis\NWD\matnwb\nwd_files"
2022-08-30 13:10:28,635 [ INFO] Loading metadata from 14 files
[Parallel(n_jobs=-1)]: Using backend LokyBackend with 12 concurrent workers.
[Parallel(n_jobs=-1)]: Done 3 out of 14 | elapsed: 4.4s remaining: 16.3s
[Parallel(n_jobs=-1)]: Done 5 out of 14 | elapsed: 4.4s remaining: 8.0s
[Parallel(n_jobs=-1)]: Done 7 out of 14 | elapsed: 4.4s remaining: 4.4s
[Parallel(n_jobs=-1)]: Done 9 out of 14 | elapsed: 4.4s remaining: 2.4s
[Parallel(n_jobs=-1)]: Done 11 out of 14 | elapsed: 4.4s remaining: 1.1s
[Parallel(n_jobs=-1)]: Done 14 out of 14 | elapsed: 4.5s finished
2022-08-30 13:10:33,207 [ INFO] Hard link support autodetected; setting files_mode='hardlink'
2022-08-30 13:10:33,230 [ INFO] Logs saved in C:\Users\Pluta lab\AppData\Local\dandi\dandi-cli\Logs\20220830171021Z-9648.log
Error: [WinError 17] The system cannot move the file to a different disk drive: 'D:\MATLAB Analysis\NWD\matnwb\nwd_files\ecephys_t_26.nwb' -> '.\sub-M26\sub-M26_ses-20200319T153237_behavior+ecephys.nwb'
```

reported in https://github.com/dandi/helpdesk/discussions/80#discussioncomment-3511318

jwodder:
Closes #1110.

### DIFF
--- a/dandi/organize.py
+++ b/dandi/organize.py
@@ -680,17 +680,15 @@ def _get_non_unique_paths(metadata):
     return non_unique
 
 
-def detect_link_type(srcdir, destdir):
+def detect_link_type(srcfile, destdir):
     """
-    Determine what type of links the filesystem will let us make from the
-    directory ``srcdir`` to the directory ``destdir``.  If symlinks are
-    allowed, returns ``"symlink"``.  Otherwise, if hard links are allowed,
-    returns ``"hardlink"``.  Otherwise, returns ``"copy"``.
+    Determine what type of links the filesystem will let us make from the file
+    ``srcfile`` to the directory ``destdir``.  If symlinks are allowed, returns
+    ``"symlink"``.  Otherwise, if hard links are allowed, returns
+    ``"hardlink"``.  Otherwise, returns ``"copy"``.
     """
-    srcfile = Path(srcdir, f".dandi.{os.getpid()}.src")
     destfile = Path(destdir, f".dandi.{os.getpid()}.dest")
     try:
-        srcfile.touch()
         try:
             os.symlink(srcfile, destfile)
         except OSError:
@@ -712,10 +710,6 @@ def detect_link_type(srcdir, destdir):
     finally:
         try:
             destfile.unlink()
-        except FileNotFoundError:
-            pass
-        try:
-            srcfile.unlink()
         except FileNotFoundError:
             pass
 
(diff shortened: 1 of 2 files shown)


dandi/organize.py
```diff
@@ -857,7 +861,7 @@ def _get_metadata(path):
         act(os.makedirs, dandiset_path)
 
     if files_mode == "auto":
-        files_mode = detect_link_type(dandiset_path)
+        files_mode = detect_link_type(srcdir, dandiset_path)
```
yarikoptic, line 864: why do we have to take directory here for `srcdir` and can't just take a `metadata[0]['path']` as the source file to test linking against as I thought about in https://github.com/dandi/dandi-cli/issues/1110#issuecomment-1244429099 ?

jwodder: @yarikoptic Before you wrote that comment, I had already modified `detect_link_type()` to take two directory paths.  I don't see an advantage to changing it to take a file path.

yarikoptic: Re advantage: User might not have write permissions to modify the `srcdir` - e.g. it belongs to read-only lab account.  Current `master` code would work ok since we create `srcdir` also in the target dandiset, so for such use cases it would work.  With the change you propose code would start to error out since user would not be able to create anything in `srcdir`, so it would be a regression we better avoid introducing.

dandi/organize.py
```diff
@@ -789,13 +783,15 @@ def act(func, *args, **kwargs):
             )
         lgr.info("We will organize %s in-place", dandiset_path)
         in_place = True
-        paths = dandiset_path
+        paths = [dandiset_path]
 
     if len(paths) == 1 and paths[0].endswith(".json"):
         # Our dumps of metadata
         metadata = load_jsonl(paths[0])
+        link_test_file = metadata[0]["path"]
```
yarikoptic, line 791: in principle AFAIK we do not even need `link_test_file` since we can simply use `metadata[0]["path"]` directly since we do load metadata records for regular paths too... but ok, let it be this way for now

yarikoptic:
Thank you @jwodder and let's proceed!

PR #1121, opened 2022-10-03
Pull latest Docker images before spinning up Compose

### LINKED ISSUE #1120: docker fixture to `docker-compose pull` first before `up`
In light of the today's need to troubleshoot testing for @djarecka, I think the fixture should first ensue that we have "compatible" docker images or troubleshooting takes a  bit too long.  So I think we should do `docker-compose pull` or at least `docker pull dandiarchive/dandiarchive-api` first.  I will leave it up to @jwodder to do full `compose` pull or just of our dandiarchive-api image.

jwodder:
Closes #1120.


yarikoptic:
ok, let's do such full pull, hopefully would be quick enough and would save us time troubleshooting. Thank you @jwodder !

PR #1123, opened 2022-10-03
adding testing instructions

djarecka:
- adding testing instruction to `DEVELOPMENT.md`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -89,7 +89,6 @@ test =
     pytest-cov
     pytest-mock
     responses
-    tox
 tools=
     boto3
 all =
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -89,6 +89,7 @@ test =
     pytest-cov
     pytest-mock
     responses
+    tox
```
jwodder, line 92: This does not belong here; for one thing, if you're installing tox, then you don't want to install the other test dependencies in the same environment, as tox will install them in one of its environments.

djarecka: ok, that's true. But do you still suggest that people use `tox` to test the source code locally (instead of `pytest` directly)? Should I include `pip install tox` in `DEVELOPMENT.md`?

jwodder: Yes.

djarecka:
@jwodder - do you want me to change something else? or can we merge it?

Issue #1130, opened 2022-10-10
Should ask for a new token to be stored in-place of the old one


yarikoptic:
I have created a new emabrgoed dandiset but download of it errors out as

```shell
lena:~/proj/dandi/dandisets-more
$> dandi download https://dandiarchive.org/dandiset/000344
2022-10-10 15:06:04,658 [ WARNING] A newer version (0.46.4) of dandi/dandi-cli is available. You are using 0.46.2
2022-10-10 15:06:06,842 [    INFO] Logs saved in /home/yoh/.cache/dandi-cli/log/20221010190604Z-3984369.log
Error: Error 401 while sending GET request to https://api.dandiarchive.org/api/auth/token: {"detail":"Invalid token."}
```

and requires me to specify `DANDI_API_KEY` env var "manually" instead of providing UI to request a new one as we do in some other places.

jwodder:
@yarikoptic I tried running the same command with dandi 0.46.4, and it prompted me for an API key.  Note that the error message you got says "Invalid token", suggesting that dandi successfully got a token from your keychain but it was invalid.

yarikoptic:
Correct, I think it is just the matter of it remembering an old no longer valid token, `dandi-cli` should reask for a new one.

jwodder:
@yarikoptic It should indeed (see below); what kind of environment were you running dandi in?

https://github.com/dandi/dandi-cli/blob/d642bccfc7819aa13aec3ae767c58ac8f9d21e00/dandi/dandiapi.py#L490-L499

yarikoptic:
> @yarikoptic It should indeed (see below); what kind of environment were you running dandi in?
> 
> https://github.com/dandi/dandi-cli/blob/d642bccfc7819aa13aec3ae767c58ac8f9d21e00/dandi/dandiapi.py#L490-L499

hm, sorry about the noise -- somehow my environment was "poisoned" -- I had DANDI_API_KEY env var set. Proposing https://github.com/dandi/dandi-cli/pull/1133 and will consider this issue resolved.

PR #1133, opened 2022-10-10
ENH: lgr.debug that we are using DANDI_API_KEY provided key

yarikoptic:
to possibly avoid confusions like
https://github.com/dandi/dandi-cli/issues/1130

or should we go for `.info`?

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -474,6 +474,7 @@ class DandiAPIClient(RESTFullAPIClient):
         # Shortcut for advanced folks
         api_key = os.environ.get("DANDI_API_KEY", None)
         if api_key:
+            lgr.debug("Using api key from DANDI_API_KEY environment variable")
             self.authenticate(api_key)
             return
         if self.api_url in known_instances_rev:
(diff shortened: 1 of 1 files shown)


dandi/dandiapi.py
```diff
@@ -474,6 +474,7 @@ def dandi_authenticate(self) -> None:
         # Shortcut for advanced folks
         api_key = os.environ.get("DANDI_API_KEY", None)
         if api_key:
+            lgr.debug("Using api key from DANDI_API_KEY env variable")
```
jwodder, line 477:
```suggestion
            lgr.debug("Using API key from DANDI_API_KEY environment variable")
```

jwodder (changes requested):
You might want to add corresponding "Using API key from keyring/user input" log messages around line 492.

dandi/dandiapi.py
yarikoptic: will do, thanks

yarikoptic:
> You might want to add corresponding "Using API key from keyring/user input" log messages around line 492.

good idea, did that+ -- please check

dandi/dandiapi.py
jwodder: @yarikoptic You didn't commit my suggested change.

jwodder (approved):

PR #1138, opened 2022-10-13
Make internal Zarr-checksumming tree types support empty Zarrs

jwodder:
@yarikoptic Please release this so I can use it in backups2datalad.

### DIFF
--- a/dandi/cli/tests/test_digest.py
+++ b/dandi/cli/tests/test_digest.py
@@ -1,3 +1,4 @@
+import os
 from pathlib import Path
 
 from click.testing import CliRunner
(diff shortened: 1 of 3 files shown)


yarikoptic:
is an empty directory is a legit zarr and we will get those to archive?  sorry if I am not seeing the connection (there is no reference) to other ongoing discussions.

jwodder:
@yarikoptic I'm not sure at the moment what the CLI would do if told to upload an empty Zarr, but it's certainly possible to create one via the API, and the backups2datalad tests include backing up an empty Zarr.

yarikoptic:
gotcha, `invalid` but still `possible`

PR #1139, opened 2022-10-18
Don't try to cancel failed Zarr uploads twice

jwodder:
Fixed a bug seen in https://github.com/dandi/dandi-cli/issues/1028#issuecomment-1282761146.

### DIFF
--- a/dandi/files/zarr.py
+++ b/dandi/files/zarr.py
@@ -450,6 +450,7 @@ class ZarrAsset(LocalDirectoryAsset[LocalZarrEntry]):
                                 f.cancel()
                             executor.shutdown()
                             client.delete(f"/zarr/{zarr_id}/upload/")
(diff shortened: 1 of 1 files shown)


dandi/files/zarr.py
```diff
@@ -450,6 +450,7 @@ def mkzarr() -> str:
                                 f.cancel()
                             executor.shutdown()
                             client.delete(f"/zarr/{zarr_id}/upload/")
+                            ZARR_UPLOADS_IN_PROGRESS.discard(upload_data)
```
yarikoptic, line 453: uff, took me awhile to locate this `ZARR_UPLOADS_IN_PROGRESS` which is apparently not a local to the function, but  bound at module level below at some line 576, uff . With that it is a global structure, it makes sense.

PR #1141, opened 2022-10-21
Add `DANDI_TESTS_PULL_DOCKER_COMPOSE` envvar

### LINKED ISSUE #1140: Add env var to disable docker-compose pull in the fixture
We already have `DANDI_TESTS_PERSIST_DOCKER_COMPOSE` and let's add `DANDI_TESTS_PULL_DOCKER_COMPOSE` (or anything else really) which would default to `1` but allow for `0` and then would be used as a bool (or some other way allowing to disable) to disable `docker compose pull` introduced in https://github.com/dandi/dandi-cli/pull/1121 to fix #1120 and which lead to dandi-archive integration testing not using freshly built docker image as discovered in https://github.com/dandi/dandi-archive/issues/1336#issuecomment-1286919392 .

jwodder:
Closes #1140.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -74,6 +74,10 @@ development command line options.
   new set on each run.  Set this environment variable to `0` to cause the
   containers to be destroyed at the end of the next run.
 
+- `DANDI_TESTS_PULL_DOCKER_COMPOSE` -- When set to an empty string or `0`, the
+  tests will not pull the latest needed Docker images at the start of a run if
+  older versions of the images are already present.
+
 ## Sourcegraph
 
 The [Sourcegraph](https://sourcegraph.com) browser extension can be used to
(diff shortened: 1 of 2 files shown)


yarikoptic (approved):
Thank you! Do we need a release to take advantage of this in dandi-archive testing? if so -- please add `release` label and merge

yarikoptic:
answering my own question since was looking at that workflow anyways -- we do want release since we test against both dev and release, thus adding the label

jwodder:
@yarikoptic Dandi Archive tests against both the latest release of dandi-cli and against the master branch, so a release isn't really needed.

yarikoptic:
;-) indeed the fact that we test against both could be treated the other way as well (that no release is needed).  But I think we better re-establish the common base now, so I will keep the `release` label and can be merged as ubuntu CI runs complete

yarikoptic:
hm, testing is failing but likely due to the dandi-archive introducing breaking change and this PR trying to allow to avoid that in the future. I do not see how this could add breakage, so will merge

PR #1143, opened 2022-10-26
Test against Python 3.11

jwodder:
Python 3.11 was released on October 24, so it's time to start testing against it.

Note: I expect this PR to fail for the next few weeks until numpy makes a 3.11-compatible release.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,6 +32,7 @@ jobs:
           - 3.8
           - 3.9
           - '3.10'  # Needs quotes so YAML doesn't think it's 3.1
+          - '3.11'
         mode:
           - normal
         include:
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,6 +16,7 @@ classifiers =
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
     Programming Language :: Python :: 3.10
+    Programming Language :: Python :: 3.11
     Topic :: Scientific/Engineering
 license = Apache 2.0
 description = Command line client for interaction with DANDI archive elements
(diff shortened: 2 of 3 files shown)


yarikoptic:
FTR according to changes in setup.py was added in around Aug but never released yet:

```
 ~/proj/numpy  main ?1 ▓▒
❯ git describe --contains --tags f06a315c965898fc8c6de144e279a161e8175a36
fatal: cannot describe 'f06a315c965898fc8c6de144e279a161e8175a36'
❯ git describe --tags f06a315c965898fc8c6de144e279a161e8175a36
v1.24.0.dev0-643-gf06a315c9
```

jwodder:
Current status: The latest version of nwbinspector requires `numpy>=1.22.0,<1.23.0` (even though the latest version of numpy is 1.23.5), and the last numpy 1.22.x release does not have any Python 3.11 wheels available, and trying to build from source fails.

bendichter:
nwbinspector requires pynwb, which requires hdmf, which requires h5py, which also does not have wheels released for py3.11 last time I checked. We have an open PR for testing hdmf against py3.11 here: https://github.com/hdmf-dev/hdmf/pull/803 but we will need to wait on numpy and h5py. In the meantime we are making the numpy requirements less strict on nwbinspector.

jwodder:
Current status: Only the the Windows + Python 3.11 tests are failing, due to the same issue affecting Windows + Python 3.10 tests described [here](https://github.com/dandi/dandi-cli/pull/1213#issuecomment-1436169590).

jwodder:
@yarikoptic All tests pass now.

yarikoptic:
Thank you @jwodder ! I wonder if we are doomed to add to https://github.com/dandi/dandi-cli#installation for now the need for such system wide dance for 3.11 to preclude possible user filed issues?

jwodder:
@yarikoptic If you're referring to the need to install hdf5, that no longer seems to be necessary, as the latest version of h5py includes wheels for Python 3.11.

yarikoptic:
oh, great, and we are still green -- let's proceed. Thank you @jwodder ! (as for added warning ignore -- ok with me, I just upgraded botocore - seems resolved it)

PR #1148, opened 2022-10-31
Fixed logic and commented a temporarily broken BIDS (lacking README) dataset

TheChymera:
Error [fixed upstream](https://github.com/bids-standard/bids-specification/pull/1346#event-7696972438) but it was the specification schema rather than the example data which was broken, so we won't have access to it until the new release; removing test dataset from whitelist until then.


yarikoptic:
Thanks!

Issue #917, opened 2022-02-18
namespace checks


satra:
this section of the pynwb CLI figures out all the cached namespaces in an NWB file and validates against them:  https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/validate.py#L70

the `validate` function only validates against the CORE namespace. this has resulted in this error: https://github.com/dandi/helpdesk/discussions/43

there needs to be some refactoring in pynwb, but that's not going to fix dandi issues immediately unless we depend on the latest and greatest version. so perhaps some of the code could be imported into the dandi validate nwb function. 

cc: @rly @oruebel

oruebel:
In https://github.com/NeurodataWithoutBorders/pynwb/pull/1432 I've extracted the code to determine the extension namespace from the file to a separate function `` pynwb.validate.get_chached_namespaces_to_validate`` so that the code can be reused to validate against cached namespaces when calling ``pynwb.validate`` directly from Python, rather than the command line. The code example below illustrates how you could fix this quickly in DANDI. Once PyNWB has been released you would then remove the ``get_chached_namespaces_to_validate`` from DANDI and simply replace it with the import from PyNWB ``from pynwb.validate import get_chached_namespaces_to_validate``. 

```
from pynwb import validate, NWBHDF5IO

def get_chached_namespaces_to_validate(path):
   """
   Determine the most specific namespace(s) (i.e., extensions) that are chached in the given
   NWB file that should be used for validation.

   :param path: Path for the NWB file
   :return: Tuple with:
     - List of strings with the most specific namespace(s) to use for validation.
     - BuildManager object for opening the file for validation
     - Dict with the full result from NWBHDF5IO.load_namespaces
   """
   # Importing locally here so we can easily switch to using pynwb.validate.get_chached_namespaces_to_validate
   # once it is part of the release
   from hdmf.spec import NamespaceCatalog
   from hdmf.build import BuildManager
   from pynwb.spec import NWBDatasetSpec, NWBGroupSpec, NWBNamespace
   from hdmf.build import TypeMap as TypeMap

   catalog = NamespaceCatalog(NWBGroupSpec, NWBDatasetSpec, NWBNamespace)
   ns_deps = NWBHDF5IO.load_namespaces(catalog, path)
   s = set(ns_deps.keys())  # determine which namespaces are the most
   for k in ns_deps:  # specific (i.e. extensions) and validate
      s -= ns_deps[k].keys()  # against those
[...]

PR #1149, opened 2022-11-01
Use cached namespace validation

CodyCBakerPhD:
Fixes https://github.com/dandi/dandi-cli/issues/917

Which came from https://github.com/dandi/helpdesk/discussions/43

Replaces https://github.com/dandi/dandi-cli/pull/1036

Newest version of PyNWB supports validation against cached namespaces (and indeed this is now the default).

I've updated the usage here in a minimal way to both support older PyNWB versions (my other thought was to bump the version in the `setup.cfg`, but IDK what your policy or view is on that) that do not offer this feature as well as the newest version that supports it as a simple adjustment to the way the `pynwb.validate` function is called.

### DIFF
--- a/dandi/pynwb_utils.py
+++ b/dandi/pynwb_utils.py
@@ -341,21 +341,21 @@ def validate(
         else:  # Fallback if an older version
             with pynwb.NWBHDF5IO(path=path, mode="r", load_namespaces=True) as reader:
                 error_outputs = pynwb.validate(io=reader)
-                # TODO: return ValidationResult structs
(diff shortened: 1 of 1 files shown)


yarikoptic:
Thank you @CodyCBakerPhD and sorry I've missed it whenever you originally filed it. Looks good to me and is Green (freshier PRs aren't as lucky due to some disturbance in the Force).  So I will not delay merging it as is.  I think we can stick to not demand upgraded pynwb for a bit longer but might soon as well beef it up.

PR #1150, opened 2022-11-03
Remove Parameters description from docstring used by click

yarikoptic:
With that numpy style Parameters --help looks ugly and duplicates click produced help.

    ❯ dandi validate --help ecephys_tutorial.nwb
    Usage: dandi validate [OPTIONS] [PATHS]...

      Validate files for NWB and DANDI compliance.

      Exits with non-0 exit code if any file is not compliant.

      Parameters ----------

      grouping : str, optional     A string which is either "", "error", or "path"
      — "error" not yet implemented.

    Options:
      -g, --grouping [none|path]  How to group error/warning reporting.
      --help                      Show this message and exit.

I remember asking for this change but it was missed.

### DIFF
--- a/dandi/cli/cmd_validate.py
+++ b/dandi/cli/cmd_validate.py
@@ -87,12 +87,6 @@ def validate(
     """Validate files for NWB and DANDI compliance.
 
     Exits with non-0 exit code if any file is not compliant.
-
-    Parameters
-    ----------
-
-    grouping : str, optional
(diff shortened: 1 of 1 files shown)


TheChymera (approved):
ok, I had just originally thought it's more convenient to have the documentation represented homogeneously in all functions — hence numpy docstring everywhere... but you're right, maybe the fact that the function is just a CLI binding warrants an exception.

PR #1156, opened 2022-11-11
Fix typing error under mypy 0.990

jwodder:
The latest release of mypy now detects an error in our type annotations, which this PR fixes.

### DIFF
--- a/dandi/files/_private.py
+++ b/dandi/files/_private.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
+from collections.abc import Mapping
 from dataclasses import dataclass
 from enum import Enum
 from pathlib import Path
+from typing import ClassVar
 import weakref
 
 from dandi.consts import (
@@ -14,6 +16,7 @@ from dandi.exceptions import UnknownAssetError
 
 from .bases import DandiFile, GenericAsset, LocalAsset, NWBAsset, VideoAsset
 from .bids import (
+    BIDSAsset,
     BIDSDatasetDescriptionAsset,
     GenericBIDSAsset,
     NWBBIDSAsset,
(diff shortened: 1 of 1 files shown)


yarikoptic:
didn't check what failures are due to. is it the same as https://github.com/dandi/dandi-archive/issues/1365 but even in the container we use?

jwodder:
@yarikoptic No, this is a type-checking error.  It has absolutely nothing to do with the API server performance.

yarikoptic:
oh right , FTR looked into one of the errors and it is
```
dandi/upload.py:15: in <module>
    from .files import (
dandi/files/__init__.py:23: in <module>
    from ._private import BIDSFileFactory, DandiFileFactory
dandi/files/_private.py:74: in <module>
    class BIDSFileFactory(DandiFileFactory):
/opt/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/dataclasses.py:1019: in dataclass
    return wrap(cls)
/opt/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/dataclasses.py:1011: in wrap
    return _process_class(cls, init, repr, eq, order, unsafe_hash, frozen)
/opt/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/dataclasses.py:861: in _process_class
    cls_fields = [_get_field(cls, name, type)
/opt/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/dataclasses.py:861: in <listcomp>
    cls_fields = [_get_field(cls, name, type)
/opt/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/dataclasses.py:745: in _get_field
    raise ValueError(f'mutable default {type(f.default)} for field '
E   ValueError: mutable default <class 'dict'> for field CLASSES is not allowed: use default_factory
```

jwodder:
@yarikoptic Now all the remaining test failures are ones that will be resolved by #1157.

PR #1161, opened 2022-11-16
No longer mark `test_rename_type_mismatch` as xfailing

jwodder:
The blocking issue, https://github.com/dandi/dandi-archive/issues/1109, was resolved.


yarikoptic:
Thank you !

PR #1166, opened 2022-11-28
Corrected reporting function logic to complete group message variable

TheChymera:
Addressing issue raised here: https://github.com/dandi/dandi-cli/pull/1165#issuecomment-1329566207


yarikoptic:
@TheChymera as PR title would be changelog entry -- please expand it slightly to become self-descriptive. Otherwise not clear what logic was fixed and why/for what reason.  And btw -- for our own sake, what it resulted in really fixed, e.g. can you show on example output in master version vs this PR fixed one?

jwodder (approved):

TheChymera:
@yarikoptic edited.

PR #1164, opened 2022-11-23
Unravelling `ls` issues

TheChymera:
Currently littered with diagnostic print calls, addressing https://github.com/dandi/dandi-cli/issues/1038

PR #1043, opened 2022-07-05
Changing `docker-compose` to docker `docker compose` subcommand

TheChymera:
Fixes https://github.com/dandi/dandi-cli/issues/1042

PR #1168, opened 2022-12-05
Changing `docker-compose` to docker `docker compose` subcommand

TheChymera:
Having another go at https://github.com/dandi/dandi-cli/pull/1043 (came up during debugging of https://github.com/dandi/dandi-cli/pull/1164 )

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -94,14 +94,14 @@ jobs:
     - name: Dump Docker Compose logs
       if: failure() && startsWith(matrix.os, 'ubuntu')
       run: |
-        docker-compose \
+        docker compose \
           -f dandi/tests/data/dandiarchive-docker/docker-compose.yml \
           logs --timestamps
 
     - name: Shut down Docker Compose
       if: startsWith(matrix.os, 'ubuntu')
       run: |
-        docker-compose \
+        docker compose \
           -f dandi/tests/data/dandiarchive-docker/docker-compose.yml \
           down -v
 
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -37,7 +37,7 @@ tox -e lint,typing
 ### dandi-api instance
 
 The [dandi-api](https://github.com/dandi/dandi-api) repository provides a
-docker-compose recipe for establishing a local instance of a fresh dandi-api.
+`docker compose` recipe for establishing a local instance of a fresh dandi-api.
 See
 [README.md:Docker](https://github.com/dandi/dandi-api#develop-with-docker-recommended-quickstart)
 for the instructions.  In a new instance, you would need to generate a new API
--- a/dandi/tests/fixtures.py
+++ b/dandi/tests/fixtures.py
@@ -335,16 +335,27 @@ def docker_compose_setup() -> Iterator[Dict[str, str]]:
     try:
         if create:
             if os.environ.get("DANDI_TESTS_PULL_DOCKER_COMPOSE", "1") not in ("", "0"):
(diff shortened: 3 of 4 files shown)


yarikoptic:
> Having another go at #1043 (came up during debugging of #1164 )

please remind me why we **need** to use `docker compose` instead of `docker-compose`?  In my case I seems to have too old docker 

```
❯ apt-cache policy docker.io
docker.io:
  Installed: 20.10.17+dfsg1-1+b1
  Candidate: 20.10.21+dfsg1-1
  Version table:
     20.10.21+dfsg1-1 900
        900 http://deb.debian.org/debian bookworm/main amd64 Packages
        600 http://http.debian.net/debian sid/main amd64 Packages
 *** 20.10.17+dfsg1-1+b1 100
        100 /var/lib/dpkg/status
     20.10.5+dfsg1-1+deb11u2 100
        100 http://deb.debian.org/debian bullseye/main amd64 Packages
❯ docker-compose --version
docker-compose version 1.29.2, build unknown
❯ docker --version
Docker version 20.10.17+dfsg1, build 100c701
❯ docker compose
docker: 'compose' is not a docker command.
See 'docker --help'
```

so I cannot "switch" to use `docker compose` locally and tests would start to fail for me (or any other system with older docker).

TheChymera:
@yarikoptic your issue is probably that it's more than a name change, but a full [re-write and partial re-design using a new plug-in system](https://github.com/docker/compose#about-update-and-backward-compatibility). This sounds counter-intuitive, but as best I understand the broader docker package architecture, the **old** `docker-compose` in spite of being its own command is part of the Docker package whereas the **new** `docker compose` *subcommand* is its own package. I looked it up, and apparently on Debian, you should run:

```console
dragonhost ~ # apt-get install docker-compose-plugin
```

As for why: `docker-compose` seems to be EOL and there seem to be [significant differences between the versions](https://github.com/docker/compose#about-update-and-backward-compatibility). Thus far, it seems we have steered clear of those and we can drop-in replace it, but the sooner we switch, the surer we are to not have to do an emergency switch later on with possible issues accumulating which prohibit drop-in replacement.

Additionally, some distros/packages seem to do some [sneaky symlinking](https://github.com/docker/compose/issues/9814#issuecomment-1239333979), whereby if you use `docker-compose` you don't know which one your're using. With `docker compose` you know you're using the newer one.

yarikoptic:
thanks for the explanation!  my issue would be then that there is no `docker-compose-plugin` package in debian proper at all: https://packages.debian.org/search?keywords=docker-compose-plugin&searchon=names&suite=all&section=all .  It is likely available from docker's APT repo https://docs.docker.com/engine/install/ubuntu/#set-up-the-repository . didn't check.

I do appreciate the desire to not delay migrations but I we should also think about DX and not unnecessarily overcomplicate it... I will be ok with us going forward to `docker compose` but allowing for old one, without spending much effort.  what about (@jwodder -- any refinements/objections)

- in fixtures.py define

```
DOCKER_COMPOSE_CMD = os.environ.get('DANDI_TESTS_DOCKER_COMPOSE', 'docker compose').split(' ')
```

and use that list instead of hardcoded `'docker', 'compose'` elements.  This would allow at least for a poor me to overcome the problem locally.  Also note its presence in `DEVELOPMENT.md`.

TheChymera:
@yarikoptic oh, sorry about that, not that familiar with debian package repositories :( Ah well, since there seems to be some resistance I just went ahead and downgraded mine, you probably use it more than me and the gentoo packages are slotted, so I can have both in parallel and (hopefully) not have them interfere. I think that makes more sense than to pressure you to rely on a hack... But I really think it would be a good idea to merge this whenever everybody else is ready :3

PR #1173, opened 2022-12-13
Add validation of filepaths for non-BIDS NWB assets

### LINKED ISSUE #1163: upload doesn't seem to be checking for valid DANDI (when not BIDS) organization of files/assets
it seems that some aspects of dandiset naming/layout may be falling through the cracks. here is another example: https://dandiarchive.org/dandiset/000168/draft

as part of healthcheck if may be worthwhile considering if files should have a different name, but doesn't. this could capture a few things: 1) dandisets like this one, 2) dandisets that were uploaded in parts, and may need name resolution before upload.

jwodder:
Closes #1163.

### DIFF
--- a/dandi/organize.py
+++ b/dandi/organize.py
@@ -1032,7 +1032,11 @@ def organize(
 
 LABELREGEX = r"[^_*\\/<>:|\"'?%@;.]+"
 ORGANIZED_FILENAME_REGEX = (
-    rf"sub-{LABELREGEX}(_(ses|tis|slice|cell)-{LABELREGEX})*(_[a-z]+(\+[a-z]+)*)?\.nwb"
+    rf"sub-{LABELREGEX}"
+    rf"(_ses-{LABELREGEX})?"
+    rf"(_(tis|slice|cell|probe|obj)-{LABELREGEX})*"
+    r"(_[a-z]+(\+[a-z]+)*)?"
+    r"\.nwb"
 )
 ORGANIZED_FOLDER_REGEX = rf"sub-{LABELREGEX}"
 
--- a/dandi/organize.py
+++ b/dandi/organize.py
@@ -2,19 +2,21 @@
 ATM primarily a sandbox for some functionality for  dandi organize
 """
 
+from __future__ import annotations
+
 import binascii
 from collections import Counter
 from copy import deepcopy
 import os
 import os.path as op
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 import re
 from typing import List
 import uuid
 
 import numpy as np
 
-from . import get_logger
+from . import __version__, get_logger
 from .dandiset import Dandiset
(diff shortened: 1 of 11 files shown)


dandi/organize.py
```diff
@@ -1025,3 +1028,75 @@ def msg_(msg, n, cond=None):
         msg_(" %d invalid not considered.", skip_invalid),
         dandiset_path.rstrip("/"),
     )
+
+
+LABELREGEX = r"[^_*\\/<>:|\"'?%@;.]+"
+ORGANIZED_FILENAME_REGEX = (
+    rf"sub-{LABELREGEX}(_(ses|tis|slice|cell)-{LABELREGEX})*(_[a-z]+(\+[a-z]+)*)?\.nwb"
```
yarikoptic, line 1035: - let's make at least `_ses` strictly following `sub-` if present
- according to https://github.com/dandi/dandi-cli/blob/HEAD/dandi/organize.py#L45 is still missing `probe` and `obj`.

yarikoptic: ideally there should be a single data structure describing entities etc we should have in the filename. In principle that `potential_fields` is the one but I guess can't be used here "as is", unless you enrich that structure with regexes for the values, in particular for extensions etc.  Then this regex  can be constructed programmatically and we would avoid duplication of this knowledge.

jwodder: @yarikoptic Can tis/slice/cell/probe/obj occur without a preceding ses, or is ses a prerequisite for them?

yarikoptic: according to `potential_fields`, `ses` is not mandatory, so other entities could occur without a preceding `ses`.

jwodder: Done: 55df999.

yarikoptic:
ok, I dislike that we have duplication of specification ATM between organize and validate but ok -- let's RF it after.

Meanwhlie filed #1174 which found while test driving, and seems to work:

```
❯ dandi validate XCaMPgf/XCaMPgf_ANM471996_cell01.nwb
[NWBI.check_subject_species_form] XCaMPgf/XCaMPgf_ANM471996_cell01.nwb — Subject species 'mus musculus' should be in latin binomial form, e.g. 'Mus musculus' and 'Homo sapiens'
[NWBI.check_intracellular_electrode_cell_id_exists] XCaMPgf/XCaMPgf_ANM471996_cell01.nwb — Please include a unique cell_id associated with this IntracellularElectrode.
[DANDI.NON_DANDI_FILENAME] XCaMPgf/XCaMPgf_ANM471996_cell01.nwb — Filename does not conform to Dandi standard
[DANDI.NON_DANDI_FOLDERNAME] XCaMPgf/XCaMPgf_ANM471996_cell01.nwb — File is not in folder at root with subject name
```

although messages are a bit cryptic I guess.

PR #1176, opened 2022-12-16
Validation results refactoring: centralize results for different reused origins, use _pydantic_errors_to_validation_results more

yarikoptic:
See individual commits messages for more information.

Please check (especially @TheChymera ) if my use of `_pydantic_errors_to_validation_results` is correct -- I am not sure any test is hitting that code path at all - can someone come up with a unittest to hit it?

### DIFF
--- a/dandi/files/bases.py
+++ b/dandi/files/bases.py
@@ -146,7 +146,7 @@ class DandisetMetadataFile(DandiFile):
                 if devel_debug:
                     raise
                 return _pydantic_errors_to_validation_results(
-                    e, str(self.filepath), dandiset_path=self.dandiset_path
+                    [e], str(self.filepath), dandiset_path=self.dandiset_path
                 )
             return []
 
@@ -198,7 +198,7 @@ class LocalAsset(DandiFile):
                 if devel_debug:
                     raise
                 return _pydantic_errors_to_validation_results(
-                    e, str(self.filepath), dandiset_path=self.dandiset_path
+                    [e], str(self.filepath), dandiset_path=self.dandiset_path
                 )
             except Exception as e:
                 if devel_debug:
@@ -733,7 +733,7 @@ def _get_nwb_inspector_version():
 
 
 def _pydantic_errors_to_validation_results(
-    errors: Any[list[dict], Exception],
+    errors: list[dict | Exception],
     file_path: str,
     dandiset_path: Optional[Path] = None,
 ) -> list[ValidationResult]:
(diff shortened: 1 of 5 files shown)


jwodder:
@yarikoptic While you're messing with `_pydantic_errors_to_validation_results()`, could you please change the `Any` in the type signature to `Union`?  (The PR diff's nowhere the function definition, so I can't make this a code comment.)

dandi/files/bases.py
```diff
@@ -745,6 +735,7 @@ def _get_nwb_inspector_version():
 def _pydantic_errors_to_validation_results(
     errors: Any[list[dict], Exception],
```
jwodder, line 736:
```suggestion
    errors: list[dict | Exception],
```

jwodder, line 736:
```suggestion
    errors: list[dict | Exception],
```

dandi/files/bases.py
```diff
@@ -180,23 +197,9 @@ def get_validation_errors(
...
-                        scope=Scope.FILE,
-                        # metadata=metadata,
-                        path=self.filepath,  # note that it is not relative .path
-                        message=str(e),
-                        # TODO? dataset_path=dataset_path,
-                        dandiset_path=self.dandiset_path,
-                    )
-                ]
+                return _pydantic_errors_to_validation_results(
+                    e, str(self.filepath), dandiset_path=self.dandiset_path
```
jwodder, line 201:
```suggestion
                    [e], str(self.filepath), dandiset_path=self.dandiset_path
```

jwodder (changes requested):
Shouldn't this PR be labelled "internal" instead of "patch"?  None of these changes appear to be externally visible.

TheChymera (approved):
Looks good to me, the `partial()` definitions are pretty convenient. As for tests not hitting those two lines, I guess the “problem” is that nothing in our test suite is triggering those exceptions. This looks more like failsafe code, in case something that should never break by some coincidence does 🤔 not exactly sure how to make it break without breaking in other places before.

yarikoptic:
> Shouldn't this PR be labelled "internal" instead of "patch"? None of these changes appear to be externally visible.

my thinking was "because it changes visible to user results" due to " use _pydantic_errors_to_validation_results more". but indeed, I guess, could just be `internal`.

yarikoptic:
> This looks more like failsafe code, in case something that should never break by some coincidence does  not exactly sure how to make it break without breaking in other places before.

hm, shouldn't that `get_validation_errors` case be triggered on any asset with invalid according to pydantic model?  But let's continue on that in a separate issue

yarikoptic:
I think it might have been closed in error.  Should be redone.

yarikoptic:
too many conflicts and will be superseded by 
- https://github.com/dandi/dandi-cli/pull/1514

PR #1186, opened 2023-01-16
Fail upload only for validation ERRORs and not necessarily WARNINGs or HINTs + log all errors

### LINKED ISSUE #1185: dandi upload fails with no hints
```
$ dandi upload
2023-01-16 08:52:02,198 [    INFO] Found 2 files to consider
PATH                      SIZE       ERRORS     UPLOAD STATUS           MESSAGE                  
dandiset.yaml             1.2 kB                       skipped          should be edited online  
sub-monk-g/sub-monk-g.nwb 18.3 MB      12              skipped          failed validation        
Summary:                  18.3 MB 1 with errors        2 skipped        1 should be edited online
                                                                        1 failed validation      
2023-01-16 08:52:02,403 [    INFO] Logs saved in /Users/bendichter/Library/Logs/dandi-cli/20230116115158Z-13440.log
```

this command fails and there is no message why. I am not able to see why from the log either. `dandi validate` does not help nor does `nwbinspector`. Can we please change the upload command to output the reason the upload failed?

[206482.zip](https://github.com/dandi/dandi-cli/files/10425511/206482.zip)

[20230116115158Z-13440.log](https://github.com/dandi/dandi-cli/files/10425519/20230116115158Z-13440.log)

yarikoptic:
Reported by a user in slack to be unable to upload a file, although validation did not report any errors.  Apparently we did not log errors and it was impossible to figure out what is going on.  So I have also added logging of those validation errors, without any custom serialization into a string so we log the most complete (and possibly the most useful since provides information on paths within file etc) representation.

Closes #1185

### DIFF
--- a/dandi/upload.py
+++ b/dandi/upload.py
@@ -155,8 +155,8 @@ def upload(
                                 strpath,
                                 len(validation_errors),
                             )
-                            for i, e in enumerate(validation_errors):
-                                lgr.warning(" Error %d: %s", i + 1, e)
+                            for i, e in enumerate(validation_errors, start=1):
+                                lgr.warning(" Error %d: %s", i, e)
                             yield skip_file("failed validation")
                             return
                     else:
--- a/dandi/upload.py
+++ b/dandi/upload.py
@@ -22,6 +22,7 @@ from .files import (
 )
 from .misctypes import Digest
 from .utils import ensure_datetime, get_instance, pluralize
+from .validate_types import Severity
 
 if TYPE_CHECKING:
     from .support.typing import TypedDict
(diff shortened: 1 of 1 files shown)


dandi/upload.py
```diff
@@ -141,11 +142,21 @@ def process_path(dfile: DandiFile) -> Iterator[dict]:
...
                     # TODO: split for dandi, pynwb errors
                     if validation_errors:
                         if validation == "require":
+                            lgr.warning(
+                                "%r had %d validation errors preventing its upload:",
+                                strpath,
+                                len(validation_errors),
+                            )
+                            for i, e in enumerate(validation_errors):
+                                lgr.warning(" Error %d: %s", i + 1, e)
```
jwodder, line 159:
```suggestion
                            for i, e in enumerate(validation_errors, start=1):
                                lgr.warning(" Error %d: %s", i, e)
```

jwodder (changes requested):

yarikoptic:
> Well this is an improvement but I'd really prefer if errors preventing upload were printed to the console as opposed to the logger. A user will want to find these every time they occur.

where should they be printed in this UI e.g. while uploading 100 files and each one of them having varying number of errors?

yarikoptic:
marking for release so we some users unblocked and also release that minor tune up for zarr files pagination.
If there would be no objections , I  will merge by the end of the work day.

bendichter:
> where should they be printed in this UI e.g. while uploading 100 files and each one of them having varying number of errors?

Well ideally we'd catch these errors on the nwbinspector step, which packages these errors into a report. If the user ignores those warnings, or if for some reason there is a misalignment between `nwbinspector`, `dandi validate`, and `dandi upload`, then yes, I would like to print out hundreds of lines of errors to the console.

satra:
@bendichter - i think what yarik is getting at is that currently the upload command reports progress, which is more useful to most users, and if there are errors they are logged and one can look at those even while uploading. these are not just validation errors, they could connection timeouts and other errors that are unrelated to validation.

one option that may be useful is to summarize the current errors either on an interruption (SIGINT) or at the end of upload to the console. going through the log file is non-trivial for most users at the moment to figure out what is happening, so a helper to summarize would be useful.

yarikoptic:
Thank you for the points -- I think they are worth discussion in dedicated issues - feel welcome to open new ones to improve UX. But we do need this PR to avoid failing uploads and do log errors -- that is complimentary to summaries or other dumps visible to the user IMHO, so I will proceed as is for now.

PR #1188, opened 2023-01-18
RF/BF: make find_parent_directory_containing operate on absolute path

### LINKED ISSUE #1187: validate - Does not determine root of dandiset  when running in subfolder
```
❯ pwd
/home/yoh/proj/dandi/dandisets/000029
❯ dandi validate su*/sub-monk-g.nwb
... warnings ...
No errors found.
```

but if I cd to that asset subfolder:

```
❯ cd sub-monk-g
❯ dandi validate sub-monk-g.nwb
...
[DANDI.NO_DANDISET_FOUND] sub-monk-g.nwb — File is not inside a Dandiset
```

Initially detected/noted while looking at https://github.com/dandi/dandi-cli/issues/1185#issuecomment-1384535352

yarikoptic:
The problem was that it would not work whenever path is relative path -- it would simply not go up the hierarchy since parents would not include parent directory above current directory so we would not be able to determine the parent above CWD given relative path from CWD.

As a side effect we change return value of find_parent_directory_containing to always be absolute path

Closes #1187 

TODOs
- [x] add dedicated test for `find_parent_directory_containing` - it seems that there is none, so not surprising we had buggy implementation.  ~~@jwodder @TheChymera  could you please enrich with the test here?~~

### DIFF
--- a/dandi/utils.py
+++ b/dandi/utils.py
@@ -408,13 +408,19 @@ def find_parent_directory_containing(
 ) -> Optional[Path]:
     """Find a directory, on the path to 'path' containing filename
 
-    if no 'path' - path from cwd
-    Returns None if no such found, pathlib's Path to the directory if found
+    if no 'path' - path from cwd. If 'path' is not absolute, absolute path
+    is taken assuming relative to cwd.
+
+    Returns None if no such found, pathlib's Path (absolute) to the directory
+    if found.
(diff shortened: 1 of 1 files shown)


TheChymera (approved):
Looks good, also, I support absolute paths wherever possible :)

PR #1191, opened 2023-01-23
Fix for newer HDMF

TheChymera:
From: https://github.com/hdmf-dev/hdmf/issues/817#issuecomment-1399080575

#### Short summary (fulls story in the issue above):

`=hdmf-3.5.0` breaks our `dandi/tests/test_files.py::test_validate_bogus` test. This is because a new deletion method was introduced, which assumes all HDF objects have `_HDF5IO__open_links`, which some constructors don't create. Upstream has fixed this, but this will only take effect in future versions. However, this will still not fix our test, because upstream would prefer to still raise a warning, which pytest detects as a failure. In order to be able to use future HDMF versions, we need this PR.

@yarikoptic

### DIFF
--- a/dandi/tests/test_files.py
+++ b/dandi/tests/test_files.py
@@ -3,6 +3,7 @@ from pathlib import Path
 import subprocess
 from typing import cast
 from unittest.mock import ANY
+import warnings
 
 from dandischema.models import get_schema_version
 import numpy as np
@@ -358,7 +359,11 @@ def test_validate_bogus(tmp_path):
     # intended to produce use-case for https://github.com/dandi/dandi-cli/issues/93
     # but it would be tricky, so it is more of a smoke test that
     # we do not crash
-    errors = dandi_file(path).get_validation_errors()
+    with warnings.catch_warnings():
+        warnings.filterwarnings(
+            "ignore", message="HDF5IO was not fully initialized before close"
+        )
+        errors = dandi_file(path).get_validation_errors()
     # ATM we would get 2 errors -- since could not be open in two places,
     # but that would be too rigid to test. Let's just see that we have expected errors
     assert any(e.message.startswith("Unable to open file") for e in errors)


dandi/tests/test_files.py
```diff
@@ -358,7 +359,11 @@ def test_validate_bogus(tmp_path):
     # intended to produce use-case for https://github.com/dandi/dandi-cli/issues/93
     # but it would be tricky, so it is more of a smoke test that
     # we do not crash
-    errors = dandi_file(path).get_validation_errors()
+    with warnings.catch_warnings():
+        warnings.filterwarnings(
+            "ignore", message="HDF5IO was not fully initialized before close"
+        )
+        errors = dandi_file(path).get_validation_errors()
```
yarikoptic, line 366: please add a comment on why needs to be filtered, point to hdmf issue.

PR should be accompanied with the revert of 6d065ec76c4516d3831353ed0faded5d9c070cd7 to ensure that we test it with current/buggy hdmf

TheChymera: Commenting improved in [9ad66c4](https://github.com/dandi/dandi-cli/pull/1191/commits/9ad66c44daf7baeb24d9edb5d49ffbe45eb8e7bc).

@yarikoptic 
As for the second part, sorry if my explanation in the first post was not clear enough, but we specifically do **not** want to revert that commit. The issue is a real bug in `3.5.0`, which we cannot easily circumvent in DANDI. This PR simply leverages the upstream patch, which will become available upon the next release, hopefully one we can ask for right now if you agree this has been properly addressed for our use case (I do).

yarikoptic: Ah, I think I got it. But is this the only place where such warning might be raised? If not, we might need to add it to https://github.com/dandi/dandi-cli/blob/master/tox.ini#L48 instead of to some other tests.

TheChymera: Well, this is the only test with that issue. Not sure if it's a good idea to move the catching to `tox.ini` — as that will not be caught when running straight-up pytest... or maybe I'm misunderstanding what you're suggesting? I predominantly use pytest directly to test locally, and none of the `tox.ini` errors ever popped up 🤔

TheChymera: In any case upstream said a new patch release for this sounds good.

yarikoptic: confusingly enough pytest reads (as well) config from tox.ini, see https://docs.pytest.org/en/7.1.x/reference/customize.html#tox-ini , that is why you didn't see any error.  Ok, if you say that it is the only test -- let's proceed.

PR #1170, opened 2022-12-05
dandi-cli readme edit

melster1010:
Edit readme for stylesheet compliance/grammar/clarity/consistency.

PR #1193, opened 2023-01-26
Further improve README.md: point to specific sections of handbook, remove lengthy `dandi ls --help` example

melster1010:
address 3 items outstanding from review in PR #1170

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,98 +35,17 @@ or
 ## CLI Tool
 
 This package provides a command line utility with a basic interface
-to help you prepare and upload your data to, or obtain data from, the [DANDI archive](http://dandiarchive.org):
+to help you prepare and upload your data to, or obtain data from, the [DANDI archive](http://dandiarchive.org). 
 
-```bash
-$> dandi
-Usage: dandi [OPTIONS] COMMAND [ARGS]...
-
-  A client to support interactions with DANDI archive
-  (http://dandiarchive.org).
-
-  To see help for a specific command, run
-
-      dandi COMMAND --help
-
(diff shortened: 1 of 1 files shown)


yarikoptic:
well, although I agree that sample output of `--help` commands can get outdated, I think I do like seeing them in README.md since it gives an immediate overview of exposed by the client functionality.  If it is outdated already -- then updates are welcome for those parts.

melster1010:
> well, although I agree that sample output of `--help` commands can get outdated, I think I do like seeing them in README.md since it gives an immediate overview of exposed by the client functionality. If it is outdated already -- then updates are welcome for those parts.

After removing all of the UI graphics it does seem empty, and I understand that it would be useful to see _something_ in the readme prior to using the tool...so I restored the first UI graphic of the initial interface but left out the other one for the help output, so a compromise! :)

waxlamp (commented):
I think we should remove the listing of the dandi CLI help output and perhaps replace it with a few examples of common use cases for the tool. Let me know what you think of that.

melster1010:
> I think we should remove the listing of the dandi CLI help output and perhaps replace it with a few examples of common use cases for the tool. Let me know what you think of that.

I like that suggestion a lot, showing the UI and a couple of examples as done [here](https://github.com/francisrstokes/readme-cli). Which commands would be best to showcase, download and upload perhaps?

yarikoptic:
Attn @melster1010  -- I changed the PR description to summarize changes since it would become a part of the changelog, so should be aiming at the user/worthy appearing in changelog.

PR #1194, opened 2023-01-30
We no longer need to catch warning as hdmf has fixed the bug

TheChymera:
As per https://github.com/hdmf-dev/hdmf/pull/820

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -37,8 +37,6 @@ install_requires =
     etelemetry >= 0.2.2
     fasteners
     fscacher
-    # Pending resolution of:
-    # https://github.com/hdmf-dev/hdmf/issues/817
     hdmf != 3.5.0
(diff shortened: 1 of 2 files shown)


TheChymera:
This partial reversion of the previous commit stems from the fact that upstream in fact decided to silence the warning, which initially they said they would not.

setup.cfg
```diff
@@ -37,9 +37,7 @@ install_requires =
     etelemetry >= 0.2.2
     fasteners
     fscacher
-    # Pending resolution of:
-    # https://github.com/hdmf-dev/hdmf/issues/817
-    hdmf != 3.5.0
+    hdmf >= 3.5.1
```
yarikoptic, line 40: why do we need hdmf minimal version 3.5.1?

PR #1198, opened 2023-02-02
Move doc/design under docs/ where the rest of docs reside

yarikoptic:
in dandi-archive we have them under doc/ so I guess that is why had them under doc/ here as well, but IMHO there is no reason to split documentation here into two folders. If we are to unify, we might want to rename doc/ to docs/ in dandi-archive as well.

Alternative -- move everything under docs/* into doc/ but we might need to tune up more (RTD etc). WDYT? ping @waxlamp @satra and @dandi/dandi-cli


jwodder:
I'm not a fan of this file being in the repo in the first place.  The file is only of relevance to developers, half of it is unimplemented, and once it's all implemented, there should be no reason to keep it around.  I think it would be better for the unimplemented parts to exist in a Google doc (while still being hashed out) and/or in in GitHub issues for finally getting everything implemented.

yarikoptic:
Many files in the repo are of relevance only to developers (CI, tox.ini, etc.).  I would be ok to remove it if you wish (submit an alternative PR) - indeed it might be confusing to find discrepancies. But I do find such files useful as to help to understand why current code is what it is without doing major code archaeological expedition. May be indeed ideally someone could eventually annotated such "design docs" with what was/or not implemented - but that is extra effort nobody likes to put in ATM and thus unlikely to happen.

yarikoptic:
ok, let's move for now, and then possibly remove/harmonize across etc.

PR #1200, opened 2023-02-02
List in documentation the "high level" interfaces corresponding to CLI

yarikoptic:
as consequence, lower the other two levels to Mid/Low levels (unless there is objection)

but how to point to the specific functions (e.g. `download.download`) instead of full modref -- @jwodder could you make it happen since I have tried some ways and keep making sphinx unhappy

### DIFF
--- a/dandi/delete.py
+++ b/dandi/delete.py
@@ -188,6 +188,11 @@ def delete(
     force: bool = False,
     skip_missing: bool = False,
 ) -> None:
+    """Delete dandisets and assets from the server.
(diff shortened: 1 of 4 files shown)


jwodder:
@yarikoptic I can't figure out how to get linking to functions to work.

jwodder (changes requested):

yarikoptic:
ok, let's proceed as is for now. May be will improve (point to specific functions) later.

PR #1203, opened 2023-02-06
Merging `bids_validate` into `validate`

TheChymera:
Long time coming, a few issues to even out, this will affect both the command line interface, and internal validation as part of other commands.


TheChymera:
@jwodder having validation solely be done from within the BIDS* objects obscures the parameters space of the upstream function. That's not necessarily an issue, except that the `report` function is very useful for debugging. Does DANDI have some sort of global debug flag which I could somehow query inside the object and pass a default-value report parameter if it's true?

TheChymera:
@jwodder could you have a look at [[a072601](https://github.com/dandi/dandi-cli/pull/1203/commits/a072601f7bcb5e7ab58450efa202fba8638116d3)](https://github.com/dandi/dandi-cli/blob/d40996241575b6563722ee7fb80602cf42134e18/dandi/validate.py#L171) ? Any idea why this happens? is there some way to make it not rely on the first pass? It fails as is with one of the three broken reference datasets, if you remove the `if` statement, it will fail with all 3.

jwodder:
@TheChymera 

> Does DANDI have some sort of global debug flag

No.  Global state is evil.

jwodder:
@TheChymera I believe the caching you're seeing is due to the fact that calling `get_validation_errors()` on a BIDS object causes the validation code in [`_validate()`](https://github.com/dandi/dandi-cli/blob/67f015ccffc944b2627c6b9c6a47860c8aa164be/dandi/files/bids.py#L66) to be run, which stores its results on the `BIDSDatasetDescriptionAsset` and will not be rerun if any validation methods are called for the dataset a second time.  Does that answer your question as to what's happening?  Would it solve things if I added a method for clearing validation data?

TheChymera:
> Would it solve things if I added a method for clearing validation data?

@jwodder yes, it certainly would. However, now I'm wondering whether that's the right thing to do. The issue I see with that, is that the validator would be re-run for each file. Unless there are some very nifty caching features in Python or you'd be willing to significantly redesign the already quite complex `BIDS*` object infrastructure, the consequence of that would be that for each file, 3 files (the file + `dataset_description.json` + `README`) are validated, so it would take thrice as long as it needs to, in addition to overhead. Currently that's not an issue since the upstream validator consists of fairly simple Python code on top of blazing fast regex, but we might end up adding more features for all the intricate ways in which BIDS files are inter-connected, and things might be a lot harder to fix once we have a bunch more features on top... Is this a valid concern from your perspective?

What I was thinking is maybe I could instead add some auto-detection logic in `validate.py` so that if BIDS directories are passed, we don't do individual file iteration. This would be pretty ad-hoc, however, and much more inelegant and potentially less robust than the very nice `find_dandi_files()` logic you've constructed in `files/__init__.py` 🤔

jwodder:
@TheChymera Let's back up a bit.  Why exactly is the caching a problem?  Are you trying to validate the same dataset more than once, and, if so, why?

TheChymera:
@jwodder I am trying to adapt `bids_validate()` to the logic of the vanilla `validate()` function. It seems to run `find_dandi_files()` inside the provided directory (or directories) and then get the errors for them individually. As I understand it's the iteration in the existing function which might require running the validator for each file separately.

jwodder:
@TheChymera 

> As I understand it's the iteration in the existing function which might require running the validator for each file separately.

Can you elaborate on this?  Are you referring to this code?

https://github.com/dandi/dandi-cli/blob/67f015ccffc944b2627c6b9c6a47860c8aa164be/dandi/validate.py#L168-L171

I don't see how this sort of thing would cause a problem with caching.

TheChymera:
@jwodder yes, that is the code I am referring to. The test function for running the validator on all our example datasets and making sure we detect errors would be called with `pytest -vvs tests/test_validate.py::test_validate_bids_errors`

If you run it on the current status of the PR, it will succeed for all example error datasets save for one — without the code that skips explicit `datset_description.json` and `README` validation it will fail on all datasets, i.e. it will not detect errors.

I don't understand why exactly happens either, but I assume it has to be due to the iteration by files, it seems as soon as `df.get_validation_errors()` is called on some files individually, the results from future calls of `df.get_validation_errors()` on files from the same dataset will be unreliable.

TheChymera:
@jwodder @yarikoptic ok, I think this is my best shot. It passes all the example-based tests, but fails the tests for the object constructor. Unsure whether I broke something or the test suite should be modified as well.

dandi/files/__init__.py
```diff
@@ -128,10 +131,17 @@ def find_dandi_files(
                     # we may add later), so traverse through it as a regular
                     # directory.
                     if (p / BIDS_DATASET_DESCRIPTION).exists():
-                        bids2 = dandi_file(p / BIDS_DATASET_DESCRIPTION, dandiset_path)
-                        assert isinstance(bids2, BIDSDatasetDescriptionAsset)
-                        bidsdd = bids2
-                    path_queue.extend((q, bidsdd) for q in p.iterdir())
+                        # No nested BIDS.
+                        if not any([i in p.parents for i in bids_roots]):
```
jwodder, line 135:
```suggestion
                        if not any(i in p.parents for i in bids_roots):
```

dandi/files/__init__.py
```diff
@@ -128,10 +131,17 @@ def find_dandi_files(
                     # we may add later), so traverse through it as a regular
                     # directory.
                     if (p / BIDS_DATASET_DESCRIPTION).exists():
-                        bids2 = dandi_file(p / BIDS_DATASET_DESCRIPTION, dandiset_path)
-                        assert isinstance(bids2, BIDSDatasetDescriptionAsset)
-                        bidsdd = bids2
-                    path_queue.extend((q, bidsdd) for q in p.iterdir())
+                        # No nested BIDS.
```
jwodder, line 134: Please explain this.  Are BIDS datasets not allowed to be contained within other BIDS datasets?  I thought they were.

dandi/tests/test_validate.py
```diff
@@ -1,34 +1,63 @@
...
+    selected_dataset, error_file = (
+        "invalid_asl003",
+        "sub-Sub1/perf/sub-Sub1_headshape.jpg",
     )
 
-    # Check if a report is being produced.
-    assert len(glob(report_path)) == 1
+    bids_file_path = os.path.join(bids_error_examples, selected_dataset, error_file)
+    bids_dataset_path = pathlib.Path(
+        os.path.join(bids_error_examples, selected_dataset)
```
jwodder, line 49:
```suggestion
        bids_error_examples, selected_dataset
```

No need to `join` inside the `Path` constructor.

dandi/tests/test_validate.py
```diff
@@ -39,9 +68,15 @@ def test_validate_bids_errors(bids_error_examples, dataset):
     # ideally make a list and erode etc.
 
     selected_dataset = os.path.join(bids_error_examples, dataset)
-    validation_result = validate_bids(selected_dataset, report=True)
+    validation_result = validate(selected_dataset)
     with open(os.path.join(selected_dataset, ".ERRORS.json")) as f:
         expected_errors = json.load(f)
+    validation_result = [i for i in validation_result]
```
jwodder, line 74: This line accomplishes nothing but wasted cycles.

jwodder (changes requested):

jwodder:
@TheChymera 

> I'm also thinking whether we might want to deprecate `BIDSDatasetDescriptionAsset.dataset_files` since there is no way for us to determine a list of files without using or leveraging the bidsschematools validator — we'd either need to reimplement the logic (which is slated to become more complex as upstream includes support for .bidsignore) or directly call the [private upstream function](https://github.com/bids-standard/bids-specification/blob/68f91ca94dc8940bba3e9949b5a4d9091c2a8bba/tools/schemacode/bidsschematools/validator.py#L25) which does file selection.

* Are we planning on excluding bidsignored-files when uploading a Dandiset?  (cc @yarikoptic)  If so, that upstream function's going to have to go public.
* Exactly what steps do you have in mind for "deprecating" `dataset_files`?  My definition of that word does not entail removing any tests until the feature is removed.  If you're planning on entirely tearing it out in this PR, you should also remove the weakref wrapper around `BIDSDatasetDescriptionAsset` in `BIDSAsset`.

yarikoptic:
> * Are we planning on excluding bidsignored-files when uploading a Dandiset?  (cc @yarikoptic)  If so, that upstream function's going to have to go public.

good question. I think we could as well indeed ignore to pick them up automatically whenever uploading if those paths are not explicitly specified and `--allow-any-path` is given.  So similarly to how we ignore non-nwb/video files by default AFAIK.

Rationale: I think it is Ok to add a little of burden upon user to go extra "mile" if they want to upload some data which is not "valid". Otherwise (if it is just easy) we end up with cases like could be found in openneuro archive where people simply ignore entire subject folders to pacify the bids-validator upon upload -- that must be really discouraged.

WDYT @satra?

dandi/files/__init__.py
TheChymera: They can, but if they are they are what's known as “derivatives”. Those have different and overall far more permissive and less agreed on rules for validation. Might be worth considering whether we want to support these at some point, but currently the upstream validator doesn't support them, so we shouldn't try to either.

TheChymera:
@jwodder @yarikoptic 

#### For this PR

>  If so, that upstream function's going to have to go public.

> Exactly what steps do you have in mind for "deprecating" dataset_files? 

Not necessarily, my idea was to maybe have `dataset_files` populated automatically after the validation is run on the BIDS root. That way we get the list of what counts as a BIDS dataset file straight from the horse's mouth. No need to first run `bst.validator._get_paths` and then again `bst.validator.validate_bids`, which runs the previous function again.

#### Ideally not for this PR

> `--allow-any-path`

There's no flag upstream for turning off `.bidsignore` support, and I don't think there should be one. You want it off, you `rm` the file. The purpose of `.bidsignore` is to tell the validator to stop nagging you about things you are aware of, most commonly the use case is that you're developing a new BEP, but indeed it can be arbitrarily abused. I think we should have a separate flag, e.g. `--allow-any-path` to just bypass validation entirely, and keep that separate from the `.bidsignore` logic. But let's maybe discuss this in another PR or issue. I just brought it up as an example for why BIDS-agnostic file detection is reaching its limits.

dandi/tests/test_validate.py
TheChymera: @jwodder I had hoped so too, but since we're now using the unified validation function with the `yield` return, it's actually needed:

```
(dev) [deco]~/src/dandi-cli/dandi ❱ pytest -vvs tests/test_validate.py::test_validate_bids_errors
=========================================================== test session starts ============================================================
platform linux -- Python 3.10.10, pytest-7.2.1, pluggy-1.0.0 -- /usr/bin/python3.10
cachedir: .pytest_cache
rootdir: /home/chymera/src/dandi-cli, configfile: tox.ini
plugins: httpserver-1.0.6, mock-3.10.0, pkgcore-0.12.19
collected 3 items

tests/test_validate.py::test_validate_bids_errors[invalid_asl003] Cloning into '/tmp/pytest-of-chymera/pytest-361/gitrepo0'...
remote: Enumerating objects: 145, done.
remote: Counting objects: 100% (145/145), done.
remote: Compressing objects: 100% (121/121), done.
remote: Total 145 (delta 20), reused 136 (delta 20), pack-reused 0
Receiving objects: 100% (145/145), 363.77 KiB | 3.06 MiB/s, done.
Resolving deltas: 100% (20/20), done.
FAILED
tests/test_validate.py::test_validate_bids_errors[invalid_eeg_cbm] FAILED
tests/test_validate.py::test_validate_bids_errors[invalid_pet001] FAILED

================================================================= FAILURES =================================================================
________________________________________________ test_validate_bids_errors[invalid_asl003] _________________________________________________
tests/test_validate.py:76: in test_validate_bids_errors
    assert len(validation_result) > 0
E   TypeError: object of type 'generator' has no len()
________________________________________________ test_validate_bids_errors[invalid_eeg_cbm] ________________________________________________
tests/test_validate.py:76: in test_validate_bids_errors
    assert len(validation_result) > 0
E   TypeError: object of type 'generator' has no len()
________________________________________________ test_validate_bids_errors[invalid_pet001] _________________________________________________
tests/test_validate.py:76: in test_validate_bids_errors
    assert len(validation_result) > 0
E   TypeError: object of type 'generator' has no len()
=========================================================== slowest 10 durations ===========================================================
0.61s setup    dandi/tests/test_validate.py::test_validate_bids_errors[invalid_asl003]
0.00s setup    dandi/tests/test_validate.py::test_validate_bids_errors[invalid_eeg_cbm]
0.00s teardown dandi/tests/test_validate.py::test_validate_bids_errors[invalid_pet001]
0.00s setup    dandi/tests/test_validate.py::test_validate_bids_errors[invalid_pet001]
0.00s call     dandi/tests/test_validate.py::test_validate_bids_errors[invalid_asl003]
0.00s teardown dandi/tests/test_validate.py::test_validate_bids_errors[invalid_asl003]
0.00s teardown dandi/tests/test_validate.py::test_validate_bids_errors[invalid_eeg_cbm]
0.00s call     dandi/tests/test_validate.py::test_validate_bids_errors[invalid_eeg_cbm]
0.00s call     dandi/tests/test_validate.py::test_validate_bids_errors[invalid_pet001]
========================================================= short test summary info ==========================================================
FAILED tests/test_validate.py::test_validate_bids_errors[invalid_asl003] - TypeError: object of type 'generator' has no len()
FAILED tests/test_validate.py::test_validate_bids_errors[invalid_eeg_cbm] - TypeError: object of type 'generator' has no len()
FAILED tests/test_validate.py::test_validate_bids_errors[invalid_pet001] - TypeError: object of type 'generator' has no len()
============================================================ 3 failed in 1.07s =============================================================
(dev) [deco]~/src/dandi-cli/dandi ❱ head -76 tests/test_validate.py | tail -5
    with open(os.path.join(selected_dataset, ".ERRORS.json")) as f:
        expected_errors = json.load(f)

    # We know that these datasets contain errors.
    assert len(validation_result) > 0
```

jwodder: So if we're just converting a generator to a list, write it as `validation_result = list(validation_result)` so it looks less ridiculous.

dandi/files/__init__.py
jwodder: I've pushed a commit (ccc8d79) that updates the code & tests to treat `dataset_description.json` files beneath other `dataset_description.json` files as generic BIDS assets that don't start new BIDS datasets.  Also, `find_bids_dataset_description()` now finds the topmost `dataset_description.json` file.

jwodder:
@TheChymera Only populating `dataset_files` via validation doesn't seem like a good idea to me.  At the very least, it'd cause problems when uploading with validation disabled.

dandi/files/__init__.py
TheChymera: Thank you :)

dandi/validate.py
```diff
@@ -150,6 +143,7 @@ def validate(
     schema_version: Optional[str] = None,
     devel_debug: bool = False,
     allow_any_path: bool = False,
+    report: bool = False,
```
jwodder, line 146: This parameter appears to be unused.

dandi/validate.py
```diff
@@ -53,12 +47,11 @@ def validate_bids(
...
 
-    if report and not report_path:
-        log_dir = appdirs.user_log_dir("dandi-cli", "dandi")
-        report_path = "{log_dir}/bids-validator-report_{{datetime}}-{{pid}}.log"
-        report_path = report_path.format(
-            log_dir=log_dir,
-        )
+    log_dir = appdirs.user_log_dir("dandi-cli", "dandi")
+    report_path = "{log_dir}/bids-validator-report_{{datetime}}-{{pid}}.log"
+    report_path = report_path.format(
```
jwodder, line 52: Do we ever notify the user about the existence of this report?

jwodder (changes requested):

dandi/validate.py
yarikoptic: as before I would also argue to make it optional. FWIW we already store the log file for the run, so if desired can log what is desired to be said.

TheChymera: @yarikoptic it's a bit more difficult now, since the validation is run by a method of the object. Not sure whether we should be burdening the object with this method parameter. The default DANDI logging does not log the validator report as that's just a dictionary produced and optionally written to a file by an external function. We could disable it entirely, but that means if we have issues, we need to ask people to run the `bst validate` command directly to get the actual report listing all the comparisons.

TheChymera:
@jwodder @yarikoptic looking good now?

jwodder (approved):

dandi/tests/fixtures.py
```diff
@@ -269,6 +269,14 @@ def fixture(tmp_path_factory: pytest.TempPathFactory) -> str:
...
         lgr.debug("Cloning %r into %r", url, path)
         run(["git", "clone", "--depth=1", url, path], check=True)
+        for i in os.listdir(path):
+            bids_dataset_path = os.path.join(path, i)
+            if os.path.isdir(bids_dataset_path):
+                dandiset_metadata_file_path = os.path.join(
+                    bids_dataset_path, dandiset_metadata_file
+                )
+                with open(dandiset_metadata_file_path, "w") as f:
+                    f.write(" \n")
```
yarikoptic, line 279: this is a generic `get_gitrepo_fixture`. I really would prefer it to do exactly that and not anyhow change that cloned git repo unconditionally !  It feels like you want to create some tuned up fixture based on this, called it some descriptive name/docstring on what side effect it creates on top of a clone.  But also not clear to me what is the goal here -- make every somefolder into a dandiset there?

or may be add it as an option to this fixture, e.g. `make_subfolders_into_dandisets=False` (or whatever would fit the purpose here) and enable thiscode block conditionally on that.

dandi/tests/fixtures.py
```diff
@@ -303,7 +311,15 @@ def fixture(
...
         run(["git", "sparse-checkout", "set"] + whitelist, cwd=path, check=True)
-        yield str(path)
+        for i in os.listdir(path):
+            bids_dataset_path = os.path.join(path, i)
+            if os.path.isdir(bids_dataset_path):
+                dandiset_metadata_file_path = os.path.join(
+                    bids_dataset_path, dandiset_metadata_file
+                )
+                with open(dandiset_metadata_file_path, "w") as f:
+                    f.write(" \n")
```
yarikoptic, line 321: 1. see comment above in another similar fixture
2. I have severe allergy to code duplication. Please make a helper function (e.g. `_make_each_folder_into_dandiset(path)` or alike) and use it in those two spots.

dandi/tests/test_files.py
```diff
@@ -236,11 +236,11 @@ def test_find_dandi_files_with_bids(tmp_path: Path) -> None:
             dandiset_path=tmp_path,
             bids_dataset_description_ref=ANY,
         ),
-        BIDSDatasetDescriptionAsset(
+        GenericBIDSAsset(
```
yarikoptic, line 239: so nested BIDS dataset would not be considered a bids dataset? (nothing in PR description hints on the change in behavior for such aspect)

dandi/validate.py
yarikoptic: > The default DANDI logging does not log the validator report as that's just a dictionary produced and optionally written to a file by an external function

we get to validation in 2 cases:
- direct use of `dandi validate` -- there we render/output results AFAIK
- optional validation during `upload` -- there I believe we do logging at https://github.com/dandi/dandi-cli/blob/master/dandi/upload.py#L161 

am I missing some other use case where we should log bids-validator output in custom/separate file ?

dandi/tests/test_validate.py
```diff
@@ -1,34 +1,61 @@
...
-    validation_result = validate_bids(selected_dataset, report=True)
+    validation_result = validate(selected_dataset)
     for i in validation_result:
         assert i.severity is None
 
 
-def test_report_path(bids_examples, tmp_path):
+def test_validate_bids_onefile(bids_error_examples, tmp_path):
+    """
+    Dedicated test using to single-file validation.
```
yarikoptic, line 31: note to myself: unlike `nose`, `pytest` doesn't use docstring as the "name" of the test while running testing so there is no obfuscation and we are indeed "open" to provide docstrings giving details on tests. good.

dandi/cli/cmd_validate.py
```diff
@@ -9,56 +9,6 @@
...
-    from ..validate import validate_bids as validate_bids_
-
-    validator_result = validate_bids_(  # Controller
-        *paths,
-        report=report,
-        report_path=report_path,
-        schema_version=schema,
-    )
-
-    _process_issues(validator_result, grouping)
```
yarikoptic, line 59: Let's do proper deprecation cycle, to assist users in not requiring to check CHANGELOG which would not tell much about disappearance of `bids-validate` command (besides that it is merged):
- here if invoked issue a DeprecationWarning alerting users to use `validate` command instead
- if `report` option is provided, say that there is no such option any longer -- people would need to just redirect output of validator to a file if desire to store it etc.

and then invoke the "merged" `validate` itself with corresponding options.

yarikoptic (changes requested):
Looks monumental and great to see tests passing, well done!  A few overall points
- even though we AFAIK do not have yet "nested BIDS" datasets, changes to logic relating to those are not clear to me
- duplication and unconditional change of fixtures behavior should be avoided
- I remain of opinion of not needing an extra log file for validation of bids
- We better make transition smoother for the users, thus retain `validate-bids` for a bit longer with a DeprecationWarning but invoking `validate` logic

dandi/files/__init__.py
```diff
@@ -115,9 +117,10 @@ def find_dandi_files(
                 lgr.warning("%s: Ignoring unsupported symbolic link to directory", p)
             elif dandiset_path is not None and p == Path(dandiset_path):
                 if (p / BIDS_DATASET_DESCRIPTION).exists():
-                    bids2 = dandi_file(p / BIDS_DATASET_DESCRIPTION, dandiset_path)
-                    assert isinstance(bids2, BIDSDatasetDescriptionAsset)
-                    bidsdd = bids2
+                    bids = dandi_file(p / BIDS_DATASET_DESCRIPTION, dandiset_path)
+                    assert isinstance(bids, BIDSDatasetDescriptionAsset)
+                    bidsdd = bids
```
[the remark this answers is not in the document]
TheChymera: yes, that is correct. The renaming was part of the refactoring to better handle multiple BIDS at the same hierarchical level, while avoiding nested BIDS (which are always derivatives). The name change just made more sense in light of that.

dandi/files/__init__.py
```diff
@@ -127,10 +130,13 @@ def find_dandi_files(
                     # (ie., it's not a Zarr or any other directory asset type
                     # we may add later), so traverse through it as a regular
                     # directory.
-                    if (p / BIDS_DATASET_DESCRIPTION).exists():
+                    if (p / BIDS_DATASET_DESCRIPTION).exists() and not any(
+                        i in p.parents for i in bids_roots
```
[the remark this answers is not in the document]
TheChymera: 1. The path queue is extended as part of the while loop, it will always be traversed top-down.
2. Nested BIDS are currently either invalid or derivatives. Support for derivatives is limited in the schema, and absent in the upstream validator. Even if we at some point support derivatives, that will require an object structure redesign. Currently if we validate them, they will be validated as normal BIDS and will fail.

dandi/files/__init__.py
```diff
@@ -105,7 +105,9 @@ def find_dandi_files(
                 raise ValueError(
                     "Path {str(p)!r} is not inside Dandiset path {str(dandiset_path)!r}"
                 )
-        path_queue.append((p, None))
+        bidsdd = None
+        path_queue.append((Path(p), bidsdd))
```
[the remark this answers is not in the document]
TheChymera: yes, you are right, fixed in d962e7d8c1656f86664fbd190771a96e6e0e2d8e

dandi/tests/fixtures.py
TheChymera: done in 6bfb52045dfcd5cef1e5ad0d681c439fce324056

dandi/tests/fixtures.py
TheChymera: ok, same as above done in 6bfb52045dfcd5cef1e5ad0d681c439fce324056

dandi/tests/test_files.py
TheChymera: True, though a lot of stuff has happened that's not in the PR description. I added that one though, since it is one of the bigger things.

dandi/tests/fixtures.py
```diff
@@ -257,6 +259,7 @@ def get_gitrepo_fixture(
     url: str,
     committish: Optional[str] = None,
     scope: Scope = "session",
+    make_subdirs_dandisets: Optional[bool] = False,
```
jwodder, line 262: `Optional` does not mean "optional argument"; it means "can be `None`," and I don't see a reason to support that here.  Get rid of the `Optional` here and below.

dandi/tests/fixtures.py
```diff
@@ -52,6 +52,8 @@
     "qmri_vfa",
 ]
 
+BIDS_ERROR_TESTDATA_SELECTION = ["invalid_asl003", "invalid_pet001"]
```
jwodder, line 55: What happened to `"invalid_eeg_cbm"`?  That was one of the `dataset` values passed to `test_validate_bids_error` and `test_validate_bids_errors` prior to 6bfb52045dfcd5cef1e5ad0d681c439fce324056.

dandi/cli/cmd_validate.py
TheChymera: @yarikoptic tried to bring it back in [60f7f97](https://github.com/dandi/dandi-cli/pull/1203/commits/60f7f972487317e3c45a448b4db09a23a0046578) . For some reason it doesn't work. I think Python might be getting confused and think that `validate` is a module:

```
(dev) [deco]~/src/dandi-cli ❱ dandi validate-bids ~/src/bids-examples/asl003/
2023-02-24 13:22:13,031 [ WARNING] A newer version (0.49.1) of dandi/dandi-cli is available. You are using 0.49.0+51.g60f7f97
(*args: Any, **kwargs: Any) -> Any
2023-02-24 13:22:13,038 [    INFO] Logs saved in /home/chymera/.cache/dandi-cli/log/20230224182212Z-16933.log
Error: Context.__init__() got an unexpected keyword argument 'grouping'
```

Though I renamed it as well, and that still happens. Again these are just a few lines but I tried a buch of different fixes. Are you sure this is worth the extra effort in figuring it out? we could just make it do nothing and issue a warning...

dandi/tests/fixtures.py
TheChymera: we don't need it, it's just wasting time on more tests for the same sort of things, and as the `bids_error_examples` repository grows, it's best to already have filtering and keep the subset we use in one place rather than further down in the test function.

dandi/cli/cmd_validate.py
TheChymera: @yarikoptic tried to bring it back in [60f7f97](https://github.com/dandi/dandi-cli/pull/1203/commits/60f7f972487317e3c45a448b4db09a23a0046578) . For some reason it doesn't work. I think Python might be getting confused and think that `validate` is a module:

```
(dev) [deco]~/src/dandi-cli ❱ dandi validate-bids ~/src/bids-examples/asl003/
2023-02-24 13:22:13,031 [ WARNING] A newer version (0.49.1) of dandi/dandi-cli is available. You are using 0.49.0+51.g60f7f97
(*args: Any, **kwargs: Any) -> Any
2023-02-24 13:22:13,038 [    INFO] Logs saved in /home/chymera/.cache/dandi-cli/log/20230224182212Z-16933.log
Error: Context.__init__() got an unexpected keyword argument 'grouping'
```

Though I renamed it as well, and that still happens. Again these are just a few lines but I tried a buch of different fixes. Are you sure this is worth the extra effort in figuring it out? we could just make it do nothing and issue a warning...

dandi/tests/fixtures.py
TheChymera: ok, fixed.

TheChymera:
@yarikoptic any idea why this test fails all of a sudden? I can't reproduce the failure locally.

dandi/cli/cmd_validate.py
```diff
@@ -40,23 +41,24 @@ def validate_bids(
...
-        schema_version=schema,
+    warnings.warn(
+        "The `dandi validate-bids` command line interface is deprecated, you can use "
+        "`dandi validate` instead. Proceeding to parse the call to `dandi validate` now.",
+        DeprecationWarning,
     )
+    import inspect
 
-    _process_issues(validator_result, grouping)
+    from .cmd_validate import validate as validate_
```
jwodder, line 58: You're importing from `cmd_validate` within `cmd_validate.py`.  There's no reason to do that.

dandi/cli/cmd_validate.py
```diff
@@ -40,23 +41,24 @@ def validate_bids(
...
+
+    print(inspect.signature(validate_))
+    validate_(paths, grouping=grouping)
```
jwodder, line 61: `validate` is a Click command object, not a normal function.  You need to invoke it as described at https://click.palletsprojects.com/en/8.1.x/advanced/#invoking-other-commands.

jwodder (changes requested):

dandi/cli/cmd_validate.py
TheChymera: this was an attempted fix for this → https://github.com/dandi/dandi-cli/pull/1203#discussion_r1117474990

dandi/cli/cmd_validate.py
TheChymera: this I think addresses the former point as well :)

TheChymera: ok, I think I solved it, thank you.

dandi/cli/cmd_validate.py
TheChymera: ok, fixed in [06e2a35](https://github.com/dandi/dandi-cli/pull/1203/commits/06e2a354df922b2c31607e1f200dc92f2ab0cde0)

TheChymera:
@yarikoptic looking good?

dandi/cli/cmd_validate.py
```diff
@@ -31,32 +32,31 @@
...
 ):
     """Validate BIDS paths.
-
     Notes
     -----
-    Used from bash, eg:
-        dandi validate-bids /my/path
+    * Used from bash, eg:
+    dandi validate-bids /my/path
+    * Deprecated
```
yarikoptic, line 50:
```suggestion
    * DEPRECATED: use  dandi validate /my/path
```

dandi/cli/cmd_validate.py
```diff
@@ -31,32 +32,31 @@
...
     """
 
-    from ..validate import validate_bids as validate_bids_
-
-    validator_result = validate_bids_(  # Controller
-        *paths,
-        report=report,
-        report_path=report_path,
-        schema_version=schema,
+    warnings.filterwarnings("default")
```
yarikoptic, line 53: interesting -- without this I was shown no `DeprecationWarning`, odd...  @jwodder - do you know why?

dandi/tests/fixtures.py
```diff
@@ -303,18 +309,34 @@ def fixture(
...
 
+def _make_subdirs_dandisets(path: str) -> None:
+    for i in os.listdir(path):
+        bids_dataset_path = os.path.join(path, i)
+        if os.path.isdir(bids_dataset_path):
+            dandiset_metadata_file_path = os.path.join(
+                bids_dataset_path, dandiset_metadata_file
+            )
+            with open(dandiset_metadata_file_path, "w") as f:
+                f.write(" \n")
```
yarikoptic, line 327: Please use pathlib more -- it should make code more concise and readable (especially if long variable names used), e.g. (untested, fix if needed) here:

```suggestion
        bids_dataset_path = Path(path) / i
        if bids_dataset_path.isdir():
            (bids_dataset_path / dandiset_metadata_file).write_text(" \n")
```
is much easier to read IMHO.

dandi/files/__init__.py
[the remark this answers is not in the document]
yarikoptic: IMHO it is ok to fail them if they do not follow the schema of common derivatives which is part of the BIDS. 

- do we have a use case which somehow demands us to pretend that nested BIDS is not BIDS?
- if not -- I would prefer for us to stay consistent and treat BIDS datasets as BIDS datasets, nested they are or not because otherwise it would just then need to treat every file under derivatives as either "I don't care legit " or "illegal" -- situation is nohow better IMHO.

yarikoptic (changes requested):
minor code tune up + bigger question again on why there is a change in treatment of (possibly) nested BIDS datasets. Is it somehow demanded by the use case or some other cause?  if not - we better keep them as bids datasets and validate them, even though indeed might fail now.

dandi/cli/cmd_validate.py
TheChymera: yep. took me a while to figure it out as well...

TheChymera:
@yarikoptic The upstream `bst` validator does not currently validate derivatives, it just ignores it. If someone has derivs (some of our test data has derivs) it will pick it up as a separate non-deriv dataset and fail. I can make this the next feature we work on, but it will have to be done on the `bst` end and the nested-ness will be handled there.

yarikoptic:
ok, let's proceed! Thank you @TheChymera and @jwodder !

PR #1204, opened 2023-02-06
Improve messages displayed when an upload fails

### LINKED ISSUE #1197: Provide warning log message with instructions if upload fails
People keep running into situations where more recent version of dandi (or dependency) might have helped. I don't think we have to generally mandate (ie error out) if client version is not the most recent - new version might not only be fixing but also could introduce bugs, and trying to address one issue people might introduce another. Also this would entirely prevent inclusion into stable distributions.
We do announce minimal compatible version in archive api and some times do boost it up.

We announce availability of new version on top of run, but it gets lost in the terminal output.

To improve feedback to the user, if any upload fails - exit with non-O. If existing with non-0, check version again - if newer one available - add warning level message suggesting to update since might potentially address the issue. It will be suggestive.

Also if validation fails during upload - log warning to check the log (may be hint which line to look for) for details.

jwodder:
Closes #1197.

### DIFF
--- a/dandi/upload.py
+++ b/dandi/upload.py
@@ -336,8 +336,9 @@ def upload(
                 if Version(latest_version) > Version(__version__):
                     lgr.warning(
                         "Upload failed, and you are not using the latest"
-                        " version of dandi.  Suggest upgrading dandi and trying"
-                        " again."
+                        " version of dandi.  We suggest upgrading dandi to v%s"
+                        " and trying again.",
+                        latest_version,
                     )
             sys.exit(1)
 
--- a/dandi/upload.py
+++ b/dandi/upload.py
@@ -3,15 +3,17 @@ from functools import reduce
 import os.path
 from pathlib import Path
 import re
+import sys
 import time
 from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional, Set, Tuple, Union
 
 import click
+from packaging.version import Version
 
-from . import lgr
+from . import __version__, lgr
 from .consts import DRAFT, dandiset_identifier_regex, dandiset_metadata_file
 from .dandiapi import RemoteAsset
-from .exceptions import NotFoundError
+from .exceptions import NotFoundError, UploadError
 from .files import (
     DandiFile,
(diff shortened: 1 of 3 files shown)


dandi/upload.py
```diff
@@ -314,9 +317,30 @@ def upload_agg(*ignored: Any) -> str:
...
+
+                latest_version = etelemetry.get_project("dandi/dandi-cli")["version"]
+            except Exception:
+                pass
+            else:
+                if Version(latest_version) > Version(__version__):
+                    lgr.warning(
+                        "Upload failed, and you are not using the latest"
+                        " version of dandi.  Suggest upgrading dandi and trying"
+                        " again."
```
yarikoptic, line 340:
```suggestion
                        "Upload failed, and you are not using the latest"
                        " version of dandi.  We suggest upgrading dandi to "
                        f" {latest_version} and trying again."
```

jwodder: Done in 5f1727d.

dandi/upload.py
```diff
@@ -314,9 +317,30 @@ def upload_agg(*ignored: Any) -> str:
...
+                    )
+            sys.exit(1)
```
yarikoptic, line 343: could we `exit(1)` at `cmd_upload.py` level, not here, which is just a "high level" Python API... I think we do not even need to exit(1) explicitly and our `click` wrapper would do it.  Let's just (re)raise `UploadError` exception here?

jwodder: The `UploadError` messages are all rather terse and don't exactly make for user-friendly exit messages.

yarikoptic: Agree, but for now I think it would suffice.  It would still be better than throwing SystemExit into user while calling `upload` function somewhere within IPython notebook.

jwodder: Done in b6b9195.

yarikoptic:
Thank you!  I will find some time later to give it a run to get "the feel" for it.

yarikoptic:
@bendichter do you have some favorite failing to validate/upload dataset to try this on?

PR #1207, opened 2023-02-07
Fix type errors; unbound mypy

jwodder:
This silences some typing errors related in some inscrutable way to the latest version of `retrying`.  It also unrestricts the maximum mypy version so that we can use the newly-released v1.0.


yarikoptic:
Thank you!

PR #1209, opened 2023-02-08
BF+RF validation

yarikoptic:
Might still be incomplete since I left  in place the odd dichotomy of results in validation depending on either schema_version is provided or not, as emphasized in the tests at https://github.com/dandi/dandi-cli/blob/HEAD/dandi/tests/test_files.py#L313 . @jwodder @TheChymera do you remember a reason why we did it this way?

meanwhile I made `LocalAsset` to always validate against schema even if a `schema_version` is not provided and that might be going against some logic which is pointed to above.  But I think we should simplify/harmonize it here one way or another: we might want to always validate against everything possible and then results just filtered, or would need to pass some *filtering* option (e.g. listing validation `id` prefixes thus to enable only some) to select which validators to use.  So if no objections/concerns would be stated, I would proceed to remove dichotomy in NWBAsset validation and always include super's validation (thus validation against schema).

Then I fixed up validation of zarrs in that their metadata records should be populated with zarr specific  dummy checksum (previously tests would not even get to that case I believe).

### DIFF
--- a/dandi/files/bases.py
+++ b/dandi/files/bases.py
@@ -470,6 +470,12 @@ class NWBAsset(LocalFileAsset):
         schema_version: Optional[str] = None,
         devel_debug: bool = False,
     ) -> list[ValidationResult]:
+        """Validate NWB asset
+
+        If `schema_version` was provided, we only validate basic metadata,
+        and completely skip validation using nwbinspector.inspect_nwb
+
+        """
         errors: list[ValidationResult] = pynwb_validate(
             self.filepath, devel_debug=devel_debug
         )
--- a/dandi/files/bases.py
+++ b/dandi/files/bases.py
@@ -168,64 +168,62 @@ class LocalAsset(DandiFile):
         schema_version: Optional[str] = None,
         devel_debug: bool = False,
     ) -> list[ValidationResult]:
-        if schema_version is not None:
-            current_version = get_schema_version()
-            if schema_version != current_version:
-                raise ValueError(
-                    f"Unsupported schema version: {schema_version}; expected {current_version}"
+        current_version = get_schema_version()
+        if schema_version is None:
+            schema_version = current_version
+        if schema_version != current_version:
+            raise ValueError(
+                f"Unsupported schema version: {schema_version}; expected {current_version}"
+            )
+        try:
+            asset = self.get_metadata(digest=DUMMY_DIGEST)
+            BareAsset(**asset.dict())
+        except ValidationError as e:
+            if devel_debug:
+                raise
+            # TODO: how do we get **all** errors from validation - there must be a way
+            return [
+                ValidationResult(
+                    origin=ValidationOrigin(
+                        name="dandischema",
+                        version=dandischema.__version__,
+                    ),
+                    severity=Severity.ERROR,
+                    id="dandischema.TODO",
+                    scope=Scope.FILE,
+                    # metadata=metadata,
+                    path=self.filepath,  # note that it is not relative .path
+                    message=str(e),
+                    # TODO? dataset_path=dataset_path,
+                    dandiset_path=self.dandiset_path,
                 )
-            try:
-                asset = self.get_metadata(digest=DUMMY_DIGEST)
-                BareAsset(**asset.dict())
-            except ValidationError as e:
-                if devel_debug:
-                    raise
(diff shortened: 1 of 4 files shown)


dandi/files/bases.py
```diff
@@ -743,12 +739,15 @@ def _get_nwb_inspector_version():
 
 
 def _pydantic_errors_to_validation_results(
-    errors: Any[list[dict], Exception],
+    errors: list[dict | Exception] | ValidationError,
     file_path: str,
```
jwodder, line 743: Since the only use of `file_path` in this function converts it to a `Path`, it would be better to make this argument a `Path` and remove the `str()` call when passing it.

dandi/files/bases.py
```diff
@@ -488,6 +476,12 @@ def get_validation_errors(
         schema_version: Optional[str] = None,
         devel_debug: bool = False,
     ) -> list[ValidationResult]:
+        """Validate NWB asset
+
+        If `schema_version` was provided, we only validate basic metadata,
```
jwodder, line 481:
```suggestion
        If ``schema_version`` was provided, we only validate basic metadata,
```

One backtick for things that can be linked to (like classes and functions), two backticks for other code.

jwodder (changes requested):

dandi/files/bases.py
yarikoptic: indeed! thanks, done in b6f81807d5c76eec8f3d5df8a22633a7dc15351b

yarikoptic:
Thanks @jwodder for the review.  Before I embark on further RF journey -- do you remember/have an idea why we had that different behavior depending on `schema_version` I mentioned/changed here for the LocalAsset?

dandi/files/zarr.py
```diff
@@ -180,6 +182,9 @@ def get_metadata(
     ) -> BareAsset:
         metadata = get_default_metadata(self.filepath, digest=digest)
         metadata.encodingFormat = ZARR_MIME_TYPE
+        # TODO: .size obtained via get_default_metadata would be the one
+        # from os.stat and thus not reflective of actual size of the zarr
+        # folder.
```
jwodder, line 187: Actually, for Zarrs, the size is extracted from the digest:

https://github.com/dandi/dandi-cli/blob/4920831257762c4bb039db2f682ae24549292c68/dandi/metadata.py#L993-L998

jwodder:
@yarikoptic I think the goal of `schema_version` was to use the old, "flat" metadata if no `schema_version` was set, and to use the dandischema classes if it was set.  It was added during the transition to the dandischema classes.

yarikoptic:
Thanks! that aligns with my observation on behavior of metadata extraction on dandiset to use `_check_required_fields`... I think I will then continue RFing in the follow up PR to make it always work with dandischema classes, and will merge this one meanwhile unless you would recommend to grow this single PR @jwodder .

dandi/files/zarr.py
yarikoptic: oh, missed that, thanks! I will remove the TODO but might RF to move this zarr specific code into zarr specific place ;)

TheChymera (approved):
Looked through the diff, nothing I can see wrong with it. However, the schema specification dichotomy wasn't my design decision, and I'm unsure why it's done this way, so with respect to that me not seeing anything wrong might be of limited value.

One thing which I do see, is a bit of extra steps for string/Path object conversion. We have a bunch of gotchas like that where some things take strings and some others Path objects. Probably not worth dealing with in this PR, just putting it out there. Maybe paths should always be strings, as they always are in GNU utils, and the objects should be created only if we ever need the advanced features like `.relative_to()` which can't be easily done with `os.path`.

jwodder:
@TheChymera 

> Maybe paths should always be strings

I disagree.  Paths should be `Path`s, in order to make it clear that they're paths and in order to take advantage of the simpler & more powerful `pathlib` API.

yarikoptic:
I tend to agree with @jwodder that `Path` is more versatile and easier to use construct, and indeed we can then benefit from type annotation.  I wish even if there was convention to encode "RelativePath" and ensuring that proper conversion is done where needed etc.

yarikoptic:
with above, let's proceed and I hope to push more work here soon.

PR #1210, opened 2023-02-08
Make `RemoteDandiset.get_version()` return a `VersionInfo` instance with validation error fields

### LINKED ISSUE #1206: allow retrieving validation errors
thank you @jwodder for https://github.com/dandi/dandi-cli/issues/1205#issuecomment-1420816824.

opening a new issue here to allow retrieving validation errors in the call. since the endpoint returns it, i'm assuming it's discarded. let's add a parameter to not discard it.

jwodder:
Closes #1206.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -650,7 +650,15 @@ class Version(APIBase):
         return self.identifier
 
 
-class ValidationError(APIBase):
+class RemoteValidationError(APIBase):
+    """
+    .. versionadded:: 0.49.0
+
+    Validation error record obtained from a server.  Not to be confused with
+    :class:`dandi.validate_types.ValidationResult`, which provides richer
+    representation of validation errors.
+    """
+
     field: str
     message: str
 
@@ -663,8 +671,8 @@ class VersionInfo(Version):
     errors
     """
 
-    asset_validation_errors: List[ValidationError]
-    version_validation_errors: List[ValidationError]
+    asset_validation_errors: List[RemoteValidationError]
+    version_validation_errors: List[RemoteValidationError]
 
 
 class RemoteDandisetData(APIBase):
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -650,6 +650,23 @@ class Version(APIBase):
         return self.identifier
 
 
+class ValidationError(APIBase):
+    field: str
+    message: str
+
+
+class VersionInfo(Version):
+    """
+    .. versionadded:: 0.49.0
+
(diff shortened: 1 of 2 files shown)


dandi/dandiapi.py
```diff
@@ -650,6 +650,23 @@ def __str__(self) -> str:
         return self.identifier
 
 
+class ValidationError(APIBase):
```
yarikoptic, line 653:
```suggestion
class RemoteValidationError(APIBase):
    """
    .. versionadded:: 0.49.0

    Validation Error record obtained from a server. To not to be confused with
    :class:`dandi.validate_types.ValidationResult` which provides richer representation
    of validation errors.
```

as most of the constructs in this file have `Remote` in them. (note that sphinx ref to class might be wrong, typing from memory of how those are done)

But I also wonder if may be `ValidationResult` could just be used here instead?

dandi/dandiapi.py
```diff
@@ -958,26 +980,21 @@ def wait_until_valid(self, max_time: float = 120) -> None:
...
             sleep(0.5)
         # TODO: Improve the presentation of the error messages
         about = {
-            "asset_validation_errors": r.get("asset_validation_errors"),
-            "version_validation_errors": r.get("version_validation_errors"),
+            "asset_validation_errors": [
+                e.json_dict() for e in v.asset_validation_errors
+            ],
+            "version_validation_errors": [
+                e.json_dict() for e in v.asset_validation_errors
```
yarikoptic, line 993:
```suggestion
                e.json_dict() for e in v.version_validation_errors
```
as a duplication is evil.  What about 

```
about = {
    f"{scope}_validation_error" : [
        e.json_dict() for e in getattr(v, f"{scope}_validation_errors"))
    for scope in ('asset', 'version')
}
```
or alike altogether ?

yarikoptic (changes requested):

dandi/dandiapi.py
yarikoptic: if switch to ValidationResult I would just keep all errors (for version and assets) in a single list but with a proper `scope` annotation within each ValidationResult.

jwodder: > if switch to ValidationResult I would just keep all errors (for version and assets) in a single list

That's not really up to us; the API returns the errors in two separate fields, and it wouldn't be worth the hassle to change that.

jwodder: Done: f1d752b

yarikoptic: well, we have those super-powers I would say, but ok -- let's stay close to the original return structure at this level.

PR #1213, opened 2023-02-20
Restore testing against Windows with Python 3.10

jwodder:
This reverts part of #315, as I believe the underlying issue has since been resolved.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,11 +41,6 @@ jobs:
           - os: ubuntu-latest
             python: 3.7
             mode: dev-deps
-        exclude:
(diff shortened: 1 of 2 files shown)


jwodder:
Testing with Python 3.10 on Windows is now failing for a completely different reason: it seems that pytest's import-hacking is not playing well with something in keyring and/or pycryptodomex, but I am unable to produce an MVCE.

jwodder:
I got the tests to pass by ignoring a warning emitted due to pywin32-ctypes (a dependency of keyring on Windows) using a deprecated Python API.  I don't understand why I couldn't reproduce this in my attempts at an MVCE, but oh well.

yarikoptic:
Thank you @jwodder !

PR #1216, opened 2023-02-22
Metadata: add extraction of strain from nwb files

satra:
extract strain from nwb

### DIFF
--- a/dandi/tests/data/metadata/metadata2asset_3.json
+++ b/dandi/tests/data/metadata/metadata2asset_3.json
@@ -82,7 +82,8 @@
         "name": "Male"
       },
       "strain": {
-        "name": "abcdef/1"
+        "name": "abcdef/1",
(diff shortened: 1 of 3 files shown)


dandi/tests/data/metadata/metadata2asset_3.json
```diff
@@ -81,6 +81,9 @@
         "identifier": "http://purl.obolibrary.org/obo/PATO_0000384",
         "name": "Male"
       },
+      "strain": {
+        "name": "abcdef/1"
```
yarikoptic, line 85:
```suggestion
        "name": "abcdef/1",
        "schemaKey": "StrainType"
```

seems needed according to failing tests

yarikoptic: Let me just try ;) committing

PR #1220, opened 2023-02-23
Filter out pyout logs from logfile by default

### LINKED ISSUE #1219: do not log  pyout in the log file unless there was an explicit
`pyout` logs are overwhelming in the interactive commands log files which log unconditionally on DEBUG level.  
I think we should
- not log `pyout` at all (filter out) **unless** DEBUG level logging was explicitly requested via DEBUG `log_level` in that dandi/cli/command.py:main
- (unrelated to original issue but while at it): log to file at `min(log_level,10)` i.e. if we specify `--log-level 2` we would go deeper in the log file too

jwodder:
Closes #1219.

### DIFF
--- a/dandi/cli/command.py
+++ b/dandi/cli/command.py
@@ -109,6 +109,10 @@ def main(ctx, log_level, pdb=False):
         datefmt="%Y-%m-%dT%H:%M:%S%z",
     )
     handler.setFormatter(fmter)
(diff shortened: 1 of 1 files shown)


jwodder:
The tests are failing because www.ontobee.org appears to be down.

yarikoptic:
tested manually -- works as desired. Thank you @jwodder !

PR #1227, opened 2023-02-27
Update `parse_purlobourl()` test for change to rat metadata

jwodder:
`test_metadata.py::test_parseobourl[http://purl.obolibrary.org/obo/NCBITaxon_10116-value1]` is currently failing because the obolibrary resource now lists "rat" rather than "Norway rat" as the first `oboInOwl:hasExactSynonym` entry.

### DIFF
--- a/dandi/tests/test_metadata.py
+++ b/dandi/tests/test_metadata.py
@@ -392,7 +392,7 @@ def test_time_extract_gest() -> None:
             "http://purl.obolibrary.org/obo/NCBITaxon_10116",
             {
                 "rdfs:label": "Rattus norvegicus",
-                "oboInOwl:hasExactSynonym": "Norway rat",
+                "oboInOwl:hasExactSynonym": "Rat",
             },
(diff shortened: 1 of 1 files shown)


satra:
while this would work, it seems that some aspect of it is not deterministic. on the webpage, rat is the last exact synonym. so this may fix this problem, but may not be stable. we may want to change the test to account for possible reordering (so using a `set` perhaps).

jwodder:
@satra When I view the actual XML (via curl), I see "rat" as the first synonym and "Norway rat" as the last.  Do we know how the XML is generated?

satra:
we don't and it's possible that the search call in this function (https://github.com/dandi/dandi-cli/blob/master/dandi/metadata.py#L528) is non-deterministic.

jwodder:
@satra I've updated the test to accept all of the "rat" synonyms currently listed on the page.

satra (approved):

Issue #1212, opened 2023-02-16
[Bug]: dandi validate failing uninformatively


CodyCBakerPhD:
As a follow-up to #1211, if I move an otherwise perfectly valid file into the dandiset using `dandi organize`, and then call `dandi validate` from inside the dandiset (which now has a single file in it) I get

```
dandi validate

> 2023-02-16 15:38:47,564 [    INFO] Logs saved in ...

Error: Path {str(p)!r} is not inside Dandiset path {str(dandiset_path)!r}
```

Log: [20230216203844Z-10624.log](https://github.com/dandi/dandi-cli/files/10760628/20230216203844Z-10624.log)

`dandi==0.49.0`

Windows - python 3.10.9

CodyCBakerPhD:
My solution for the user was to simply downgrade to `dandi==0.48.1` and everything worked as expected

CodyCBakerPhD:
This also uninformatively prevents upload

```
dandi upload

> ...t-upload/sub-test-upload.nwb 181.9 kB   0           ERROR        Error 404...
Summary:                        183.5 kB               1 skipped    1 should ...
                                                       1 ERROR      1 Error 4...
2023-02-16 15:46:33,558 [    INFO] Logs saved in C:\Users\Raven\AppData\Local\dandi\dandi-cli\Logs\20230216204628Z-19680.log
Error: Error 404 while sending POST request to https://api.dandiarchive.org/api/uploads/initialize/: {"detail":"Not found."}
```

jwodder:
@CodyCBakerPhD My best guess as to what happened with the `"detail":"Not found."` upload is that you're trying to upload to a Dandiset that doesn't exist — most likely, you meant to upload to staging with `dandi upload -i dandi-staging`, but doing `dandi upload` will try to upload to production.

yarikoptic:
I will send quick PR to provide `f` for that string, but the underlying issue should be resolved I guess -- but we would need to know which path and where is the dandiset boundary (i.e. where dandiset.yaml is).

> My solution for the user was to simply downgrade to dandi==0.48.1 and everything worked as expected

and what was expected - to which dandiset it uploaded?

CodyCBakerPhD:
> My best guess as to what happened with the "detail":"Not found."

That makes sense, but would it be possible to make it return a more informative error?


> and what was expected - to which dandiset it uploaded?

[...]

PR #1228, opened 2023-02-27
BF: add missing f- for a string

yarikoptic:
Ref: https://github.com/dandi/dandi-cli/issues/1212


yarikoptic:
no brainer and green -- will just merge

PR #1231, opened 2023-02-28
download: support for downloading multiple URLs at once; dandiset path is now included in the report

### LINKED ISSUE #1217: enable downloading in parallel from asset URL list
`download()` in `dandi/download.py` takes the argument `jobs`. I thought it would allow downloading from multiple asset URLs at the same time, but it seems that either it's not implemented yet or `jobs` is for something else. Any way parallel downloading by URLs can be enabled?

Would allow users to efficiently download a specific subset of a dataset.

jwodder:
Closes #1217.

Note that I'm not sure about whether the aggregating iterator works correctly, as when I tried downloading multiple URLs at once, the aggregate stats were only shown once the program finished.  At least one problem is that `it.finished` will be briefly true between URLs.

### DIFF
--- a/dandi/cli/cmd_download.py
+++ b/dandi/cli/cmd_download.py
@@ -34,7 +34,7 @@ class ChoiceList(click.ParamType):
 # click doesn't use it unless we explicitly assign it to `help`:
 @click.command(
     help=f"""\
-Download a file or entire folder from DANDI.
+Download files or entire folders from DANDI.
 
 \b
 {_dandi_url_parser.known_patterns}
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -183,6 +183,7 @@ class Downloader:
 
     url: ParsedDandiURL
     output_dir: InitVar[str | Path]
+    output_prefix: Path = field(init=False)
     output_path: Path = field(init=False)
     existing: str
     get_metadata: bool
@@ -201,9 +202,10 @@ class Downloader:
         # under
         if isinstance(self.url, DandisetURL):
             assert self.url.dandiset_id is not None
-            self.output_path = Path(output_dir, self.url.dandiset_id)
+            self.output_prefix = Path(self.url.dandiset_id)
         else:
-            self.output_path = Path(output_dir)
+            self.output_prefix = Path()
+        self.output_path = Path(output_dir, self.output_prefix)
 
     def download_generator(self) -> Iterator[dict]:
         """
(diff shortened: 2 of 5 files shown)


yarikoptic:
> Note that I'm not sure about whether the aggregating iterator works correctly, as when I tried downloading multiple URLs at once, the aggregate stats were only shown once the program finished. At least one problem is that `it.finished` will be briefly true between URLs.

yeah, I am not sure yet either, might have a look tomorrow with a fresh'ier eye.  But we have another conundrum to resolve I ran into right away.  When pointed to multiple datasets, we have `path` within dandiset, and not showing to which dandiset it belongs. As a result we do not only somewhat incompletely inform user, but since path is the "identifier" for pyout, files with identical path (e.g. `dandiset.yaml`) overwrite within in the same row: 

```shell
❯ dandi download DANDI:00002{7,9}
PATH                                             SIZE     DONE            DONE% CHECKSUM STATUS          MESSAGE   
dandiset.yaml                                                                            done            updated   
sub-RAT123/sub-RAT123.nwb                        18.8 kB  18.8 kB          100%    ok    done                      
sub-anm369964/sub-anm369964_behavior+ecephys.nwb 7.7 MB   7.7 MB           100%    ok    done                      
sub-anm369962/sub-anm369962_behavior+ecephys.nwb 6.6 MB   6.6 MB           100%    ok    done                      
sub-anm369963/sub-anm369963_behavior+ecephys.nwb 6.4 MB   6.4 MB           100%    ok    done                      
Summary:                                         20.7 MB  20.7 MB                        5 done          1 updated 
                                                 +18.8 kB 99.91%                                                   
                                                          ETA: 6 seconds                                           
```

I guess we are doomed to uniformize (regardless if multiple URLs are given or not) this to the paths as they are to appear on local drive, hence in this case with dandiset prefixed, i.e. smth like

```
❯ dandi download DANDI:00002{7,9}
PATH                                             SIZE     DONE            DONE% CHECKSUM STATUS          MESSAGE   
000027/dandiset.yaml                                                                            done            updated   
000027/sub-RAT123/sub-RAT123.nwb                        18.8 kB  18.8 kB          100%    ok    done                      
000029/dandiset.yaml                                                                            done            updated   
000029/sub-anm369964/sub-anm369964_behavior+ecephys.nwb 7.7 MB   7.7 MB           100%    ok    done                      
000029/sub-anm369962/sub-anm369962_behavior+ecephys.nwb 6.6 MB   6.6 MB           100%    ok    done                      
000029/sub-anm369963/sub-anm369963_behavior+ecephys.nwb 6.4 MB   6.4 MB           100%    ok    done          
```

yarikoptic:
BTW, while at it, if you feel you see how to  "just" RF it to parallelize within out code instead of `max_workers=jobs` for pyout and rely on it being fed generators, and rather just parallelize, and feed pyout "serially" from the parallel threads - might simplify this beast and make it parallelizable not only via pyout.

PR #1237, opened 2023-03-02
`validate`: Yield error if path is not inside a Dandiset

### LINKED ISSUE #1229: validate: if path does not contain (and is not part of) any Dandiset
```
❯ dandi --version
0.49.1+33.ga5a98fa2


❯ ls -l /tmp/test
total 4
lrwxrwxrwx 1 yoh yoh 108 Feb 20 16:16 blah -> .git/annex/objects/pF/Zf/MD5E-s4--ba1f2511fc30423bdbb183fe33f3dd0f/MD5E-s4--ba1f2511fc30423bdbb183fe33f3dd0f

❯ dandi validate /tmp/test
No errors found.
2023-02-27 18:14:47,705 [    INFO] Logs saved in /home/yoh/.cache/dandi-cli/log/20230227231446Z-2749729.log

```

and I expect validation error.

jwodder:
Closes #1229.

### DIFF
--- a/dandi/validate.py
+++ b/dandi/validate.py
@@ -4,6 +4,7 @@ import os
 from pathlib import Path
 from typing import Iterator, Optional, Union
 
+from . import __version__
 from .consts import dandiset_metadata_file
 from .files import find_dandi_files
 from .utils import find_parent_directory_containing
@@ -149,6 +150,15 @@ def validate(
     """
     for p in paths:
         dandiset_path = find_parent_directory_containing(dandiset_metadata_file, p)
+        if dandiset_path is None:
+            yield ValidationResult(
+                id="DANDI.NO_DANDISET_FOUND",
+                origin=ValidationOrigin(name="dandi", version=__version__),
+                severity=Severity.ERROR,
+                scope=Scope.DANDISET,
+                path=Path(p),
+                message="Path is not inside a Dandiset",
+            )
         for df in find_dandi_files(
             p, dandiset_path=dandiset_path, allow_all=allow_any_path
         ):
--- a/dandi/cli/tests/test_command.py
+++ b/dandi/cli/tests/test_command.py
@@ -20,9 +20,10 @@ def test_smoke(organized_nwb_dir, command):
     # But we must cd to the temp directory since current directory could
     # have all kinds of files which could trip the command, e.g. validate
     # could find some broken test files in the code base
-    with runner.isolated_filesystem():
-        r = runner.invoke(command, [])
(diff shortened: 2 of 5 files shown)


dandi/validate.py
```diff
@@ -149,6 +150,15 @@ def validate(
...
         dandiset_path = find_parent_directory_containing(dandiset_metadata_file, p)
+        if dandiset_path is None:
+            yield ValidationResult(
+                id="DANDI.NO_DANDISET_FOUND",
+                origin=ValidationOrigin(name="dandi", version=__version__),
+                severity=Severity.ERROR,
+                scope=Scope.DANDISET,
+                path=Path(p),
+                message="Path is not inside a Dandiset",
+            )
```
yarikoptic, line 161: I wonder if we should issue this only if nothing below yields a `ValidationResult` with such an id, or may be completely remove adding such one at https://github.com/dandi/dandi-cli/blob/HEAD/dandi/files/bases.py#L561 since otherwise we can provide two nearly identical ones:

```shell
❯ dandi validate RAT123.nwb
[DANDI.NO_DANDISET_FOUND] RAT123.nwb — Path is not inside a Dandiset
[pywnb.GENERIC] RAT123.nwb — error: nwb_version '2.0b' is not a proper semantic version. See http://semver.org
[NWBI.check_subject_age] RAT123.nwb — Subject age, '12 mo', does not follow ISO 8601 duration format, e.g. 'P2Y' for 2 years or 'P23W' for 23 weeks. You may also specify a range using a '/' separator, e.g., 'P1D/P3D' for an age range somewhere from 1 to 3 days. If you cannot specify the upper bound of the range, you may leave the right side blank, e.g., 'P90Y/' means 90 years old or older.
[DANDI.NO_DANDISET_FOUND] RAT123.nwb — File is not inside a Dandiset
```

jwodder: Either way is fine with me.

yarikoptic: ok, let's just remove at  https://github.com/dandi/dandi-cli/blob/HEAD/dandi/files/bases.py#L561 and stay with this centralized one added here

yarikoptic: what I have not realized before saw affect on the tests that dandifile instances would not complain then about not being in dandi datasets, kinda odd... let me think - may be the other thought (filtering) was actually better although I feel that filtering is kinda odd thing to do.

yarikoptic:
ok, let's keep it this way for now although might need to redo it later on

PR #1243, opened 2023-03-07
Bidsschematools 0.7.0 API usage update

TheChymera:
To be merged after the prospective `0.7.0` release of BST.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,8 +1,10 @@
 version: 2
-formats: all
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.8"
 python:
-    version: 3
-    install:
+   install:
         - requirements: docs/requirements.txt
         - method: pip
           path: .
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -4,7 +4,7 @@ build:
   tools:
     python: "3.8"
 python:
-   install:
+    install:
         - requirements: docs/requirements.txt
         - method: pip
           path: .
--- a/dandi/validate.py
+++ b/dandi/validate.py
@@ -26,8 +26,6 @@ def validate_bids(
     ----------
     paths : list(str)
         Paths to validate.
-    schema_version : str, optional
-        BIDS schema version to use, this setting will override the version specified in the dataset.
     devel_debug : bool, optional
         Whether to trigger debugging in the BIDS validator.
 
@@ -46,10 +44,7 @@ def validate_bids(
     import bidsschematools
     from bidsschematools.validator import validate_bids as validate_bids_
 
-    validation_result = validate_bids_(
-        paths,
-        schema_version=schema_version,
-    )
+    validation_result = validate_bids_(paths, exclude_files=["dandiset.yaml"])
     our_validation_result = []
     origin = ValidationOrigin(
         name="bidsschematools",
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,7 +30,7 @@ project_urls =
 python_requires = >=3.8
 install_requires =
     appdirs
-    bidsschematools >= 0.7.0
+    bidsschematools ~= 0.7.0
     click
     click-didyoumean
     dandischema ~= 0.8.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,7 +31,7 @@ project_urls =
 python_requires = >=3.7
 install_requires =
     appdirs
-    bidsschematools >= 0.5.0
+    bidsschematools >= 0.7.0
     click
     click-didyoumean
     dandischema ~= 0.8.0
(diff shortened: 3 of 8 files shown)


yarikoptic:
what's the plan for this PR @TheChymera ?

TheChymera:
@yarikoptic pending BST release.

jwodder:
@yarikoptic @TheChymera bidsschematools 0.7.0 is now out, and dandi-cli's tests are failing due to the lack of this PR.

setup.cfg
```diff
@@ -28,10 +27,10 @@ project_urls =
     Documentation = https://dandi.readthedocs.io
 
 [options]
-python_requires = >=3.7
+python_requires = >=3.8
 install_requires =
     appdirs
-    bidsschematools >= 0.5.0
+    bidsschematools >= 0.7.0
```
jwodder, line 33: Should `< 0.8.0` be made an upper bound so we don't have to make another emergency PR like #1293?

.readthedocs.yaml
```diff
@@ -1,8 +1,10 @@
 version: 2
-formats: all
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.8"
 python:
-    version: 3
-    install:
+   install:
```
jwodder, line 7: You dedented this by a space.  I'd prefer it if the line were kept at a four-space indent.

jwodder (commented):
If we're committing to dropping 3.7 support, I suppose you'll also want to get rid of [this placeholder](https://github.com/dandi/dandi-cli/blob/9b7ce308e09333c3ae3259348599efa4c78413ee/dandi/tests/fixtures.py#L58)?

.readthedocs.yaml
yarikoptic:
```suggestion
    install:
```

setup.cfg
yarikoptic: +1 on that , so 
```suggestion
    bidsschematools ~= 0.7.0
```
?

TheChymera: Wasn't really an emergency PR:

(1) it existed for over a month
(2) the mistake was me not [limiting the version](https://github.com/dandi/dandi-cli/pull/1293) when I made it so that this could have been handled without temporarily breaking CI.

Adding artificial upper bounds I think is more dangerous, since we might end up forgetting to update.
In the case of BST, which is just me in another repo, I think either is fine.

dandi/validate.py
```diff
@@ -46,10 +44,7 @@ def validate_bids(
     import bidsschematools
     from bidsschematools.validator import validate_bids as validate_bids_
 
-    validation_result = validate_bids_(
-        paths,
-        schema_version=schema_version,
-    )
+    validation_result = validate_bids_(paths, exclude_files=["dandiset.yaml"])
```
yarikoptic, line 47: if this is the one incompatibility, in the future for such ones we might even better stay compatible with both versions for a while (e.g. base on . That would allow to instruct users to e.g. downgrade back to prior version of dependency if new version introduced some relevant regression etc.  So please keep that in mind @TheChymera in the future.

.readthedocs.yaml
TheChymera: @jwodder yarik already fixed this in fb2dc42207077109373601ab75e00f709e089c60

setup.cfg
jwodder: I was referring to #1293 (which you created today) as the "emergency PR," and if you say that the mistake was not limiting the version, isn't that an argument in support of limiting the version going forwards?

.readthedocs.yaml
jwodder: @TheChymera That happened after I commented.

dandi/validate.py
yarikoptic: specifically here actually we could have stayed then compatible with 3.7 and this PR could have concentrated really on bidsschematools compatibility aspect. 
3.7 support Ends in 1 month and 1 week (27 Jun 2023)

ok, this time let's do it, but in the future -- see above.

setup.cfg
TheChymera: not really, since we don't know for a fact that there will be any API-breaking changes in 0.8, or 0.9, or 0.10 etc. On the contrary I actually suspect there won't. If I start suspecting there will, yes, I will remember to add the limit this time :3

TheChymera: Honestly It's fine if you really want this....

Just that there's an epidemic of people artificially setting upper bounds and they're almost always pointless and needlessly confuse package management.

yarikoptic:
I am marking this as a `release`. I think we then simply can skip merging #1293 . agree @jwodder ?

jwodder:
@yarikoptic Agreed.

TheChymera:
@jwodder yeah, didn't include the delete because I didn't want to make the BST fix hinge on that bit.

TheChymera:
But if you want to do a release just after this PR, might as well include it.

PR #1244, opened 2023-03-08
Creating test for file-wise NWB error output listing

TheChymera:
Do not review yet, this is just a rough draft, needs [clarification upstream](https://github.com/NeurodataWithoutBorders/nwbinspector/issues/345) before anything interesting happens here:

### DIFF
--- a/dandi/cli/cmd_validate.py
+++ b/dandi/cli/cmd_validate.py
@@ -141,6 +141,7 @@ def _process_issues(
     grouping: str,
     ignore: Optional[str] = None,
 ) -> None:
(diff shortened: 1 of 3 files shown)


yarikoptic:
well, there is no really need in extra "integration" tests with generating files and running validation on them. It would have been better (faster, more explicit since seeing what is to be rendered) to feed rendering with some prepared `ValidationResult` records and just vary parameters for rendering and match exactly what I expect to see from their rendering with different grouppings.

TheChymera:
@yarikoptic looking good now?

yarikoptic (commented):
ok, just 1 more comment possibly needing addressing

PR #1245, opened 2023-03-08
Corrected docstring

TheChymera:
Really minor fix I noticed while working on something else, hence the unfitting branch name.


TheChymera:
The type check failure has nothing to do with this PR.

yarikoptic:
Thanks! type checking clearly must not relate!

PR #1246, opened 2023-03-08
Restrict mypy to pre-v1.1.1

jwodder:
Type-checking of pydantic types is broken in many respects under mypy 1.1.1 (See https://github.com/pydantic/pydantic/issues/5148), so until pydantic publishes a new release with a fix, we need to use an older mypy in order to get type-checking to pass.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -28,7 +28,7 @@ commands =
 
 [testenv:typing]
 deps =
-    mypy
+    mypy<1.1.1
     types-appdirs
     types-python-dateutil
     types-requests


tox.ini
```diff
@@ -28,7 +28,7 @@ commands =
 
 [testenv:typing]
 deps =
-    mypy
+    mypy<1.1.1
```
yarikoptic, line 31: looking forward for a fix, and presumably already fixed in https://github.com/pydantic/pydantic/pull/5111 (could you check may be), should we just
```suggestion
    mypy!=1.1.1
```
?

jwodder: `<` is slightly better, in the case where mypy releases a new version before pydantic does.

yarikoptic: Ok, let's proceed but later change when know which needed or to be avoided

PR #1248, opened 2023-03-09
[ENH] Add severity restriction option to `dandi-validate`

TheChymera:


TheChymera:
@jwodder any ideas what I could do in order to fix the typing issue? I tried  https://github.com/dandi/dandi-cli/pull/1248/commits/2fab6f2df5d4ab5331895926d586852c7435bf63 but the assert and the list filtering don't seem to work. I was alternatively thinking of creating an `INFO = 0` level for `Severity` and make the severity non-optional. Does that sound like a good idea, or is there some other way to do this without adding a new level we don't really need otherwise?

jwodder:
@TheChymera Just do:

```python
validator_result = [
         i for i in validator_result if i.severity is not None and i.severity.value >= min_severity_value
]
```

dandi/cli/cmd_validate.py
```diff
@@ -79,13 +79,20 @@ def validate_bids(
     default="none",
 )
 @click.option("--ignore", metavar="REGEX", help="Regex matching error IDs to ignore")
+@click.option(
+    "--min-severity",
```
jwodder, line 83: Also document this in `docs/source/cmdline/validate.rst`.

jwodder (changes requested):

dandi/cli/cmd_validate.py
```diff
@@ -117,7 +124,15 @@ def validate(
...
         allow_any_path=allow_any_path,
     )
-    _process_issues(validator_result, grouping, ignore)
+
+    min_severity_value = Severity[min_severity].value
+
+    filtered_results = [
+        i for i in validator_result
+        if i.severity is not None and i.severity.value >= min_severity_value
+    ]
```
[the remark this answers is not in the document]
TheChymera: Yes, in fact we previously did just that in the `_process_issues()` function.

dandi/cli/cmd_validate.py
TheChymera: ok, done.

jwodder (approved):

PR #1254, opened 2023-03-18
enh: add cellline parsing

### LINKED ISSUE #1253: add celline parsing from strain field in NWB
we are asking folks to add `cellline: <description_string>` to the `strain` field of NWB when dealing with recordings from cell lines. this needs to change the strain parser to detect and add to cellline metadata when relevant.

satra:
closes #1253


dandi/metadata.py
```diff
@@ -455,11 +455,22 @@ def extract_sex(metadata: dict) -> Optional[models.SexType]:
...
+        if value.lower().startswith("cellline:"):
+            return None
         return models.StrainType(name=value)
     else:
         return None
 
 
+def extract_cellLine(metadata: dict) -> Optional[str]:
+    value = metadata.get("strain", None)
+    if value is not None and value != "" and value.lower().startswith("cellline:"):
```
yarikoptic, line 468: simplify and robustify?

```suggestion
    if isinstance(value, str) and value.lower().startswith("cellline:"):
```

since `.lower()` is safe only if it is a string... if just to simplify, could be:

```suggestion
    if value and value.lower().startswith("cellline:"):
```

dandi/metadata.py
```diff
@@ -455,11 +455,22 @@ def extract_sex(metadata: dict) -> Optional[models.SexType]:
...
+        return value.split("cellline: ")[1]
```
yarikoptic, line 469: To be inline with above check (no trailing space) and to ensure absent additional splitting and no leading/trailing whitespaces

```suggestion
        return value.split("cellline:", 1)[1].strip()
```

dandi/metadata.py
```diff
@@ -455,11 +455,22 @@ def extract_sex(metadata: dict) -> Optional[models.SexType]:
 def extract_strain(metadata: dict) -> Optional[models.StrainType]:
     value = metadata.get("strain", None)
     if value is not None and value != "":
```
yarikoptic, line 457: while at it, could be simplified similarly to below
```suggestion
    if value:
```

yarikoptic (changes requested):
minor to harmonize/simplify/robustify

yarikoptic:
I added unittest and made it a bit more robust, remains green -- let's proceed.

PR #1260, opened 2023-04-03
validate: resolved issues with relative path input

TheChymera:
This is what could happen before this:

```
(dev) [deco]~/src/bids-examples/asl003 ❱ dandi validate .
2023-04-03 12:43:56,852 [    INFO] Note: NumExpr detected 12 cores but "NUMEXPR_MAX_THREADS" not set, so enforcing safe limit of 8.
2023-04-03 12:43:56,852 [    INFO] NumExpr defaulting to 8 threads.
2023-04-03 12:43:58,230 [    INFO] Logs saved in /home/chymera/.cache/dandi-cli/log/20230403164355Z-7402.log
Error: Path '.' is not inside Dandiset path '/home/chymera/src/bids-examples/asl003'
```

### DIFF
--- a/dandi/tests/test_validate.py
+++ b/dandi/tests/test_validate.py
@@ -16,6 +16,17 @@ def test_validate_nwb_error(simple3_nwb: Path) -> None:
     assert len([i for i in validation_result if i.severity]) > 0
 
 
+def test_validate_relative_path(
+    bids_examples: Path,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    selected_dataset_path = bids_examples / "asl003"
+    monkeypatch.chdir(selected_dataset_path)
+    # improper relative path handling would fail with:
+    # ValueError: Path '.' is not inside Dandiset path '/tmp/.../asl003'
+    validate(".")
+
+
 def test_validate_empty(tmp_path: Path) -> None:
     assert list(validate(tmp_path)) == [
         ValidationResult(
--- a/dandi/validate.py
+++ b/dandi/validate.py
@@ -149,7 +149,7 @@ def validate(
       errors for a path
     """
     for p in paths:
-        p = os.path.abspath(os.path.expanduser(p))
+        p = os.path.abspath(p)
         dandiset_path = find_parent_directory_containing(dandiset_metadata_file, p)
         if dandiset_path is None:
             yield ValidationResult(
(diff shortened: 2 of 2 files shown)


dandi/validate.py
```diff
@@ -149,6 +149,7 @@ def validate(
       errors for a path
     """
     for p in paths:
+        p = os.path.abspath(os.path.expanduser(p))
```
jwodder, line 152:
```suggestion
        p = os.path.abspath(p)
```

Unexpanded paths are not valid paths and should not be supported.

jwodder (changes requested):

dandi/validate.py
TheChymera: @jwodder so better address this in the CLI wrapper? Clearly we want to support people doing stuff like `dandi validate .`, right?

jwodder: The `~/somedir` in `dandi validate` should be expanded by the shell.  If the user does something funky that prevents shell expansion, they deserve to get an error.

yarikoptic: my thinking is along with @jwodder's on this.  If ppl use from within Python -- it would be for them to expand user. After all, `~` is a legit symbol in a filename, so we must not blindly expand it:

```shell
❯ touch '~yoh'
❯ lst
-rw-------  1 yoh yoh      0 Apr  3 16:11 ~yoh
```

TheChymera: ok

yarikoptic:
could you add a test which does just what you did to demonstrate the issue -- runs `validate .` and produces expected results and doesn't fail.  Can just extend some existing test with that.

grep for `monkeypatch.chdir` to see how to temporarily chdir to run that invocation.

Otherwise, without a test, there is no guarantee that we would not "fix it back" some time.

dandi/tests/test_validate.py
```diff
@@ -16,6 +17,14 @@ def test_validate_nwb_error(simple3_nwb: Path) -> None:
     assert len([i for i in validation_result if i.severity]) > 0
 
 
+def test_validate_relative_path(bids_examples: Path) -> None:
+    selected_dataset_path = bids_examples / "asl003"
+    os.chdir(selected_dataset_path)
```
jwodder, line 22: Use pytest's `monkeypatch` fixture instead of changing the PWD for all tests.

TheChymera: ok.

TheChymera:
@yarikoptic done.

PR #1262, opened 2023-04-04
Add CI run testing against NFS

### LINKED ISSUE #1261: Add tests run whenever temporary files are generated on NFS
We have a number of issues reported (below) 
- https://github.com/dandi/dandi-cli/issues/764
- https://github.com/dandi/dandi-cli/issues/1033
- https://github.com/dandi/dandi-cli/issues/1257

which all have NFS as the commonality in the setup and an odd behavior from filesystem being the most logical explanation.

Let's add a run to our test setup which would use NFS mounted partition as the TEMPDIR. We do similar already in https://github.com/datalad/git-annex/blob/master/.github/workflows/template/build-%7B%7Bostype%7D%7D.yaml.j2#L443 so recipe could be adopted here.

Even though it is unlikely to hit the same issue(s), it would be good to have such a run since eventually we might get some NFS specific spurious failures 

edit 1: make `$HOME` to also reside on NFS. We use fscacher which keeps cache at HOME and most likely on all those failing cases HOME was also under NFS, so yet another factor.

jwodder:
Closes #1261.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,6 +42,9 @@ jobs:
           - os: ubuntu-latest
             python: 3.7
             mode: dev-deps
+          - os: ubuntu-latest
+            python: 3.7
+            mode: nfs
 
     steps:
     - name: Set up environment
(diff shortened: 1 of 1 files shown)


yarikoptic:
Looks great, thank you @jwodder.  (Un)fortunately does not trigger any issue, oh well.

PR #1264, opened 2023-04-04
Log more information about retried HTTP requests

jwodder:
ref: https://github.com/dandi/dandi-cli/issues/1257


yarikoptic:
Ok, let's log more details and see if any come handy.

PR #1267, opened 2023-04-05
Log results of `requests.utils.super_len()` when `DANDI_DEVEL_INSTRUMENT_REQUESTS_SUPERLEN` is set

jwodder:
Part of #1257.

### DIFF
--- a/dandi/files/zarr.py
+++ b/dandi/files/zarr.py
@@ -3,13 +3,14 @@ from __future__ import annotations
 from base64 import b64encode
 from collections.abc import Generator, Iterator
 from concurrent.futures import Future, ThreadPoolExecutor, as_completed
-from contextlib import closing
+from contextlib import ExitStack, closing
 from dataclasses import dataclass, field, replace
 from datetime import datetime
 import os
 from pathlib import Path
 from time import sleep
 from typing import Any, Optional
+from unittest.mock import patch
 
 from dandischema.digests.zarr import get_checksum
 from dandischema.models import BareAsset, DigestType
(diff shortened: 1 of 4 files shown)


dandi/files/zarr.py
```diff
@@ -540,14 +541,38 @@ def mkzarr() -> str:
...
-    with item.filepath.open("rb") as fp:
-        storage_session.put(
-            upload_url,
-            data=fp,
-            json_resp=False,
-            retry_if=_retry_zarr_file,
-            headers={"Content-MD5": item.base64_digest},
-        )
+    with ExitStack() as stack:
+        if os.environ.get("DANDI_DEVEL_INSTRUMENT_REQUESTS_SUPERLEN"):
```
yarikoptic, line 545: please add to CONTRIBUTING.md where we describe such variables

dandi/files/zarr.py
```diff
@@ -540,14 +541,38 @@ def mkzarr() -> str:
...
+            stack.enter_context(patch("requests.models.super_len", new_super_len))
+
+        with item.filepath.open("rb") as fp:
+            storage_session.put(
+                upload_url,
+                data=fp,
+                json_resp=False,
+                retry_if=_retry_zarr_file,
+                headers={"Content-MD5": item.base64_digest},
+            )
```
yarikoptic, line 575: nice design pattern on many aspects!  

@asmacdo @candleindark - please review/take into your toolbelt on how to 
- add optional context managers to the piece of code without duplication
- instrument/patch some functionality (outside or not) temporarily

dandi/files/zarr.py
jwodder: We don't have a `CONTRIBUTING.md`.  Envvars are described in `DEVELOPMENT.md`.

jwodder:
@yarikoptic I just realized that this approach will be a disaster, as Zarr entries are uploaded in a thread pool run within another set of threads for parallel asset uploads, yet monkeypatching is process-wide; hence, the patches to `super_len()` will all interfere with each other, with the most likely outcome involving the wrong filepaths being logged for `super_len()` return values.

yarikoptic:
- don't patch within that method but rather "process wide" at e.g.  import of dandi.upload stage. No need even to unpatch
- just log that `o` within super_len, no need for any exterior variables.

with this  patch to requests

```patch
diff --git a/requests/utils.py b/requests/utils.py
index a367417f..dc4df067 100644
--- a/requests/utils.py
+++ b/requests/utils.py
@@ -131,6 +131,7 @@ def dict_to_sequence(d):
 
 
 def super_len(o):
+    print(f"super_len on str={o} repr={o!r}")
     total_length = None
     current_position = 0
 

```

I see 
```
super_len on str=<_io.BufferedReader name='/home/yoh/.tmp/pytest-of-yoh/pytest-39/test_upload_zarr0/example.zarr/arr_0/0'> repr=<_io.BufferedReader name='/home/yoh/.tmp/pytest-of-yoh/pytest-39/test_upload_zarr0/example.zarr/arr_0/0'>
```
while running our tests -- so either of those (str or repr) would be good enough to identify the location.

this way patched function would need only the content which is given to it.

yarikoptic:
somehow the run in that instrumented nfs environment resulted in

```
_________________ test_dandi_file_zarr_with_excluded_dotfiles __________________
dandi/tests/test_files.py:309: in test_dandi_file_zarr_with_excluded_dotfiles
    zf = dandi_file(zarr_path)
dandi/files/__init__.py:195: in dandi_file
    return factory(filepath, path, dandiset_path)
dandi/files/_private.py:71: in __call__
    return self.CLASSES[DandiFileType.classify(filepath)](
dandi/files/_private.py:42: in classify
    raise UnknownAssetError("Empty directories cannot be Zarr assets")
E   dandi.exceptions.UnknownAssetError: Empty directories cannot be Zarr assets
```
I didn't check if that is a flaky or whatnot, just restarting for a good check on that

PR #1268, opened 2023-04-05
Give `DandiAPIClient.authenticate()` a `save_to_keyring` parameter

### LINKED ISSUE #1259: Make it possible to provide/store API key for `dandi_authenticate`
While preparing the https://github.com/dandi/dandi-cli/issues/1159#issuecomment-1488638627 answer, I could not come up with an easy solution where via API we are provided and store in keyring credentials (API key) for an instance so it could be later reused by high level `download()` function.   I think the easiest/best would be to extend `dandi_authenticate` which takes care about storing token in keyring, with `token: Optional[str]=None` parameter which would provide the token to be used **and** be stored if provided.  No other mechanisms (keyring/env var) should be queried for a key in this case.

jwodder:
Closes #1259.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -49,7 +49,7 @@ from .consts import (
     known_instances_rev,
 )
 from .exceptions import HTTP404Error, NotFoundError, SchemaVersionError
-from .keyring import keyring_lookup
+from .keyring import keyring_lookup, keyring_save
 from .misctypes import Digest, RemoteReadableAsset
 from .utils import (
     USER_AGENT,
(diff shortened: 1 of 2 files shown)


dandi/dandiapi.py
```diff
@@ -457,15 +457,32 @@ def for_dandi_instance(
...
+            ``save_to_keyring`` added
         """
         # Fails if token is invalid:
         self.get("/auth/token", headers={"Authorization": f"token {token}"})
+        if save_to_keyring:
+            if self.api_url in known_instances_rev:
+                client_name = known_instances_rev[self.api_url]
+            else:
+                raise NotImplementedError("TODO client name derivation for keyring")
+            app_id = f"dandi-api-{client_name}"
```
yarikoptic, line 483: please avoid duplication for this code block which is also used in `dandi_authenticate`. Seems asking for something like `def _get_keyring_app_id` or alike to contain above lines to be used here and in dandi_authenticate

yarikoptic:
nfs -- known #1269

dandi/dandiapi.py
jwodder: Done: acb9038

Issue #1252, opened 2023-03-18
upload doesn't work when pointed to a file inside a dandiset


satra:
inside a bids dandiset, we did `dandi upload README`, it found 0 files to consider.

TheChymera:
I think I located this one. It seems to work with `--allow-any-path` and this is by design because [we don't just want to accept `GenericAsset`s](https://github.com/dandi/dandi-cli/blob/47e9d97e2535d66d89edafa3d0a83603f3966dac/dandi/files/__init__.py#L146-L147). @jwodder  Should I create a built-in whitelist for known `GenericAsset`s, or is the solution to just make sure they're detected as something else? 

```
(dev) [deco]~/src/bids-examples/asl003 ❱ DANDI_DEVEL=1 dandi upload -i dandi-staging --allow-any-path README
2023-04-03 12:53:01,245 [    INFO] Note: NumExpr detected 12 cores but "NUMEXPR_MAX_THREADS" not set, so enforcing safe limit of 8.
2023-04-03 12:53:01,245 [    INFO] NumExpr defaulting to 8 threads.
/home/chymera/src/bids-examples/asl003/README
GenericAsset(filepath=PosixPath('/home/chymera/src/bids-examples/asl003/README'), dandiset_path=PosixPath('/home/chymera/src/bids-examples/asl003'), path='README')
2023-04-03 12:53:03,101 [    INFO] Found 1 files to consider
PATH                 SIZE      ERRORS UPLOAD STATUS                MESSAGE
README               232 Bytes   0           ERROR                 Error 404 while sending POST request to https://api-staging.dandiarchi...
Summary:             232 Bytes               1 ERROR               1 Error 404 while sending POST request to https://api-staging.dandiarc...
```

satra:
@TheChymera - given bids support, we should no longer need to use either `DANDI_DEVEL`, `--allow-any-path` or `--validation ignore`. thus in this case i expect the system to automatically detect that the `README` file is within a dandiset, and that it's a BIDS dandiset.

yarikoptic:
Given that we have odd behavior of silently ignoring non-nwb/video files in DANDI I think we should change/generalize behavior of `validate` and `upload` to be
- walk the dandiset, possibly just the provided directory within, for all non-dotdirectories (i.e. `.dandi/`, `.git/`, `.datalad/` ATM) and discover all paths
- validate those paths according to either DANDI or BIDS (as already implemented) and 
[...]

PR #1271, opened 2023-04-07
Support for `dandi upload README`

TheChymera:
Closes: https://github.com/dandi/dandi-cli/issues/1252

### DIFF
--- a/dandi/files/__init__.py
+++ b/dandi/files/__init__.py
@@ -140,6 +140,10 @@ def find_dandi_files(
                 else:
                     yield df
         else:
+            if (p.parent / BIDS_DATASET_DESCRIPTION).exists() and not bidsdd:
+                bids = dandi_file(p.parent / BIDS_DATASET_DESCRIPTION, dandiset_path)
+                assert isinstance(bids, BIDSDatasetDescriptionAsset)
+                bidsdd = bids
             df = dandi_file(p, dandiset_path, bids_dataset_description=bidsdd)
             # Don't use isinstance() here, as GenericBIDSAsset's should still
             # be returned
--- a/dandi/files/__init__.py
+++ b/dandi/files/__init__.py
@@ -105,7 +105,13 @@ def find_dandi_files(
                 raise ValueError(
                     f"Path {str(p)!r} is not inside Dandiset path {str(dandiset_path)!r}"
                 )
-        path_queue.append((Path(p), None))
+            else:
+                bidsdd = find_bids_dataset_description(p, dandiset_path)
+                if bidsdd:
+                    yield bidsdd
+        else:
+            bidsdd = None
+        path_queue.append((Path(p), bidsdd))
     bids_roots = []
     while path_queue:
         p, bidsdd = path_queue.popleft()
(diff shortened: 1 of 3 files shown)


TheChymera:
@yarikoptic so the use case of `dandi upload /path/to/dandiset-bids/README` (still needs full path pending ) works as of [5047f7f](https://github.com/dandi/dandi-cli/pull/1271/commits/5047f7f5e0006b4103862de47c7d8fe565b1056b).

Sadly if I try to formulate a test for this (see [e97aa99](https://github.com/dandi/dandi-cli/pull/1271/commits/e97aa99a7f7d1c2bce471aba9f944a41ab058834)) it fails with:

```
dandi/files/bids.py:168: in bids_dataset_description
    assert bdd is not None
E   AssertionError
```

The issue seems to be that the `bids_dataset_description_ref` of the `GenericBIDSAsset` is dead.
Inserting a `print(self)` [here](https://github.com/dandi/dandi-cli/blob/eef8443a16f0968891f4ddfca43d663df3f07f2b/dandi/files/bids.py#L167) gives me:

```
GenericBIDSAsset(filepath=PosixPath('/tmp/pytest-of-chymera/pytest-29/dandiset0/README'), dandiset_path=PosixPath('/tmp/pytest-of-chymera/pytest-29/dandiset0'), path='README', bids_dataset_description_ref=<weakref at 0x7f7a742ad3f0; dead>)
```

@jwodder any idea what I could do about it?

jwodder:
@TheChymera

* The problem is that the BIDS assets code assumes that a `BIDSDatasetDescriptionAsset` will always be stored somewhere other than in the weakref of a `BIDSAsset`, which is not the case for the `bidsdd` created by this PR; since the `bidsdd` is never yielded by `find_dandi_files()`, it's soon garbage collected, and the weakrefs in the assets that were yielded end up dead.
* Disregarding the weakref issue, this PR is flawed on multiple levels:
   * It should check all directories between `dandiset_path` and `p` for a `BIDS_DATASET_DESCRIPTION`, not just `p.parent`.
   * It does not handle directories in `paths` that are inside directories containing `BIDS_DATASET_DESCRIPTION`.
   * If `BIDS_DATASET_DESCRIPTION` is not found anywhere beneath a given element of `paths`, the code will end up checking for `BIDS_DATASET_DESCRIPTION` for every file at every level beneath the path.
   
    It would be better to insert above line 108 a search for a `BIDS_DATASET_DESCRIPTION` between `dandiset_path` & `p` and use the result in place of the `None` on line 108.

TheChymera:
@jwodder better? :)

jwodder:
@TheChymera Somewhat, except that now `find_dandi_files()` can yield assets outside of `paths`, which it shouldn't, and which will result in `dandi upload` also uploading `dataset_description.json` even when that's not under the given paths.

TheChymera:
@jwodder any ideas about this? seems however I try it if `dataset_description.json` isn't uploaded the wekref is dead, and it won't validate :(

jwodder:
Closing in favor of #1286.

PR #1272, opened 2023-04-07
Fix `test_validate_relative_path`

jwodder:
`validate()` is a generator function, and so calling it does not run any code unless the returned value is iterated over.

### DIFF
--- a/dandi/tests/test_validate.py
+++ b/dandi/tests/test_validate.py
@@ -24,7 +24,7 @@ def test_validate_relative_path(
     monkeypatch.chdir(selected_dataset_path)
     # improper relative path handling would fail with:
     # ValueError: Path '.' is not inside Dandiset path '/tmp/.../asl003'
-    validate(".")
+    list(validate("."))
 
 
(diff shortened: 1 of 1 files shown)


dandi/tests/test_validate.py
```diff
@@ -24,7 +24,7 @@ def test_validate_relative_path(
     monkeypatch.chdir(selected_dataset_path)
     # improper relative path handling would fail with:
     # ValueError: Path '.' is not inside Dandiset path '/tmp/.../asl003'
-    validate(".")
+    list(validate("."))
```
yarikoptic, line 27: oh, nice catch.

@TheChymera did you actually check that this test failed on master without the fix?  it is a nice "habbit" to have/test that way.  You can `git checkout master` even without committing your test thus to check if your test catching original problem, and then `git checkout -`, make sure that now it passes, and then `git commit`

yarikoptic:
Thank you @jwodder, let's proceed. I left a comment for @TheChymera

PR #1275, opened 2023-04-11
Accept Other sex as o abbreviation

bendichter:
Add the ability for Subject.sex to be "O", indicating "Other" sex.


dysprague:
I am working on an extension to support optophysiology experiments in C Elegans. C elegans have two sexes, male (XO) and hermaphrodite (XX). It would be very helpful to be able to use the 'other' categorization for this.

yarikoptic:
Adjusted the title since Other actually seems was supported already.

PR #1276, opened 2023-04-11
C elegans

bendichter:
Add support for C. elegans

### DIFF
--- a/dandi/metadata.py
+++ b/dandi/metadata.py
@@ -527,7 +527,7 @@ species_map = [
         "Danio rerio - Zebra fish",
     ),
     (
-        ["c. elegans", "Caenorhabditis elegans"],
+        ["c. elegans", "caenorhabditis elegans"],
         "caenorhabditis",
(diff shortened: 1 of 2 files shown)


dandi/metadata.py
```diff
@@ -526,6 +526,12 @@ def extract_cellLine(metadata: dict) -> Optional[str]:
         "http://purl.obolibrary.org/obo/NCBITaxon_7955",
         "Danio rerio - Zebra fish",
     ),
+    (
+        ["c. elegans", "Caenorhabditis elegans"],
```
yarikoptic, line 530: this is the first one with a capital letter in alternative names. 

Looking at https://github.com/dandi/dandi-cli/blob/HEAD/dandi/metadata.py#L598 we need

```suggestion
        ["c. elegans", "caenorhabditis elegans"],
```

But also I think we better also add `.lower()`  there to `key` (this value) to altogether ignore case  , will do both now for this PR.

yarikoptic (approved):
looks good, if no surprises from CI - let's merge

PR #1282, opened 2023-04-21
Explicitly open h5py.File as read-only where we read it

yarikoptic:
ref: https://github.com/dandi/helpdesk/discussions/95#discussioncomment-5687405

apparently we started to fail to extract metadata due to some warning being spitted out that file is not open for writing whenever we do not even expect it to open for writing.  Not yet entirely sure what was happening there since user reported that the issue was resolved by upgrade of h5py. But here it is worth addressing the warning anyways:

    H5pyDeprecationWarning: The default file mode will change to 'r' (read-only) in h5py 3.0. To suppress this warning, pass the mode you need to h5py.File(), or set the global default h5.get_config().default_file_mode, or set the environment variable H5PY_DEFAULT_READONLY=1. Available modes are: 'r', 'r+', 'w', 'w-'/'x', 'a'. See the docs for details.

@jwodder  - can you spot any other instance were we might need it?

### DIFF
--- a/dandi/pynwb_utils.py
+++ b/dandi/pynwb_utils.py
@@ -115,7 +115,7 @@ def get_nwb_version(
         def _sanitize(v: Any) -> str:
             return str(v)
 
-    with open_readable(filepath) as fp, h5py.File(fp) as h5file:
+    with open_readable(filepath) as fp, h5py.File(fp, "r") as h5file:
         # 2.x stored it as an attribute
         try:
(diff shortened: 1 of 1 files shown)


jwodder (approved):
There don't appear to be any other instances in need of updating.

yarikoptic:
marking for release so to avoid that warning to pop up for users, and also release glob supports and c elegans into the wild.

PR #1285, opened 2023-05-10
Use `inspect_nwbfile()` instead of `inspect_nwb()`; address bug in recent `ruamel.yaml` versions

jwodder:
The test suite has been failing for the past few days with the following causes:

- Calls to `nwbinspector.inspect_nwb()` (used to validate NWB files) were returning an error message reading:

    > The API function 'inspect_nwb' has been deprecated and will be removed in a future release! To remove ambiguity, please call either 'inspect_nwbfile' giving a path to the unopened file on a system, or 'inspect_nwbfile_object' passing an already open pynwb.NWBFile object.

  and this ended up in the set of validation errors.  I have replaced `inspect_nwb()` with `inspect_nwbfile()`.

- [A bug in recent versions of `ruamel.yaml`](https://sourceforge.net/p/ruamel-yaml/tickets/461/) was causing a certain spot in `organize.py` to fail.  I addressed this by removing the `age.pop("units", None)` line, which seems to be unnecessary (at least in ruamel.yaml 0.17.21, released 2022-02-12, the last release before this month).

### DIFF
--- a/dandi/files/bases.py
+++ b/dandi/files/bases.py
@@ -19,7 +19,7 @@ from dandischema.models import BareAsset, CommonModel
 from dandischema.models import Dandiset as DandisetMeta
 from dandischema.models import get_schema_version
 from etelemetry import get_project
-from nwbinspector import Importance, inspect_nwb, load_config
+from nwbinspector import Importance, inspect_nwbfile, load_config
 from nwbinspector.utils import get_package_version
 from packaging.version import Version
 from pydantic import ValidationError
@@ -492,11 +492,11 @@ class NWBAsset(LocalFileAsset):
         schema_version: Optional[str] = None,
         devel_debug: bool = False,
     ) -> list[ValidationResult]:
-        """Validate NWB asset
+        """
+        Validate NWB asset
 
         If ``schema_version`` was provided, we only validate basic metadata,
-        and completely skip validation using nwbinspector.inspect_nwb
-
+        and completely skip validation using nwbinspector.inspect_nwbfile
         """
         errors: list[ValidationResult] = pynwb_validate(
             self.filepath, devel_debug=devel_debug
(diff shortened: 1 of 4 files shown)


jwodder:
Note that, at time of writing, GitHub Actions (and other GitHub services) is experiencing issues, and so the tests did not run.

CodyCBakerPhD:
Huh, were they actual errors? I thought it was just a simple `DeprecationWarning` for the time being

dandi/files/bases.py
```diff
@@ -19,7 +19,7 @@
 from dandischema.models import Dandiset as DandisetMeta
 from dandischema.models import get_schema_version
 from etelemetry import get_project
-from nwbinspector import Importance, inspect_nwb, load_config
+from nwbinspector import Importance, inspect_nwbfile, load_config
```
yarikoptic, line 22: Do we need to boost minimal version for nwbinspector ?

dandi/organize.py
```diff
@@ -604,7 +604,6 @@ def is_undefined(d, f):
         age["minimum"] = min(uvs["age"])
         age["maximum"] = max(uvs["age"])
         if age.get("units", None) in (None,) + DEFAULT_VALUES:  # template
-            age.pop("units", None)
```
yarikoptic, line 607: Don't we need to pop off it isn't None?

yarikoptic (commented):
Two questions

dandi/organize.py
jwodder: The `insert()` on the next line overwrites any previous value of `age["units"]`.

dandi/files/bases.py
jwodder: I don't know what version `inspect_nwbfile()` was added in.  CC @CodyCBakerPhD

dandi/organize.py
yarikoptic: :facepalm:

dandi/files/bases.py
CodyCBakerPhD: It was added in the latest version, so yes I would bump minimal.

CodyCBakerPhD: latest = `0.4.28`

Keep in mind `etelemetry` complains (just a warning I think) if running validation when not using the latest release too, but upgrading the requirement is never a bad idea either

yarikoptic:
Thank you @jwodder for being proactive with this one!

PR #1290, opened 2023-05-17
Retry test that intermittently fails on NFS

jwodder:
`test_files.py::test_dandi_file_zarr_with_excluded_dotfiles` keeps intermittently failing on NFS, so this PR uses pytest-rerunfailures to rerun it up to ten times until it succeeds.


yarikoptic:
heh heh. reminds me of good old times of PyMVPA and our `@labile` helper  https://github.com/PyMVPA/PyMVPA/blob/master/mvpa2/testing/tools.py#L403 . Ok, let's hope it helps 

image

PR #1292, opened 2023-05-18
Don't treat certain broken symlinks as absent files

### LINKED ISSUE #1289: make .exists( check of pathlib into the one to support broken symlinks
came up during review of https://github.com/dandi/dandi-cli/pull/1286/files#diff-cdc58989d7a5c424994a24e3e9a85a7a03a4c7c2594e0feb139d7ebea7d387a6R55 

<details>
<summary>it seems we are making a good number of checks using `.exists`</summary> 

```shell
❯ git grep -p '\.exists(' | grep -v -e test -e versionee
dandi/dandiset.py=    def __init__(self, path: Union[str, Path], allow_empty: bool = False) -> None:
dandi/dandiset.py:        if not allow_empty and not (self.path_obj / dandiset_metadata_file).exists():
dandi/dandiset.py=    def _load_metadata(self) -> None:
dandi/dandiset.py:        if self._metadata_file_obj.exists():
dandi/dandiset.py=    def update_metadata(self, meta: dict) -> None:
dandi/dandiset.py:        if self._metadata_file_obj.exists():
dandi/download.py=    def __enter__(self) -> DownloadDirectory:
dandi/download.py:            if not chkpath.exists():
dandi/files/__init__.py=def find_dandi_files(
dandi/files/__init__.py:                if (p / BIDS_DATASET_DESCRIPTION).exists():
dandi/files/__init__.py:                    if (p / BIDS_DATASET_DESCRIPTION).exists() and not any(
dandi/files/bids.py=    def _validate(self) -> None:
dandi/files/bids.py:                        readme_candidate.exists()
dandi/files/zarr.py=    def exists(self) -> bool:
dandi/files/zarr.py:        return self.filepath.exists()
dandi/keyring.py=def keyring_op(
dandi/keyring.py:        if op.exists(kb.file_path):
dandi/keyring.py:            if keyring_cfg.exists():
dandi/move.py=    def get_path(self, path: str, is_src: bool = True) -> File | Folder:
dandi/move.py:        if not p.exists():
dandi/organize.py=def organize_external_files(
dandi/organize.py:                if not op.exists(name_old_str):
dandi/organize.py=def populate_dataset_yml(filepath, metadata):
dandi/organize.py:    if not op.exists(filepath):
dandi/organize.py=        def act(func, *args, **kwargs):
dandi/organize.py:        if op.exists(dandiset_path):
dandi/organize.py=        def _get_metadata(path):
dandi/organize.py:    if not op.exists(dandiset_path):
dandi/organize.py:        if op.exists(dandi_fullpath):
dandi/organize.py:            if not op.exists(dandi_dirpath):
dandi/organize.py:            if op.exists(d):
dandi/upload.py=        def upload_agg(*ignored: Any) -> str:
dandi/upload.py:                    and not Path(dandiset.path, asset.path).exists()
dandi/utils.py=def find_parent_directory_containing(
dandi/utils.py:        if (path / filename).exists():
tools/update-assets-on-server=def process_dandiset(dandiset, update, api_key):
tools/update-assets-on-server:        if not os.path.exists(localpath):

```
</details>

as those files could be under `git-annex` control and exist as symlink but not really available with content, we should code those paths with that awareness. E.g. in many cases we should not make logic based on `.exists` but rather on `.lexists` to e.g. try to load from a file or not.  So it is desired to go through these and either replace with `.lexists` in case of using `os.path` or some create/use helper for checks of [pathlib.Paths (since there is no .lexists there).

jwodder:
Closes #1289.

@yarikoptic I can't tell which uses of `exists()` in `dandi/organize.py` should be replaced with `lexists()` and which rely on the file actually being present/openable.  Could you go through the file and check the uses of `exists()` there?

### DIFF
--- a/dandi/dandiset.py
+++ b/dandi/dandiset.py
@@ -38,8 +38,10 @@ class Dandiset:
                 )
         self.path = str(path)
         self.path_obj = Path(path)
-        if not allow_empty and not (self.path_obj / dandiset_metadata_file).exists():
-            raise ValueError(f"No Dandiset at {path}")
+        if not allow_empty and not os.path.lexists(
+            self.path_obj / dandiset_metadata_file
+        ):
+            raise ValueError(f"No dandiset at {path}")
         self.metadata: Optional[dict] = None
         self._metadata_file_obj = self.path_obj / dandiset_metadata_file
         self._load_metadata()
--- a/dandi/files/__init__.py
+++ b/dandi/files/__init__.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 
 from collections import deque
 from collections.abc import Iterator
+import os.path
 from pathlib import Path
 from typing import Optional
 
@@ -115,7 +116,7 @@ def find_dandi_files(
             if p.is_symlink():
                 lgr.warning("%s: Ignoring unsupported symbolic link to directory", p)
             elif dandiset_path is not None and p == Path(dandiset_path):
-                if (p / BIDS_DATASET_DESCRIPTION).exists():
+                if os.path.lexists(p / BIDS_DATASET_DESCRIPTION):
                     bids = dandi_file(p / BIDS_DATASET_DESCRIPTION, dandiset_path)
                     assert isinstance(bids, BIDSDatasetDescriptionAsset)
                     bidsdd = bids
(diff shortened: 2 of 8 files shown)


yarikoptic:
>  Could you go through the file and check the uses of `exists()` there?

done, pushed

yarikoptic:
I've merged another PR and we need conflicts resolved now @jwodder

jwodder:
@yarikoptic Resolved.

yarikoptic:
ok, let's proceed with this and see if any side-effects come up.  Thank you @jwodder !

PR #1293, opened 2023-05-19
Limit bidsschematools version

TheChymera:
Before we merge https://github.com/dandi/dandi-cli/pull/1243 , dandi-cli is not compatible with newest BST.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,7 +31,7 @@ project_urls =
 python_requires = >=3.7
 install_requires =
     appdirs
-    bidsschematools >= 0.5.0
+    bidsschematools >= 0.5.0, <0.7.0
     click
     click-didyoumean
     dandischema ~= 0.8.0


TheChymera:
@jwodder @yarikoptic quick merge on this one? The update should also be done in an hour or so, just feels nicer to signal the transition in the version constraints.

jwodder:
@TheChymera Are you expecting this to be released immediately as well?

TheChymera:
@jwodder what do you mean by “this”?

yarikoptic:
> @jwodder yes, but not between this PR and the BST 0.7 migration PR.

might be friday effect -- but why not?  as @jwodder hinted, we likely do want to kick out such release so the most recent version of dandi-cli precludes coinstallation with most recent bidsschematools?

but since pypi doesn't have a way to describe "breaks" (within bidsschematools) we are screwed now anyways, and better concentrate on addressing compatibility issue(s)

jwodder:
@yarikoptic 

> but since pypi doesn't have a way to describe "breaks" (within bidsschematools)

What do you mean by "breaks"?  If you mean breaking changes, those are supposed to be communicated by semantic versioning, and reverse dependencies should guard against them with upper version bounds.

TheChymera:
Seems from the other discussion we no longer want this.

PR #1298, opened 2023-05-23
Accept URLs as Dandi instance specifiers

### LINKED ISSUE #1085: Allow to operate on an instance provided by URL, based on /server-info
With https://github.com/dandi/dandi-archive/pull/1190 in dandi-archive we would (re-?)gain usable /server-info from the local instances.  But ATM `dandi` is not designed to be able to work with some arbitrary instance of dandi archive it doesn't know so without #1003 or replacing it #1084 it wouldn't even be able to download a dandiset from it:

```
$> curl http://localhost:8085/server-info/
{"schema_version":"0.6.3","schema_url":"https://raw.githubusercontent.com/dandi/schema/master/releases/0.6.3/dandiset.json","version":"0.2.39.post19+g9b92512f","services":{"api":{"url":"http://localhost:8000/api"},"webui":{"url":"http://localhost:8085"},"jupyterhub":{"url":"https://hub.dandiarchive.org/"}},"cli-minimal-version":"0.14.2","cli-bad-versions":[]}

$> dandi download http://localhost:8085/dandiset/000001
2022-07-26 17:00:36,014 [    INFO] Logs saved in /home/yoh/.cache/dandi-cli/log/20220726210034Z-1548312.log
Error: No such Dandiset: '000001'
``` 

But `download` is not a single command which needs to know about instance, and we have `-i|--dandi-instance` to specify the instance.  I think we should 
- if instance cannot be deduced from URL, get that URLs top `/server-info` and create an instance with that URL as the name internally for the duration of that process to be used as the instance_name.
- allow for a URL (such as `http://localhost:8085/`) to be provided/used as the instance name, thus allowing for any other operation , such as `upload -i http://localhost:8085/`.

jwodder:
Closes #1085.

To do:

- [x] Permit the `--dandi-instance` option to take URL values
- [x] Update URL parsing to accept arbitrary base URLs
- [x] Update docs
- [x] Add/update tests
- [x] Remove all the redirector stuff?
- [x] What if an API URL ending in `/api` is passed as an instance specifier?
- [x] Give `DandiInstance` a `name` attribute that defaults to `host:port` for non-builtin instances
    - Use this to replace `DandiAPIClient._instance_id`
- [x] Update [this spot](https://github.com/dandi/dandi-cli/blob/e74c5ac88d4a3af917c0ac9d773325058c09a185/dandi/cli/cmd_download.py#L125) in the "download" command
- [x] Update keyring client name derivation to handle arbitrary instances
- [x] Make `dandi/delete.py` preserve a supplied `DandiInstance` instance
- [x] Mark any tests that now need network access
- [x] Cache `get_instance()` requests


yarikoptic (approved):
Have not spotted anything of concern.  Let's see it in use!  I will not add `release` label just yet.
Thank you @jwodder !

PR #1288, opened 2023-05-15
Update `asset_validation_errors` structure

jwodder:
Part of/blocked by https://github.com/dandi/dandi-archive/pull/1579.

PR #1303, opened 2023-06-08
Update `asset_validation_errors` structure

jwodder:
Blocked by https://github.com/dandi/dandi-archive/pull/1579.

Replaces #1288.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -693,6 +693,10 @@ class RemoteValidationError(APIBase):
     message: str
 
 
+class RemoteAssetValidationError(RemoteValidationError):
+    path: str
+
+
 class VersionInfo(Version):
     """
     .. versionadded:: 0.49.0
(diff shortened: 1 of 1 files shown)


dandi/dandiapi.py
```diff
@@ -693,6 +693,10 @@ class RemoteValidationError(APIBase):
     message: str
 
 
+class RemoteAssetValidationError(RemoteValidationError):
+    path: str
```
yarikoptic, line 697: can we make pydantic assign path to be `None` (make it `Optional`?) if not present in current version? That is to not need to align 100% to version of server switch since there would be period of time when staging would be having it new way while main one old way.  And only whenever current version of dandi-archive is gone from use - would make it "mandatory".

jwodder: Done: 43d1d27

yarikoptic:
thanks! Let's proceed so dandi-cli is ready for what is coming in dandi-archive. We can always get back if so would be needed.  Filed https://github.com/dandi/dandi-cli/issues/1304 so we do not forget to make non-optional.

PR #1311, opened 2023-07-18
Fix wait_until_valid to properly wait for a publishable version

danlamanna:
This brings the CLI up to parity with https://github.com/dandi/dandi-archive/pull/1548. The tldr is that it's not enough to check the version status to know if a version is publishable since the assets must also be valid.

cc @AlmightyYakob @jwodder

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -1023,7 +1023,7 @@ class RemoteDandiset:
         start = time()
         while time() - start < max_time:
             v = self.get_version(self.version_id)
-            if v.status is VersionStatus.VALID:
+            if v.status is VersionStatus.VALID and not v.asset_validation_errors:
                 return
             sleep(0.5)
         # TODO: Improve the presentation of the error messages


AlmightyYakob:
LGTM, I'll defer to @jwodder for the full approval

dandi/dandiapi.py
```diff
@@ -1023,7 +1023,7 @@ def wait_until_valid(self, max_time: float = 120) -> None:
         start = time()
         while time() - start < max_time:
             v = self.get_version(self.version_id)
-            if v.status is VersionStatus.VALID:
+            if v.status is VersionStatus.VALID and not v.asset_validation_errors:
```
yarikoptic, line 1026: @danlamanna and @AlmightyYakob -- would you be so kind to digest for me here why having "Valid" indicator is not sufficient any longer and we also check that there is no errors?  related: https://github.com/dandi/dandisets/issues/356#issuecomment-1687321702

AlmightyYakob: "Valid" used to be returned only if both the version metadata and all of the assets were valid. The behavior has since been changed to only represent the version metadata, and so the extra check is required.

PR #1315, opened 2023-08-09
ENH: Add support for _desc- entity to label different processings of a file

yarikoptic:
ref: #1314 

TODO outside of this PR:  check/decide on possibility of `organize` to actually automagically assign such `desc`... unlikely since the space is huge...

PR #1316, opened 2023-08-09
Fix typing errors

jwodder:


jwodder:
@yarikoptic

* The keyring-related typing errors were due to changes in mypy and/or keyring (I'm not going to try to figure out which).  I don't believe keyring changed in any way that affects the functioning of dandi.
* I believe the `cv2` typing errors occurred because the library (which is written in C++) only recently added Python type annotations, and these annotations describe interfaces that are stricter than what actually works (e.g., `apiPreference` is declared as an `int`, but the tests still succeed when a `None` is passed instead).  This PR updates the code to use `cv2` the way its annotations say it should be used.

yarikoptic:
coolio, I will then merge this one whenever turns green and then https://github.com/dandi/dandi-cli/pull/1315 to release. Thank you!

Issue #1317, opened 2023-08-10
downloading dandiset.yaml without downloading the full dandiset


magland:
Is it possible to download or somehow obtain the dandiset.yaml file without downloading the entire dandiset? This would be useful for amending a dandiset using the cli without downloading all the files. Alternatively, is there a more straightforward way to upload a file to an existing dandiset other than using the dandiset.yaml?

yarikoptic:
- `dandi download` can do that for you

```shell
❯ dandi download --help | grep -e --download -A2
  --download [dandiset.yaml,assets,all]
                                  Comma-separated list of elements to download
                                  [default: all]
```

so e.g.
```shell
❯ dandi download --download dandiset.yaml DANDI:000026
2023-08-10 09:41:54,008 [ WARNING] Request to https://api.dandiarchive.org/api failed (404 Client Error: Not Found for url: https://api.dandiarchive.org/api/api/info/)
2023-08-10 09:41:54,008 [ WARNING] Using hard-coded URLs
PATH                 SIZE DONE    DONE% CHECKSUM STATUS MESSAGE   
000026/dandiset.yaml                             done   updated   
Summary:                  0 Bytes                1 done 1 updated 
                          <0.00%                                  
2023-08-10 09:41:54,438 [    INFO] Logs saved in /home/yoh/.cache/dandi-cli/log/20230810134152Z-616416.log
dandi download --download dandiset.yaml DANDI:000026  1.92s user 3.13s system 197% cpu 2.559 total
```

- all dandisets are also provided as DataLad datasets from https://github.com/dandisets with super-dataset at https://github.com/dandi/dandisets to identify the versions of the entire archive.  You can simply git clone any of those to get to dandiset.yaml ;-)

satra:
> Alternatively, is there a more straightforward way to upload a file to an existing dandiset other than using the dandiset.yaml?

you can do it programmatically using the python library (or replicate the functionality using the REST API in any language).

PR #1318, opened 2023-08-10
Check if URL matches the API URL so we do not hardcode wrong "False" for is_api

yarikoptic:
I have spotted odd log msg while preparing example for https://github.com/dandi/dandi-cli/issues/1317#issuecomment-1673250340:

```shell
❯ dandi download --download dandiset.yaml DANDI:000026
2023-08-10 09:41:54,008 [ WARNING] Request to https://api.dandiarchive.org/api failed (404 Client Error: Not Found for url: https://api.dandiarchive.org/api/api/info/)
2023-08-10 09:41:54,008 [ WARNING] Using hard-coded URLs
...
```

which was apparently due to our assumption that provided URL (here from DANDI: identifiers.org redirect) is not API URL.  I have not looked deeper into that logic, but rather just fixed checking for known instances in case if URL does match a known API end point.

### DIFF
--- a/dandi/utils.py
+++ b/dandi/utils.py
@@ -596,6 +596,7 @@ class ServerInfo(BaseModel):
 
 def get_instance(dandi_instance_id: str | DandiInstance) -> DandiInstance:
     dandi_id = None
+    is_api = True
     redirector_url = None
     if isinstance(dandi_instance_id, DandiInstance):
         instance = dandi_instance_id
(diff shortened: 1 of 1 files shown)


yarikoptic:
I just now merged typing fixes PR #1316 , typing errors here seems to be of that PR fixes kind, so unrelated to this PR

Issue #1332, opened 2023-10-16
test_update_dandiset_from_doi started to fail


yarikoptic:
seems started to fail a week ago

```
FAILED ../../../../../opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/dandi/cli/tests/test_service_scripts.py::test_update_dandiset_from_doi[10.1101/2020.01.17.909838-biorxiv] - AssertionError: assert 1 == 0
368
 +  where 1 = <Result JSONDecodeError('Expecting value: line 1 column 1 (char 0)')>.exit_code
369
FAILED ../../../../../opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/dandi/cli/tests/test_service_scripts.py::test_update_dandiset_from_doi[10.1523/JNEUROSCI.6157-08.2009-jneurosci] - AssertionError: assert 1 == 0
370
 +  where 1 = <Result JSONDecodeError('Expecting value: line 1 column 1 (char 0)')>.exit_code
371
FAILED ../../../../../opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/dandi/cli/tests/test_service_scripts.py::test_update_dandiset_from_doi[10.1016/j.neuron.2019.10.012-neuron] - AssertionError: assert 1 == 0
372
 +  where 1 = <Result JSONDecodeError('Expecting value: line 1 column 1 (char 0)')>.exit_code
373
FAILED ../../../../../opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/dandi/cli/tests/test_service_scripts.py::test_update_dandiset_from_doi[10.7554/eLife.48198-elife] - AssertionError: assert 1 == 0
374
 +  where 1 = <Result JSONDecodeError('Expecting value: line 1 column 1 (char 0)')>.exit_code
375
FAILED ../../../../../opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/dandi/cli/tests/test_service_scripts.py::test_update_dandiset_from_doi[10.1038/s41467-023-37704-5-nature] - AssertionError: assert 1 == 0
376
 +  where 1 = <Result JSONDecodeError('Expecting value: line 1 column 1 (char 0)')>.exit_code
377
= 5 failed, 261 passed, 425 deselected, 1 xfailed, 5 warnings in 631.20s (0:10:31) =
```

[...]

PR #1333, opened 2023-10-16
Ignore unclosed connection resource warning from VCR

yarikoptic:
locally (vcr 5.1.0) it causes tests to fail, possibly masking another issue ref: https://github.com/dandi/dandi-cli/issues/1332 which I otherwise fail to reproduce locally


yarikoptic:
ok, it doesn't resolve #1332 anyhow.  I will proceed with it anyways to simplify troubleshooting

PR #1335, opened 2023-10-16
typing: Account for the fact that requests.HTTPError .response migth be None now

yarikoptic:


yarikoptic:
with bad internet come duplicate PRs... see #1336 with a label

PR #1336, opened 2023-10-16
typing: Account for the fact that requests.HTTPError .response migth be None now

yarikoptic:


yarikoptic:
hm, there might have been functionality effects from the proposed here changes since seems some new tests start to fail then... or may be it is coinciding with fresh fixes in dandi-archive which might relate (failing tests are here about moves and uploads)...

yarikoptic:
if fresh run fails with anything but `test_update_dandiset_from_doi` in 3.10 and 3.11 -- it is caused by the changes here and should be addressed

yarikoptic:
oh, I guess there is a deeper meaning of `bool(response)` to be True if it not errorneous or smth like that

```
(Pdb) p bool(e.response)
False
*(Pdb) p e.response
<Response [400]>
```

yarikoptic:
ok, only coverage is "affected". proceeding with a release

PR #1337, opened 2023-10-16
Re-record VCR tapes using newer versions of libraries

yarikoptic:
I believe that somehow some upgrade among

```
❯ wdiff -3 /tmp/f1 /tmp/f2

====================================================================== [-06/github/cron/20231006T060312/0c1d4a0/Tests/5197/test-]{+07/github/cron/20231007T060245/0c1d4a0/Tests/5198/test+} ======================================================================
 [-dependencies.txt:2023-10-06T06:05:18.6948668Z-] {+dependencies.txt:2023-10-07T06:04:51.1091526Z+}
======================================================================
 [-boto3-1.28.61 botocore-1.31.61-] {+boto3-1.28.62 botocore-1.31.62+}
======================================================================
 [-entrypoints-0.4-]
======================================================================
 [-numcodecs-0.11.0-] {+numcodecs-0.12.0+}
======================================================================
 [-urllib3-1.26.17-] {+urllib3-2.0.6+}
======================================================================
```

(so likely urllib3) lead to drastically different behavior which triggered VCR to try to not reuse prior tape but then subsequent dialog started to fail -- we got no json.  As you can see in the diff we observe that now we do get json (as requested) recorded although previously it was some "binary".

Closes #1332 although I am afraid that we might need to really identify the library lead to breakage and then set explicit minimal version of tests so we have consistent testing across distributions etc

PS `pre-commit` refused to commit since it kept codespell'ing the .vcr.yaml files although ignored within codespell config of `setup.cfg`. I guess codespell does not consult that `skip` config if explicit file names are given.  Might need to add those to excludes of pre-commit

PR #1338, opened 2023-10-20
Minor documentation tune up

yarikoptic:
PR is primarily to test current state of CI given number of fails here and there.


yarikoptic:
ok, 3.10 and 3.11 fails are known #1332 to be fixed in #1337 (WiP).

PR #1341, opened 2023-10-31
Expose number of jobs to organize

CodyCBakerPhD:
Recent usage of the DANDI API in the NWB GUIDE revealed a sporadic `joblib` issue that results in segmentation faults - no clue what actually causes it for some users, but the best solution we have for now is to just expose job usage directly

Where should tests for this go?

### DIFF
--- a/dandi/cli/cmd_organize.py
+++ b/dandi/cli/cmd_organize.py
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-from typing import Optional
-
 import click
 
 from .base import dandiset_path_option, devel_debug_option, map_to_click_exceptions
--- a/dandi/cli/cmd_organize.py
+++ b/dandi/cli/cmd_organize.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from typing import Optional
+
 import click
 
 from .base import dandiset_path_option, devel_debug_option, map_to_click_exceptions
@@ -59,6 +61,13 @@ from ..consts import dandi_layout_fields, file_operation_modes
     ),
 )
 @click.argument("paths", nargs=-1, type=click.Path(exists=True))
+@click.option(
+    "--number-of-jobs",
+    "number_of_jobs",
+    type=int,
+    default=None,
+    help="The number of jobs to use during organization.",
+)
 @devel_debug_option()
 @map_to_click_exceptions
 def organize(
@@ -70,6 +79,7 @@ def organize(
     devel_debug=False,
     update_external_file_paths=False,
     media_files_mode=None,
+    number_of_jobs: Optional[int] = None,
 ):
     """(Re)organize files according to the metadata.
 
@@ -115,4 +125,5 @@ def organize(
         update_external_file_paths=update_external_file_paths,
         media_files_mode=media_files_mode,
         required_fields=required_fields,
+        number_of_jobs=number_of_jobs,
     )
--- a/dandi/cli/cmd_organize.py
+++ b/dandi/cli/cmd_organize.py
@@ -73,7 +73,7 @@ def organize(
     devel_debug=False,
     update_external_file_paths=False,
     media_files_mode=None,
-    jobs: Optional[int] = None,
+    jobs=None,
 ):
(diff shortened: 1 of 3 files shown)


dandi/cli/cmd_organize.py
```diff
@@ -59,6 +61,13 @@
     ),
 )
 @click.argument("paths", nargs=-1, type=click.Path(exists=True))
+@click.option(
+    "--number-of-jobs",
+    "number_of_jobs",
+    type=int,
+    default=None,
+    help="The number of jobs to use during organization.",
+)
```
yarikoptic, line 70:
```suggestion
@click.option("-J", "--jobs", type=int, help="Number of jobs during organization")
```
for consistency with other commands? (this one from `move` "as is")

yarikoptic:
more info on segfault? I usually try to run those through valgrind to get a sense where/why might be happening although IIRC it was tricky with python stuff.

re tests - just make one of the invocations in dandi/tests/test_organize.py with `jobs=1` to cover that codepath, not sure if we should sweat too much on really testing beyond smoke test on that aspect

CodyCBakerPhD:
> more info on segfault? I usually try to run those through valgrind to get a sense where/why might be happening although IIRC it was tricky with python stuff.

Sadly no. It was just something random that occurred in certain user environments during first round of alpha testing - the local traceback led back to `joblib` called by DANDI organize - we've been unable to reproduce it - so the best workaround we can think on is to disable `joblib` usage in those situations

dandi/cli/cmd_organize.py
```diff
@@ -70,6 +73,7 @@ def organize(
     devel_debug=False,
     update_external_file_paths=False,
     media_files_mode=None,
+    number_of_jobs: Optional[int] = None,
```
CodyCBakerPhD, line 76:
```suggestion
    jobs: Optional[int] = None,
```

dandi/cli/cmd_organize.py
```diff
@@ -115,4 +119,5 @@ def organize(
         update_external_file_paths=update_external_file_paths,
         media_files_mode=media_files_mode,
         required_fields=required_fields,
+        number_of_jobs=number_of_jobs,
```
CodyCBakerPhD, line 122:
```suggestion
        jobs=jobs,
```

dandi/organize.py
```diff
@@ -721,8 +721,10 @@ def organize(
     update_external_file_paths=False,
     media_files_mode=None,
     required_fields=None,
+    number_of_jobs: Optional[int] = None,
```
CodyCBakerPhD, line 724:
```suggestion
    jobs: Optional[int] = None,
```

dandi/organize.py
```diff
@@ -721,8 +721,10 @@ def organize(
...
 ):
     in_place = False  # If we deduce that we are organizing in-place
+    number_of_jobs = number_of_jobs or -1
```
CodyCBakerPhD, line 727:
```suggestion
    jobs= jobsor -1
```

dandi/organize.py
```diff
@@ -812,12 +814,12 @@ def _get_metadata(path):
             meta["path"] = path
             return meta
 
-        if not devel_debug:
+        if not devel_debug and number_of_jobs != 1:  # Do not use joblib at all if number_of_jobs=1
```
CodyCBakerPhD, line 817:
```suggestion
        if not devel_debug and jobs != 1:  # Do not use joblib at all if number_of_jobs=1
```

dandi/organize.py
```diff
@@ -812,12 +814,12 @@ def _get_metadata(path):
...
             # Note: It is Python (pynwb) intensive, not IO, so ATM there is little
             # to no benefit from Parallel without using multiproc!  But that would
             # complicate progress bar indication... TODO
             metadata = list(
-                Parallel(n_jobs=-1, verbose=10)(
+                Parallel(n_jobs=number_of_jobs, verbose=10)(
```
CodyCBakerPhD, line 822:
```suggestion
                Parallel(n_jobs=jobs, verbose=10)(
```

PR #1346, opened 2023-11-08
Remove unused configuration settings

jwodder:


tox.ini
```diff
@@ -49,32 +49,14 @@ markers =
...
     # <https://github.com/h5py/h5py/issues/1765>
     # <https://github.com/dandi/dandi-cli/pull/275>
     ignore:.* size changed, may indicate binary incompatibility.*:RuntimeWarning
-    # <https://github.com/hdmf-dev/hdmf/issues/547>
-    ignore:\s*safe_load will be removed.*:PendingDeprecationWarning:hdmf
-    ignore:\s*load will be removed.*:PendingDeprecationWarning:ruamel.yaml
-    ignore:Passing None into shape arguments.*:DeprecationWarning:h5py
-    ignore:the imp module is deprecated:DeprecationWarning
-    ignore:`Unit` has been deprecated:DeprecationWarning:humanize
-    ignore:The distutils package is deprecated:DeprecationWarning:joblib
```
yarikoptic, line 65: are you sure we have minimal versions provided so that we no longer can run into those warningns?

jwodder: The warnings don't occur in testing anymore (as the tests pass), and this configuration only affects testing.

yarikoptic: Correct, but testing might be done where we do not have most recent versions of those dependencies (conda, or some linux distribution) thus causing fails.

jwodder: Most of our runtime dependencies don't have minimum versions specified at the moment, so theoretically testing elsewhere could use arbitrarily old packages with all sorts of warnings.  Exactly what situation are you trying to avoid?

yarikoptic: pretty much those I described -- testing package build ATM on conda, but later could be on debian/ubuntu systems.  If we don't have minimal versions specified, we have no guarantees that we would not hit those warnings which are to be ignored. So why not to keep them around for now?

jwodder: I've removed the commit that deleted the "ignores" and created #1360.

yarikoptic:
type checking was fixed elsewhere

PR #1355, opened 2023-11-09
find_files: Simplify check for VCS and DataLad paths

jwodder:


yarikoptic:
looks legit, couldn't think of changes in behavior.

PR #1357, opened 2023-11-10
Replace magic string arguments with enums

jwodder:


dandi/tests/test_download.py
```diff
@@ -88,7 +100,7 @@ def test_download_000027_metadata_only(url: str, tmp_path: Path) -> None:
 @mark.skipif_no_network
 @pytest.mark.parametrize(
     "url",
-    [  # Should go through API
+    [  # Should go through AP
```
yarikoptic, line 103: what is AP?

yarikoptic:
cool!

dandi/tests/test_download.py
jwodder: I did not mean to make that change, and I don't know how it happened.

yarikoptic: ok, pushed 10d076e7bd9625840d8a4e1cab8a49d986c415ea directly to master to revert this one

PR #1359, opened 2023-11-16
Fix typing error

jwodder:
I suspect the error in question only started being emitted when mypy 1.7.0 was released on the 10th.

### DIFF
--- a/dandi/cli/cmd_move.py
+++ b/dandi/cli/cmd_move.py
@@ -44,7 +44,7 @@ from .base import devel_debug_option, instance_option, map_to_click_exceptions
 @devel_debug_option()
 @map_to_click_exceptions
 def move(
-    paths: tuple[str],
+    paths: tuple[str, ...],
     dandiset: str | None,
     dry_run: bool,
     existing: str,


dandi/cli/cmd_move.py
```diff
@@ -44,7 +44,7 @@
 @devel_debug_option()
 @map_to_click_exceptions
 def move(
-    paths: tuple[str],
+    paths: tuple[str, ...],
```
yarikoptic, line 47: for my own sake, so the old one is a tuple of a single `str` and new one is tuple of strings or tuple with the first element a `str`?

jwodder: `tuple[str]` is a tuple of length 1 containing a `str`.  `tuple[str, ...]` is a tuple of any length where all elements are `str`s.

Mypy was complaining because we we doing this inside the function:

https://github.com/dandi/dandi-cli/blob/cb86c9cb3fad0b8735ad86eb4ba43569d4a0a425/dandi/cli/cmd_move.py#L88-L89

and `len()` of a one-element tuple is always less than 2, making the line after the `raise` unreachable.

yarikoptic: so it is the 
```
dandi/cli/cmd_move.py:90: error: Statement is unreachable  [unreachable]
        move_mod.move(
        ^
Found 1 error in 1 file (checked 77 source files)
```
one. got it. thank you for the explanation

PR #1365, opened 2023-11-27
Update for `zarr_checksum` 0.2.12

jwodder:
- We can't update to `zarr_checksum` 0.3.0 yet, as that requires Pydantic v2.

- As `zarr_checksum` gained type annotations in dandi/zarr_checksum#21, we can remove its override from the mypy config.

- Since dandi/zarr_checksum#17 was merged, it's now safe to import `zarr_checksum` at the top level of a module.

- Fixed a typing error caused by importing `ZarrChecksumTree` from the top level of `zarr_checksum` despite it not being listed in `zarr_checksum.__all__`, which triggered a typing failure due to our mypy configuration containing `implicit_reexport = False`.  (This error could only be detected after `zarr_checksum` started shipping a `py.typed` file.)

### DIFF
--- a/dandi/files/zarr.py
+++ b/dandi/files/zarr.py
@@ -15,6 +15,7 @@ from typing import Any
 from dandischema.digests.zarr import get_checksum
 from dandischema.models import BareAsset, DigestType
 import requests
+from zarr_checksum.tree import ZarrChecksumTree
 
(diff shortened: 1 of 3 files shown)


yarikoptic:
Thank you @jwodder ! as we are many commits after release, let's release while at it.

PR #1370, opened 2023-11-29
Repeatedly double-check return values of zero when spying on `super_len()`

jwodder:
Part of #1257.

### DIFF
--- a/dandi/upload.py
+++ b/dandi/upload.py
@@ -5,11 +5,13 @@ from collections.abc import Iterator, Sequence
 from contextlib import ExitStack
 from enum import Enum
 from functools import reduce
+import io
 import os.path
 from pathlib import Path
 import re
 import time
-from typing import Any, TypedDict
+from time import sleep
+from typing import Any, TypedDict, cast
 from unittest.mock import patch
 
 import click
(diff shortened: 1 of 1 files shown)


dandi/upload.py
```diff
@@ -108,7 +170,16 @@ def new_super_len(o):
                     raise
                 else:
                     lgr.debug("requests.utils.super_len() reported %d for %r", n, o)
-                    return n
+                    if (
+                        n == 0
+                        and isinstance(o, io.IOBase)
+                        and (name := getattr(o, "name", None)) not in (None, "")
```
yarikoptic, line 176: Previously we didn't have those checks of io and name, would we add at least a debug message if we get 0 for those cases?

jwodder: > Previously we didn't have those checks of io and name

Previously, we weren't checking for anything.

> would we add at least a debug message if we get 0 for those cases?

I can't tell what you're trying to say here.

yarikoptic: I thought that we had some dedicated debug message for size 0 only but now I see that it was for all sizes... 

My point was that what if we still get super_len to report 0 but then it would not be matching those two other conditions and thus we silently would not be testing using `check_len` and it might not be clear from the debug message above on either we should have or should have not... but I also see that we should be reporting some more debug lines from within check_len so we could tell between those cases anyways, so probably no more logging is needed, let's proceed.

PR #1371, opened 2023-12-08
Adopt `dandischema.digests.zarr.get_checksum()` to dandi-cli: use zarr_checksum library constructs instead of copies in dandischema

jwodder:
And rewrite it in terms of `zarr_checksum`

Part of dandi/dandi-schema#209

### DIFF
--- a/dandi/files/zarr.py
+++ b/dandi/files/zarr.py
@@ -12,7 +12,6 @@ from pathlib import Path
 from time import sleep
 from typing import Any
 
-from dandischema.digests.zarr import get_checksum
 from dandischema.models import BareAsset, DigestType
 import requests
 from zarr_checksum.tree import ZarrChecksumTree
@@ -153,25 +152,25 @@ class ZarrAsset(LocalDirectoryAsset[LocalZarrEntry]):
 
         def dirstat(dirpath: LocalZarrEntry) -> ZarrStat:
             # Avoid heavy import by importing within function:
-            from dandi.support.digests import md5file_nocache
+            from dandi.support.digests import checksum_zarr_dir, md5file_nocache
 
             size = 0
-            dir_md5s = {}
-            file_md5s = {}
+            dir_info = {}
+            file_info = {}
             files = []
             for p in dirpath.iterdir():
                 if p.is_dir():
                     st = dirstat(p)
                     size += st.size
-                    dir_md5s[p.name] = (st.digest.value, st.size)
+                    dir_info[p.name] = (st.digest.value, st.size)
                     files.extend(st.files)
                 else:
                     size += p.size
-                    file_md5s[p.name] = (md5file_nocache(p.filepath), p.size)
+                    file_info[p.name] = (md5file_nocache(p.filepath), p.size)
                     files.append(p)
             return ZarrStat(
                 size=size,
-                digest=Digest.dandi_zarr(get_checksum(file_md5s, dir_md5s)),
+                digest=Digest.dandi_zarr(checksum_zarr_dir(file_info, dir_info)),
                 files=files,
             )
 
--- a/dandi/support/digests.py
+++ b/dandi/support/digests.py
@@ -24,6 +24,7 @@ from pathlib import Path
 
 from dandischema.digests.dandietag import DandiETag
 from fscacher import PersistentCache
+from zarr_checksum.checksum import ZarrChecksum, ZarrChecksumManifest
 from zarr_checksum.tree import ZarrChecksumTree
 
 from .threaded_walk import threaded_walk
@@ -134,3 +135,31 @@ def md5file_nocache(filepath: str | Path) -> str:
     present in Zarrs
     """
     return Digester(["md5"])(filepath)["md5"]
+
+
+def checksum_zarr_dir(
+    files: dict[str, tuple[str, int]], directories: dict[str, tuple[str, int]]
+) -> str:
+    """
+    Calculate the Zarr checksum of a directory only from information about the
+    files and subdirectories immediately within it.
+
+    :param files:
+        A mapping from names of files in the directory to pairs of their MD5
+        digests and sizes
+    :param directories:
+        A mapping from names of subdirectories in the directory to pairs of
+        their Zarr checksums and the sum of the sizes of all files recursively
+        within them
+    """
+    manifest = ZarrChecksumManifest(
+        files=[
+            ZarrChecksum(digest=digest, name=name, size=size)
+            for name, (digest, size) in files.items()
+        ],
+        directories=[
+            ZarrChecksum(digest=digest, name=name, size=size)
+            for name, (digest, size) in directories.items()
+        ],
+    )
+    return manifest.generate_digest().digest
--- a/dandi/support/tests/test_digests.py
+++ b/dandi/support/tests/test_digests.py
@@ -7,12 +7,15 @@
 #
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 
+from __future__ import annotations
+
 from pathlib import Path
 
+import pytest
 from pytest_mock import MockerFixture
 
 from .. import digests
-from ..digests import Digester, get_zarr_checksum
+from ..digests import Digester, checksum_zarr_dir, get_zarr_checksum
 
 
 def test_digester(tmp_path):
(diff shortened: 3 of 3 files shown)


dandi/files/zarr.py
```diff
@@ -153,25 +152,25 @@ def stat(self) -> ZarrStat:
...
                     files.extend(st.files)
                 else:
                     size += p.size
-                    file_md5s[p.name] = (md5file_nocache(p.filepath), p.size)
+                    file_info[p.name] = (md5file_nocache(p.filepath), p.size)
                     files.append(p)
             return ZarrStat(
                 size=size,
-                digest=Digest.dandi_zarr(get_checksum(file_md5s, dir_md5s)),
+                digest=Digest.dandi_zarr(checksum_zarr_dir(file_info, dir_info)),
```
yarikoptic, line 173: do we really need to "manually" loop here via `dirpath.iterdir()` or could make use of https://github.com/dandi/zarr_checksum#python

```python
compute_zarr_checksum(yield_files_local("local_path"))
```
? i.e. could we reuse `yield_files_local` to collect `files` and then just pass to `compute_zarr_checksum` instead of reimplementing it as `checksum_zarr_dir`?

yarikoptic: hm, that sounds like a possible cause of differences/confusion which we ideally clarify/unify while at it.

I started writing items below, but then looked at the zarr-python implementation of traversal at https://github.com/zarr-developers/zarr-python/blob/HEAD/zarr/storage.py#L1174  and it seems to list all files really. Although "not guaranteed", I guess, but are you sure that it actually excludes any "extraneous" files @jwodder?

some thoughts I started to compose:

- in dandi-cli we typically (unless DEVEL option allow_any_path or bids datasets) upload only recognized by us files too, i.e. by default we ignore "extraneous" files, so `zarr_checksum`'s default (and only ATM) behavior is somewhat inline with that, while our own support of `zarr` diverges from it
- in principle we should not silently ignore extraneous files within `.zarr` -- but I guess zarr validation doesn't fail (does not care) about them right?
- do we have any knowledge of such extraneous files uploaded already? might be good to know and since we have dandizarrs datalad datasets on drogon -- we might be able to "quickly" (nothing is quick when dealing with that many files) check.

AlmightyYakob: It does seem that `NestedDirectoryStore.keys` essentially calls `os.walk` on the local file path. I also assumed more was at play.

yarikoptic: Before any other decision making: @AlmightyYakob @jwodder do you think it might be valuable to suggest `zarr_checksum` to become **the** implementation of checksumming for zarr within their library?  then we could harmonize/centralize all that to that library instead of maintaining our own code. WDYT?

jwodder: Aside from the bit that this PR fixes, dandi-cli already uses `zarr_checksum` for assembling digests of individual Zarr entries into a checksum for the entire Zarr.  The Zarr-digesting code in dandi-cli itself is largely concerned with iterating over Zarr trees efficiently using threads, including [this thing](https://github.com/dandi/dandi-cli/blob/a51e0a7abb281204e58789facd6ee8914f75beb5/dandi/files/zarr.py#L577), which doesn't really belong in `zarr_checksum`.

AlmightyYakob: > Before any other decision making: @AlmightyYakob @jwodder do you think it might be valuable to suggest `zarr_checksum` to become **the** implementation of checksumming for zarr within their library? then we could harmonize/centralize all that to that library instead of maintaining our own code. WDYT?

That would mean that we're once again requiring the `zarr` library, which has heavy dependencies, and whose concern is almost entirely disjoint with our owns (e.g. we use none of the zarr library tools as is for checksumming). So I think it's a bad idea.

jwodder: @yarikoptic 

> do you think it might be valuable to suggest zarr_checksum to become the implementation of checksumming for zarr within their library?

No.  The Zarr checksumming algorithm is something the Dandi project came up with for Dandi's purposes, and I am not aware of the Zarr project expressing any interest in it.

---

What exactly are your remaining concerns, if any, about this PR?

yarikoptic: > > Before any other decision making: @AlmightyYakob @jwodder do you think it might be valuable to suggest `zarr_checksum` to become **the** implementation of checksumming for zarr within their library? then we could harmonize/centralize all that to that library instead of maintaining our own code. WDYT?
> 
> That would mean that we're once again requiring the `zarr` library, which has heavy dependencies, and whose concern is almost entirely disjoint with our owns (e.g. we use none of the zarr library tools as is for checksumming).

Just for clarity, @jjnesbitt -- didn't you just [above](https://github.com/dandi/dandi-cli/pull/1371/files#r1421063729) confirm using zarr library's construct within zarr_checksum, which is

```
zarr_checksum/generators.py:    from zarr.storage import NestedDirectoryStore
```
which is indeed just a delayed import, but anyways `zarr_checksum` depends on `zarr` anyways and just for this "traverse" functionality?   My point is that ATM `zarr_checksum` does depend on `zarr` and its heavy dependencies, correct?

jjnesbitt: It _currently_ does depend on the `zarr` library, but as we've learned in this discussion, there's no good reason for it. It only depends on the `zarr` library to discover local files to checksum, which could be trivially replaced with `os.walk`.

dandi/support/digests.py
```diff
@@ -134,3 +135,31 @@ def md5file_nocache(filepath: str | Path) -> str:
...
+        A mapping from names of files in the directory to pairs of their MD5
+        digests and sizes
+    :param directories:
+        A mapping from names of subdirectories in the directory to pairs of
+        their Zarr checksums and the sum of the sizes of all files recursively
+        within them
+    """
+    manifest = ZarrChecksumManifest(
+        files=[
+            ZarrChecksum(digest=digest, name=name, size=size)
```
yarikoptic, line 157: note to myself here: that instead of using dandi-schema's ZarrJSONChecksumSerializer as it was used in original defined in dandi-schema `get_checksum` we would start using `zarr_checksum` defined `ZarrChecksumManifest` and `ZarrChecksum`.
That would allow to prune that `dandischema.digests.zarr`

dandi/support/digests.py
```diff
@@ -134,3 +135,31 @@ def md5file_nocache(filepath: str | Path) -> str:
...
+            for name, (digest, size) in files.items()
```
yarikoptic, line 158: note to myself (worth adding as a comment???): if previously we explicitly sorted in

https://github.com/dandi/dandi-schema/blob/bdb7dd681435f1092814041eb01eb4fabe87bf57/dandischema/digests/zarr.py#L150

now it is not needed since ZarrChecksumManifest.generate_digest would do it: https://github.com/dandi/zarr_checksum//blob/HEAD/zarr_checksum/checksum.py#L80

dandi/files/zarr.py
jwodder: @yarikoptic 

> more efficient threaded walk. Ideally IMHO `zarr_checksum` should just become as efficient, ie carry such an implementation. @jwodder - do you see why that should not or could not be done?

Which efficient threaded walk are you referring to?  [`get_zarr_checksum()`](https://github.com/dandi/dandi-cli/blob/72816ab51a4f168bd9629cc6ac9a6a4edd585bee/dandi/support/digests.py#L101) or the digesting-for-uploading machinery?  I'm fine with the former being in `zarr_checksum`, but the latter currently does not belong there.

> Shouldn't we centralize the walker for zarr, e.g. within zarr_checksum?

Are you going to create an issue about this?  If not, your concern is just going to go into the void.

> So, for now I guess we are doomed to have this duplicated implementation but I think in the longer run we might want to see how we could avoid it.

The entire `ZarrAsset.stat()` code is currently unused by the library.  (I'd look up what commit removed the last use, but I can't get `git log -S` to honor any word boundary syntax I know of.)

yarikoptic: > The entire `ZarrAsset.stat()` code is currently unused by the library. (I'd look up what commit removed the last use, but I can't get `git log -S` to honor any word boundary syntax I know of.)

something like 47c524c60e4987d36719842645305ecbdaee4627  which seemed was a big refactoring which stopped using `.stat().`?

jwodder: No, that commit only got rid of stat'ing Zarr entries on the Archive.

yarikoptic: > > more efficient threaded walk. Ideally IMHO `zarr_checksum` should just become as efficient, ie carry such an implementation. @jwodder - do you see why that should not or could not be done?
> 
> Which efficient threaded walk are you referring to? [`get_zarr_checksum()`](https://github.com/dandi/dandi-cli/blob/72816ab51a4f168bd9629cc6ac9a6a4edd585bee/dandi/support/digests.py#L101) or the digesting-for-uploading machinery? I'm fine with the former being in `zarr_checksum`, but the latter currently does not belong there.

We are on the same page here. filed https://github.com/dandi/zarr_checksum/issues/52 

> > Shouldn't we centralize the walker for zarr, e.g. within zarr_checksum?
> 
> Are you going to create an issue about this? If not, your concern is just going to go into the void.

- https://github.com/dandi/dandi-cli/issues/1387

PR #1374, opened 2023-12-11
Double-check file sizes before & after uploading

### LINKED ISSUE #1373: Make file upload  be robust to file size changes
Per request of @jwodder in https://github.com/dandi/dandi-cli/issues/1257#issuecomment-1850127241 filing it as a separate issue.

Background: As discovered [in that issue](https://github.com/dandi/dandi-cli/issues/1257#issuecomment-1839101911) we caught NFS (or some layer above on the way to Python) lying about file size to be 0 whenever it is not: the reported size gets changed within few seconds.  We also spotted S3 failing some uploads due to "Content-MD5 you specified did not match what we received", an exact reason for which we do not know.
So, apparently we cannot really rely on file system to provide us robust file size, and thus we should 1. provide mitigation to the 0-size file issue and to possibly help isolate any other incorrect file size reporting add a safe-guard.  

I am filing it as a single issue since IMHO two aspects are very related and likely to be implemented in a concerted effort.   I will provide only high level description of desired behavior instead of mandating any particular implementation.

- Should be in effect for *any* file upload,  be it a blob or a file within zarr folder upload
- Upon initiation of upload do a *filesize check*: a file size reported to be 0, loop for up to 2 seconds (sleeping may be 100ms between) re-checking the file size. If it changes from 0 - can immediately proceed forward. If it remains 0 - proceed forward with 0 file size after those 2 seconds.
- Upon completion of upload, be it successful or failed with an error, compare filesize with the one obtained in previous *filesize check* step. If size differs: if upload finished    successfully - do raise an (`IOError` or other you find more appropriate) exception summarizing size changes, if upload errorred out - just log an ERROR level log message summarizing size changes.

jwodder:
Closes #1373.

### DIFF
--- a/dandi/utils.py
+++ b/dandi/utils.py
@@ -852,15 +852,11 @@ def post_upload_size_check(path: Path, pre_check_size: int, erroring: bool) -> N
     # More checks for NFS flakiness
     size = path.stat().st_size
     if size != pre_check_size:
+        msg = (
+            f"Size of {path} was {pre_check_size} at start of upload but is"
+            f" now {size} after upload"
+        )
         if erroring:
-            lgr.error(
-                "Size of %s was %d at start of upload but is now %d after upload",
-                path,
-                pre_check_size,
-                size,
-            )
+            lgr.error(msg)
         else:
-            raise RuntimeError(
-                f"Size of {path} was {pre_check_size} at start of upload but is"
-                f" now {size} after upload"
-            )
+            raise RuntimeError(msg)
--- a/dandi/utils.py
+++ b/dandi/utils.py
@@ -19,6 +19,7 @@ import re
 import shutil
 import subprocess
 import sys
+from time import sleep
 import traceback
 import types
 from typing import IO, Any, List, Optional, Protocol, TypeVar, Union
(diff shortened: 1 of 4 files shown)


dandi/utils.py
```diff
@@ -834,3 +835,32 @@ def _prepare_path_parts(paths: Iterable[str | PurePath]) -> list[tuple[str, ...]
...
+                "Size of %s was %d at start of upload but is now %d after upload",
+                path,
+                pre_check_size,
+                size,
+            )
+        else:
+            raise RuntimeError(
+                f"Size of {path} was {pre_check_size} at start of upload but is"
+                f" now {size} after upload"
+            )
```
yarikoptic, line 866: Now that we know it is safe for logger to just have a string possibly with `%`s, let's just please anti-duplication gods with:

```suggestion
        msg = f"Size of {path} was {pre_check_size} at start of upload but is" \
            f" now {size} after upload"
        if erroring:
            lgr.error(msg)
        else:
            raise RuntimeError(msg)
```

?

yarikoptic: having a little tiny test to cover this codepath would be also great.

jwodder: @yarikoptic How exactly am I supposed to change the filesize in the middle of an upload for a test?

yarikoptic: unittest at the level of `post_upload_size_check` function itself -- just provide a wrong filesize thus to trigger logging or exception

yarikoptic: if you want for the entire upload -- just as a thought excercise, I thought it could be via instrumentation of that `super_len` function so it changes the file after returning the size ;-)

PR #1381, opened 2023-12-18
Update for Pydantic v2

### LINKED ISSUE #1352: Update for Pydantic 2.0
Blocked by:

- https://github.com/dandi/zarr_checksum/issues/16
- https://github.com/dandi/dandi-schema/issues/176

jwodder:
Closes #1352.

~Blocked by https://github.com/dandi/dandi-archive/issues/1820.~

~Do not merge until after https://github.com/dandi/dandi-schema/pull/203 is merged & released.  Once that is done, the "[TEMP] Use Pydantic 2.0 branch of dandischema" commit in this PR should be replaced with a commit that sets the `dandischema` version requirement to `~= 0.9.0` instead.~ [Done]


yarikoptic (approved):
Looks LGTM to me.

candleindark (approved):
LGTM

CodyCBakerPhD:
@yarikoptic What was the timeline on getting this merged/released?

jwodder:
@yarikoptic Various tests are currently failing because the Archive doesn't yet accept schema version 0.6.5.

yarikoptic:
@jwodder - could we TEMP test against version in 
- https://github.com/dandi/dandi-archive/pull/1823 
before it is merged/release so we know that all changes are ok with it? (also PR requires rebase)

dandi/utils.py
```diff
@@ -860,3 +867,17 @@ def post_upload_size_check(path: Path, pre_check_size: int, erroring: bool) -> N
...
+
+def joinurl(base: str, path: str) -> str:
+    """
+    Append a slash-separated ``path`` to a base URL ``base``.  The two
+    components are separated by a single slash, removing any excess slashes
+    that would be present after naïve concatenation.
+
+    If ``path`` is already an absolute URL, it is returned unchanged.
+    """
+    if path.lower().startswith(("http://", "https://")):
```
yarikoptic, line 880: Function name and description suggests being applicable to URL of any transport, not just https so better be smth like

```suggestion
    if re.match('[a-z]+://', path.lower()):
```

to avoid surprises.

yarikoptic: but also how it is different from a standard urllib's join?

```
In [3]: urllib.parse.urljoin('http://a.c/', '/1/2/3')
Out[3]: 'http://a.c/1/2/3'

In [4]: urllib.parse.urljoin('http://a.c/', 's3://1/2/3')
Out[4]: 's3://1/2/3'
```

jwodder: * The function is only meant for HTTP(S) URLs.

* It differs from `urljoin()` for URLs that have a path that doesn't end in `/`:

    ```
    >>> urljoin("https://example.com/foo/bar", "quux")
    'https://example.com/foo/quux'
    ```

    `joinurl()` would instead give `https://example.com/foo/bar/quux`.

yarikoptic: * then better name it `httpurljoin` or at least describe it in its description
* also please add into description  this difference in behavior as rationale

jwodder: Done: 728882f1

PR #1384, opened 2023-12-21
`dandi download dandi://…/dandiset.yaml` now downloads `dandiset.yaml`

### LINKED ISSUE #1383: provide support for download of dandiset.yaml alone
ATM we seems cannot point to `dandiset.yaml` directly since it is not a "true asset":

```
❯ dandi download dandi://dandi/000027@draft
PATH                             SIZE     DONE    DONE% CHECKSUM STATUS          MESSAGE   
000027/dandiset.yaml                                             done            updated   
000027/sub-RAT123/sub-RAT123.nwb 18.8 kB  18.8 kB  100%    ok    done                      
Summary:                         18.8 kB  18.8 kB                2 done          1 updated 
                                          100.00%                                          
2023-12-21 09:26:39,359 [    INFO] Logs saved in /home/yoh/.local/state/dandi-cli/log/20231221142637Z-16640.log

❯ dandi download dandi://dandi/000027@draft/dandiset.yaml
2023-12-21 09:26:43,473 [    INFO] Logs saved in /home/yoh/.local/state/dandi-cli/log/20231221142642Z-16681.log
Error: No asset at path 'dandiset.yaml'
```

but being able to download that file alone could be of use in cases e.g. when people are preparing a new dandiset for upload, so I think we should provide special case handling here, at least at the top level `download` command.

jwodder:
Closes #1383.

Note that this feature only applies to `dandi://` URLs, as GUI and API URLs cannot refer to a specific asset by path in the first place.

Note that this re-interpretation of `dandi://` URLs only happens when downloading and not when a `dandi://…/dandiset.yaml` URL is supplied to a `move`, `delete`, `ls`, or `reextract-metadata` command (I think those are all the places that `dandi://` URLs are accepted), as it wouldn't make sense for most of those.

### DIFF
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -28,7 +28,13 @@ import requests
 from . import get_logger
 from .consts import RETRY_STATUSES, dandiset_metadata_file
 from .dandiapi import AssetType, BaseRemoteZarrAsset, RemoteDandiset
-from .dandiarchive import DandisetURL, ParsedDandiURL, SingleAssetURL, parse_dandi_url
+from .dandiarchive import (
+    AssetItemURL,
+    DandisetURL,
+    ParsedDandiURL,
+    SingleAssetURL,
+    parse_dandi_url,
+)
 from .dandiset import Dandiset
 from .exceptions import NotFoundError
 from .files import LocalAsset, find_dandi_files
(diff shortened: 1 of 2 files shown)


yarikoptic:
Thank you @jwodder !

PR #1385, opened 2024-01-02
Fix service script tests for change in autogenerated date

jwodder:
The `update-dandiset-from-doi` tests are currently failing because they test against the Dandisets' final `citation` fields (among others), which the Archive automatically populates based on the year the Dandisets were created (among other things), and so our expected output is no longer accurate as of the new year.


yarikoptic:
Thank you for the quick fix.  I am not yet verse in `anys` so have difficulty figuring the patch out but that is ok -- CI is happy! ;)

PR #1393, opened 2024-01-23
Update zarr_checksum dependency to `~= 0.4.0`

jwodder:


jwodder:
@yarikoptic It removes some heavy dependencies from dandi-cli's dependency tree.

yarikoptic:
I get that, but
- for new user -- it would already install newer lighter version
- for existing user -- it would do nothing besides updating zarr_checksum -- those dependencies are already installed and would not even be removed.

so what is the point?

jwodder:
@yarikoptic 

> for new user -- it would already install newer lighter version

No, it wouldn't.  `dandi-cli`'s current `master` commit depends on `zarr_checksum ~= 0.3.2`, which has an upper bound of `< 0.4.0`.

Another point in favor of this change: `zarr_checksum`'s dependency on `boto3` has twice caused dependency issues with `backups2datalad` and its `aiobotocore` dependency.  In order to permanently avoid this problem, `zarr_checksum` must be updated to 0.4.0+ for both `backups2datalad` and `dandi` (the latter being a dependency of `backups2datalad`).

PR #1398, opened 2024-01-31
Add notes to docs about `get_metadata()` vs. `get_raw_metadata()`

### LINKED ISSUE #1363: [Bug] .get_metadata() from embargoed dandiset
```python
from dandi.dandiapi import DandiAPIClient

client = DandiAPIClient(token="my token")  # need token for embargo

dandiset_620 = client.get_dandiset(dandiset_id="000620")  # embargoed

dandiset_620_metadata = dandiset_620.get_metadata()

> ValidationError: 1 validation error for Dandiset
> access -> 0 -> __root__
>   An embargo end date is required for NIH awards to be in compliance with NIH resource sharing policy. (type=value_error)
```

I'd be happy to set this value on the dandiset, but I don't know how to and it wasn't required to create the dandiset either

jwodder:
Closes #1363.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -18,7 +18,7 @@ from urllib.parse import quote_plus, urlparse, urlunparse
 
 import click
 from dandischema import models
-from pydantic import BaseModel, Field, PrivateAttr, ValidationError
+from pydantic import BaseModel, Field, PrivateAttr
 import requests
 import tenacity
 
@@ -978,15 +978,7 @@ class RemoteDandiset:
             metadata.  Consider using `get_raw_metadata()` instead in order to
             fetch unstructured, possibly-invalid metadata.
         """
-        try:
-            return models.Dandiset.parse_obj(self.get_raw_metadata())
-        except ValidationError as e:
-            raise ValueError(
-                f"{type(e).__name__}: {e}\n\nNote: Only metadata for published"
-                " Dandiset versions can be expected to be valid.  Use"
-                " get_raw_metadata() instead to get unstructured,"
-                " possibly-invalid metadata."
-            )
+        return models.Dandiset.parse_obj(self.get_raw_metadata())
 
     def get_raw_metadata(self) -> dict[str, Any]:
         """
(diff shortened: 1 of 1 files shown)


dandi/dandiapi.py
```diff
@@ -971,8 +971,22 @@ def get_metadata(self) -> models.Dandiset:
...
+
+            Only published Dandiset versions can be expected to have valid
+            metadata.  Consider using `get_raw_metadata()` instead in order to
+            fetch unstructured, possibly-invalid metadata.
         """
-        return models.Dandiset.parse_obj(self.get_raw_metadata())
+        try:
+            return models.Dandiset.parse_obj(self.get_raw_metadata())
+        except ValidationError as e:
+            raise ValueError(
```
yarikoptic, line 984: IMHO we should not adapt into another type of exception here -- [ValidationError](https://github.com/pydantic/pydantic/blob/HEAD/pydantic/v1/error_wrappers.py#L50) has `.errors()` etc...

could you please subclass it locally with `__init__` to copy over errors and model and adjust its `__str__` to get that note ?

jwodder: While subclassing looks like it would work in Pydantic 1.x, it won't work in 2.x, as there `ValidationError` is implemented in Rust, and the only public Python method I see for creating a `ValidationError` is a static method that can only ever return a `ValidationError`, not a custom subclass.

yarikoptic: oh! I have missed `**v1**` in that `pydantic/v1/error_wrappers.py` path... what could possibly go wrong by mixing Python and Rust? I guess we are yet to see and for now just enjoy public classes/interfaces like

```
In [5]: ValidationError?
Init signature: ValidationError(self, /, *args, **kwargs)
Docstring:      Inappropriate argument value (of correct type).
File:           ~/venvs/dev3.11/lib/python3.11/site-packages/pydantic_core/_pydantic_core.cpython-311-x86_64-linux-gnu.so
Type:           type
Subclasses:   
```

which cannot even

```
In [6]: ValidationError()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[6], line 1
----> 1 ValidationError()

TypeError: No constructor defined
```

heh...  I really do not see then a way we could improve UX/DX here then... we should not issue warnings etc. I guess let's just leave the `note` in the doc string and be done with it.

yarikoptic:
Thank you @jwodder

PR #1404, opened 2024-02-12
Draft for BIDS organize

TheChymera:
Early work in progress.


yarikoptic:
I think this could be safely closed , but attn @CodyCBakerPhD as working on this outside of dandi-cli codebase

PR #1414, opened 2024-02-28
Add arguments for API query parameters when fetching all Dandisets; support creating embargoed Dandisets

### LINKED ISSUE #1413: get_dandisets: Expose options (embargoed, empty, draft) of the /dandisets/ endpoint
I thought that as long as I authenticate I would also get embargoed datasets listed, as I believe is done in web UI.  But I found no Embargoed dandisets in returned list. Looking at our `get_dandisets` - apparently we do not expose any of the query parameters
image
for 
image

Immediate need is to get `embargoed` dandisets. I generally would not mind even adding "search" parameter.

jwodder:
Closes #1413.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -613,10 +613,24 @@ class DandiAPIClient(RESTFullAPIClient):
         ):
             yield RemoteDandiset.from_data(self, data)
 
-    def create_dandiset(self, name: str, metadata: dict[str, Any]) -> RemoteDandiset:
-        """Creates a Dandiset with the given name & metadata"""
+    def create_dandiset(
+        self, name: str, metadata: dict[str, Any], *, embargo: bool = False
+    ) -> RemoteDandiset:
+        """
+        Creates a Dandiset with the given name & metadata.  If ``embargo`` is
+        `True`, the resulting Dandiset will be embargoed.
+
+        .. versionchanged:: 0.61.0
+
+            ``embargo`` argument added
+        """
         return RemoteDandiset.from_data(
-            self, self.post("/dandisets/", json={"name": name, "metadata": metadata})
+            self,
+            self.post(
+                "/dandisets/",
+                json={"name": name, "metadata": metadata},
+                params={"embargo": "true" if embargo else "false"},
+            ),
         )
 
     def check_schema_version(self, schema_version: str | None = None) -> None:
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -554,13 +554,63 @@ class DandiAPIClient(RESTFullAPIClient):
                     return d.for_version(version_id)
             return d
 
-    def get_dandisets(self) -> Iterator[RemoteDandiset]:
+    def get_dandisets(
+        self,
+        *,
+        draft: bool | None = None,
+        embargoed: bool | None = None,
+        empty: bool | None = None,
+        mine: bool | None = None,
+        order: str | None = None,
+        search: str | None = None,
+    ) -> Iterator[RemoteDandiset]:
         """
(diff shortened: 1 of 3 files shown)


dandi/dandiapi.py
```diff
@@ -554,13 +554,62 @@ def get_dandiset(
...
+            "/dandisets/",
+            params={
+                "draft": draft,
+                "embargoed": embargoed,
+                "empty": empty,
+                "ordering": order,
+                "search": search,
+                "user": "me" if mine else None,
+            },
+        ):
```
yarikoptic, line 612: could we avoid specifying it at all if it was not specified in the call, i.e. `None`?

jwodder: @yarikoptic If you mean omitting the fields from `params`, that's unnecessary, as setting a param's value to `None` in `requests` is the same as not setting that param.

dandi/dandiapi.py
```diff
@@ -554,13 +554,63 @@ def get_dandiset(
...
+        """
+        for data in self.paginate(
+            "/dandisets/",
+            params={
+                "draft": draft,
+                "embargoed": embargoed,
+                "empty": empty,
+                "ordering": order,
+                "search": search,
+                "user": "me" if mine else None,
```
yarikoptic, line 611: should we stick to API's interface here and expose as `user`?  May be admins (later?) might be able to query for some other users?

jwodder: No, I don't think having a parameter that can only take one value makes much sense.

yarikoptic (changes requested):
Let's add a test or two. 

ATM we do not really test/deal much with embargoed dandisets seems to me, primarily since we do not even expose `embargo` option:

```
dandi/dandiapi.py:    def create_dandiset(self, name: str, metadata: dict[str, Any]) -> RemoteDandiset:
```

So let's as part of this PR expose that option, create also embargoed dandiset and test listing here for a set of query parameters to see that they map correctly to the API call (as you mentioned -- requests would simply ignore if we do not change them from `None`)

dandi/dandiapi.py
yarikoptic: I agree... as we are to RF order vs ordering, please file another issue on this particular one. May be indeed the intent would be to expose querying for other users as well. If not -- we better recommend to refactor it to `mine`.

jwodder: Issue filed: https://github.com/dandi/dandi-archive/issues/1879

jwodder:
@yarikoptic 

> test listing here for a set of query parameters to see that they map correctly to the API call

What exactly should be tested?  Should there just be a test for each `bool` argument that checks that a matching Dandiset is/isn't returned when the argument is `true`/`false`?

Note that there can't be a test that checks the entirety of the `get_dandisets()` return value, as the test Dandisets created for most tests persist in later tests, and it's possible for tests to be skipped or run out of order, so there's no guarantee about what Dandisets will exist in the test Archive when a test runs.

yarikoptic:
> Note that there can't be a test that checks the entirety of the `get_dandisets()` return value, as the test Dandisets created for most tests persist in later tests, and it's possible for tests to be skipped or run out of order, so there's no guarantee about what Dandisets will exist in the test Archive when a test runs.

I didn't realize that -- I thought that we just have a fixture to populate a number of dandisets and then test on them.


> What exactly should be tested? Should there just be a test for each `bool` argument that checks that a matching Dandiset is/isn't returned when the argument is `true`/`false`?

I think testing for that would suffice. Fixture could ensure to populate at least one dandiset to satisfy each criteria, then (with the light of above) - we could test that there is at least one dandiset satisfying the desired criteria.

yarikoptic:
Thank you!

PR #1415, opened 2024-03-01
Use `yarl` to clean up some code

### LINKED ISSUE #1380: Use a dedicated URL-handling library
There are various places in the code that would benefit from the use of a modern URL-handling library, such as [furl](https://github.com/gruns/furl), [purl](https://github.com/codeinthehole/purl), or [yarl](https://github.com/aio-libs/yarl).  These include:

* [x] [`is_same_url()` in `delete.py`](https://github.com/dandi/dandi-cli/blob/cb00a7cb8e92d86dd78a9330780fc4c1eab0cfea/dandi/delete.py#L248)
* ~[`extract_species()` in `metadata/util.py`](https://github.com/dandi/dandi-cli/blob/cb00a7cb8e92d86dd78a9330780fc4c1eab0cfea/dandi/metadata/util.py#L437-L443)~ *(yarl doesn't help here; see #1416 instead)*
* [x] [`get_content_url()` in `dandiapi.py`](https://github.com/dandi/dandi-cli/blob/cb00a7cb8e92d86dd78a9330780fc4c1eab0cfea/dandi/dandiapi.py#L1449-L1450)
* [x] [`get_instance()` in `utils.py`](https://github.com/dandi/dandi-cli/blob/cb00a7cb8e92d86dd78a9330780fc4c1eab0cfea/dandi/utils.py#L568-L569)
* [x] [`is_page2_url()` in `utils.py`](https://github.com/dandi/dandi-cli/blob/cb00a7cb8e92d86dd78a9330780fc4c1eab0cfea/dandi/utils.py#L779)
* [x] [`_get_instance()` in `utils.py`](https://github.com/dandi/dandi-cli/blob/f256a55389033f7655d4fd905e6dc973cd785dfc/dandi/utils.py#L620-L630)
* ~[`joinurl()` in `utils.py`](https://github.com/dandi/dandi-cli/blob/f256a55389033f7655d4fd905e6dc973cd785dfc/dandi/utils.py#L872)~ *(yarl doesn't help here)*
* ~Most of `dandi/dandiarchive.py`?~ *(I think rewriting this would be more trouble than it's worth)*
* implementing the URL validation needed for #1308

We should use such a library to improve the code.

jwodder:
Closes #1380.

### DIFF
--- a/dandi/delete.py
+++ b/dandi/delete.py
@@ -6,6 +6,7 @@ from operator import attrgetter
 from pathlib import Path
 
 import click
+from yarl import URL
 
 from .consts import DRAFT, ZARR_EXTENSIONS, DandiInstance, dandiset_metadata_file
 from .dandiapi import DandiAPIClient, RemoteAsset, RemoteDandiset
@@ -246,5 +247,8 @@ def find_local_asset(filepath: str) -> tuple[str, str]:
 
 
 def is_same_url(url1: str, url2: str) -> bool:
-    # TODO: Use a real URL library like furl, purl, or yarl
-    return url1.rstrip("/") == url2.rstrip("/")
+    u1 = URL(url1)
+    u1 = u1.with_path(u1.path.rstrip("/"))
+    u2 = URL(url2)
+    u2 = u2.with_path(u2.path.rstrip("/"))
+    return u1 == u2
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -14,13 +14,13 @@ import re
 from time import sleep, time
 from types import TracebackType
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
-from urllib.parse import quote_plus, urlparse, urlunparse
 
 import click
 from dandischema import models
 from pydantic import BaseModel, Field, PrivateAttr
 import requests
 import tenacity
+from yarl import URL
 
 from . import get_logger
 from .consts import (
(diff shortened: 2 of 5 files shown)


dandi/delete.py
```diff
@@ -246,5 +247,8 @@ def find_local_asset(filepath: str) -> tuple[str, str]:
 
 
 def is_same_url(url1: str, url2: str) -> bool:
-    # TODO: Use a real URL library like furl, purl, or yarl
-    return url1.rstrip("/") == url2.rstrip("/")
+    u1 = URL(url1)
+    u1 = u1.with_path(u1.path.rstrip("/"))
+    u2 = URL(url2)
+    u2 = u2.with_path(u2.path.rstrip("/"))
+    return u1 == u2
```
yarikoptic, line 254: so how is this better? will it reorder query params or smth to find equality? if so -- would be nice to add a testcase which would solidify adding such capability which was not present before.

jwodder: It normalizes various URL bits, such as lowercasing the scheme & host and normalizing internationalized domain names.

jwodder: Tests added: e58df625

yarikoptic:
I filed an issue
- https://github.com/OntoZoo/ontobee/issues/196
but given that the last commit was 6 years ago, I wonder if anyone cares???  @satra - do you know any of those folks somehow?  `whois` record is all privatized so nobody to ask really.

yarikoptic:
fwiw ontobee is back online, tests passed after merging

PR #1416, opened 2024-03-01
Clean up URL parsing in `extract_species()`

jwodder:
yarl doesn't really seem like it would help with this bit, but I came up with a different way to make the code arguably cleaner.

### DIFF
--- a/dandi/metadata/util.py
+++ b/dandi/metadata/util.py
@@ -329,68 +329,68 @@ def extract_cellLine(metadata: dict) -> str | None:
         return None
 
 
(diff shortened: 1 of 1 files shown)


dandi/metadata/util.py
```diff
@@ -329,66 +329,68 @@ def extract_cellLine(metadata: dict) -> str | None:
         return None
 
 
+NCBITAXON_URI_TEMPLATE = "http://purl.obolibrary.org/obo/NCBITaxon_{}"
```
yarikoptic, line 332: pushed a renaming for that template variable. If really don't like it -- revert, but I think that it is better when more specific like this

yarikoptic (approved):

PR #1417, opened 2024-03-07
Convert DOI URLs in `related_publications` to related resources

### LINKED ISSUE #1308: publications in asset metadata
In https://github.com/dandi/helpdesk/discussions/100, @GaelleChapuis wanted to know how to query for NWB files within [the big IBL dandiset](https://dandiarchive.org/dandiset/000409) that are associated with a given publication. This dandiset contains data that are relevant to several different publications, and each NWB file lists which publications it is associated with under the field `nwbfile.related_publications`. It looks like the dandi cli currently extracts these values and adds them to the dandiset-level metadata, but not the asset-level metadata. As a result, our best approach to query for these files currently is it iterate over all of the NWB files in the dandiset, open them in stream mode, and read this value. This is very slow compared to iterating over the asset-level metadata.

The solution I would like is to use these extracted values to also populate the asset-level metadata so they can be differentiated by this attribute. Since `RelatedResource` is part of the `CommonModel` that `Asset` inherits, it looks like the schema can support this already. It's just a matter of having the dandi cli extract this information. Note that NWB files do not always contain reference to publications, and they should not be required to, since sometimes data is not associated with any publication, so this extraction logic should be robust to the case where `nwbfile.related_publication is None`.

jwodder:
Closes #1308.

### DIFF
--- a/dandi/metadata/util.py
+++ b/dandi/metadata/util.py
@@ -11,6 +11,7 @@ from xml.dom.minidom import parseString
 from dandischema import models
 import requests
 import tenacity
+from yarl import URL
 
 from .. import __version__
 from ..utils import ensure_datetime
@@ -583,6 +584,29 @@ def extract_digest(metadata: dict) -> dict[models.DigestType, str] | None:
         return None
 
 
+def extract_related_resource(metadata: dict) -> list[models.Resource] | None:
+    pubs = metadata.get("related_publications")
+    if not isinstance(pubs, (list, tuple)):
+        return None
+    related = []
+    for v in pubs:
+        if not isinstance(v, str):
+            continue
+        try:
+            u = URL(v)
+        except ValueError:
+            continue
+        if u.scheme not in ("http", "https") or u.host != "doi.org":
+            continue
+        related.append(
+            models.Resource(
+                identifier=v,
+                relation=models.RelationType.IsDescribedBy,
+            )
+        )
+    return related
+
+
 FIELD_EXTRACTORS: dict[str, Callable[[dict], Any]] = {
     "wasDerivedFrom": extract_wasDerivedFrom,
     "wasAttributedTo": extract_wasAttributedTo,
@@ -595,6 +619,7 @@ FIELD_EXTRACTORS: dict[str, Callable[[dict], Any]] = {
     "anatomy": extract_anatomy,
     "digest": extract_digest,
     "species": extract_species,
+    "relatedResource": extract_related_resource,
 }
 
 
--- a/dandi/tests/data/metadata/metadata2asset_3.json
+++ b/dandi/tests/data/metadata/metadata2asset_3.json
@@ -92,5 +92,12 @@
         "name": "Cyperus bulbosus"
       }
     }
+  ],
+  "relatedResource": [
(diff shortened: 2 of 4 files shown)


bendichter:
LGTM!

bendichter:
It's currently marked as WIP

jwodder:
@bendichter How about now?

jwodder:
@yarikoptic Ping.

dandi/metadata/util.py
```diff
@@ -583,6 +584,29 @@ def extract_digest(metadata: dict) -> dict[models.DigestType, str] | None:
...
+        try:
+            u = URL(v)
+        except ValueError:
+            continue
+        if u.scheme not in ("http", "https") or u.host != "doi.org":
+            continue
+        related.append(
+            models.Resource(
+                identifier=v,
+                relation=models.RelationType.IsDescribedBy,
```
yarikoptic, line 604: I think this one is is too specific to assume that any reference is the description of the data we find in the file.
What we can say only that the file references that publication, and hence I would better go with 

```suggestion
                relation=models.RelationType.References,
```

as the default.  WDYT @bendichter @satra ?

bendichter: I see your point that we cannot necessarily infer the relationship between the data and the paper.

The [DataCite definition for "References"](https://datacite-metadata-schema.readthedocs.io/en/4.5/appendices/appendix-1/relationType/#references) is "indicates B is used as a source of information for A." Here, this would mean that the paper is used as a source of information for the Dandiset. I don't know if this really fits here.

I have been wondering what the best way to associate papers with datasets using these relations. What are the different types of ways a paper and a dataset can be associated? So far, I have been using [isDescribedBy](https://datacite-metadata-schema.readthedocs.io/en/4.5/appendices/appendix-1/relationType/#isdescribedby) (indicates that A describes B) for everything, assuming that the publication is the primary publication in which the data is introduced and described. Another type of relationship might be that a paper reuses a dataset. In this case, [IsSupplementTo](https://datacite-metadata-schema.readthedocs.io/en/4.5/appendices/appendix-1/relationType/#issupplementto) (indicates that A is a supplement to B ) might be a better choice, since it applies to both. Are there other relationship types that describe how a paper and dandiset might be related? What are the different use-cases here?

This is challenging in part because there is currently no aspect of the dandi schema that allows us to describe what the resource is (see issue [here](https://github.com/dandi/dandi-schema/issues/231)), so that needs to be inferred at least somewhat by the relation type.

jwodder: @bendichter Now that dandi/dandi-schema#231 has been resolved, how can we move this PR forwards?

bendichter: Yes, exactly. My preference would be to change to schema to make this optional. Now that we have a resource type I am much less reliant on this particular field, though I do appreciate the caveat that optional fields are very rarely populated. I would also be happy with defaulting to IsDescribedBy or IsSupplementTo.

yarikoptic: ok, to expedite, let's proceed with IsDescribedBy and see where it takes us ;-)

PR #1423, opened 2024-03-22
Xfail flaky ontobee tests, unless running daily tests

### LINKED ISSUE #1422: Ontobee: mark the tests, allow for that mark to xfail atm
We know that ontobee is down ATM. We need to exclude those tests but without requiring any downstream testing to adjust. So we need to xfail such tests.

Ideally we would still fail cron jobs so we remain aware that ontobee still has issues and ping to get it resolved

jwodder:
Closes #1422.

### DIFF
--- a/dandi/tests/test_metadata.py
+++ b/dandi/tests/test_metadata.py
@@ -29,6 +29,7 @@ from dandischema.models import Dandiset as DandisetMeta
 from dateutil.tz import tzutc
 from pydantic import ByteSize
 import pytest
+import requests
 from semantic_version import Version
 
 from .fixtures import SampleDandiset
@@ -234,6 +235,12 @@ def test_timedelta2duration(td: timedelta, duration: str) -> None:
     assert timedelta2duration(td) == duration
 
 
+@pytest.mark.xfail(
+    condition="not config.getoption('--scheduled')",
+    reason="Flaky ontobee site",
+    strict=False,
+    raises=requests.RequestException,
+)
 @mark.skipif_no_network
 @pytest.mark.parametrize(
     "filename, metadata",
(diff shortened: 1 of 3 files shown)


dandi/tests/test_metadata.py
```diff
@@ -234,6 +235,12 @@ def test_timedelta2duration(td: timedelta, duration: str) -> None:
     assert timedelta2duration(td) == duration
 
 
+@pytest.mark.xfail(
+    condition="not config.getoption('--scheduled')",
+    reason="Flaky ontobee site",
+    strict=False,
+    raises=requests.RequestException,
+)
```
yarikoptic, line 243: could you please place this into `@mark.xfail_ontobee` similarly to `@mark.skipif_no_network` to avoid duplication/clutter?

yarikoptic: @jwodder ping on this one so we could bring our tests back into greenland

jwodder: @yarikoptic The custom `mark` machinery only seems to support custom `skipif` decorators, not `xfail`.  I think creating a custom `mark.xfail_ontobee` would be more trouble than it's worth.

yarikoptic: Hmm, ok

What about just binding decorator into a helper "variable" eg `@mark_xfail_ontobee` or that wouldn't work?

jwodder: It seems to work, but now the tests are failing due to #1424.

yarikoptic: ok, since seems to be unrelated -- let's proceed. Thank you @jwodder

Issue #1407, opened 2024-02-15
`DONE%` value when downloading a Zarr asset is not correct


kabilar:
## Description

When downloading a single Zarr asset or downloading an entire Dandiset that contains a Zarr asset, I have found that the `DONE%` value for the Zarr is constantly set as 98% or 99%, whereas the `Summary` displays a seemingly more accurate progress (e.g. 1.95% in the screenshot below).

![image](https://github.com/dandi/dandi-cli/assets/871137/e2228a69-b5af-4ab6-93a9-477e916f3252)

## Reproducibility

1. Version: dandi==0.59.1
1. Command to download Zarr asset from Dandiset 000108:
  ```
  dandi download https://api.dandiarchive.org/api/assets/a25835b8-2293-4bf3-81f0-fbb3b22196fd/download/
  ```

Thank you

PR #1443, opened 2024-05-14
OPT+RF of zarr downloads: do not wait for full files listing + compute %done from total zarr size

yarikoptic:
Individual commits have more description.

Closes: #1407 . @kabilar could you please check if works for you ok'ish?

edit 1: I think initial OPT commit is not entirely complete -- somewhere in the sandwich of all the generators there is still waiting  for an iterable to exhaust itself.

### DIFF
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -1038,7 +1038,7 @@ class ProgressCombiner:
         )
         return {
             "done": total_downloaded,
-            "done%": total_downloaded / self.zarr_size * 100,
+            "done%": total_downloaded / self.zarr_size * 100 if self.zarr_size else 0,
         }
 
     def set_status(self, statusdict: dict) -> None:
@@ -1082,7 +1082,8 @@ class ProgressCombiner:
                 self.maxsize += size
             self.set_status(out)
             yield out
-            yield self.get_done()
+            if self.zarr_size:
+                yield self.get_done()
         elif keys == ["size"]:
             self.files[path].size = size
             self.maxsize += status["size"]
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -878,33 +878,40 @@ def _download_zarr(
     # Avoid heavy import by importing within function:
     from .support.digests import get_zarr_checksum
 
-    download_gens = {}
-    entries = list(asset.iterfiles())
+    # we will collect them all while starting the download
+    # with the first page of entries received from the server.
+    entries = []
     digests = {}
+    pc = ProgressCombiner(zarr_size=asset.size)
 
     def digest_callback(path: str, algoname: str, d: str) -> None:
         if algoname == "md5":
             digests[path] = d
 
-    for entry in entries:
-        etag = entry.digest
-        assert etag.algorithm is DigestType.md5
-        download_gens[str(entry)] = _download_file(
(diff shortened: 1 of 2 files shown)


dandi/download.py
```diff
@@ -988,7 +995,7 @@ class DownloadProgress:
 @dataclass
 class ProgressCombiner:
     zarr_size: int
-    file_qty: int
+    file_qty: int = -1  # set to specific known value whenever full sweep is complete
```
jwodder, line 998:
```suggestion
    file_qty: int | None = None  # set to specific known value whenever full sweep is complete
```

dandi/download.py
```diff
@@ -1021,18 +1028,25 @@ def get_done(self) -> dict:
...
-            "done%": total_downloaded / self.maxsize * 100,
+            "done%": total_downloaded / self.zarr_size * 100 if self.zarr_size else 0,
         }
 
     def set_status(self, statusdict: dict) -> None:
         state_qtys = Counter(s.state for s in self.files.values())
         total = len(self.files)
         if (
-            total == self.file_qty
+            self.file_qty >= 0  # if already known
```
jwodder, line 1048:
```suggestion
            self.file_qty is not None  # if already known
```

dandi/download.py
```diff
@@ -878,33 +878,40 @@ def _download_zarr(
...
+                    digest_callback=partial(digest_callback, str(entry)),
+                ),
+            )
+        pc.file_qty = len(entries)
 
-    pc = ProgressCombiner(zarr_size=asset.size, file_qty=len(download_gens))
     final_out: dict | None = None
     with interleave(
-        [pairing(p, gen) for p, gen in download_gens.items()],
+        downloads_gen(),
```
jwodder, line 914: `interleave()` evaluates/consumes the iterable in its entirety before returning, so you haven't accomplished anything by making `downloads_gen()` into a generator.

In order to add downloads to `interleave()` as they're yielded by `asset.iterfiles()`, you'll need to:

* Replace the call to `interleave()` with a directly-constructed `Interleaver` instance
* In separate thread, call the `Interleaver`'s `submit()` method whenever a new file is yielded by `asset.iterfiles()`, and then call its `finalize()` method at end of iteration.

[See `interleave`'s docs for more information.](https://github.com/jwodder/interleave?tab=readme-ov-file#api)

jwodder (changes requested):

dandi/download.py
yarikoptic: would you be so kind to approach this or otherwise address this issue the way you like it while also ideally reducing that delay / not demanding full listing to arrive to start the download?

yarikoptic: or may be we could merge this (as resolves the issue) and then improve on top of it?

jwodder: Getting the error handling right for the solution I suggested would be very difficult, so we should probably just merge the current version for now.

yarikoptic: ok, let's proceed!  Do you want me to create a dedicated issue or you keep this on your plate?

jwodder: Create a dedicated issue.

kabilar:
Thank you @yarikoptic and @jwodder.

PR #1445, opened 2024-05-23
Separate time components in logfile names with periods

jwodder:
The logfile names are currently too hard to read; the names need to be broken up more.

### DIFF
--- a/dandi/cli/command.py
+++ b/dandi/cli/command.py
@@ -99,7 +99,7 @@ def main(ctx, log_level, pdb=False):
 
     logdir = platformdirs.user_log_dir("dandi-cli", "dandi")
     logfile = os.path.join(
-        logdir, f"{datetime.now(timezone.utc):%Y%m%d%H%M%SZ}-{os.getpid()}.log"
+        logdir, f"{datetime.now(timezone.utc):%Y.%m.%d.%H.%M.%SZ}-{os.getpid()}.log"
     )
     os.makedirs(logdir, exist_ok=True)
     handler = logging.FileHandler(logfile, encoding="utf-8")


dandi/cli/command.py
```diff
@@ -99,7 +99,7 @@ def main(ctx, log_level, pdb=False):
 
     logdir = platformdirs.user_log_dir("dandi-cli", "dandi")
     logfile = os.path.join(
-        logdir, f"{datetime.now(timezone.utc):%Y%m%d%H%M%SZ}-{os.getpid()}.log"
+        logdir, f"{datetime.now(timezone.utc):%Y.%m.%d.%H.%M.%SZ}-{os.getpid()}.log"
```
yarikoptic, line 102: so instead of 
```
20240503020100Z-1853999.log
```
it would become
```
2024.05.03.02.01.00Z-1853999.log
```

which also doesn't help in separating the date from time - IMHO the most important one for readability. What about middle ground of just
```suggestion
        logdir, f"{datetime.now(timezone.utc):%Y%m%d.%H%M%SZ}-{os.getpid()}.log"
```
ie.

```
20240503.020100Z-1853999.log
```

jwodder: @yarikoptic I think just separating the individual components is the most important part for readability.  How about keeping the component separators but changing the date-time separator to a hyphen, e.g., `2024.05.03-02.01.00Z-1853999.log`?

yarikoptic: looks better!  one more possible iteration:  in duct (attn @asmacdo ) ATM we take iso8601 but replacing `:` with `-` as well to avoid `:` in filename (but also lacking `Z` for timezone depiction)... so it then could be `2024-05-03T02-01-00Z-1853999.log`. Frankly using `.` makes it cleaner. So may be let's go with this last iteration and consider doing the same in `duct` @asmacdo

jwodder:
@yarikoptic Can this be merged now?

PR #1449, opened 2024-06-05
Script to get bibliography for (all) dandisets (prototype)

yarikoptic:
a follow up to @bendichter  's 
- https://github.com/dandi/dandi-archive/issues/1555

I think this is a viable prototype with some TODOs (provided in the script) for us to collate/update an ultimate DANDI bibliography (as bibtex).  WDYT?  

Also includes the `dandi.bib` which script produced.

Related: 
- https://github.com/dandi/dandi-archive/issues/1709#issuecomment-1924227635 -- use Zenodo's approach and have DOI per dandiset pointing to most recent published (or draft until then).  Meanwhile I just populate per dandi for the most recent version

### DIFF
--- a/sandbox/dandi.bib
+++ b/sandbox/dandi.bib
@@ -1,0 +1,4058 @@
+# DANDISET 000003
+@misc{dandi.000003/0.210812.1448,
+  doi = {10.48324/DANDI.000003/0.210812.1448},
+  url = {https://dandiarchive.org/dandiset/000003/0.210812.1448},
+  author = {Senzai, Yuta and Fernandez-Ruiz, Antonio and Buzsáki, György},
+  keywords = {cell types, current source density, laminar recordings, oscillations, mossy cells, granule cells, optogenetics},
+  title = {Physiological Properties and Behavioral Correlates of Hippocampal Granule Cells and Mossy Cells},
+  publisher = {DANDI Archive},
+  year = {2021}
+}
+
+
+@misc{dandi.000003/0.230629.1955,
+  doi = {10.48324/DANDI.000003/0.230629.1955},
+  url = {https://dandiarchive.org/dandiset/000003/0.230629.1955},
+  author = {Senzai, Yuta and Fernandez-Ruiz, Antonio and Buzsáki, György},
+  keywords = {cell types, current source density, laminar recordings, oscillations, mossy cells, granule cells, optogenetics},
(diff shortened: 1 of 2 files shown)


sandbox/dandi.bib
```diff
@@ -0,0 +1,4058 @@
...
+  author = {Halchenko, Yaroslav O.},
+  keywords = {development},
+  title = {Test dataset for testing dandi-cli.},
+  publisher = {DANDI Archive},
+  year = {2021}
+}
+
+
+# DANDISET 000029
+<!DOCTYPE html>
```
yarikoptic, line 403: Ho ho -- we might be identifying problems for dandisets where we have releases and DOI minting has failed! someone might need to look into that!

yarikoptic: filed 
- https://github.com/dandi/dandi-archive/issues/1953

Meanwhile will adjust script to skip those

sandbox/dandi.bib
```diff
@@ -0,0 +1,4058 @@
...
+}
+
+
+# DANDISET 000404
+@misc{dandi.000404/0.230605.2024,
+  doi = {10.48324/DANDI.000404/0.230605.2024},
+  url = {https://dandiarchive.org/dandiset/000404/0.230605.2024},
+  author = {Athalye, Vivek R and Khanna, Preeya and Gowda, Suraj and Orsborn, Amy L and Costa, Rui M and Carmena, Jose M},
+  keywords = {neural population dynamics, motor cortex, motor control, brain-machine interface, neuroprosthetics, optimal feedback control, motor commands, movement representations, dynamical systems},
+  title = {Monkey 2D cursor BMI},
```
yarikoptic, line 3562: apparently `data.crosscite.org` is not doing that great of a job here -- for BibTeX it ideally should have made it into

```suggestion
  title = {Monkey {2D} cursor {BMI}},
```

to prevent LaTeX lower casing them... now we just need to figure out where is that code to file a complaint :-/

yarikoptic: the easiest of cause IIRC is just to double `{}` it ie
 ```suggestion
  title = {{Monkey 2D cursor BMI}},
```
and in principle we could do that ourselves here after someone confirms that my concern here is correct 

attn @TheChymera as who might know and @bendichter as who might care ;-)

PR #1454, opened 2024-06-14
Normalize path while requesting list of assets from the server

### LINKED ISSUE #1452: RemoteDandiset.download_directory doesn't appear to work?
Running `download_directory` in my notebook which demonstrates downloading from Dandi, appears not to work, 
described in this issue
https://github.com/AllenInstitute/openscope_databook/issues/398

I've attempted to replicate it and receive the same results. The function call does not yield an error, but simply seems to not actually run any download. I'm using dandi==0.61.2

yarikoptic:
Closes #1452

See individual commits for more description/rationale.

Related:
- https://github.com/AllenInstitute/openscope_databook/pull/399

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -937,6 +937,29 @@ class RemoteDandiset:
             client=client, identifier=data["identifier"], version=version, data=data
         )
 
+    @staticmethod
+    def _normalize_path(path: str) -> str:
+        """
+        Helper to normalize path before passing it to the server.
+
+        We and API call it "path" but it is really a "prefix" with inherent
+        semantics of containing directory divider '/' and emphasizing with
+        trailing '/' to point to a directory.
+        """
+        # Server (now) expects path to be a proper prefix, so to account for user
+        # possibly specifying ./ or some other relative paths etc, let's normalize
+        # the path.
+        # Ref: https://github.com/dandi/dandi-cli/issues/1452
+        path_normed = os.path.normpath(path)
+        if path_normed == ".":
+            path_normed = ""
+        elif path.endswith("/"):
+            # we need to make sure that we have a trailing slash if we had it before
+            path_normed += "/"
+        if path_normed != path:
+            lgr.debug("Normalized path %r to %r", path, path_normed)
+        return path_normed
+
     def json_dict(self) -> dict[str, Any]:
         """
         Convert to a JSONable `dict`, omitting the ``client`` attribute and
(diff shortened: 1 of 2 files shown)


dandi/dandiapi.py
```diff
@@ -1154,6 +1177,8 @@ def get_assets_with_path_prefix(
         Returns an iterator of all assets in this version of the Dandiset whose
         `~RemoteAsset.path` attributes start with ``path``
 
+        ``path`` is normalized first to possibly remove leading `./` or relative
+        paths (e.g., `../`) within it.
```
jwodder, line 1181: Add a blank line after this line to force a paragraph break.

Also, this paragraph should probably also be added to the docs for `get_asset_by_path()` and `download_directory()`, since those pass their path arguments to this method.

jwodder, line 1181:
```suggestion
        ``path`` is normalized first to possibly remove leading ``./`` or relative
        paths (e.g., ``../``) within it.
```

dandi/dandiapi.py
```diff
@@ -937,6 +937,29 @@ def from_data(cls, client: DandiAPIClient, data: dict[str, Any]) -> RemoteDandis
             client=client, identifier=data["identifier"], version=version, data=data
         )
 
+    @staticmethod
+    def _normalize_path(path: str) -> str:
+        """
+        Helper to normalize path before passing it to the server.
+
+        We and API call it "path" but it is really a "prefix" with inherent
+        semantics of containing directory divider '/' and emphasizing with
```
jwodder, line 946: I don't think "emphasizing" is the word you want here.  Maybe "and referring to a directory when terminated with '/'."

dandi/dandiapi.py
```diff
@@ -937,6 +937,29 @@ def from_data(cls, client: DandiAPIClient, data: dict[str, Any]) -> RemoteDandis
...
+        trailing '/' to point to a directory.
+        """
+        # Server (now) expects path to be a proper prefix, so to account for user
+        # possibly specifying ./ or some other relative paths etc, let's normalize
+        # the path.
+        # Ref: https://github.com/dandi/dandi-cli/issues/1452
+        path_normed = os.path.normpath(path)
```
jwodder, line 953: Use `posixpath` instead of `os.path` here so that forward slashes will remain forward slashes when running on Windows.

jwodder (changes requested):

dandi/dandiapi.py
yarikoptic: ok, will add use to `get_asset_by_path` as well

jwodder (approved):

yarikoptic:
ha -- I have missed that my change to `test_upload` (to actually use `./` in download_directory) which I accidentally committed to test locally broke tests!  and rightfully so since normalizing the path lead to side effect in that function... not happy about my solution now but let's proceed by fixing up `download_directory` for now.

yarikoptic:
windows -- flaky

```
________________________ test_IteratorWithAggregation _________________________
dandi\support\tests\test_iterators.py:72: in test_IteratorWithAggregation
    assert got in ([], [0])
E   assert [0, 1] in ([], [0])

---------- coverage: platform win32, python 3.10.11-final-0 ----------
Coverage XML written to file coverage.xml

```

PR #1458, opened 2024-07-12
Temporarily restrict dandischema requirement to `< 0.10.2`

jwodder:
Per @jjnesbitt's advice [here](https://github.com/dandi/dandi-archive/issues/1975#issuecomment-2223977835)

Note that I have added the "release" tag to this PR in order to get this restriction out to users soon.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,7 +33,9 @@ install_requires =
     bidsschematools ~= 0.7.0
     click >= 7.1
     click-didyoumean
-    dandischema >= 0.9.0, < 0.11
+    # Don't allow v0.10.2 of the schema until
+    # <https://github.com/dandi/dandi-archive/issues/1975> is resolved:
+    dandischema >= 0.9.0, < 0.10.2
     etelemetry >= 0.2.2
     fasteners
     fscacher >= 0.3.0


jjnesbitt (approved):

yarikoptic:
Thank you @jwodder !  

We really should see if we could somehow reduce such close tie to metadata library in terms of versioning.  Hopefully having model expressed as linkml would provide us a remedy so that we could just use corresponding version of the model (like now could be done with jsonschema serialization) supported/requested by the server.  Right @candleindark ?

candleindark:
> Thank you @jwodder !
> 
> We really should see if we could somehow reduce such close tie to metadata library in terms of versioning. Hopefully having model expressed as linkml would provide us a remedy so that we could just use corresponding version of the model (like now could be done with jsonschema serialization) supported/requested by the server. Right @candleindark ?

I am not sure LinkML can solve the particular problem we are facing. The current problem is that the GUI is not accepting the new generated JSON schema generated from Pydantic models. If we have the models expressed in LinkML, the JSON schema will be generated from these models expressed in LinkML instead. You will still have to test the JSON schema against the GUI.

yarikoptic:
I meant more of being able to test against specific version of schema. We already have jsonschema dumps in https://github.com/dandi/schema/tree/master/releases so in principle, anything which relies on jsonschema alone and is fully compatible with it -- should be able to validate even against versions it didn't see yet before.  But with pydantic we cannot really load/use different versions of schema, only "current one", hence rely on "most recent one" and upgrades into it.

candleindark:
> I meant more of being able to test against specific version of schema. We already have jsonschema dumps in https://github.com/dandi/schema/tree/master/releases so in principle, anything which relies on jsonschema alone and is fully compatible with it -- should be able to validate even against versions it didn't see yet before. But with pydantic we cannot really load/use different versions of schema, only "current one", hence rely on "most recent one" and upgrades into it.

Yes, in principle that should work. In the backend, assuming you are still using Pydantic models, generated from LinkML, you will need to load different versions of these models from different versions of LinkML schema.

PR #1478, opened 2024-08-08
Authenticate user upon the move command for an embargoed Dandiset

kabilar:
I received the error below when trying to rename an asset for an embargoed Dandiset.

Command
```
dandi move sub-mr261FZ/mr261FZ.tiff sub-mr261FZ/sub-mr261.tiff
```

Error
```
2024-08-08 17:08:52,143 [   ERROR] Error 401 while sending GET request to https://api.dandiarchive.org/api/dandisets/001089/: {"detail":"Authentication credentials were not provided."}
2024-08-08 17:08:52,144 [    INFO] Logs saved in /Users/kabilar/Library/Logs/dandi-cli/2024.08.08-22.08.51Z-13496.log
Error: Error 401 while sending GET request to https://api.dandiarchive.org/api/dandisets/001089/: {"detail":"Authentication credentials were not provided."}
```

We encountered the same error on the LINC project and @aaronkanzer provided a fix which I have submitted in this pull request.

- Original LINC issue: https://github.com/lincbrain/linc-cli/issues/41
- Original LINC fix: https://github.com/lincbrain/linc-cli/pull/42

I wasn't able to test these changes to the dandi-cli locally due to some versioneer issue.  So please double check this work.  Thank you.

PR #1480, opened 2024-08-09
move: Immediately authenticate client after creation

jwodder:
In order to deal with embargoed Dandisets

Closes #1478.

### DIFF
--- a/dandi/move.py
+++ b/dandi/move.py
@@ -812,6 +812,7 @@ def move(
                 raise TypeError("`dandiset` must be a Path when work_on='both'")
             local_ds, subpath = find_dandiset_and_subpath(dandiset)
             client = DandiAPIClient.for_dandi_instance(dandi_instance)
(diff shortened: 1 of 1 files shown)


jwodder:
@yarikoptic Ready for review.

dandi/move.py
```diff
@@ -812,6 +812,7 @@ def move(
                 raise TypeError("`dandiset` must be a Path when work_on='both'")
             local_ds, subpath = find_dandiset_and_subpath(dandiset)
             client = DandiAPIClient.for_dandi_instance(dandi_instance)
+            client.dandi_authenticate()
```
yarikoptic, line 815: in principle we do not need to authenticate if doing `dry_run` do we?

Would it be feasible in general to provide logic to authenticate when receiving 401 and haven't been authenticated before?

yarikoptic:
Thank you, let's proceed!

PR #1486, opened 2024-08-14
Dump audit table to file specified by `DANDI_TESTS_AUDIT_CSV` envvar

### LINKED ISSUE #1484: Add an option to dump content of `Audit` table from Docker env upon completion of testing
Now that we have `audit` functionality merged into master of dandi-archive, it would be great to 
- do some stress testing of it
- inspect what kind of records it records, and either with desired properties etc

Could we hackup quickly the following :

- If `DANDI_TESTS_AUDIT_JSON` env var is set:
  - Do not just use docker image, install `master` of dandi-archive into it (I think we do smth similar already)
  - Before cleaning up the env in the fixture taking care about docker compose, connect to postgres DB and dump `Audit` table as a list of json records into the file pointed to by `DANDI_TESTS_AUDIT_JSON` env var

jwodder:
Closes #1484.

### DIFF
--- a/dandi/tests/fixtures.py
+++ b/dandi/tests/fixtures.py
@@ -488,6 +488,27 @@ def docker_compose_setup() -> Iterator[dict[str, str]]:
                 raise RuntimeError("Django container did not start up in time")
         yield {"django_api_key": django_api_key}
     finally:
+        if auditfile := os.environ.get("DANDI_TESTS_AUDIT_CSV", ""):
+            with open(auditfile, "wb") as fp:
+                run(
+                    [
+                        "docker",
+                        "compose",
+                        "exec",
+                        "postgres",
+                        "psql",
+                        "-U",
+                        "postgres",
+                        "-d",
+                        "django",
+                        "--csv",
+                        "-c",
+                        "SELECT * FROM api_auditrecord;",
+                    ],
+                    stdout=fp,
+                    cwd=str(LOCAL_DOCKER_DIR),
+                    check=True,
+                )
         if persist in (None, "0"):
             run(
                 ["docker", "compose", "down", "-v"],
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -110,7 +110,17 @@ jobs:
     - name: Run Dandi API tests only
       if: matrix.mode == 'dandi-api'
       run: |
+        export DANDI_TESTS_AUDIT_CSV=/tmp/audit.csv
         python -m pytest -s -v --cov=dandi --cov-report=xml --dandi-api dandi
+        if [ ! -e /tmp/audit.csv ]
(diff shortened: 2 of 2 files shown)


yarikoptic (changes requested):
This is awesome, thank you!

That added code ATM not covered on CI. So we do not miss when code starts to rot (e.g. changes to dandi-archive etc), please add setting this env var to some tmp file in a single CI run, and verifying that file is created and has some representative number of records/lines.

note: I am trying to run tests locally with `DANDI_TESTS_AUDIT_CSV=/tmp/audit.csv python -m pytest -v dandi/tests/test_dandiapi.py` but they lead to 100% CPU utilization by celery although nothing alarming in that container's logs, very slow progress through tests, and some tests even failing... odd, likely unrelated, but may be worth checking run times of CI before / after this monday when I believe audit was merged...

dandi/tests/fixtures.py
```diff
@@ -488,6 +488,27 @@ def docker_compose_setup() -> Iterator[dict[str, str]]:
                 raise RuntimeError("Django container did not start up in time")
         yield {"django_api_key": django_api_key}
     finally:
+        if auditfile := os.environ.get("DANDI_TESTS_AUDIT_CSV", ""):
+            with open(auditfile, "wb") as fp:
+                run(
+                    [
```
yarikoptic, line 494: codecov still reports this block not covered by tests -- any idea why?

jwodder: I think that's just due to the annotations on GitHub not being updated.  If I go to [the Codecov site](https://app.codecov.io/github/dandi/dandi-cli/commit/9abc18ff93caa0e6d006ab3137ee1256c1eef40e/blob/dandi/tests/fixtures.py), the lines are marked covered.

yarikoptic:
3.10 failed with unrelated (just making note)

```
=========================== short test summary info ============================
ERROR dandi/tests/test_fixtures.py::test_organized_nwb_dir2 - pytest.PytestUnhandledThreadExceptionWarning: Exception in thread ExecutorManagerThread

  File "/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/site-packages/joblib/externals/loky/process_executor.py", line 602, in run
    self.process_result_item(result_item)
  File "/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/site-packages/joblib/externals/loky/process_executor.py", line 752, in process_result_item
    warnings.warn(
UserWarning: A worker stopped while some jobs were given to the executor. This can be caused by a too short worker timeout or by a memory leak.
```

PR #1487, opened 2024-08-14
Fix host used in minio URLs

jwodder:
Using `localhost` was causing test failures for me locally when fsspec was involved, likely due to `localhost` also resolving to `::1`, which the minio port was not exposed on.

### DIFF
--- a/dandi/tests/data/dandiarchive-docker/docker-compose.yml
+++ b/dandi/tests/data/dandiarchive-docker/docker-compose.yml
@@ -29,7 +29,7 @@ services:
       DJANGO_MINIO_STORAGE_ENDPOINT: minio:9000
       DJANGO_MINIO_STORAGE_SECRET_KEY: minioSecretKey
       DJANGO_STORAGE_BUCKET_NAME: django-storage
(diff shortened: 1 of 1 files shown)


dandi/tests/data/dandiarchive-docker/docker-compose.yml
```diff
@@ -29,7 +29,7 @@ services:
       DJANGO_MINIO_STORAGE_ENDPOINT: minio:9000
       DJANGO_MINIO_STORAGE_SECRET_KEY: minioSecretKey
       DJANGO_STORAGE_BUCKET_NAME: django-storage
-      DJANGO_MINIO_STORAGE_MEDIA_URL: http://localhost:9000/django-storage
+      DJANGO_MINIO_STORAGE_MEDIA_URL: http://127.0.0.1:9000/django-storage
```
yarikoptic, line 32: it is odd that it would need it for MINIO but nothing else... but ok, if that resolves it for you I do not mind, but please add a brief comment above on why it is needed. ATM not even commit message provides an explanation -- would be hard to figure out why it was needed and thus might get reverted etc.

jwodder: Comment added: 85f6dd8e

Issue #1489, opened 2024-08-19
1 out of 296 paths are not unique. We will try adding _obj- based on crc32 of object_id


nrsc:
Hello all,

Running into an issue while trying to organize my dandi set using the cli. Getting `[    INFO] 1 out of 296 paths are not unique. We will try adding _obj- based on crc32 of object_id`, and I can't seem to locate any more information about the non-unique path in question. I've been checking the logs but there are no further details on which path may be not unique. 

Build info as follows. 
```
[INFO    ] dandi 647577:124894109286400 dandi v0.63.0+5.g37b63509, hdmf v3.14.3, pynwb v2.8.1, h5py v3.11.0
```
I am using the `dandi v0.63.0+5.g37b63509` build to address metadata problem I was facing [previously](https://github.com/dandi/dandi-cli/issues/1481), but this issue comes up whether I am using the +5.g37...09 build or not. I have added more files to the dataset, so I assume that somewhere along the way I added a file that trips the `not unique` issue. Unfortunately I do not know where to look to identify where the issue stems from, as the "non-unique path" is not written into the log.

Another error that shows up: `Error: 'numpy.bytes_' object has no attribute 'encode'`

I will attach the log as well.
[2024.08.19-16.50.59Z-647577.log](https://github.com/user-attachments/files/16663583/2024.08.19-16.50.59Z-647577.log)

Other details regarding build information
<details>/
2024-08-19 09:55:08,798 [    INFO] Loading metadata from 296 files
[Parallel(n_jobs=-1)]: Using backend LokyBackend with 20 concurrent workers.
/home/nrsc/.local/share/r-miniconda/envs/dandi/lib/python3.12/site-packages/hdmf/utils.py:668: UserWarning: Ignoring cached namespace 'hdmf-common' version 1.1.3 because version 1.8.0 is already loaded.
  return func(args[0], **pargs)

/home/nrsc/.local/share/r-miniconda/envs/dandi/lib/python3.12/site-packages/hdmf/utils.py:668: UserWarning: Ignoring cached namespace 'core' version 2.2.4 because version 2.7.0 is already loaded.
  return func(args[0], **pargs)

/home/nrsc/.local/share/r-miniconda/envs/dandi/lib/python3.12/site-packages/hdmf/utils.py:668: UserWarning: Ignoring cached namespace 'hdmf-common' version 1.1.3 because version 1.8.0 is already loaded.
  return func(args[0], **pargs)

[...]

PR #1490, opened 2024-08-20
ENH: organize - show files with conflicts requiring adding _obj-

yarikoptic:
Logging check is currently not really functioning since IIRC we do enable logging into a file at higher level than INFO. So we might need to tune decision making here

Should provide UX side to partially address:
- https://github.com/dandi/dandi-cli/issues/1489

example:

```
❯ dandi organize --dandiset-path /tmp/outds /home/yoh/proj/dandi/dandisets/000004/sub-P53CS/*nwb /home/yoh/proj/dandi/dandisets/000004/sub-P54CS/*nwb
...
2024-08-19 20:03:17,448 [    INFO] Symlink support autodetected; setting files_mode='symlink'
2024-08-19 20:03:17,452 [    INFO] 2 out of 5 paths are not unique.
 2 paths 'compete' for the path 'sub-P53CS/sub-P53CS_ses-20171101_ecephys+image.nwb':
  /home/yoh/proj/dandi/dandisets/000004/sub-P53CS/sub-P53CS_ses-20171101_obj-108aqix_ecephys+image.nwb
  /home/yoh/proj/dandi/dandisets/000004/sub-P53CS/sub-P53CS_ses-20171101_obj-lj04dr_ecephys+image.nwb
 3 paths 'compete' for the path 'sub-P54CS/sub-P54CS_ses-20180101_ecephys+image.nwb':
  /home/yoh/proj/dandi/dandisets/000004/sub-P54CS/sub-P54CS_ses-20180101_obj-1ru98u8_ecephys+image.nwb
  /home/yoh/proj/dandi/dandisets/000004/sub-P54CS/sub-P54CS_ses-20180101_obj-1ukzk61_ecephys+image.nwb
  /home/yoh/proj/dandi/dandisets/000004/sub-P54CS/sub-P54CS_ses-20180101_obj-1vx1a5w_ecephys+image.nwb
 We will try adding _obj- based on crc32 of object_id
```

TODOs
- [ ] check what leads to tests failing ATM
- [ ] extend test covering this code to have multiple groups requiring `_obj`

### DIFF
--- a/dandi/organize.py
+++ b/dandi/organize.py
@@ -308,7 +308,7 @@ def _assign_obj_id(metadata, non_unique):
     # Avoid heavy import by importing within function:
     from .pynwb_utils import get_object_id
 
-    msg = "%d out of %d paths are not unique." % (len(non_unique), len(metadata))
+    msg = f"{len(non_unique)} out of {len(metadata)} paths are not unique."
     non_unique_paths = sorted(non_unique)
 
     # provide more information to the user
@@ -329,7 +329,7 @@ def _assign_obj_id(metadata, non_unique):
         else:
             for ex_path in non_unique_paths[1:]:
                 msg += "\n " + get_msg(ex_path)
-    lgr.info(msg + "\n We will try adding _obj- based on crc32 of object_id")
+    lgr.info("%s\n We will try adding _obj- based on crc32 of object_id", msg)
     seen_obj_ids = {}  # obj_id: object_id
     seen_object_ids = {}  # object_id: path
     recent_nwb_msg = "NWB>=2.1.0 standard (supported by pynwb>=1.1.0)."
--- a/dandi/organize.py
+++ b/dandi/organize.py
@@ -9,7 +9,6 @@ from collections import Counter
 from collections.abc import Sequence
 from copy import deepcopy
 from enum import Enum
-import logging
 import os
 import os.path as op
 from pathlib import Path, PurePosixPath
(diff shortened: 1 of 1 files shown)


dandi/organize.py
```diff
@@ -307,9 +308,28 @@ def _assign_obj_id(metadata, non_unique):
...
+    def get_msg(path, indent="  "):
+        in_paths = non_unique[path]
+        return (
+            f"{len(in_paths)} paths 'compete' for the path {path!r}:"
+            + f"\n{indent}".join([""] + in_paths)
+        )
+
+    msg += "\n " + get_msg(non_unique_paths[0])
+    if len(non_unique) > 1:
+        if not lgr.isEnabledFor(logging.DEBUG):
```
yarikoptic, line 324: @jwodder 
this condition is not effective -- we seems always enabled and effectiveLevel is 0.  I guess due to having multiple backends etc. Is there a way to make it dependent on what user specifies in cli?

jwodder: I don't think this condition is a good idea in the first place; the content of log messages should not depend on whether another log messages would be shown.  If you really want to display some paths at INFO and the full paths at DEBUG, then always log two messages.

dandi/organize.py
```diff
@@ -307,9 +308,28 @@ def _assign_obj_id(metadata, non_unique):
     # Avoid heavy import by importing within function:
     from .pynwb_utils import get_object_id
 
-    msg = "%d out of %d paths are not unique" % (len(non_unique), len(metadata))
+    msg = "%d out of %d paths are not unique." % (len(non_unique), len(metadata))
```
jwodder, line 311: Use f-strings here.

dandi/organize.py
```diff
@@ -307,9 +308,28 @@ def _assign_obj_id(metadata, non_unique):
...
+    non_unique_paths = sorted(non_unique)
 
-    lgr.info(msg + ". We will try adding _obj- based on crc32 of object_id")
+    # provide more information to the user
+    def get_msg(path, indent="  "):
```
jwodder, line 315: `get_msg()` never seems to be called with a custom `indent`, so why make it a parameter?

dandi/organize.py
```diff
@@ -307,9 +308,28 @@ def _assign_obj_id(metadata, non_unique):
...
+    if len(non_unique) > 1:
+        if not lgr.isEnabledFor(logging.DEBUG):
+            msg += (
+                " Rerun with logging at DEBUG level '-l debug' "
+                "to see {len(non_unique) - 1} more cases."
+            )
+        else:
+            for ex_path in non_unique_paths[1:]:
+                msg += "\n " + get_msg(ex_path)
+    lgr.info(msg + "\n We will try adding _obj- based on crc32 of object_id")
```
jwodder, line 332:
```suggestion
    lgr.info("%s\n We will try adding _obj- based on crc32 of object_id", msg)
```

This will prevent crashes if a nonunique path contains `%`.

jwodder (changes requested):

dandi/organize.py
yarikoptic: yeah, should be consistent.  
```suggestion
    msg = f"{len(non_unique)} out of {len(metadata)} paths are not unique."
```

dandi/organize.py
yarikoptic: yeah, in initial version I had different indents, will change

dandi/organize.py
yarikoptic: I think we figured that without `args` provided, it would not try to do interpolation, but I agree that this would be a safer pattern, I will apply.

PR #1496, opened 2024-09-05
Disallow hdmf 3.14.4, make organize to not parallelize for a single file, log information about all exceptions while reading metadata for organize

### LINKED ISSUE #1494: Test suite failure on HDMF 3.14.4
Found this in the NWB Inspector

We use the following sequence in the actions: https://github.com/NeurodataWithoutBorders/nwbinspector/blob/a7e7172d1e1cabcba703c0c1ada343ab0864f79a/.github/workflows/dandi-dev.yml#L26-L35, usually minus line 34

But without line 34 (forcible pin to previous HDMF version) this results in an error in `dandi/tests/test_organize.py:test_ambiguous`: https://github.com/NeurodataWithoutBorders/nwbinspector/actions/runs/10712486753/job/29702975698

Unsure what the true source of the problem is though, and if its a problem to be concerned about in general, just wanted to let y'all know

yarikoptic:
Individual commits have more information in their description.

Overall it 
- Fixes #1494 
- Improves logging of problems in organize while loading metadata

I have  bundled all the fixes etc so we could return back our CI to green (happen there is more to fix) and release!

### DIFF
--- a/dandi/organize.py
+++ b/dandi/organize.py
@@ -859,7 +859,7 @@ def organize(
             return meta
 
         if (
-            not devel_debug and jobs != 1
+            not devel_debug and jobs != 1 and not len(paths) == 1
         ):  # Do not use joblib at all if number_of_jobs=1
             # Note: It is Python (pynwb) intensive, not IO, so ATM there is little
             # to no benefit from Parallel without using multiproc!  But that would
--- a/dandi/organize.py
+++ b/dandi/organize.py
@@ -13,6 +13,7 @@ import os
 import os.path as op
 from pathlib import Path, PurePosixPath
 import re
+import traceback
 import uuid
 
 import ruamel.yaml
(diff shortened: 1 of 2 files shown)


dandi/organize.py
```diff
@@ -841,44 +842,59 @@ def act(func, *args, **kwargs):
...
-                # pbar.desc = "Files (%d failed)" % len(failed)
-                lgr.debug("Failed to get metadata for %s: %s", path, exc)
+            except Exception as e:
+                exc = (
+                    e.__class__,
+                    str(e),
+                    traceback.TracebackException.from_exception(e),
+                )
+                # meta = {}
+                # lgr.debug("Failed to get metadata for %s: %s", path, exc)
```
asmacdo, line 860:
```suggestion
```

dandi/organize.py
```diff
@@ -841,44 +842,59 @@ def act(func, *args, **kwargs):
...
+                if not e:
+                    continue
+                lgr.debug(
+                    "Loading metadata for path %s resulted in following exception:\n%s",
+                    m["path"],
+                    "\n".join(e[-1].format()),
+                )
 
-    metadata, skip_invalid = filter_invalid_metadata_rows(metadata)
+    metadata, skip_invalid = filter_invalid_metadata_rows([m for m, e in metadata_excs])
```
asmacdo, line 897:
```suggestion
    metadata, skip_invalid = filter_invalid_metadata_rows([m for m, _ in metadata_excs])
```

yarikoptic:
well, `dev-deps` would keep failing until HDMF is fixed up!  I think we are ready and should proceed.  I will leave it up until tomorrow for a possible review by @asmacdo or @candleindark .

yarikoptic:
well, 3.12 failed with

```
=========================== short test summary info ============================
FAILED dandi/tests/test_metadata.py::test_nwb2asset_remote_asset - Failed: Timeout >300.0s
```

since remote -- may be some connection? but it cannot relate to the changes here

PR #1497, opened 2024-09-09
Use "cache_spec:False" in our copy_nwb_file helper while exporting NWB copy

yarikoptic:
- [ ] Stopped since ideally we should make it dependent on either original file contains the spec.  Question posted

### DIFF
--- a/dandi/pynwb_utils.py
+++ b/dandi/pynwb_utils.py
@@ -524,7 +524,7 @@ def copy_nwb_file(src: str | Path, dest: str | Path) -> str:
             # do not export spec since
             **(
                 {"cache_spec": False}
-                if Version(pynwb.__version__) > Version("2.8.2.dev11")
+                if Version(pynwb.__version__) > Version("2.8.2")
                 else {}
             ),
         )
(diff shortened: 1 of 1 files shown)


dandi/pynwb_utils.py
```diff
@@ -518,7 +518,16 @@ def copy_nwb_file(src: str | Path, dest: str | Path) -> str:
...
         data = ior.read()
         data.generate_new_id()
-        iow.export(ior, nwbfile=data)
+        iow.export(
+            ior,
+            nwbfile=data,
+            # do not export spec since
+            **(
+                {"cache_spec": False}
+                if Version(pynwb.__version__) > Version("2.8.2.dev11")
```
rly, line 527: pynwb 2.8.2 has been released now with this feature, so you could just do 
```suggestion
                if Version(pynwb.__version__) > Version("2.8.2")
```

rly (approved):

yarikoptic:
filed
- https://github.com/dandi/dandi-cli/issues/1507
for the NFS fails which apparently do happen. Otherwise should be good, let's proceed

PR #1499, opened 2024-09-12
ENH: log dandischema version and ensure we log (consistently) path for log messages in download

yarikoptic:
hopefully would be in assist to troubleshoot issues with downloads, like the one now reported by @Kevancic on slack

### DIFF
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -685,7 +685,9 @@ def _download_file(
             if digester is not None:
                 break
         if digester is None:
-            lgr.warning("Found no digests in hashlib for any of %s", str(digests))
+            lgr.warning(
+                "%s - found no digests in hashlib for any of %s", path, str(digests)
+            )
 
     # TODO: how do we discover the total size????
     # TODO: do not do it in-place, but rather into some "hidden" file
@@ -718,7 +720,8 @@ def _download_file(
                             warned = True
                             # Yield ERROR?
                             lgr.warning(
-                                "Downloaded %d bytes although size was told to be just %d",
+                                "%s - downloaded %d bytes although size was told to be just %d",
+                                path,
                                 downloaded,
                                 size,
                             )
@@ -746,14 +749,15 @@ def _download_file(
                     *RETRY_STATUSES,
                 )
             ):
-                lgr.debug("Download failed: %s", exc)
+                lgr.debug("%s - download failed: %s", path, exc)
                 yield {"status": "error", "message": str(exc)}
                 return
             # if is_access_denied(exc) or attempt >= 2:
             #     raise
             # sleep a little and retry
             lgr.debug(
-                "Failed to download on attempt #%d: %s, will sleep a bit and retry",
+                "%s - failed to download on attempt #%d: %s, will sleep a bit and retry",
+                path,
                 attempt,
                 exc,
             )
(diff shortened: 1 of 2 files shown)


dandi/download.py
```diff
@@ -825,16 +834,22 @@ def __enter__(self) -> DownloadDirectory:
...
             if not chkpath.exists():
-                lgr.debug("Starting new download in new download directory")
+                lgr.debug(
+                    "%s - starting new download in new download directory", self.dirpath
+                )
             else:
                 lgr.debug(
-                    "Download directory found, but digests do not match; starting new download"
+                    "%s - download directory found, but digests do not match;"
+                    " starting new download",
```
asmacdo, line 851: Would digests.keys() and self. digests.keys() be useful to know here?

dandi/download.py
```diff
@@ -853,12 +868,33 @@ def __exit__(
...
+            )
         self.fp.close()
         try:
             if exc_type is None:
                 try:
                     self.writefile.replace(self.filepath)
                 except IsADirectoryError:
+                    import pdb
+
+                    pdb.set_trace()
```
asmacdo, line 893:
```suggestion
```

dandi/tests/test_download.py
```diff
@@ -1038,3 +1042,81 @@ def test_pyouthelper_time_remaining_1339():
...
+        assert dl.writefile.stat().st_size == 0
+        assert dl.offset == 0
+
+        dl.append(b"123")
+        assert dl.fp is not None
+        dl.fp.flush()  # appends are not flushed automatically
+        assert dl.writefile.stat().st_size == 3
+        assert dl.offset == 0  # doesn't change
+
+        dl.append(b"456")
```
asmacdo, line 1060: Does context manager exit automatically flush?

dandi/download.py
yarikoptic: dang, thanks!

dandi/tests/test_download.py
yarikoptic: it closes the file, so that should flush.  in part it is a "poor man" (since sizes are tiny in comparison) testing for odd behavior observed in 
- #1500

dandi/download.py
yarikoptic: I felt like I wondered the same and then dismissed as it would be "too much noise". but now not sure -- it might be... yet to decide

dandi/download.py
```diff
@@ -853,12 +869,35 @@ def __exit__(
         exc_tb: TracebackType | None,
     ) -> None:
         assert self.fp is not None
+        if exc_type is not None or exc_val is not None or exc_tb is not None:
```
candleindark, line 872: `exc_tb` is not used in the log here.

yarikoptic: first I had it but then kicked out from actual log -- the point is that if any of them provided, exception happened.

dandi/download.py
```diff
@@ -825,16 +835,22 @@
...
+                "%s - download directory exists and has matching checksum(s) %s; resuming download",
+                self.dirpath,
+                matching_algs,
             )
             self.fp = self.writefile.open("ab")
         else:
             # Delete the file (if it even exists) and start anew
             if not chkpath.exists():
-                lgr.debug("Starting new download in new download directory")
+                lgr.debug(
```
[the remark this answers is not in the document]
yarikoptic: in principle - you are right. Eg. if below `self.writefile.unlink()` failed for e.g. permissionerror, we would have never written chkpath... or even if somehow `self.writefile` worked out and then magically crashed right before dumping chkpath...   so in principle more precise would be to say that no prior checksum file found, did in 56690baffa4a0dc5e901ddffe71b6dd8d4abc35b

yarikoptic:
This grew too big and unwildly.  I will separate latter commits into a separate PR or PRs.

yarikoptic:
I shrunk to 1 commit, if passes, would merge

yarikoptic:
3.8 windows:
```
________________________ test_IteratorWithAggregation _________________________
dandi\support\tests\test_iterators.py:72: in test_IteratorWithAggregation
    assert got in ([], [0])
E   assert [0, 1] in ([], [0])
```

anyways -- did not spot related fails, let's proceed

Issue #1503, opened 2024-09-27
Python 3.8 tests failing 


asmacdo:
We are installing nwbinspector from source, and upcoming release 0.6.0 drops support for 3.8.

`ERROR: Package 'nwbinspector' requires a different Python: 3.8.18 not in '>=3.9'`

yarikoptic:
Just switch to use 3.9 in that CI run.

asmacdo:
@yarikoptic that CI run is the python 3.8 run.

PR #1504, opened 2024-09-27
Drop claiming support and testing against EOL Python 3.8

asmacdo:
Fixes https://github.com/dandi/dandi-cli/issues/1503

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -13,8 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         python:
-          - 3.8
-          #- 3.9
+          - 3.9
           #- 3.10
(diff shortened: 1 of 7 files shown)


yarikoptic:
not yet, in a month or so whenever it actually EOLs

asmacdo:
The 3.12 Test breaking is unrelated, its just a flake. 

I bumped the timeout [here](https://github.com/dandi/dandi-cli/pull/1505) fwiw.

asmacdo:
I thought it was EOL on October 14, but its gone as of October 7. https://devguide.python.org/versions/

yarikoptic:
Thank you! I am also going through similar fun activities in other corners

PR #1506, opened 2024-09-27
Switch to python 3.9 in dev-deps CI matrix run since dev deps are now beginning to drop Python 3.8 which is EOL soon

asmacdo:
Fixes https://github.com/dandi/dandi-cli/issues/1503


yarikoptic:
added a label for `auto` , and adjusted PR title so it is descriptive of the change whenever it gets into CHANGELOG.  dev-deps is green, let's proceed

PR #1508, opened 2024-09-27
Add ability to setup tmate debug session

yarikoptic:
Current desire/need is to troubleshoot
   https://github.com/dandi/dandi-cli/pull/1499
but moreover -- Windows.  Not yet even sure if would work for windows but still -- would be good to have generally

May be later we should parametrize more, not for every debug run


yarikoptic:
not sure what was up/whether was anything to finish.  When needed, someone could again work that out I guess

PR #1509, opened 2024-09-27
download: more consistent and exhaustive logging, new `DANDI_DEVEL_AGGRESSIVE_RETRY` mode, respect (?) Retry-After

yarikoptic:
It was originally developed as part of the PR but derailed into fixing/robustifying download functionality. So sits on top of 
- #1499

TODOs
- [x] make sure testing passes (added test revealed some odd behavior on Windows)
-  ~~possibly fold in here~~ fixup for download to do full file checksumming if download was resumed: https://github.com/dandi/dandi-cli/issues/1528

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -85,6 +85,10 @@ development command line options.
   function will patch `requests` to log the results of calls to
   `requests.utils.super_len()`
 
+- `DANDI_DOWNLOAD_AGGRESSIVE_RETRY` -- When set, would make `download()` retry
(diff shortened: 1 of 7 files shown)


yarikoptic:
FTR: rebased/force-pushed to get a fresh run of tests after recent changes and drop of 3.8 support.

yarikoptic:
ok
- windows:  filed https://github.com/pytest-dev/pytest/issues/12964 . Unfortunately I think we are doomed to skip it on windows for now to not cause that issue until it is resolved one way or another. 
- vcr -- delayed import and not used on 3.9

PR #1512, opened 2024-09-27
Add xfail marks for some tests: test_nwb2asset_remote_asset, test_reextract_metadata, test_IteratorWithAggregation

yarikoptic:
Rationale: we need to bring CI into more stable operation and then see how to address them properly

TODOs
- [ ] constraint test_nwb2asset_remote_asset xfail only for timeout, otherwise way too wide


yarikoptic:
Let's just proceed for now

PR #1513, opened 2024-10-04
Enhance pynwb validation parsing to report the path within the file, fix pynwb typo

rly:
re: https://github.com/con/validation/issues/1#issuecomment-2391975295

PyNWB reports a list of [HDMF validator Error objects](https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/validate/errors.py#L19).

- Enhance pynwb validator output to report the path within the file and separate the reason (message) from the rest of the error.
- Fix typo "pywnb" -> "pynwb"


rly:
Note that NWB inspector runs and parses the pynwb validator, so if you run that first, then running the pynwb validator is unnecessary.

PR #1514, opened 2024-10-18
Overhauling validation results to get them closer to cover different types of validators

yarikoptic:
Ref:
- https://github.com/con/validation/issues/1

and result of our chat with @candleindark while also reviewing results of linkml validation over dandisets:
- https://github.com/dandi/dandisets-linkml-status/blob/temp-with-error-locations/dandi-reports/summary.md

Notes/TODOs:
- BREAKING CHANGE: 
   - we replaced `ValidationOrigin.bids_version` to more generic `Origin.standard` + `Origin.standard_version`
   - We renamed `ValidationOrigin.name` to `Origin.validator` and `ValidationOrigin.version` to `Origin.validator_version`
   - `Severity.HINT`, `Severity.WARNING`, and `Severity.ERROR` now have value of `20`, `30`, and `40` respectively instead of `1`, `2`, and `3`.
   - Introduce `Origin.type` to indicate the type of the origin of a validation result.
   - We made the `ValidationResult` model into a Pydantic model so that it is readily to serialization to JSON.
- [x] going through the code base and harmonizing construction of those ValidationResults
- [ ] ~~redo #1176 in here~~

Please see individual commit messages for more details of the change.

**Note**:
This PR needs the changes in #1601 to pass the tests. Rebase/merge with master once #1601 is merged into master.

### DIFF
--- a/dandi/cli/tests/test_cmd_validate.py
+++ b/dandi/cli/tests/test_cmd_validate.py
@@ -6,7 +6,14 @@ import pytest
 
 from ..cmd_validate import _process_issues, validate
 from ...tests.fixtures import BIDS_ERROR_TESTDATA_SELECTION
-from ...validate_types import Scope, Severity, ValidationOrigin, ValidationResult
+from ...validate_types import (
+    Origin,
+    OriginType,
+    Scope,
+    Severity,
+    ValidationResult,
+    Validator,
+)
 
 
 @pytest.mark.parametrize("dataset", BIDS_ERROR_TESTDATA_SELECTION)
@@ -71,9 +78,10 @@ def test_process_issues(capsys):
     issues = [
         ValidationResult(
             id="NWBI.check_data_orientation",
-            origin=ValidationOrigin(
-                name="nwbinspector",
-                version="",
+            origin=Origin(
+                type=OriginType.VALIDATION,
+                validator=Validator.nwbinspector,
+                validator_version="",
             ),
             scope=Scope.FILE,
             message="Data may be in the wrong orientation.",
@@ -82,9 +90,10 @@ def test_process_issues(capsys):
         ),
         ValidationResult(
             id="NWBI.check_data_orientation",
-            origin=ValidationOrigin(
-                name="nwbinspector",
-                version="",
+            origin=Origin(
+                type=OriginType.VALIDATION,
+                validator=Validator.nwbinspector,
+                validator_version="",
             ),
             scope=Scope.FILE,
             message="Data may be in the wrong orientation.",
@@ -93,9 +102,10 @@ def test_process_issues(capsys):
         ),
         ValidationResult(
             id="NWBI.check_missing_unit",
-            origin=ValidationOrigin(
-                name="nwbinspector",
-                version="",
+            origin=Origin(
+                type=OriginType.VALIDATION,
+                validator=Validator.nwbinspector,
+                validator_version="",
             ),
             scope=Scope.FILE,
             message="Missing text for attribute 'unit'.",
--- a/dandi/cli/tests/test_cmd_validate.py
+++ b/dandi/cli/tests/test_cmd_validate.py
@@ -75,14 +75,16 @@ def test_validate_nwb_path_grouping(organized_nwb_dir4: Path) -> None:
 
 
 def test_process_issues(capsys):
+    origin_validation_nwbinspector = Origin(
+        type=OriginType.VALIDATION,
+        validator=Validator.nwbinspector,
+        validator_version="",
+    )
+
     issues = [
         ValidationResult(
             id="NWBI.check_data_orientation",
-            origin=Origin(
-                type=OriginType.VALIDATION,
-                validator=Validator.nwbinspector,
-                validator_version="",
-            ),
+            origin=origin_validation_nwbinspector,
             scope=Scope.FILE,
             message="Data may be in the wrong orientation.",
             path=Path("dir0/sub-mouse004/sub-mouse004.nwb"),
@@ -90,11 +92,7 @@ def test_process_issues(capsys):
         ),
         ValidationResult(
             id="NWBI.check_data_orientation",
-            origin=Origin(
-                type=OriginType.VALIDATION,
-                validator=Validator.nwbinspector,
-                validator_version="",
-            ),
+            origin=origin_validation_nwbinspector,
             scope=Scope.FILE,
             message="Data may be in the wrong orientation.",
             path=Path("dir1/sub-mouse001/sub-mouse001.nwb"),
@@ -102,11 +100,7 @@ def test_process_issues(capsys):
         ),
         ValidationResult(
             id="NWBI.check_missing_unit",
-            origin=Origin(
-                type=OriginType.VALIDATION,
-                validator=Validator.nwbinspector,
-                validator_version="",
-            ),
+            origin=origin_validation_nwbinspector,
             scope=Scope.FILE,
             message="Missing text for attribute 'unit'.",
             path=Path("dir1/sub-mouse001/sub-mouse001.nwb"),
--- a/dandi/files/bases.py
+++ b/dandi/files/bases.py
@@ -29,6 +29,8 @@ from dandi.metadata.core import get_default_metadata
 from dandi.misctypes import DUMMY_DANDI_ETAG, Digest, LocalReadableFile, P
 from dandi.utils import post_upload_size_check, pre_upload_size_check, yaml_load
 from dandi.validate_types import (
+    ORIGIN_INTERNAL_DANDI,
+    ORIGIN_VALIDATION_DANDI,
     Origin,
     OriginType,
     Scope,
@@ -210,11 +212,7 @@ class LocalAsset(DandiFile):
             )
             return [
                 ValidationResult(
-                    origin=Origin(
-                        type=OriginType.INTERNAL,
-                        validator=Validator.dandi,
-                        validator_version=dandi.__version__,
-                    ),
+                    origin=ORIGIN_INTERNAL_DANDI,
                     severity=Severity.ERROR,
                     id="dandi.SOFTWARE_ERROR",
                     scope=Scope.FILE,
@@ -537,7 +535,7 @@ class NWBAsset(LocalFileAsset):
         else:
             # make sure that we have some basic metadata fields we require
             try:
-                origin = Origin(
+                origin_validation_nwbinspector = Origin(
                     type=OriginType.VALIDATION,
                     validator=Validator.nwbinspector,
                     validator_version=str(_get_nwb_inspector_version()),
@@ -561,7 +559,7 @@ class NWBAsset(LocalFileAsset):
                         }
                     errors.append(
                         ValidationResult(
-                            origin=origin,
+                            origin=origin_validation_nwbinspector,
                             severity=severity,
                             id=f"NWBI.{error.check_function_name}",
                             scope=Scope.FILE,
@@ -707,11 +705,7 @@ def _check_required_fields(
             message = f"Required field {f!r} has no value"
             errors.append(
                 ValidationResult(
-                    origin=Origin(
-                        type=OriginType.VALIDATION,
-                        validator=Validator.dandi,
-                        validator_version=dandi.__version__,
-                    ),
+                    origin=ORIGIN_VALIDATION_DANDI,
                     severity=Severity.ERROR,
                     id="dandischema.requred_field",
                     scope=Scope.FILE,
@@ -723,11 +717,7 @@ def _check_required_fields(
             message = f"Required field {f!r} has value {v!r}"
             errors.append(
                 ValidationResult(
-                    origin=Origin(
-                        type=OriginType.VALIDATION,
-                        validator=Validator.dandi,
-                        validator_version=dandi.__version__,
-                    ),
+                    origin=ORIGIN_VALIDATION_DANDI,
                     severity=Severity.WARNING,
                     id="dandischema.placeholder_value",
                     scope=Scope.FILE,
--- a/dandi/files/bases.py
+++ b/dandi/files/bases.py
@@ -14,6 +14,7 @@ from typing import IO, Any, Generic
 from xml.etree.ElementTree import fromstring
 
 import dandischema
+from dandischema.consts import DANDI_SCHEMA_VERSION
 from dandischema.digests.dandietag import DandiETag
 from dandischema.models import BareAsset, CommonModel
 from dandischema.models import Dandiset as DandisetMeta
@@ -27,7 +28,15 @@ from dandi.dandiapi import RemoteAsset, RemoteDandiset, RESTFullAPIClient
 from dandi.metadata.core import get_default_metadata
 from dandi.misctypes import DUMMY_DANDI_ETAG, Digest, LocalReadableFile, P
 from dandi.utils import post_upload_size_check, pre_upload_size_check, yaml_load
-from dandi.validate_types import Scope, Severity, ValidationOrigin, ValidationResult
+from dandi.validate_types import (
+    Origin,
+    OriginType,
+    Scope,
+    Severity,
+    Standard,
+    ValidationResult,
+    Validator,
+)
 
 lgr = dandi.get_logger()
 
@@ -201,13 +210,15 @@ class LocalAsset(DandiFile):
             )
             return [
                 ValidationResult(
-                    origin=ValidationOrigin(
-                        name="dandi",
-                        version=dandi.__version__,
+                    origin=Origin(
+                        type=OriginType.INTERNAL,
+                        validator=Validator.dandi,
+                        validator_version=dandi.__version__,
                     ),
                     severity=Severity.ERROR,
                     id="dandi.SOFTWARE_ERROR",
                     scope=Scope.FILE,
+                    origin_result=e,
                     # metadata=metadata,
                     path=self.filepath,  # note that it is not relative .path
                     message=f"Failed to read metadata: {e}",
@@ -526,9 +537,10 @@ class NWBAsset(LocalFileAsset):
         else:
             # make sure that we have some basic metadata fields we require
             try:
-                origin = ValidationOrigin(
-                    name="nwbinspector",
-                    version=str(_get_nwb_inspector_version()),
+                origin = Origin(
+                    type=OriginType.VALIDATION,
+                    validator=Validator.nwbinspector,
+                    validator_version=str(_get_nwb_inspector_version()),
                 )
 
                 for error in inspect_nwbfile(
@@ -553,6 +565,7 @@ class NWBAsset(LocalFileAsset):
                             severity=severity,
                             id=f"NWBI.{error.check_function_name}",
                             scope=Scope.FILE,
+                            origin_result=error,
                             path=Path(error.file_path),
                             message=error.message,
                             # Assuming multiple sessions per multiple subjects,
@@ -694,9 +707,10 @@ def _check_required_fields(
             message = f"Required field {f!r} has no value"
             errors.append(
                 ValidationResult(
-                    origin=ValidationOrigin(
-                        name="dandischema",
-                        version=dandischema.__version__,  # type: ignore[attr-defined]
+                    origin=Origin(
+                        type=OriginType.VALIDATION,
+                        validator=Validator.dandi,
+                        validator_version=dandi.__version__,
                     ),
                     severity=Severity.ERROR,
                     id="dandischema.requred_field",
@@ -709,9 +723,10 @@ def _check_required_fields(
             message = f"Required field {f!r} has value {v!r}"
             errors.append(
                 ValidationResult(
-                    origin=ValidationOrigin(
-                        name="dandischema",
-                        version=dandischema.__version__,  # type: ignore[attr-defined]
+                    origin=Origin(
+                        type=OriginType.VALIDATION,
+                        validator=Validator.dandi,
+                        validator_version=dandi.__version__,
                     ),
                     severity=Severity.WARNING,
                     id="dandischema.placeholder_value",
@@ -793,13 +808,17 @@ def _pydantic_errors_to_validation_results(
             message = e.get("message", e.get("msg", None))
         out.append(
             ValidationResult(
-                origin=ValidationOrigin(
-                    name="dandischema",
-                    version=dandischema.__version__,  # type: ignore[attr-defined]
+                origin=Origin(
+                    type=OriginType.VALIDATION,
+                    validator=Validator.dandischema,
+                    validator_version=dandischema.__version__,  # type: ignore[attr-defined]
+                    standard=Standard.DANDI_SCHEMA,
+                    standard_version=DANDI_SCHEMA_VERSION,
                 ),
                 severity=Severity.ERROR,
                 id=id,
                 scope=scope,
+                origin_result=e,
                 path=file_path,
                 message=message,
                 # TODO? dataset_path=dataset_path,
--- a/dandi/files/zarr.py
+++ b/dandi/files/zarr.py
@@ -17,7 +17,7 @@ from dandischema.models import BareAsset, DigestType
 import requests
 from zarr_checksum.tree import ZarrChecksumTree
 
-from dandi import __version__, get_logger
+from dandi import get_logger
 from dandi.consts import (
     MAX_ZARR_DEPTH,
     ZARR_DELETE_BATCH_SIZE,
@@ -43,6 +43,7 @@ from dandi.utils import (
 
 from .bases import LocalDirectoryAsset
 from ..validate_types import (
+    ORIGIN_VALIDATION_DANDI_ZARR,
     Origin,
     OriginType,
     Scope,
@@ -269,20 +270,12 @@ class ZarrAsset(LocalDirectoryAsset[LocalZarrEntry]):
             )
             data = None
 
-        origin_internal_zarr = Origin(
-            type=OriginType.VALIDATION,
-            validator=Validator.dandi_zarr,
-            validator_version=__version__,
-        )
-        if data is not None:
-            origin_internal_zarr.standard_version = get_zarr_format_version(data)
-
         if isinstance(data, zarr.Group) and not data:
             errors.append(
                 ValidationResult(
-                    origin=origin_internal_zarr,
+                    origin=ORIGIN_VALIDATION_DANDI_ZARR,
                     severity=Severity.ERROR,
-                    id="zarr.empty_group",
+                    id="dandi_zarr.empty_group",
                     scope=Scope.FILE,
                     path=self.filepath,
                     message="Zarr group is empty.",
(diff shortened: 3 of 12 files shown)


dandi/validate_types.py
```diff
@@ -1,35 +1,53 @@
...
+    name: str  # Validator name
+    version: str  # Validator version
+    standard: str | None = (
+        None  # Standard being validated against # TODO: Enum for the standards??
+    )
+    standard_version: str | None = None  # Version of the standard
 
 
-class Severity(Enum):
+# TODO: decide on the naming consistency -- either prepend all with Validation or not
```
candleindark, line 19: @yarikoptic Are you talking about naming of the high level entities in this module, such as `Severity` and `ValidationOrigin`, or are you talking about the naming of the enum values within `Severity`?

dandi/validate_types.py
```diff
@@ -1,35 +1,53 @@
...
+class Severity(IntEnum):
     HINT = 1
-    WARNING = 2
-    ERROR = 3
+    INFO = 2  # new/unused, available in linkml
+    WARNING = 3
+    ERROR = 4
+    CRITICAL = 5  # new/unused, linkml has FATAL
```
candleindark, line 25: I see three issues with this though they are small issues.

1. We may include all the severity levels in the outputs of all validators, However, for a particular validator, only a subset of these may be used.
2. The same severity level may mean two different things in different validators.
3. Different validator may also use different severity level name to mean the same level.

dandi/validate_types.py
```diff
@@ -1,35 +1,53 @@
...
 
 
 class Scope(Enum):
     FILE = "file"
     FOLDER = "folder"
+    # Isaac: make it/add "dandiset-metadata" to signal specific relation to metadata
```
candleindark, line 31: Need clarification for this.

dandi/validate_types.py
```diff
@@ -1,35 +1,53 @@
...
 
 @dataclass
 class ValidationOrigin:
-    name: str
-    version: str
-    bids_version: str | None = None
+    name: str  # Validator name
+    version: str  # Validator version
+    standard: str | None = (
+        None  # Standard being validated against # TODO: Enum for the standards??
```
candleindark, line 14: @yarikoptic We can create an enum class to define the supported standards. Do you know what those those standards are ATM?

candleindark:
In updating the use of `ValidationOrigin`, it is not clear to me what should be provided for the `standard` and `standard_version` attributes of many `ValidationOrigin` objects. For example, this is not clear in `ValidationOrigin` objects in `organized.py`.

https://github.com/dandi/dandi-cli/blob/45db66c34cd6ded26fd1d8c871a7957df73ad38a/dandi/organize.py#L1122

Do we have a dandi standard? If we do, it doesn't seem that we version this standard.

I think we should indeed define an [enum class](https://github.com/dandi/dandi-cli/pull/1514/files#r1967035125) for the supported standard, and also talk about how to retrieve the version of those standards.

dandi/files/zarr.py
```diff
@@ -209,17 +209,20 @@ def get_validation_errors(
         import zarr
 
         errors: list[ValidationResult] = []
+        origin: ValidationOrigin = ValidationOrigin(
+            name="zarr",
+            version=zarr.__version__,
```
candleindark, line 214: Is this line intentional? The original one was `version=zarr.version.version`.

dandi/validate_types.py
yarikoptic: 1. agree
2. better not
3. better not

But you are pointing to the ambiguous (not clearly defined across validators) semantic . Then ideally we should be able to establish "mapping" if there is difference in agreement.

I think we should just clear up here what **we** mean for each of the levels. E.g.

- `HINT`: data is correct but could be improved
- `INFO`: ... ???
- `WARNING`: when "SHOULD" or "SHOULD NOT" level of requirement is violated
- `ERROR`: when "MUST" or "MUST NOT" level of requirement is violated
- `CRITICAL`: when makes given data completely unusable

may be we might want

- `INTERNAL`: signals about the problem with validator itself. although @candleindark rightfuly notices that it is still an ERROR but about not data, but validator itself... so we might want to explicitly point to the object of the error (not data  but validator)

dandi/validate_types.py
yarikoptic: Sure. For starters:

- BIDS
- NWB
- HED
- OME-ZARR

Could be also at the level of the "file format" standard:

- JSON
- YAML
- TSV

dandi/validate_types.py
yarikoptic: may be forget about it for now -- seems to combine with `ValidationObject` below but we might need to rethink/define some additional levels

yarikoptic: as for the folder an example could be an unknown to BIDS folder name under `sub-.../ses-.../` folder.  I am not sure ATM if BIDS validator complains and how

dandi/files/zarr.py
```diff
@@ -209,17 +209,20 @@ def get_validation_errors(
...
+            standard="zarr",
```
yarikoptic, line 215: here to add `standard_version` on what version of zarr it is .

dandi/validate_types.py
yarikoptic: might want also to create Enum for validators since we also have a set of them.

candleindark: I added the standard `ZARR` as well since you have already started using it.

dandi/validate_types.py
candleindark: Here are a few more suggestions.

1. Let's define `INFO` as not an indication of problem but information of status or confirmation. However, this `INFO` maybe different than the `INFO` from `LinkML`, but that's perfectly fine (we will have a proper mapping).
2. Using the definition of `INFO` in the last point. `INFO` should be the lowest level of severity amount those listed.
3. Mimic the definition of [logging levels](https://docs.python.org/3/library/logging.html#logging-levels) in Python, have the values of the levels be multiple of 10 instead of 1 to allow greater flexibility (to allow addition of levels in the future or allow custom levels).
4. The above definitions of `WARNING` and `ERROR` are too vague. Together with `CRITICAL`, we can draw inspiration from https://docs.python.org/3/library/logging.html#logging-levels in defining them.

candleindark: Severity levels have now been rearranged and more clearly defined.

dandi/validate_types.py
```diff
@@ -1,35 +1,97 @@
...
+    JSON = auto()
+    TSV = auto()
+    YAML = auto()
+
+
+@unique
+class Validator(StrEnum):
+    """Validators that are used to do validation"""
+
+    bidsschematools = auto()
```
yarikoptic, line 33: now soon we would have `bids_validator_deno` or alike 

and we could provision also for `hed_python_validator`

dandi/validate_types.py
```diff
@@ -1,35 +1,97 @@
...
+    WARNING = 30
+    """Data is not recognized as correct. Changes are needed to ensure correctness"""
+
+    ERROR = 40
+    """Data is recognized as incorrect"""
+
+    CRITICAL = 50
+    """
+    A serious incorrectness in data.
+    E.g., an incorrectness that prevents validation of other aspects of the data
```
yarikoptic, line 68: may be an example could be an absent or misspecified `BIDSVersion` so becomes unclear against which bids version to validate, although typically we just validate against the latest.

candleindark:
> In updating the use of `ValidationOrigin`, it is not clear to me what should be provided for the `standard` and `standard_version` attributes of many `ValidationOrigin` objects. For example, this is not clear in `ValidationOrigin` objects in `organized.py`.
> 
> https://github.com/dandi/dandi-cli/blob/45db66c34cd6ded26fd1d8c871a7957df73ad38a/dandi/organize.py#L1122
> 
> Do we have a dandi standard? If we do, it doesn't seem that we version this standard.
> 
> I think we should indeed define an [enum class](https://github.com/dandi/dandi-cli/pull/1514/files#r1967035125) for the supported standard, and also talk about how to retrieve the version of those standards.

We now, with the latest commit, have a subclass of  `Enum`, `Standard`, to represent a list of standards. Some standard is `informal`, e.g., `Standard.DANDI_SCHEMA` and have no versioning. In the case, of a standard with no version, we will just let the `standard_version` attribute of `Origin` to default to `None`.

dandi/validate_types.py
```diff
@@ -1,35 +1,53 @@
...
 
 class Scope(Enum):
     FILE = "file"
     FOLDER = "folder"
+    # Isaac: make it/add "dandiset-metadata" to signal specific relation to metadata
     DANDISET = "dandiset"
     DATASET = "dataset"
 
 
+# new/unused, may be should be gone
```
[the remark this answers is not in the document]
candleindark: done

dandi/validate_types.py
candleindark: Example is added to comment to further demo the meaning of `CRITICAL`

dandi/validate_types.py
candleindark: Marking this as resolved since the new `Enum` subclasses have been reviewed in a meeting.

dandi/validate_types.py
candleindark: The new levels have been reviewed in a meeting. Marking resolved.

candleindark:
@yarikoptic I think a `ValidationResult` object should represent the result of how well a "data" instance is validated. It should inform area/aspect of invalidity.

https://github.com/dandi/dandi-cli/blob/386e50db53a8fc0236b666927402ec1bee3e7759/dandi/files/bases.py#L193-L217
The above snippet uses `ValidationResult` to represent an error other than a `ValidationError`. The captured error doesn't tell how valid a data instance is. It has nothing to do with the data being validated against but how our software is written. If that error is raised it is either a bug in the software, which should be propagated, or a user's mistake, which should be handled gracefully in another manner. The current arrangement is conflating error handling with the purpose of `ValidationResult`.

dandi/files/zarr.py
candleindark: They are essentially the same, so it doesn't matter.

dandi/files/zarr.py
```diff
@@ -209,32 +216,44 @@
         import zarr
 
         errors: list[ValidationResult] = []
+        origin: Origin = Origin(
+            validator=Validator.zarr,
+            validator_version=zarr.__version__,
+            standard=Standard.ZARR,
+            # TODO: standard_version=...,
```
candleindark, line 223: Per https://zarr.readthedocs.io/en/stable/user-guide/v3_migration.html#migration-guide, it seems that the Python `zarr` pacakge can support more than one Zarr format 3 specification at the same time. What is the way to determine which Zarr format specification is being used in just opening a file? My bet is that possibly all support format specifications are used, so the `standard_version` field may be irrelevant and should be skipped in this case.

dandi/files/zarr.py
```diff
@@ -209,32 +216,44 @@ def get_validation_errors(
...
                     message="Error opening file.",
                 )
             )
             data = None
+
+        origin = Origin(
+            validator=Validator.dandi_zarr,
+            validator_version=__version__,
+            standard=Standard.ZARR,
+            # TODO: standard_version=...,
```
candleindark, line 248: Per https://zarr.readthedocs.io/en/stable/user-guide/v3_migration.html#migration-guide, it seems that the Python zarr pacakge can support more than one Zarr format 3 specification at the same time. What is the way to determine which Zarr format specification is being used in just opening a file? My bet is that possibly all support format specifications are used, so the standard_version field may be irrelevant and should be skipped in this case.

dandi/files/zarr.py
```diff
@@ -42,11 +42,38 @@
...
+    Severity,
+    Standard,
+    ValidationResult,
+    Validator,
+)
 
 lgr = get_logger()
 
 
+def get_zarr_storage_specification_version(data: Any) -> str:
```
yarikoptic, line 57: Let's shorten

```suggestion
def get_zarr_format_version(data: Any) -> str:
```

dandi/pynwb_utils.py
```diff
@@ -365,13 +372,16 @@ def validate(path: str | Path, devel_debug: bool = False) -> list[ValidationResu
...
             errors.append(
                 ValidationResult(
-                    origin=ValidationOrigin(
-                        name="pynwb",
-                        version=pynwb.__version__,
+                    origin=Origin(
+                        validator=Validator.pynwb,
+                        validator_version=pynwb.__version__,
+                        standard=Standard.NWB,
+                        # TODO: standard_version=...,
```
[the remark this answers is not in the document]
yarikoptic: just use `get_nwb_version` from below -- just move it up here so you would already know then what nwb version is in that file.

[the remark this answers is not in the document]
yarikoptic: a good point that internally pynwb is actually likely to do migration and then validation against latest. But that is the internal logic which we do not know about unless validation error record would have target version of nwb.   So 
- please check if error record contains a version, if not
   - file an issue requesting NWB or the particular relevant namespace (e.g. from extension) to be added
- if not known from error -- the version of nwb file is the best one we know of

dandi/pynwb_utils.py
```diff
@@ -382,13 +392,15 @@ def validate(path: str | Path, devel_debug: bool = False) -> list[ValidationResu
             raise
         errors.append(
             ValidationResult(
-                origin=ValidationOrigin(
-                    name="pynwb",
-                    version=pynwb.__version__,
+                origin=Origin(
+                    validator=Validator.pynwb,
+                    validator_version=pynwb.__version__,
+                    standard=Standard.NWB,
```
yarikoptic, line 398: standard_version here too

dandi/files/zarr.py
```diff
@@ -209,32 +236,43 @@ def get_validation_errors(
...
                 ValidationResult(
-                    origin=ValidationOrigin(
-                        name="zarr",
-                        version=zarr.version.version,
-                    ),
+                    origin=origin,
                     severity=Severity.ERROR,
                     id="zarr.cannot_open",
                     scope=Scope.FILE,
+                    origin_result=e,
```
yarikoptic, line 257: we might want to adopt later the [CapturedException](https://github.com/datalad/datalad/blob/HEAD/datalad/support/exceptions.py#L25) to avoid dragging too much of context around from the original location of that exception

dandi/pynwb_utils.py
yarikoptic, line 398: per discussion, we might not even be able to claim it has anything to do with NWB:
```suggestion
```

dandi/validate_types.py
```diff
@@ -1,21 +1,83 @@
...
+    """The validator conducting the validation"""
+
+    validator_version: str
+    """The version of the validator"""
+
+    standard: Standard | None = None
+    """Standard being validated against"""
+
+    standard_version: str | None = None
+    """Version of the standard"""
```
yarikoptic, line 56: may be here it could be `source` which could be "standard" or "internal" ATM, but then may be smth else too

dandi/files/zarr.py
candleindark: done

dandi/validate_types.py
candleindark: Since this is a field in the `Origin` object. I think naming the field `source` is a bit confusing.

Since we already use an `Origin` object to represent the "source" of a validation result, and, at this point, we want to indicate rather the "source" is an "external" validator invoked by the `dandi-cli` or some "internal" code of the `dandi-cli` itself, I think we can name the field in the `Origin` object `type` and allow the value of the field to be `internal`, `external`, or `undetermined`. The `undetermined` value is needed because there are cases where source of the error is hard/impossible to determine.

candleindark: I have decided on types of the origin to be 

```py
class Type(StrEnum):
    """Types of validation result origins"""

    INTERNAL = auto()
    """Validation result is originated from the code of this project"""

    VALIDATOR = auto()
    """Validation result is originated from a validator"""

    UNDETERMINED = auto()
    """
    Validation result is originated from a source that is not determined to
    be another value in this enum class
    
    This is usually used when the origin of the validation result is unclear
    """
```

I used `VALIDATOR` instead of your suggestion of `STANDARD` because not all validator implements a standard (nwbinspector, for example).

dandi/validate_types.py
```diff
@@ -1,21 +1,107 @@
...
+    nwbinspector = auto()
+    pynwb = auto()
+    zarr = auto()
+
+
+class Type(StrEnum):
+    """Types of validation result origins"""
+
+    INTERNAL = auto()
+    """Validation result is originated from the code of this project"""
```
yarikoptic, line 49:
```suggestion
    """Validation result is originated from the validator and not necessarily relating to the standard"""
```

dandi/files/zarr.py
```diff
@@ -209,32 +236,43 @@ def get_validation_errors(
...
         errors: list[ValidationResult] = []
+        origin: Origin = Origin(
+            validator=Validator.zarr,
+            validator_version=zarr.__version__,
+            standard=Standard.ZARR,
+            # TODO: standard_version=...,
+        )
+
         try:
             data = zarr.open(str(self.filepath))
```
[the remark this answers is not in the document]
candleindark: No, I couldn't find any, https://github.com/zarr-developers/zarr-python/blob/v2.18.2/zarr/errors.py.

candleindark:
FTR: The following look very problematic though not directly related to this PR. 

https://github.com/dandi/dandi-cli/blob/386e50db53a8fc0236b666927402ec1bee3e7759/dandi/pynwb_utils.py#L406-L408

I don't know its intend either.

dandi/pynwb_utils.py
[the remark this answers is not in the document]
candleindark: `pynwb.validate()` doesn't raise validation errors, it returns them. Errors that are raised are rightfully considered `OriginType.INTERNAL`, so both `Origin.standard` and `Origin.standard_version` are actually irrelevant in cases of those errors.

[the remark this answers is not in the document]
candleindark: Feature request, https://github.com/NeurodataWithoutBorders/pynwb/issues/2062, made.

setup.cfg
```diff
@@ -35,6 +35,8 @@ install_requires =
     click-didyoumean
     dandischema >= 0.11.0, < 0.12.0
     etelemetry >= 0.2.2
+    # For pydantic to be able to use type annotations like `X | None`
```
jwodder, line 38: Is there a reason you can't just use `Union` and `Optional` instead?

dandi/validate_types.py
```diff
@@ -25,11 +105,20 @@ class Scope(Enum):
...
 
 
-@dataclass
-class ValidationResult:
+class ValidationResult(BaseModel):
     id: str
-    origin: ValidationOrigin
+
+    origin: Origin
+    """Origin of the validation result as validator and standard used in producing it"""
```
jwodder, line 112:
```suggestion
    #: Origin of the validation result as validator and standard used in producing it
    origin: Origin
```

And likewise for other attributes.

dandi/files/zarr.py
```diff
@@ -42,11 +42,39 @@
...
+    Severity,
+    Standard,
+    ValidationResult,
+    Validator,
+)
 
 lgr = get_logger()
 
 
+def get_zarr_format_version(data: Any) -> str:
```
jwodder, line 58: What strings does this return?  That should be documented.

dandi/files/zarr.py
```diff
@@ -42,11 +42,39 @@
...
+    """
+    import zarr  # Delay heavy import
+
+    if isinstance(data, zarr.Group):
+        meta = json.loads(data.store.get(".zgroup"))
+    elif isinstance(data, zarr.Array):
+        meta = json.loads(data.store.get(".zarray"))
+    else:
+        raise TypeError("`data` must be a `zarr.core.Array` or `zarr.hierarchy.Group`")
+    return str(meta["zarr_format"])
```
jwodder, line 75: What type is `meta["zarr_format"]` before you cast it to a string?  If it's already a string and the point of the cast is just to satisfy the type checker, you should instead do:

```python
zf = meta["zarr_formar"]
assert isinstance(zf, str)
return zf
```

setup.cfg
candleindark: No, but the code base already had many uses of `|` to express `Optional` and `Union` types. I only made the models in `validate_types.py` to Pydantic models, and since Pydantic uses the types at runtime, the `eval_type_backport` package is needed. To avoid the `eval_type_backport` dependency, some code unrelated to the purpose of this PR will need to change.

dandi/validate_types.py
candleindark: Using a triple-quoted string in this context will gain IDE support and other documentation tool support.

dandi/files/zarr.py
candleindark: It is actually an `int`.

dandi/files/zarr.py
candleindark: It is the Zarr storage specification version used by the give Zarr data object. The possible versions are documented at https://zarr-specs.readthedocs.io/en/latest/specs.html.

I have added the documentation for the return value.

Thanks.

candleindark:
This PR needs the changes in https://github.com/dandi/dandi-cli/pull/1601 to pass the tests. It will be merged with master once https://github.com/dandi/dandi-cli/pull/1601 is merged into master.

candleindark:
The Change from #1601 is merged. This PR is ready for a final review.

dandi/organize.py
```diff
@@ -1151,7 +1169,12 @@ def validate_organized_path(
             errors.append(
                 ValidationResult(
                     id="DANDI.METADATA_MISMATCH_SUBJECT",
-                    origin=ValidationOrigin(name="dandi", version=__version__),
+                    origin=Origin(
+                        type=OriginType.VALIDATION,
+                        validator=Validator.dandi,
+                        validator_version=__version__,
+                        standard=Standard.DANDI_LAYOUT,
+                    ),
```
yarikoptic, line 1177: 3 "uses" for the same `origin` . Similarly to other similar cases -- just define `origin` a tthe top of the function and reuse here?  would make code shorter and thus easier to read.

But then -- this and other functions will be invoked in a loop across paths. So I wonder if we should centrally define some of those origins, like here `origin_dandi_layout` and alike? this would save some cycles/bytes across files. WDYT?

- we should at least define / reuse

dandi/files/zarr.py
```diff
@@ -209,32 +244,43 @@ def get_validation_errors(
...
                     message="Error opening file.",
                 )
             )
             data = None
+
+        origin = Origin(
+            type=OriginType.VALIDATION,
+            validator=Validator.dandi_zarr,
+            validator_version=__version__,
+        )
```
yarikoptic, line 276: here we seems to overwrite prior origin which was like `origin_internal_zarr` and this one `origin_dandi_zarr` or alike.  So might better give this somewhat more descriptive name and avoid overwrite so we do not by mistake assign the wrong one?

dandi/files/bases.py
```diff
@@ -526,9 +537,10 @@ def get_validation_errors(
         else:
             # make sure that we have some basic metadata fields we require
             try:
-                origin = ValidationOrigin(
-                    name="nwbinspector",
-                    version=str(_get_nwb_inspector_version()),
+                origin = Origin(
```
yarikoptic, line 540:
```suggestion
                origin_nwbinspector = Origin(
```

dandi/files/bases.py
```diff
@@ -201,13 +210,15 @@ def get_validation_errors(
             )
             return [
                 ValidationResult(
-                    origin=ValidationOrigin(
-                        name="dandi",
-                        version=dandi.__version__,
+                    origin=Origin(
```
yarikoptic, line 213: could be named/reused as origin_internal_dandi

dandi/cli/tests/test_cmd_validate.py
```diff
@@ -71,9 +78,10 @@ def test_process_issues(capsys):
     issues = [
         ValidationResult(
             id="NWBI.check_data_orientation",
-            origin=ValidationOrigin(
-                name="nwbinspector",
-                version="",
+            origin=Origin(
+                type=OriginType.VALIDATION,
+                validator=Validator.nwbinspector,
+                validator_version="",
```
yarikoptic, line 84: why no version for nwbinspector?    it seems to be available

```
>>> nwbinspector.__version__
'0.6.2'
```

dandi/files/bases.py
```diff
@@ -694,9 +707,10 @@ def _check_required_fields(
             message = f"Required field {f!r} has no value"
             errors.append(
                 ValidationResult(
-                    origin=ValidationOrigin(
-                        name="dandischema",
-                        version=dandischema.__version__,  # type: ignore[attr-defined]
+                    origin=Origin(
+                        type=OriginType.VALIDATION,
+                        validator=Validator.dandi,
+                        validator_version=dandi.__version__,
```
yarikoptic, line 713: could be `origin_validation_dandi`

dandi/cli/tests/test_cmd_validate.py
yarikoptic: and so -- this could  be `origin_validation_nwbinspector`

yarikoptic (commented):
a few spots where to define local `origin` to be reused.

But also I wonder if generally we should pre-define all those origins so they get reused across all the paths.  It smells a little as premature optimization to me though, so may be not worth it. But at least I think it is worth renaming them locally so they become a little more descriptive than just `origin` variables

dandi/cli/tests/test_cmd_validate.py
yarikoptic: ah, that was in the test, so ok to keep as is I guess

dandi/files/bases.py
candleindark: Renamed to `origin_validation_nwbinspector` to be consistent with other names.

dandi/files/bases.py
candleindark: Done with a constant

dandi/files/bases.py
candleindark: done with a defined constant

dandi/files/zarr.py
candleindark: Rename the first `origin` to `origin_internal_zarr`. The second `origin` is eliminated entirely because I think the subsequent `ValidationResult` objects are labeled incorrectly. Please see https://github.com/dandi/dandi-cli/pull/1514/commits/c09cab4e82947b94da391c757c012dd67e3ee759 for details.

candleindark:
This one is ready for a final review.

yarikoptic:
> FTR: The following look very problematic though not directly related to this PR.
> 
> https://github.com/dandi/dandi-cli/blob/386e50db53a8fc0236b666927402ec1bee3e7759/dandi/pynwb_utils.py#L406-L408
> 
> I don't know its intent either.

yeah, better to be reanalyzed/improved. I think we burnt there a few times already

dandi/pynwb_utils.py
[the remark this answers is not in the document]
yarikoptic: followed up on that con/validation issue 1, thanks

dandi/files/zarr.py
```diff
@@ -42,11 +42,47 @@
...
+        The Zarr data object from which to extract the storage specification version
+
+    Returns
+    -------
+    str
+        The Zarr storage specification version,
+        https://zarr-specs.readthedocs.io/en/latest/specs.html, used in the Zarr data
+        object
+    """
+    import zarr  # Delay heavy import
```
yarikoptic, line 75: hm, apparently we have no single test on validation of zarrs? that's odd :-/  

filed dedicated
- https://github.com/dandi/dandi-cli/issues/1605

for that

yarikoptic:
Thank you @candleindark , let's proceed with continuation in
- https://github.com/dandi/dandi-cli/issues/1605

PR #1524, opened 2024-11-13
fix: replace dandi-api link with dandi-archive

asmacdo:
dandi-api does forward to dandi-archive, but still nice to have correct


yarikoptic:
Thank you!

PR #1532, opened 2024-11-20
add one photon to metadata dict

bendichter:
OnePhotonSeries is a newer data type, which is why it was not included in the original list


yarikoptic:
thank you!

PR #1536, opened 2024-11-23
Do fail (raise Exception, CLI has non-0 exit) download if any of downloads fail

yarikoptic:
Sits on top of #1535 

- [x] ensure that CLI experience is ok and we do get non-0 exit
- [x] fixup tests

### DIFF
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -153,6 +153,13 @@ def download(
 
     gen_ = (r for dl in downloaders for r in dl.download_generator())
 
+    errors = []
+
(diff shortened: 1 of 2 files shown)


kabilar:
Thanks @yarikoptic.  I am not sure that I am the best person to review these changes since I am not too familiar with this codebase.  I will request a review from @jwodder.

yarikoptic:
John wouldn't be able to review until Wed or may be even later... to not delay -- let's just proceed, I am feeling it should be ok.

PR #1539, opened 2024-11-25
Ignore some type-checking errors under Python 3.13

jwodder:


yarikoptic:
could you please add a comment on why it is needed -- what are we overriding?

jwodder:
@yarikoptic Comment added.

PR #1543, opened 2024-11-27
Test dev-deps against latest supported by us version of Python (3.12 ATM)

yarikoptic:


yarikoptic:
relevant CI run is green

PR #1546, opened 2024-11-29
Add Meriones unguiculatus to species list

kabilar:
Reference: http://purl.obolibrary.org/obo/NCBITaxon_10047

### DIFF
--- a/dandi/metadata/util.py
+++ b/dandi/metadata/util.py
@@ -395,7 +395,7 @@ species_map = [
         "Macaca nemestrina",
     ),
     (
-        ["Mongolian gerbil", "Mongolian jird"],
+        ["mongolian gerbil", "mongolian jird"],
         None,
         NCBITAXON_URI_TEMPLATE.format("10047"),
         "Meriones unguiculatus",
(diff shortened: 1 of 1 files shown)


dandi/metadata/util.py
```diff
@@ -394,6 +394,12 @@ def extract_cellLine(metadata: dict) -> str | None:
         NCBITAXON_URI_TEMPLATE.format("9545"),
         "Macaca nemestrina",
     ),
+    (
+        ["Mongolian gerbil", "Mongolian jird"],
```
jwodder, line 398: These need to be lowercase.

jwodder (changes requested):

dandi/metadata/util.py
yarikoptic:
```suggestion
        ["mongolian gerbil", "mongolian jird"],
```

yarikoptic: ideally IMHO we should make code robust to specification here and just `.lower()` then when matching. Or do you see an issue with that @jwodder ?

kabilar:
Thank you, @yarikoptic @jwodder.  Would you be able to make a release this week so that I can notify the user who requested this change?

yarikoptic:
Sure.
I am looking into hopefully finalize
- https://github.com/dandi/dandi-cli/pull/1536

with which we could trigger release... if not - I will generate empty PR to trigger it

PR #1547, opened 2024-12-03
BF: do allow for user to specify a proper full name for "species" in metadata 

yarikoptic:
Addressing user report on Slack.

Apparently (thus checking with @satra) we were checking for URLs and only for "common names", not proper long names. It is surprising that the problem manifested only now then!


kabilar (approved):
Thank you, Yarik.

kabilar:
Looks like that there are a couple of tests failing that are related to these changes.

yarikoptic:
d'oh -- I should get into a habit of running type checks before pushing ;) should be resolved now.  But I still wonder how come we run into this just now???

kabilar (approved):

PR #1554, opened 2024-12-16
Reduce messaging to pyout on progress of downloads

yarikoptic:
In an attempt to help with
- #1549 

Also includes small tune up for summary over messages


dandi/download.py
```diff
@@ -358,13 +358,34 @@ def download_generator(self) -> Iterator[dict]:
...
+                    """
+                    prior_time = 0
+                    warned = False
+                    for rec in gen:
+                        current_time = time.time()
+                        if done_perc := rec.get("done%", 0):
+                            if isinstance(done_perc, (int, float)):
+                                if current_time - prior_time < 2 and done_perc != 100:
+                                    continue
+                            elif not warned:
```
jwodder, line 377: Don't you want to set `warned = True` here?

dandi/download.py
```diff
@@ -358,13 +358,34 @@ def download_generator(self) -> Iterator[dict]:
...
+                                lgr.warning("Received non numeric done%: %r", done_perc)
```
jwodder, line 378:
```suggestion
                                lgr.warning("Received non numeric done%%: %r", done_perc)
```

jwodder (changes requested):

dandi/download.py
yarikoptic: thanks! will do

jwodder (approved):

PR #1559, opened 2024-12-20
Improve retrying downloads code and testing

yarikoptic:
Follow up to
- #1558 

- retry on 429 as well
- handle Retry-After as a Date not just seconds (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Date)
- adjust logging of sleeping upon retry - do in a single place

We are into holidays already, so it is understood if review does not come. I will reserve the "right" to merge if there is user demand but likely we could just wait. If anyone sees "clearly" how to improve -- feel welcome to push directly.

### DIFF
--- a/dandi/consts.py
+++ b/dandi/consts.py
@@ -141,7 +141,7 @@ known_instances = {
         "linc-staging",
         "https://staging.lincbrain.org",
         "https://staging-api.lincbrain.org/api",
-    )
+    ),
 }
 # to map back url: name
 known_instances_rev = {
@@ -160,7 +160,7 @@ DRAFT = "draft"
 
 #: HTTP response status codes that should always be retried (until we run out
 #: of retries)
-RETRY_STATUSES = (500, 502, 503, 504)
+RETRY_STATUSES = (429, 500, 502, 503, 504)
 
 VIDEO_FILE_EXTENSIONS = [".mp4", ".avi", ".wmv", ".mov", ".flv", ".mkv"]
 VIDEO_FILE_MODULES = ["processing", "acquisition"]
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -1088,7 +1088,7 @@ def _check_if_more_attempts_allowed(
     downloaded_in_attempt: int,
 ) -> int | None:
     """Check if we should retry the download, return potentially adjusted 'attempts_allowed'"""
-    sleep_amount = random.random() * 5 * attempt
+    sleep_amount: float = -1.0
     if os.environ.get("DANDI_DOWNLOAD_AGGRESSIVE_RETRY"):
         # in such a case if we downloaded a little more --
         # consider it a successful attempt
@@ -1143,17 +1143,9 @@ def _check_if_more_attempts_allowed(
                 sleep_amount,
                 exc,
             )
-        else:
-            lgr.debug(
-                "%s - download failed on attempt #%d: %s, will sleep a bit and retry",
-                path,
-                attempt,
-                exc,
-            )
-    # if is_access_denied(exc) or attempt >= 2:
-    #     raise
-    # sleep a little and retry
-    else:
+    if sleep_amount < 0:
+        # it was not Retry-after set, so we come up with random duration to sleep
+        sleep_amount = random.random() * 5 * attempt
         lgr.debug(
             "%s - download failed on attempt #%d: %s, will sleep a bit and retry",
             path,
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -3,7 +3,8 @@ from __future__ import annotations
 from collections import Counter, deque
 from collections.abc import Callable, Iterable, Iterator, Sequence
 from dataclasses import InitVar, dataclass, field
-from datetime import datetime
+from datetime import UTC, datetime
+from email.utils import parsedate_to_datetime
 from enum import Enum
 from functools import partial
 import hashlib
@@ -1085,7 +1086,7 @@ def _check_if_more_attempts_allowed(
     exc: requests.RequestException,
     attempt: int,
     attempts_allowed: int,
-    downloaded_in_attempt: int,
+    downloaded_in_attempt: int = 0,
 ) -> int | None:
     """Check if we should retry the download, return potentially adjusted 'attempts_allowed'"""
     sleep_amount: float = -1.0
@@ -1122,19 +1123,36 @@ def _check_if_more_attempts_allowed(
             return None
         elif retry_after := exc.response.headers.get("Retry-After"):
             # playing safe
-            if not str(retry_after).isdigit():
-                # our code is wrong, do not crash but issue warning so
-                # we might get report/fix it up
-                lgr.warning(
-                    "%s - download failed due to response %d with non-integer"
-                    " Retry-After=%r: %s",
-                    path,
-                    exc.response.status_code,
-                    retry_after,
-                    exc,
-                )
-                return None
-            sleep_amount = int(retry_after)
+            retry_after = str(retry_after)
+            if retry_after.isdigit():
+                sleep_amount = int(retry_after)
+            else:
+                # else if it is a datestamp like "Wed, 21 Oct 2015 07:28:00 GMT"
+                # we could parse it and calculate how long to sleep
+                try:
+                    retry_after_date = parsedate_to_datetime(retry_after)
+                    current_date = datetime.now(UTC)
+                    difference = retry_after_date - current_date
+                    sleep_amount = int(difference.total_seconds())
+                    if sleep_amount < 0:
+                        lgr.warning(
+                            "%s - date in Retry-After=%r is in the past (current is %r)",
+                            path,
+                            retry_after,
+                            current_date,
+                        )
+                except ValueError as exc_ve:
+                    # our code or response is wrong, do not crash but issue warning
+                    # and continue with "default" sleep logic
+                    lgr.warning(
+                        "%s - download failed due to response %d with non-integer or future date"
+                        " Retry-After=%r: %s with %s upon converting assuming it is a date",
+                        path,
+                        exc.response.status_code,
+                        retry_after,
+                        exc,
+                        exc_ve,
+                    )
             lgr.debug(
                 "%s - download failed due to response %d with "
                 "Retry-After=%d: %s, will sleep and retry",
--- a/dandi/tests/test_download.py
+++ b/dandi/tests/test_download.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 from collections.abc import Callable
 from contextlib import nullcontext
+from email.utils import parsedate_to_datetime
+from functools import partial
 from glob import glob
 import json
 import logging
@@ -12,10 +14,13 @@ from pathlib import Path
 import re
 from shutil import rmtree
 import time
+from unittest import mock
 
 import numpy as np
 import pytest
 from pytest_mock import MockerFixture
+import requests
+from requests.exceptions import HTTPError
 import responses
 import zarr
 
@@ -32,6 +37,7 @@ from ..download import (
     PathType,
     ProgressCombiner,
     PYOUTHelper,
+    _check_if_more_attempts_allowed,
     download,
 )
 from ..exceptions import NotFoundError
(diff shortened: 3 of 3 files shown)


dandi/download.py
```diff
@@ -1143,17 +1161,9 @@ def _check_if_more_attempts_allowed(
...
-            lgr.debug(
-                "%s - download failed on attempt #%d: %s, will sleep a bit and retry",
-                path,
-                attempt,
-                exc,
-            )
-    # if is_access_denied(exc) or attempt >= 2:
-    #     raise
-    # sleep a little and retry
-    else:
```
jwodder, line 1156: By removing this `else`, the `lgr.debug()` call on line 1138/1156 will be followed by the call on line 1157/1167, resulting in two similar log messages.

jwodder (changes requested):

dandi/download.py
```diff
@@ -1122,19 +1123,36 @@ def _check_if_more_attempts_allowed(
...
-                    "%s - download failed due to response %d with non-integer"
-                    " Retry-After=%r: %s",
-                    path,
-                    exc.response.status_code,
-                    retry_after,
-                    exc,
-                )
-                return None
-            sleep_amount = int(retry_after)
+            retry_after = str(retry_after)
```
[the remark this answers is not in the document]
yarikoptic: I think it originally was but seems no longer needed... will remove, thanks!

dandi/download.py
yarikoptic: i think it would only be  in the case of non-conformant "retry-after" we would get 3 messages "on the matter"
- "to response %d with non-integer or future date"
- "to response %d with Retry-After"
- the one below.

if there is a good Retry-after, we would get a non-negative value for sleep_amount, and log only at 1138/1156 (but we might want it into `else:` for exception, to not log with /1147) as sleep_amount condition on /1164 would not trigger that other message where we would also define the random sleep.

But for above "3 msgs" I will adjust... thanks!

dandi/download.py
```diff
@@ -1122,19 +1123,36 @@ def _check_if_more_attempts_allowed(
...
+            if retry_after.isdigit():
+                sleep_amount = int(retry_after)
```
candleindark, line 1128: If we are assuming that this client always connect to our trusted DANDI API, I think this is fine, but if we want to want to play even "safer" I think we may want to ensure the `retry_after` string is really a non-negative integer in a decimal expression.

If  `retry_after` is `"²³"` will crash this client at this point. 

This safe guard can be excessive though since we have control of both the client and server.

I think use of `str.isdecimal()` and ensuring the return of `int()` to be non-negative is better.

dandi/download.py
```diff
@@ -1122,19 +1123,36 @@ def _check_if_more_attempts_allowed(
...
+                    # and continue with "default" sleep logic
+                    lgr.warning(
+                        "%s - download failed due to response %d with non-integer or future date"
+                        " Retry-After=%r: %s with %s upon converting assuming it is a date",
+                        path,
+                        exc.response.status_code,
+                        retry_after,
+                        exc,
+                        exc_ve,
+                    )
```
candleindark, line 1155:
```suggestion
                try:
                    retry_after_date = parsedate_to_datetime(retry_after)
                except (ValueError, TypeError) as exc_ve:
                    # our code or response is wrong, do not crash but issue warning
                    # and continue with "default" sleep logic
                    lgr.warning(
                        "%s - download failed due to response %d with non-integer or future date"
                        " Retry-After=%r: %s with %s upon converting assuming it is a date",
                        path,
                        exc.response.status_code,
                        retry_after,
                        exc,
                        exc_ve,
                    )
                else:
                    current_date = datetime.now(timezone.utc)
                    difference = retry_after_date - current_date
                    sleep_amount = int(difference.total_seconds())
                    if sleep_amount < 0:  # todo: `sleep_amount < 0` may not be an error. It the server set a `Retry-After` datetime extremely close and the network is slow, the client may receive the response after the `Retry-After` datetime.
                        lgr.warning(
                            "%s - date in Retry-After=%r is in the past (current is %r)",
                            path,
                            retry_after,
                            current_date,
                        )
```

This suggestion make the try clause 
1. more targeted
2. catches `TypeError`s from `parsedate_to_datetime`. (Try `parsedate_to_datetime('Wed, 21 O3t 2015 07:280 GMT')`)
3. adds a note with a todo comment regarding `sleep_amount < 0`

dandi/download.py
```diff
@@ -1122,19 +1123,36 @@ def _check_if_more_attempts_allowed(
...
             lgr.debug(
                 "%s - download failed due to response %d with "
                 "Retry-After=%d: %s, will sleep and retry",
```
candleindark, line 1158: the "Retry-After=[]" in this log is now insistent and `sleep_amount` may not be a meaningful number if `retry_after` failed to be parsed into a `datetime` object

dandi/download.py
```diff
@@ -1122,19 +1123,36 @@ def _check_if_more_attempts_allowed(
             return None
         elif retry_after := exc.response.headers.get("Retry-After"):
```
candleindark, line 1124: This is not about the changes in this PR. However, since are trying to improve this function, we may as well consider replace `elif` with `if` here. The last `if` clause ended with a `return` statement, so the `elif` here is about confusing.

candleindark: So can be said about the containing `elif` clause at line 1111 at which I can't place a comment.

dandi/download.py
```diff
@@ -1143,17 +1161,9 @@ def _check_if_more_attempts_allowed(
...
-                path,
-                attempt,
-                exc,
-            )
-    # if is_access_denied(exc) or attempt >= 2:
-    #     raise
-    # sleep a little and retry
-    else:
+    if sleep_amount < 0:
+        # it was not Retry-after set, so we come up with random duration to sleep
```
candleindark, line 1165: In the current arrangement, the execution flow can reach here event if `Retry-After` is set. It can be set to an extremely close datetime or some string can't be parsed properly in to a `datetime` object

candleindark (commented):
At this point, `_check_if_more_attempts_allowed` is equally, if not more, about sleeping/waiting before sending out another requests. I think it would be beneficial to rename it, or in some other way, to indicate this nature.

If we want to improve the retrying further, I think we can consider incorporating some well-known and recommended algorithms for retrying requests to a server while avoiding overwhelming it. https://chatgpt.com/share/676f8290-86cc-800a-9aa8-1ad952282d27.

dandi/download.py
yarikoptic: cool, will do `isdecimal`! thanks

dandi/download.py
yarikoptic: why is it confusing?  

for me `elif` is an indicator that the boolean condition could not be satisfied with prior one and is a part of the "chain" of conditions, unlike multiple `if` statements where handling of one could effect condition of the next one.

dandi/download.py
yarikoptic: Thanks!
- added TypeError
- moved unrelated to exception code into `else`
- in my code refactoring I added explicit handling of "near 0" (up to 2 sec) negative sleeps.  if more -- I would consider it to be unlikely a network error, and would issue warning and handle as no Retry-After was provided (helper returns `None`).

dandi/download.py
yarikoptic: sorry, I do not get what `[]` you have in mind since it can't be a list... please rereview when I push changes -- may be the issue is addressed.

dandi/download.py
yarikoptic: "extremely close" should not happen after refactoring.

dandi/download.py
candleindark: I think this is a matter of opinion. I used to prefer the way the code current is. However, I encountered this, https://docs.astral.sh/ruff/rules/superfluous-else-return/, and I started to feel that I prefer the other way. `elif`s of often make me feel that the body of the preceding `if` clause is not ending the entire `if` statement, and `elif` is needed to route the flow of the execution to the body in the `elif` clause based on the conditional expression in the `if` clause. Of course, the feeling is not always right, but I feel that replacing some of the `elif`s with `if` makes the code easier to read.

dandi/consts.py
```diff
@@ -160,7 +160,7 @@ def urls(self) -> Iterator[str]:
 
 #: HTTP response status codes that should always be retried (until we run out
 #: of retries)
-RETRY_STATUSES = (500, 502, 503, 504)
+RETRY_STATUSES = (429, 500, 502, 503, 504)
```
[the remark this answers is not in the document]
yarikoptic: I have centralized analysis of Retry-After and added that to other locations where RETRY_STATUSES is used

dandi/download.py
yarikoptic: ok, replaced with `if` since do not feel too strongly on that here.

dandi/download.py
```diff
@@ -1080,15 +1081,17 @@ def downloads_gen():
...
     attempts_allowed: int,
-    downloaded_in_attempt: int,
+    downloaded_in_attempt: int = 0,
 ) -> int | None:
-    """Check if we should retry the download, return potentially adjusted 'attempts_allowed'"""
-    sleep_amount = random.random() * 5 * attempt
+    """Check if we should retry the download, sleep if still allowed,
+    and return potentially adjusted 'attempts_allowed'
+    """
+    sleep_amount: float = -1.0
```
jwodder, line 1094:
```suggestion
    sleep_amount: float | None = None
```

dandi/download.py
```diff
@@ -1120,45 +1122,26 @@ def _check_if_more_attempts_allowed(
...
-            # playing safe
-            if not str(retry_after).isdigit():
-                # our code is wrong, do not crash but issue warning so
-                # we might get report/fix it up
-                lgr.warning(
-                    "%s - download failed due to response %d with non-integer"
-                    " Retry-After=%r: %s",
+        elif sleep_amount_ := get_retry_after(exc.response):
+            if sleep_amount_ is not None:  # could be 0
+                sleep_amount = sleep_amount_
```
jwodder, line 1127:
```suggestion
        elif (sleep_amount := get_retry_after(exc.response)) is not None:
```

dandi/download.py
```diff
@@ -1120,45 +1122,26 @@ def _check_if_more_attempts_allowed(
...
-                "%s - download failed on attempt #%d: %s, will sleep a bit and retry",
-                path,
-                attempt,
-                exc,
-            )
-    # if is_access_denied(exc) or attempt >= 2:
-    #     raise
-    # sleep a little and retry
-    else:
+    if sleep_amount < 0:
```
jwodder, line 1136:
```suggestion
    if sleep_amount is None:
```

dandi/utils.py
```diff
@@ -890,3 +891,66 @@ def joinurl(base: str, path: str) -> str:
...
+    (over a week), would return None.
+    """
+    if_unparsable = None
+    retry_after = response.headers.get("Retry-After")
+    if retry_after is None:
+        return None
+    sleep_amount: int | None
+    if retry_after.isdecimal():
+        sleep_amount = int(retry_after)
+    else:
```
jwodder, line 912:
```suggestion
    try:
        sleep_amount = int(retry_after)
    except ValueError:
```

dandi/utils.py
```diff
@@ -890,3 +891,66 @@ def joinurl(base: str, path: str) -> str:
...
+def get_retry_after(response: requests.Response) -> Optional[int]:
+    """If provided and parsed ok, returns duration in seconds to sleep before retry.
+
+    If not provided in the response header `Retry-After`, would
+    return None.
+    If parsing fails, or provided date/sleep does not make sense
+    since either too far in the past (over 2 seconds) or in the future
+    (over a week), would return None.
+    """
+    if_unparsable = None
```
jwodder, line 905: This is basically a constant alias.  What's the point of this assignment?  Just get rid of it and change `sleep_amount = if_unparsable` to `sleep_amount = None` (or even `return None`).

dandi/utils.py
```diff
@@ -890,3 +891,66 @@ def joinurl(base: str, path: str) -> str:
...
+                retry_after,
+                exc_ve,
+                if_unparsable,
+            )
+            sleep_amount = if_unparsable
+        else:
+            difference = retry_after_date - current_date
+            sleep_amount = int(difference.total_seconds())
+
+        if sleep_amount:
```
jwodder, line 933: Shouldn't this block be dedented so that integer Retry-Afters also get checked against a week?

dandi/tests/test_download.py
```diff
@@ -1169,3 +1175,91 @@ def test_DownloadDirectory_exc(
     assert dl.dirpath.exists()
     assert dl.fp is None
     assert dl.writefile.read_bytes() == b"456"
+
+
+def test__check_if_more_attempts_allowed():
```
jwodder, line 1180:
```suggestion
def test__check_attempts_and_sleep() -> None:
```

dandi/tests/test_download.py
```diff
@@ -1169,3 +1175,91 @@ def test_DownloadDirectory_exc(
...
+            assert (
+                f(HTTPError(), attempt=2, attempts_allowed=2, downloaded_in_attempt=1)
+                == 3
+            )
+            mock_sleep.assert_called_once()
+            assert mock_sleep.call_args.args[0] >= 0
+
+
+@pytest.mark.parametrize("status_code", [429, 503])
+def test__check_if_more_attempts_allowed_retries(status_code):
```
jwodder, line 1231:
```suggestion
def test__check_attempts_and_sleep_retries(status_code: int) -> None:
```

dandi/tests/test_download.py
```diff
@@ -1169,3 +1175,91 @@ def test_DownloadDirectory_exc(
...
+        # shifted by 2 minutes
+        mock_datetime.datetime.now.return_value = parsedate_to_datetime(
+            "Wed, 21 Oct 2015 07:26:00 GMT"
+        )
+        assert f(HTTPError(response=response), attempt=1, attempts_allowed=2) == 2
+        mock_sleep.assert_called_once()
+        assert mock_sleep.call_args.args[0] == 120
+
+    # we would still sleep some time if Retry-After is not decypherable
+    response.headers["Retry-After"] = "nondecypherable"
```
jwodder, line 1256:
```suggestion
    response.headers["Retry-After"] = "indecipherable"
```

dandi/utils.py
yarikoptic: any benefit?  I could potentially see how it would be more robust but not really "pragmatically"

by doing in `else:` we would avoid potential "while handling exception ... another exception" spaghetti of exceptions.

dandi/utils.py
yarikoptic: yes! will do

dandi/utils.py
yarikoptic: FTR: initially I had it as an option, then saw that I never used it, so moved "into body". Ok, will change as recommended to simplify

dandi/utils.py
jwodder: The purpose of the code is to parse `retry_after` as an integer if possible, and you can do that just with `try: sleep_amount = int(retry_after); except ...`.  Doing a "look before you leap" test with `isdecimal()` wastes CPU cycles and doesn't even exactly align with the set of strings accepted by `int()` (e.g., the latter accepts leading whitespace and negative numbers).

It's just better, more Pythonic code.

yarikoptic:
rerunning windows which puked with

```
FAILED dandi/tests/test_dandiarchive.py::test_parse_api_url_glob[https://gui.dandiarchive.org/#/dandiset/001001/draft/files?location=sub-RAT123/*.nwb-parsed_url1] - requests.exceptions.ConnectionError: ('Connection aborted.', ConnectionResetError(10054, 'An existing connection was forcibly closed by the remote host', None, 10054, None))
```

dandi/download.py
candleindark: I actually think this log statement should be removed entirely. Because of the containing if-statement, this log statement doesn't always get executed for all responses that have a `Retry-After` header. However, at the same time, I think the original intent is to execute this log statement for all responses with a `Retry-After` header. 

We are better off if we just replace the entire if statement:

```python
        if (sleep_amount := get_retry_after(exc.response)) is not None:
            lgr.debug(
                "%s - download failed due to response %d with "
                "Retry-After=%d: %s, will sleep and retry",
                path,
                exc.response.status_code,
                sleep_amount,
                exc,
            )
   ```
with
```python
sleep_amount = get_retry_after(exc.response)
```

and unindent the next log statement as I recommended below.

jwodder (approved):

dandi/utils.py
```diff
@@ -890,3 +891,65 @@ def joinurl(base: str, path: str) -> str:
...
+
+    if sleep_amount:
+        if -2 < sleep_amount < 0:
+            # allow for up to a few seconds delay in us receiving/parsing etc
+            # but otherwise assume abnormality and just return if_unparsable
+            sleep_amount = 0
+        elif sleep_amount < 0:
+            sleep_amount = None
+            lgr.warning(
+                "date in Retry-After=%r is in the past (current is %r). "
```
yarikoptic, line 940:
```suggestion
                "date in Retry-After=%s is in the past (current is %r). "
```

dandi/utils.py
```diff
@@ -890,3 +891,65 @@ def joinurl(base: str, path: str) -> str:
...
+                "Returning %r",
+                retry_after,
+                current_date,
+                sleep_amount,
+            )
+        elif sleep_amount > 7 * 24 * 60 * 60:  # week
+            sleep_amount = None
+            lgr.warning(
+                "date in Retry-After=%r is over a week in the future (current is %r). "
```
yarikoptic, line 949:
```suggestion
                "date in Retry-After=%s is over a week in the future (current is %r). "
```

dandi/download.py
yarikoptic: thanks! done that, pushed

dandi/download.py
```diff
@@ -1143,22 +1131,15 @@ def _check_if_more_attempts_allowed(
...
+    if sleep_amount is None:
+        # it was not Retry-after set, so we come up with random duration to sleep
+        sleep_amount = random.random() * 5 * attempt
         lgr.debug(
-            "%s - download failed on attempt #%d: %s, will sleep a bit and retry",
+            "%s - download failed on attempt #%d: %s, will sleep %f and retry",
             path,
             attempt,
             exc,
+            sleep_amount,
```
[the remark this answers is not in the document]
yarikoptic: ok, makes sense, dedenting the log so we always log how long we are to sleep, not only if random amount due to no Retry-After. I did remove that other log

PR #1560, opened 2025-01-07
Support dandischema v0.11.0

jwodder:
Blocked by https://github.com/dandi/dandi-archive/pull/2132


yarikoptic:
docs issue is known/unrelated. Let's proceed and see where it would get us, thank you!

PR #1561, opened 2025-01-10
Update URL for DANDI Docs

kabilar:
- [x] Depends on https://github.com/dandi/handbook/pull/165

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,10 +78,10 @@ Run `dandi --help` or `dandi <subcommand> --help` (e.g. `dandi upload --help`) t
 ## Resources
 
 * To learn how to interact with the DANDI Archive and for examples on how to use the DANDI Client in various use cases,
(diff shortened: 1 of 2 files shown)


jwodder (approved):

yarikoptic (approved):
Makes sense! We still have a little confusing `dandi` docs on RTD for the dandi-cli component, but it is what it is. Thank you @kabilar!

yarikoptic:
for type checking 
- #1562 should fix it 

and we do have problem installing deps while building docs now which is also  unrelated -- let's proceed!  Thank you @kabilar

kabilar:
> Makes sense! We still have a little confusing `dandi` docs on RTD for the dandi-cli component, but it is what it is. Thank you @kabilar!

Hi @yarikoptic, thats a good point.  I have filed https://github.com/dandi/dandi-cli/issues/1563 to hopefully address this issue.

PR #1562, opened 2025-01-10
Don't use version 0.25.5 of `responses`

jwodder:
The latest release of `responses` somehow omitted its `py.typed` file, resulting in type-checking errors from projects that use it.  [[Bug report](https://github.com/getsentry/responses/issues/751)]


yarikoptic:
docs build is apparently not new and 4 last builds did fail and another one had similar fail, so unrelated to these changes somehow.  Thank you @jwodder, let's proceed!

PR #1570, opened 2025-01-30
Pin `dandischema` to require the latest `schema` version

### LINKED ISSUE #1463: Schema mismatch between server and client
## Description

Hi team, the following steps recreate an issue I have run into today and in the past:

1. I haven't uploaded to DANDI in a while, so upgraded the DANDI Client (`pip install dandi --upgrade`) to 0.62.4.
2. Ran `dandi -l debug upload -i dandi-staging`, which returned the following error:
    ```
    Error: Server requires schema version 0.6.8; client only supports 0.6.6.  You may need to upgrade dandi and/or dandischema.
    ```
3. The installed `dandischema` version was `0.10.0`.  When I upgraded `dandischema` to `0.10.2`, I no longer received the schema mismatch error.

## Possible solution

1. Could we start to pin the `dandischema` version more strictly in the DANDI Client's requirements so that this mismatch is avoided?  Currently it is set to a range (i.e. `dandischema >= 0.9.0, < 0.11`) and so it does not necessarily get upgraded when `dandi` is upgraded.

Thank you.

kabilar:
Fixes #1463

Edit:
~We encountered this issue again today with `dandi==0.66.4` so I suggest that we pin to the latest `dandischema` version.~

~Error message:~
~$ dandi upload -i linc <path_to_ome_zarr>~
~Error: Server requires schema version 0.6.8; client only supports 0.6.9. You may need to upgrade dandi and/or dandischema~

cc @aaronkanzer @dstansby

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,7 @@ install_requires =
     bidsschematools ~= 0.7.0
     click >= 7.1
     click-didyoumean
-    dandischema >= 0.11.0
+    dandischema == 0.11.0
     etelemetry >= 0.2.2
     fasteners
     fscacher >= 0.3.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,7 @@ install_requires =
     bidsschematools ~= 0.7.0
     click >= 7.1
     click-didyoumean
-    dandischema >= 0.9.0, < 0.12
+    dandischema >= 0.11.0
     etelemetry >= 0.2.2
     fasteners
     fscacher >= 0.3.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,7 @@ install_requires =
     bidsschematools ~= 0.7.0
     click >= 7.1
     click-didyoumean
-    dandischema == 0.11.0
+    dandischema >= 0.11.0, < 0.12.0
     etelemetry >= 0.2.2
     fasteners
     fscacher >= 0.3.0


jwodder:
@kabilar This seems to be solving the wrong problem.  If the client's schema version is newer than the server's — as the given error message states — then pinning to the latest schema version will just make it so users can't downgrade their `dandischema` to an older version to solve the problem.

kabilar:
We will still upgrade the `linc-archive` `dandischema` version.

@jwodder Would it be worth it to go ahead with this pull request to pin the `dandischema` version as the `dandi-archive` does to address #1463?

From `dandi-archive` [setup.py line 45](https://github.com/dandi/dandi-archive/blob/741b5ac387b2f033d0ff2a928df7241bc43e1764/setup.py#L45):
```python
'dandischema==0.11.0',  # schema version 0.6.9
```

jwodder:
@kabilar I don't like the idea of pinning dandischema.  What if a new dandischema version is released that doesn't change the schema version but still fixes some bug?  If you really need dandischema to *only* be for the latest schema version, make the requirement be a range with the lower end being the point when that schema version was added.

kabilar:
> @kabilar I don't like the idea of pinning dandischema. What if a new dandischema version is released that doesn't change the schema version but still fixes some bug? 
Thats fair.

> If you really need dandischema to _only_ be for the latest schema version, make the requirement be a range with the lower end being the point when that schema version was added.

So in this case would `dandischema>=0.11.0` work?

jwodder:
@kabilar I think that should be `dandischema>=0.11.0,<0.12.0`.

kabilar:
> @kabilar I think that should be `dandischema>=0.11.0,<0.12.0`.

Sounds good.  I have updated it.

jwodder (approved):

yarikoptic:
tests are failing, fomr a sample run

```
FAILED dandi/tests/test_dandiarchive.py::test_parse_api_url[https://gui.dandiarchive.org/#/dandiset/000001/0.201104.2302-parsed_url2] - requests.exceptions.ConnectionError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
FAILED dandi/tests/test_dandiarchive.py::test_follow_redirect - requests.exceptions.ConnectionError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
```

looks like unrelated but that should be figured out (e.g. separate PR showing the same fails), and ideally fixed... FWIW yesterday's run seems had failed only in a single run https://github.com/dandi/dandi-cli/actions/runs/13025543642 , so more analysis should be done... assuming it was connection issue, I would restart failing tests now, so see if that would be gone

kabilar:
Not sure what changed, but it looks like all tests are passing now.

yarikoptic:
@kabilar probably they just flaked out due to connection issues or other external factors (redirects etc... as we were observing similar ones in upptime checks)

PR #1577, opened 2025-02-12
Enforce usage of auto (release) labels

asmacdo:
Ready for review, ~but not ready for merge.~

### DIFF
--- a/.github/workflows/labels.yml
+++ b/.github/workflows/labels.yml
@@ -12,7 +12,8 @@ jobs:
         uses: actions/github-script@v6
         with:
           script: |
-            const allowedPattern = /^(major|minor|patch)$/;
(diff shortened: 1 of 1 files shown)


.github/workflows/labels.yml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  check_labels:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check PR Labels
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const allowedPattern = /^(major|minor|patch)$/;
```
asmacdo, line 15: I assume we need more allowed labels here-- ie test, internal and/or similar

jwodder: The other labels are "performance", "internal", "documentation", "tests", and "dependencies".

PR #1588, opened 2025-03-06
Add Dandi admin email setting

asmacdo:
Related: https://github.com/dandi/dandi-archive/pull/2205


yarikoptic:
coolio. As soon as the name is agreed upon there, let's merge!

PR #1589, opened 2025-03-06
Add ability to release upon workflow dispatch

yarikoptic:
Bits are taken from http://github.com/brechtm/citeproc-py for which bits were taken from somewhere else as well.

Needed to release and realized that we do not do that upon dispatch yet here

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - master
+  workflow_dispatch:
 
 jobs:
   auto-release:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -31,7 +32,12 @@ jobs:
         run: python -m pip install build twine
 
       - name: Create release
-        run: ~/auto shipit -vv
+        run: |
+          if [ "${{ github.event_name }}" = workflow_dispatch ]
+          then opts=
+          else opts=--only-publish-with-release-label
```
jwodder, line 38: You also need to remove `"onlyPublishWithReleaseLabel": true` from `.autorc`.

jwodder (changes requested):

.github/workflows/release.yml
yarikoptic: thanks! now this makes more sense to me ;-)

PR #1590, opened 2025-03-07
EMBER named it sandbox already not staging

yarikoptic:
Although there is still no https://dandi-sandbox.emberarchive.org/ yet -- better than it is now and I need release, so I will release if no test fails


yarikoptic:
3.9 fail is spurious and/or unrelated:
```
_ test_parse_api_url[https://gui.dandiarchive.org/#/dandiset/000001-parsed_url0] _
```

PR #1596, opened 2025-03-24
Support for partial download of zarrs

yarikoptic:
implemented by adding "subpath" within an Asset record and adding support to assess it for .zarrs.

Note also that it adjusts logic from try/except to explicit checking -- much better since avoids swallowing unrelated exceptions.

Tried with

   dandi -l debug download -f debug --preserve-tree dandi://dandi/000108/sub-MITU01/ses-20210521h17m17s06/micr/sub-MITU01_ses-20210521h17m17s06_sample-178_stain-LEC_run-1_chunk-1_SPIM.ome.zarr/0/0/0/0/0

which seems succeeded downloading, we seems still need to

- [ ] disable checksumming for such partial downloads (since there is no partial checksums ATM at API level)
- [ ] add tests and verify that works on full file paths
- [ ] verify behavior upon redownloads

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -1240,16 +1240,29 @@ class RemoteDandiset:
         paths (e.g., ``../``) within it.
         """
         path = self._normalize_path(path)
-        try:
-            # Weed out any assets that happen to have the given path as a
-            # proper prefix:
-            (asset,) = (
-                a for a in self.get_assets_with_path_prefix(path) if a.path == path
-            )
-        except ValueError:
-            raise NotFoundError(f"No asset at path {path!r}")
-        else:
-            return asset
+        # Weed out any assets that happen to have the given path as a
+        # proper prefix:
+        assets = [a for a in self.get_assets_with_path_prefix(path) if a.path == path]
+        if assets:
+            if len(assets) > 1:
+                lgr.warning(
+                    "Multiple assets found at path %r; returning the first one",
+                    path,
+                )
+            asset = assets[0]
+        elif not assets:
+            zarr_suf = ".zarr/"
+            if (zarr_suf in path) and (zarr_suffix_idx := path.index(zarr_suf)):
+                # If path ends with .zarr/, we might have a zarr asset without
+                # a trailing slash in the path.  Try again:
+                zarr_path_len = zarr_suffix_idx + len(zarr_suf)
+                zarr_path = path[: zarr_path_len - 1]  # -1 for trailing /
+                subpath = path[len(zarr_path) :]
+                asset = self.get_asset_by_path(zarr_path)
+                asset.subpath = subpath
+            else:
+                raise NotFoundError(f"No asset at path {path!r}")
+        return asset
 
     def download_directory(
         self,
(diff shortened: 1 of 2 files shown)


dandi/dandiapi.py
```diff
@@ -1240,16 +1240,29 @@ def get_asset_by_path(self, path: str) -> RemoteAsset:
...
+        assets = [a for a in self.get_assets_with_path_prefix(path) if a.path == path]
+        if assets:
+            if len(assets) > 1:
+                lgr.warning(
+                    "Multiple assets found at path %r; returning the first one",
+                    path,
+                )
+            asset = assets[0]
+        elif not assets:
+            zarr_suf = ".zarr/"
```
jwodder, line 1254: What about `.ngff/`?

dandi/dandiapi.py
```diff
@@ -1240,16 +1240,29 @@ def get_asset_by_path(self, path: str) -> RemoteAsset:
...
+            if (zarr_suf in path) and (zarr_suffix_idx := path.index(zarr_suf)):
+                # If path ends with .zarr/, we might have a zarr asset without
+                # a trailing slash in the path.  Try again:
+                zarr_path_len = zarr_suffix_idx + len(zarr_suf)
+                zarr_path = path[: zarr_path_len - 1]  # -1 for trailing /
+                subpath = path[len(zarr_path) :]
```
jwodder, line 1260: I have several issues with this:

- This code accepts paths containing a component that's just the Zarr suffix, e.g., `foo/.zarr/bar`.
- The behavior differs from dandidav, which tests each `.zarr`/`.ngff`-suffixed leading portion for Zarr-ness.
- Instead of futzing around with strings, I think it'd better to convert `path` to a `pathlib.PurePosixPath` and check `path.parts` to see if any part ends with `.zarr` (or `.ngff`, and also that the suffix is not the whole of the part).

dandi/dandiapi.py
```diff
@@ -1378,6 +1391,8 @@ class BaseRemoteAsset(ABC, APIBase):
     identifier: str = Field(alias="asset_id")
     #: The asset's (forward-slash-separated) path
     path: str
+    #: The path within the asset, as e.g. path in a zarr/
+    subpath: str | None = Field(default=None)
```
jwodder, line 1395: I don't like the idea of a `BaseRemoteAsset` object pointing to a resource inside an asset; the class is supposed to be for assets only.  It'd be better to introduce a separate class for Zarr prefixes (and possibly also add a variant of `get_asset_by_path()` for returning this separate class).

jwodder, line 1395:
```suggestion
    subpath: Optional[str] = None
```

Pydantic can't handle `str | None` in Python versions below 3.10, and `Field(default=None)` can be simplified to just `None`.

jwodder (changes requested):

dandi/dandiapi.py
yarikoptic: yes, should also be supported

dandi/dandiapi.py
yarikoptic: ok, agree that dealing with `Path` instance might be more robust

dandi/dandiapi.py
yarikoptic: Yeah, I also disliked it a little but it is already quite a hierarchy of classes there leading to nearly combinatorial explosion.  If you see a sensible way, e.g. through adding a single extra class - please go ahead.

yarikoptic:
- #1816 is superseding

PR #1614, opened 2025-04-16
Add support for validating Zarr folders with format V3 using custom walk-through and tensorstore

### LINKED ISSUE #1609: zarr 3 support
When validating/uploading a zarr-3 archive. The command failed right away. dandi now uses zarr 2 library, which cannot read zarr 3

```
(dandi) megaera[0]:001412$ dandi validate  --min-severity HINT
[zarr.empty_group] /home/scratch/001412/test.zarr — Zarr group is empty.
2025-04-10 15:09:25,914 [    INFO] Logs saved in /homes/5/kc1708/.local/state/dandi-cli/log/2025.04.10-19.09.24Z-3985259.log
ERROR conda.cli.main_run:execute(125): `conda run dandi validate --min-severity HINT` failed. (See above for error)
```

### LINKED ISSUE #1610: Make use of get_zarr_format_version to provide version of zarr
defined but not used, I guess missed that during PR review

```shell
❯ git grep get_zarr_format_version
dandi/files/zarr.py:def get_zarr_format_version(data: Any) -> str:

```

detected while reviewing code for 
- #1609

candleindark:
This PR brings the following changes
1. It provides a temporary workaround to support validation of Zarr format V3 objects, Zarr arrays or groups, before zarr-python 3.x upgrade is done. It closes #1609.
2. In achieving point 1, this PR provides a new definition of `get_zarr_format_version()` provided and uses this definition. It closes #1610. 
3. Additionally, this PR corrects the use of `zarr.open()` in validation. `zarr.open()` has a default mode of `a`, read and write. In this mode, `zarr.open()` would attempt to modify a Zarr store if it deems the store as an invalid one, corrupting the store. This PR changes the use of `zarr.open()` from `a` mode to `r` mode, read-only.

**Note**:
Zarr format 3 objects/stores are added to the git repo for testing since with zarr-python 2.x, we can't generate such stores dynamically.

Remaining TODOs:

- [x] Add/update tests


dandi/files/zarr.py
```diff
@@ -56,31 +57,122 @@
...
+    Note
+    ----
+        This will not be needed once the upgrade to zarr-python 3.x is done and
+        should be removed.
+    """
+
+    node_type: str
+
+
+def get_zarr_format_version(ds_path: Path) -> str:
```
yarikoptic, line 80: keep variable as just path to avoid confusion with datasets / dandisets.
```suggestion
def get_zarr_format_version(path: Path) -> str:
```

candleindark: done

candleindark:
@yarikoptic Is a Zarr store, the `*.zarr` directory in the filesystem the represent a Zarr object, array or group, considered as a file in dandi-cli? I see code such as https://github.com/dandi/dandi-cli/blob/6518334b60c541890d83e293d000f350280dfbad/dandi/files/zarr.py#L265C21-L265C37 treat a directory as file by using `Scope.FILE` to mark a validation result. 

Currently,  I used `Scope.FOLDER` to mark validation error regarding Zarr objects. Should I be consistent and use `Scope.FILE` intead?

yarikoptic:
File scope was used to reflect the fact that it applies to that single "Zarr dataset", for which we have a single asset in DANDI, and not a directory which might contain multiple files/assets... Let's go with File for now, although we might want to improve or change semantic here later on

candleindark:
@yarikoptic This PR is ready for review. The remaining lines to be tested are either too simple or too tricky to worth investing time for testing this temporary workaround.

PR #1615, opened 2025-04-17
Remove obsolete (without effect) piece of code in bids.py for validation

yarikoptic:
we already demand bidsschematools ~1.0 so if we are to follow the comment -- we should be fine.  But also -- bids_paths seems to be not used (since 45209e2ae8b65492521838632237fd0bec3000c0 awhile back) so it was dead code to start with.

### DIFF
--- a/dandi/files/bids.py
+++ b/dandi/files/bids.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 from collections import defaultdict
 from dataclasses import dataclass, field
 from datetime import datetime
-import os.path
 from pathlib import Path
 from threading import Lock
 import weakref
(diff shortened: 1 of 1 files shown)


candleindark (approved):
I don't see any affect from this piece of code either. Possibly, its use has been long removed (don't see to need to investigate more now).

yarikoptic:
Fail is unrelated

```
test_parse_api_url[https://gui.dandiarchive.org/#/dandiset/000001-parsed_url0] _
/opt/hostedtoolcache/Python/3.12.10/x64/lib/python3.12/site-packages/urllib3/connectionpool.py:787: in urlopen
    response = self._make_request(
/opt/hostedtoolcache/Python/3.12.10/x64/lib/python3.12/site-packages/urllib3/connectionpool.py:488: in _make_request
    raise new_e
/opt/hostedtoolcache/Python/3.12.10/x64/lib/python3.12/site-packages/urllib3/connectionpool.py:464: in _make_request
    self._validate_conn(conn)
```

Issue #1617, opened 2025-04-23
Duplicate errors in `dandi validate` output


candleindark:
Output from `dandi validate` contains duplicates of the same errors.

Given 

<details>
<summary>dandiset.yaml</summary>

```yml
# DO NOT EDIT THIS FILE LOCALLY. ALL LOCAL UPDATES WILL BE LOST.
# It can be edited online at https://dandiarchive.org/dandiset/000027
# and obtained from the dandiarchive.
'@context': https://raw.githubusercontent.com/dandi/schema/master/releases/0.4.4/context.json
about: []
access:
- contactPoint:
    email: team@dandiarchive.org
    schemaKey: ContactPoint
    url: https://dandiarchive.org
  schemaKey: AccessRequirements
  status: dandi:OpenAccess
assetsSummary:
  approach: []
  dataStandard:
  - identifier: RRID:SCR_015242
    name: Neurodata Without Borders (NWB)
    schemaKey: StandardsType
  measurementTechnique: []
  numberOfBytes: 18792
  numberOfFiles: 1
  numberOfSubjects: 2
  schemaKey: AssetsSummary
  species:
  - identifier: http://purl.obolibrary.org/obo/NCBITaxon_10116
    name: Brown rat
    schemaKey: SpeciesType
  variableMeasured: []
citation: Halchenko, Yaroslav O. (2022) Test dataset for testing dandi-cli (Version
  0.210831.2033) [Data set]. DANDI archive. https://doi.org/10.48324/dandi.000027/0.210831.2033
contributor:
- affiliation:
  - identifier: https://ror.org/049s0rh22
    name: Dartmouth College
    schemaKey: Affiliation
  email: yoh@dartmouth.edu
  identifier: 0000-0003-3456-2493
  includeInCitation: true
  name: Halchenko, Yaroslav O.
  roleName:
  - dcite:Conceptualization
  - dcite:Author
  - dcite:Maintainer
  - dcite:ProjectMember
  - dcite:ContactPerson
  schemaKey: Person
  url: http://centerforopenneuroscience.org/
dateCreated: '2020-07-08T21:54:42.543000+00:00'
datePublished: '2021-08-31T20:33:59.583352+00:00'
description: 'Should be ignored by regular mortals.


  ATM contains only a few files from http://github.com/dandi-datasets/nwb_test_data
  which more or less appropriate (do not lack critical metadata) for testing'
doi: 10.48324/dandi.000027/0.210831.2033
ethicsApproval: []
[...]

PR #1599, opened 2025-03-28
Replace the use of `bidsschematools` with the use of deno-compiled BIDS validator in obtaining validation result

candleindark:
This PR addresses #1597 partly. It doesn't replace the use of `bidsschematools` completely with the use of the deno-compiled BIDS validator. `bidsschematools` is still needed in obtaining BIDS related metadata for individual assets. However, BIDS validation results for a dataset is now obtained completely through the deno-compiled BIDS validator.

Additionally, by modifying the validation error message regarding the `dandiset.yaml` file in BIDS validation, the PR includes changes to inform the user about `dandiset.yaml` in BIDS validation in the context of DANDI. This PR thus also closes #1602.

Remaining TODOs:

- [x] Have the `bids_validate()` func support the `--ignoreNiftiHeaders` option in the underlying deno-compiled BIDS validator
- [x] Have the `bids_validate()` func support the `--recursive` option in the underlying deno-compiled BIDS validator
- [x] Ensure the validation result of the deno-compiled BIDS validator is output to a file before it is extracted. (This way will eliminate the pollution of the result by other output to `stdout`).
- [x] Have the `bids_validate()` func support the `--config` option in the underlying deno-compiled BIDS validator
- [x] Ensure no validation error in the test of validation against the selected examples in https://github.com/bids-standard/bids-examples 
- [x] Ensure existence of validation errors in the test of validation against the selected examples in https://github.com/bids-standard/bids-error-examples
- [x] Port the use of BIDS validation by the deno-compiled validator to the codebase.

**Note**: 

- [X] This PR is based on https://github.com/dandi/dandi-cli/pull/1514. To get the relevant commits, rebase/merge with master after https://github.com/dandi/dandi-cli/pull/1514 is merged into master.
- The parsing of the validation results from the deno-compiled validator is done in accordance to the definitions in https://github.com/bids-standard/bids-validator/blob/main/src/types/validation-result.ts
- Nifti files in a BIDS dataset is not assumed to be empty and checked fully.
- This PR also contains some minor refactoring of related code to make to code more readable and manageable (see individual commits for details).
- [ ] In review, please pay particular attention to the todos in `bids_validator_deno/_validator.py`. Those todos denote options that may need to be adjusted.

PR #1619, opened 2025-04-24
Deduplicate validation results yield from `validate.validate()`

candleindark:
This PR closes https://github.com/dandi/dandi-cli/issues/1617.

This solution is a bit awkward in my opinion, but it fixes the problem without overhaul of how validation is done in dandi-cli.

### DIFF
--- a/dandi/validate.py
+++ b/dandi/validate.py
@@ -156,6 +156,16 @@ def validate(
     path, errors
       errors for a path
     """
+    # Archive of unique `ValidationResult` objects obtained through
+    # `DandiFile.get_validation_errors()`
+    # Note: This is needed to hold on to the unique `ValidationResult` objects
+    #   so that they don't get garbage collected, ensuring that the same ID in
+    #   `df_result_ids` is always associated with the same object.
+    df_results: list[ValidationResult] = []
+
+    # The ids of the objects in `df_results` obtain through the `id()` built-in function
+    df_result_ids: set[int] = set()
+
     for p in paths:
         p = os.path.abspath(p)
         dandiset_path = find_parent_directory_containing(dandiset_metadata_file, p)
@@ -171,6 +181,11 @@ def validate(
         for df in find_dandi_files(
             p, dandiset_path=dandiset_path, allow_all=allow_any_path
         ):
-            yield from df.get_validation_errors(
+            for r in df.get_validation_errors(
                 schema_version=schema_version, devel_debug=devel_debug
-            )
+            ):
+                r_id = id(r)
+                if r_id not in df_result_ids:
+                    df_results.append(r)
+                    df_result_ids.add(r_id)
+                    yield r


candleindark:
@yarikoptic Need to have this one approved quick. It is needed for https://github.com/dandi/dandi-cli/pull/1599.

yarikoptic:
"not sure", since this would just address the symptom, potentially hiding away some code inefficiency resulting in duplicate records. Did you look **why** we get duplicate records?

candleindark:
> "not sure", since this would just address the symptom, potentially hiding away some code inefficiency resulting in duplicate records. Did you look **why** we get duplicate records?

There is some inefficiency but not terrible, and this inefficiency is not the result of a bug but somewhat by design. Let's look at the code without the changes in this PR.

By NOT terrible inefficiency I mean that no repetitive validations have been done, and no duplicate `ValidationResult` objects are actually created. The duplication is resulting from repeated inclusions of the same `ValidationResult` objects at https://github.com/dandi/dandi-cli/blob/4f466a89d706f4c0c579a0cf7969383870a9360d/dandi/validate.py#L171-L176

`df` here is a `DandiFile` object that represents a "file" in the dandiset. For a dandiset that is also a BIDS dataset, `df` can be a `BIDSDatasetDescriptionAsset` or `BIDSAsset`. 

When `df` is a `BIDSDatasetDescriptionAsset`, its `get_validation_errors()` method returns all the `ValidationResult`s pertaining to the dataset and all individual assets, https://github.com/dandi/dandi-cli/blob/4f466a89d706f4c0c579a0cf7969383870a9360d/dandi/files/bids.py#L112-L124

When `df` is a `BIDSAsset`, its `get_validation_errors()` method returns `ValidationResult`s  pertaining to the dataset and the particular asset through its attached `BIDSDatasetDescriptionAsset` object, https://github.com/dandi/dandi-cli/blob/4f466a89d706f4c0c579a0cf7969383870a9360d/dandi/files/bids.py#L167-L172

https://github.com/dandi/dandi-cli/blob/4f466a89d706f4c0c579a0cf7969383870a9360d/dandi/files/bids.py#L96-L104

There reason that there is no repetition in validation and creation of the same validation result objects is that all `BIDSAsset` objects share an instance of `BIDSDatasetDescriptionAsset` https://github.com/dandi/dandi-cli/blob/4f466a89d706f4c0c579a0cf7969383870a9360d/dandi/files/bids.py#L142, and `BIDSDatasetDescriptionAsset._validate()` actually only do validation once in its life time https://github.com/dandi/dandi-cli/blob/4f466a89d706f4c0c579a0cf7969383870a9360d/dandi/files/bids.py#L64-L94

In sum, the repetition is only in the inclusions of the same `ValidationResult` objects (in `validate.validate()`) and results from how files in dandiset are currently represented.

Issue #1627, opened 2025-05-12
dandi download enums


CodyCBakerPhD:
While reviewing some PRs I decided to download a full dataset to test on

I used the syntax

```bash
dandi download DANDI:001350
```

but recieved

```bash
Usage: dandi download [OPTIONS] [URL]...
Try 'dandi download --help' for help.

Error: Invalid value for '-e' / '--existing': 'error' is not one of 'ERROR', 'SKIP', 'OVERWRITE', 'OVERWRITE_DIFFERENT', 'REFRESH'.
```

and so on for various flags. To get it to work as expected I had to do instead

```bash
dandi download DANDI:001350 -e ERROR -f PYOUT --path-type EXACT
```

which worked as I expected

Is this behavior intended? Can others reproduce it? If not assign me and I believe I know where and how it goes wrong

yarikoptic:
that is odd, are you sure you do not have some alias for `dandi` somewhere in your shell config?  here is what I see

![Image](https://github.com/user-attachments/assets/66bf2d74-565f-4349-9464-8cb578bdb558)

```
❯ dandi --version
0.68.0+2.ga7743558
```

CodyCBakerPhD:
```bash
dandi --version

> 0.68.0
```

In a fresh Conda environment (`dandi` was an unrecognized entrypoint prior to fresh install from PyPI)

(Using Windows 11)

stephprince:
I have not looked into it in detail yet, but this error also popped up in the NWBInspector [workflow logs](https://github.com/NeurodataWithoutBorders/nwbinspector/actions/runs/14974340747/job/42062713533)

CodyCBakerPhD:
Thanks for convincing me I that I have not (yet) gone completely insane @stephprince 

The logs are very helpful in that they indicate the issue is on every Python version above 3.9

I had forgotten to indicate that I was using Python 3.13 before

I just tried it again with 3.9 and verified the issue does not occur there

The new mystery is why this suddenly changed a mere 2 days ago when no DANDI related package has cut a new release in that span...

jwodder:
Hint: https://github.com/pallets/click/issues/2911

CodyCBakerPhD:
Thanks for figuring it out @jwodder 

I noticed the DANDI CLI testing suite has actually been catching this: https://github.com/dandi/dandi-cli/actions/runs/14964999717/job/42033240845#step:8:1034

But it was very hard to notice amongst all the obo lib failures

[...]

PR #1628, opened 2025-05-13
BF(workaround): disallow click 8.2.0 due to regression

yarikoptic:
refs:
- https://github.com/pallets/click/issues/2911
- https://github.com/dandi/dandi-cli/issues/1627


yarikoptic:
Windows is the known and unrelated connection issue, let's proceed to help avoid user struggles.

PR #1630, opened 2025-05-21
Restrict to specific 0.11.0 dandischema so we do not upgrade to 0.11.1 which provided new dandi schema release, and limit click before 8.2.0 for now

yarikoptic:


yarikoptic:
Windows connection issue, let's proceed

Issue #1626, opened 2025-05-12
Handle new exit behavior in deno-compiled BIDS validator


candleindark:
The exit behavior of the deno-compiled BIDS validator will change after the current version, `2.0.5`, per https://github.com/bids-standard/bids-validator/pull/196. Augment the solution provided in #1599 either before or after it is merged to master so that the behavior is of the BIDS validator at a new version is handled correctly.

Expectation:

- A few tests are needed to be update as well.

PR #1642, opened 2025-05-27
Fix daily tests by excluding new deno validator

CodyCBakerPhD:
Discovered with @asmacdo, seeing if this works


asmacdo:
dandi-archive's cli-integration tests are failing but only 1 failure, instead of the 17 i encountered.
 https://github.com/dandi/dandi-archive/pull/2391

CodyCBakerPhD:
Temporary solution until @candleindark addresses test adjustments in solution to #1626

PR #1646, opened 2025-05-28
Add superuser name for creation

asmacdo:
This avoids creating dandisets with invalid metadata,
     ie contributor name = ", "

This will require changes in dandi-archive to pass.

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -1106,6 +1106,10 @@ class RemoteDandiset:
             v = self.get_version(self.version_id)
             if v.status is VersionStatus.VALID and not v.asset_validation_errors:
                 return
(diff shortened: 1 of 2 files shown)


dandi/dandiapi.py
```diff
@@ -1106,6 +1106,10 @@ def wait_until_valid(self, max_time: float = 120) -> None:
             v = self.get_version(self.version_id)
             if v.status is VersionStatus.VALID and not v.asset_validation_errors:
                 return
+            # TODO(asmacdo) can we fail fast?
```
asmacdo, line 1109: This should be removed from this PR-- but without this each invalid dandiset waits for 2 minutes before failing.

PR #1647, opened 2025-05-28
Add superuser name for creation

asmacdo:
This avoids creating dandisets with invalid metadata, which is caused by [this bugfix PR](https://github.com/dandi/dandi-archive/pull/2377). Please see the description of that one for details.
     ie contributor name = ", "

@yarikoptic this is ready for review, but I am leaving as a draft because:
 - this should not merge unless dandi-archive/2377 merges
 - these tests are failing because they are using dandi-archive image without the bugfix. (I could easily add cross-repo PR checkout logic for PRs like I did in the dandi-archive PR)
     
- [x] Remove failfast hack

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -1106,6 +1106,10 @@ class RemoteDandiset:
             v = self.get_version(self.version_id)
             if v.status is VersionStatus.VALID and not v.asset_validation_errors:
                 return
+            # TODO(asmacdo) can we fail fast?
+            if v.status is VersionStatus.INVALID:
+                break
+
             sleep(0.5)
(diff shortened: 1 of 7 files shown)


dandi/dandiapi.py
```diff
@@ -1106,6 +1106,10 @@ def wait_until_valid(self, max_time: float = 120) -> None:
             v = self.get_version(self.version_id)
             if v.status is VersionStatus.VALID and not v.asset_validation_errors:
                 return
+            # TODO(asmacdo) can we fail fast?
```
asmacdo, line 1109: This should be removed from this PR-- but without this each invalid dandiset waits for 2 minutes before failing.

PR #1651, opened 2025-06-12
Fix bug when running metadata extraction on BIDS with NWB

CodyCBakerPhD:
Took a while to pin down the problem when uploading the BEP32 example data

I think there is still some recursion that occurs bouncing between NWB and BIDS validation, but it safely terminates without issue

Should still try to remove it, but the outright blocking issue was this line

### DIFF
--- a/dandi/metadata/nwb.py
+++ b/dandi/metadata/nwb.py
@@ -72,7 +72,7 @@ def get_metadata(
             path_metadata = df.get_metadata(digest=digest)
             meta["bids_version"] = df.get_validation_bids_version()
             # there might be a more elegant way to do this:
-            if path_metadata.wasAttributedTo is not None:
+            if path_metadata.wasAttributedTo is not None and len(path_metadata.wasAttributedTo) > 0:
                 attributed = path_metadata.wasAttributedTo[0]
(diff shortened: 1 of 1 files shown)


dandi/metadata/nwb.py
```diff
@@ -72,7 +72,7 @@ def get_metadata(
             path_metadata = df.get_metadata(digest=digest)
             meta["bids_version"] = df.get_validation_bids_version()
             # there might be a more elegant way to do this:
-            if path_metadata.wasAttributedTo is not None:
+            if path_metadata.wasAttributedTo is not None and len(path_metadata.wasAttributedTo) > 0:
```
yarikoptic, line 75: more generic would be then just

```suggestion
            if path_metadata.wasAttributedTo:
```

I think.

CodyCBakerPhD: Sure, it is more generic by abstracting to broader 'truthiness' principles, but less specific to the assertions of that particular variable (technically, we could still encounter failures in the lines below if the attribute is not a list at all)

I leave it to you if you want to be general or specific (feel free to commit suggestion if you prefer)

Issue #1639, opened 2025-05-26
Make tests independent of external services


yarikoptic:
Our testing ATM can be a subject to avalanche of failed runs due to having tests triggering communication with external services, such as ontobee , identifiers.org, etc.  Multiple issues pointing to instabilities of those services were filed:
 
- https://github.com/OntoZoo/ontobee/issues/211
- https://github.com/dandi/upptime/issues/567

I think we should generally mock away such interactions centrally and only in a single test run target testing integration with those services.

CodyCBakerPhD:
Consider how [NeuroConv](https://github.com/catalystneuro/neuroconv/blob/main/.github/workflows/live-service-testing.yml), [NWB Inspector](https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/.github/workflows/dailies.yml#L24-L25), and [NWB GUIDE](https://github.com/NeurodataWithoutBorders/nwb-guide/blob/main/.github/workflows/daily_tests.yml#L24-L28) all handle this

By splitting the testing suite into 'core' and 'those dependent on external or live serverces' (which may go down for reasons unrelated to us) we can still receive some notice that something might be awry (e.g., perhaps the service altered the way its API interactions work) it is not as high of a priority to investigate unless it persists for more than a few days

If you like that solution, then I can take this on

yarikoptic:
thank you @CodyCBakerPhD . let's walk through it when we zoom since I fail to grasp full setup there. FWIW -- I would prefer annotation / tags at pytest level so for CI it is just a matter of providing selectors, like we already have some e.g. in https://github.com/dandi/dandi-cli/blob/HEAD/dandi/conftest.py or via markers like have via helpers `@mark.skipif_no_network` and so on.

PR #1652, opened 2025-06-17
Adjust CI to differentiate obolibrary tests

CodyCBakerPhD:
Turns out the tests were quite well marked - only the CI wasn't leveraging that as fully as it could

Fixes https://github.com/dandi/dandi-cli/issues/1639 by specifying only to run obolibrary related tests on separate matrix while suppressing these during normal run

This isn't all network tests, but obolib is the main offender so starting with this for now

### DIFF
--- a/dandi/files/bases.py
+++ b/dandi/files/bases.py
@@ -799,9 +799,9 @@ def _pydantic_errors_to_validation_results(
                 )
             )
             message = e.get("message", e.get("msg", None))
-        # else:
-        #     message = f"Unhandled Pydantic error type: {e} ({type(e)})"
-        #     raise ValueError(message=message)
+
+
+
 
         out.append(
             ValidationResult(


.github/workflows/run-tests.yml
```diff
@@ -32,13 +33,14 @@ jobs:
           - ubuntu-latest
           - macos-latest
         python:
-          - 3.9
+          - 3.9  # Reaching EOL in October 2025
```
CodyCBakerPhD, line 36: Added reminder note about this

tox.ini
```diff
@@ -46,6 +46,7 @@ addopts = --tb=short --durations=10 --timeout=300
 markers =
     integration
     obolibrary
+    flaky
```
CodyCBakerPhD, line 49: I ran into an error without adding this

dandi/files/bases.py
```diff
@@ -781,19 +781,18 @@ def _pydantic_errors_to_validation_results(
...
-                    ),
-                )
+            loc = e.get("loc", [])
+            loc_str = "+".join(str(x) for x in loc)
+            id_ = ".".join(
+                filter(None, (
+                    "dandischema",
+                    e.get("type", "UNKNOWN"),
+                    loc_str,
+                ))
```
CodyCBakerPhD, line 795: Second of two MyPy chore fixes

dandi/metadata/util.py
```diff
@@ -435,7 +435,10 @@ def parse_purlobourl(
         elchild = elfound.getElementsByTagName(key)
         if elchild:
             elchild0 = elchild[0]
-            values[key] = elchild0.childNodes[0].nodeValue.capitalize()
+            node_value = elchild0.childNodes[0].nodeValue
+
+            if node_value is not None:
+                values[key] = node_value.capitalize()
```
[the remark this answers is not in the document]
CodyCBakerPhD: Moved to #1654

PR #1657, opened 2025-06-17
Adjust CI to differentiate obolibrary tests

CodyCBakerPhD:
Turns out the tests were quite well marked - only the CI wasn't leveraging that as fully as it could

Fixes https://github.com/dandi/dandi-cli/issues/1639 by specifying only to run obolibrary related tests on separate matrix while suppressing these during normal run

This isn't all network tests, but obolib is the main offender so starting with this for now

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -29,9 +29,11 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - windows-2019
+          - windows-2022
+          - windows-latest
           - ubuntu-latest
-          - macos-latest
+          - macos-latest  # M1 architecture
+          - macos-13  # non-M1 architecture
         python:
           - 3.9  # Reaching EOL in October 2025
           - '3.10'  # Needs quotes so YAML doesn't think it's 3.1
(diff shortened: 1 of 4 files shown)


yarikoptic:
@CodyCBakerPhD please resolve conflict - unfortunately can't do in browser although it is trivial one due to

image

.github/workflows/run-tests.yml
```diff
@@ -1,6 +1,7 @@
 name: Tests
 
 on:
+  workflow_dispatch:
```
CodyCBakerPhD, line 4: Went ahead and added this trigger since it is nice to be able to run CI on command

.github/workflows/run-tests.yml
```diff
@@ -28,11 +29,13 @@ jobs:
...
         os:
+          - windows-2022
           - windows-latest
           - ubuntu-latest
-          - macos-latest
+          - macos-latest  # M1 architecture
+          - macos-13  # non-M1 architecture
         python:
-          - 3.9
+          - 3.9  # Reaching EOL in October 2025
```
CodyCBakerPhD, line 38: Added reminder note about this

tox.ini
```diff
@@ -46,6 +46,7 @@ addopts = --tb=short --durations=10 --timeout=300
 markers =
     integration
     obolibrary
+    flaky
```
CodyCBakerPhD, line 49: I ran into an error without adding this

yarikoptic (approved):

PR #1664, opened 2025-07-22
Add 403 status for retry

### LINKED ISSUE #1662: Allow Retrying when Uploading Big Assets
Image
At MGB we’re subject to a network filter that terminates any HTTP connection that either exceeds a certain duration or transfers a file above a threshold size. As a result, uploads of large assets to Dandi frequently fail midway through, even though the client itself is fully capable of completing the transfer if given another chance.

# Current Behavior
- Attempting to upload large datasets or assets via the Dandi client routinely triggers our network filter.

- The HTTP connection is cut, and the client reports a failure without retrying.

- Our network team has confirmed that they cannot adjust the filter parameters on their end.

# Expected Behavior
When an upload is interrupted by a network-level timeout or size-based cutoff, the Dandi client should automatically retry (or continue) that upload until it succeeds, rather than erroring out after a single attempt.

# Proposed Solution
- New CLI flag, e.g. --retry-on-fail or --max-retries=N, which automatically resumes or re‑issues the upload request up to N times (with sensible backoff).

- Default behavior remains unchanged (no retries) to preserve backward compatibility.

CodyCBakerPhD:
Trying to fix #1662

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -314,12 +314,14 @@ class RESTFullAPIClient:
         """
         Convenience method to call `request()` with the 'POST' HTTP method.
         """
+        kwargs["retry_statuses"] = kwargs.get("retry_statuses", []) + [403]
         return self.request("POST", path, **kwargs)
 
     def put(self, path: str, **kwargs: Any) -> Any:
         """
         Convenience method to call `request()` with the 'PUT' HTTP method.
         """
+        kwargs["retry_statuses"] = kwargs.get("retry_statuses", []) + [403]
         return self.request("PUT", path, **kwargs)
 
     def delete(self, path: str, **kwargs: Any) -> Any:
--- a/dandi/consts.py
+++ b/dandi/consts.py
@@ -170,7 +170,14 @@ DRAFT = "draft"
 
 #: HTTP response status codes that should always be retried (until we run out
 #: of retries)
-RETRY_STATUSES = (429, 500, 502, 503, 504)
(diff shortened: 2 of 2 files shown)


CodyCBakerPhD:
@yarikoptic fair point; see #1666

yarikoptic:
- I think this one is superseded by #1666 then

CodyCBakerPhD:
@yarikoptic #1666 was a sub-branch showcasing what you suggested in your earlier comment; it was intended to be squashed into here to avoid scrambling commits

This branch can be merged if @calvinchai can confirm it works (since no clue how to write tests for this)

dandi/dandiapi.py
```diff
@@ -314,12 +314,14 @@ def post(self, path: str, **kwargs: Any) -> Any:
...
         Convenience method to call `request()` with the 'POST' HTTP method.
         """
+        kwargs["retry_statuses"] = kwargs.get("retry_statuses", []) + [403]
         return self.request("POST", path, **kwargs)
 
     def put(self, path: str, **kwargs: Any) -> Any:
         """
         Convenience method to call `request()` with the 'PUT' HTTP method.
         """
+        kwargs["retry_statuses"] = kwargs.get("retry_statuses", []) + [403]
```
CodyCBakerPhD, line 324:
```suggestion
        kwargs["retry_statuses"] = kwargs.get("retry_statuses", []) + [400, 403]
```

CodyCBakerPhD:
replaced by #1674

PR #1665, opened 2025-07-23
Add ndtypes from ndx-pose, ndx-photometry, ndx-fiber-photometry

rly:
In response to a Slack thread, this PR adds mappings for some neurodata types from [ndx-pose](https://github.com/rly/ndx-pose), [ndx-photometry](https://github.com/catalystneuro/ndx-photometry) (deprecated), and [ndx-fiber-photometry](https://github.com/catalystneuro/ndx-fiber-photometry). I wasn't sure whether there would be issues with the same neurodata type name from two different modules being mapped. I also made up the approaches and techniques...

It also fixes the casing of "OptogeneticStimulusSIte" -> OptogeneticStimulusSite"


rly:
@satra pointed me to:
https://github.com/BICCN/TMN/blob/main/templates/approaches_template.csv for approaches

I did not find a generic controlled vocabulary for techniques. Are you using https://github.com/BICCN/TMN/blob/main/templates/SPARC%20Modalities.csv ?

dandi/metadata/util.py
```diff
@@ -840,6 +840,36 @@ class Neurodatum(TypedDict):
         "technique": None,
         "approach": "optogenetic approach",
     },
+    "PoseEstimationSeries": {
+        "module": "ndx-pose",
```
satra, line 844: @rly - the module is what gets attached at the end of the filename. so this is where the classification would go in, not which extension it is from. should this be `behavior` instead? see earlier for some examples.

CodyCBakerPhD, line 844:
```suggestion
        "module": "behavior",
```

@rly As I understand, this is all that is necessary to get this through? (and also applying to the others)

dandi/metadata/util.py
```diff
@@ -840,6 +840,36 @@ class Neurodatum(TypedDict):
...
+        "neurodata_type": "PoseEstimation",
+        "technique": "pose estimation technique",
+        "approach": "behavioral approach",
+    },
+    "PoseTraining": {
+        "module": "ndx-pose",
+        "neurodata_type": "PoseTraining",
+        "technique": "pose estimation technique",
+        "approach": "behavioral approach",
```
satra, line 859: on a more general note, i'm a bit confused about why pose training is a data type, as that doesn't seem to reflect behavior per se, and also why it is a pose estimation technique.

some of these are direct measurement techniques or estimation techniques, but this one seems odd.

yarikoptic (changes requested):
agreeing with @satra on avoiding `-` and also overall we should aim to make them shorter

dandi/metadata/util.py
rly: PoseTraining is a group consisting of training frames and source videos. A training frame refers to a frame from a source video and an the locations of each node of 1 or more skeletons, often from human labeling. So PoseTraining isn't exactly a data type but indicative of the NWB file containing training data for pose estimation methods. Should I pick out a different data type to detect instead?

satra: we should consider data type as a broad umbrella of what data is generated from or used for (for example there is fourier analysis above). it could be associated with a "pose estimation training technique"

rly:
It's probably worthwhile also to reindex all the neurodata types observed on DANDI and add those to the map, but this PR is a start and addresses the issue raised over Slack.

yarikoptic:
there are errors from docker compose but also smth which is relevant to this PR to be addressed
```
FAILED dandi/tests/test_metadata.py::test_ndtypes[ndtypes34-asset_dict34] - AssertionError: assert 'calcium imag...ation imaging' == 'fiber photom...cal technique'
  
  - fiber photometry technique; optical technique
  + calcium imaging; cell population imaging
```

re @rly's

> It's probably worthwhile also to reindex all the neurodata types observed on DANDI and add those to the map, but this PR is a start and addresses the issue raised over Slack.

do we have enough extracted in @magland's lindi's to just grep through them?

rly:
From https://neurosift.app/dandi > "Neurodata Types Search", you can see a dropdown menu of all of the parsed neurodata types.

That is populated from parsing https://lindi.neurosift.org/dandi/neurodata_types_index.json.gz

Generated by https://github.com/magland/neurosift-kerchunker/blob/main/workflow_scripts/create_neurodata_types_index.py

I believe these come from the first 100 assets of each dandiset, only from dandisets where at least one of the first 20 assets is NWB, and only if LINDI files have been generated for the NWB files (and other edge case handling).

rly:
NWB files with sorted single units in the NWB `Units` table does not get a filename suffix, but if the data has raw ecephys data, it gets the "ecephys" suffix. I see that in the `dandi/metadata/util.py` file modified in this PR, the module is listed as "misc" and "misc" is later filtered out from appended suffixes. Is there any particular reason to not label data with a `Units` table with the "ecephys" suffix?

This confused a user when looking at https://dandiarchive.org/dandiset/001533/draft/files?location=sub-CSH-ZAD-001&page=1 which has NWB files with suffix "behavior" and has a Units table. They thought the data has only behavior and no ecephys data.

yarikoptic:
I think it would be great to add annotation on having `Units`! But I would like to be more specific than `misc` -- what is the best way to identify having them?

Also I would prefer to be more specific with them, since they are "quite derived" in a sense... what if we dedicate for them something like `eceunits` whenever we can ensure that they are associated with ecephys... or just `units` even?

In BIDS it will likely be an analogous dedicated suffix (e.g. `_units`)     for such files which would be stored under `ecephys` folder.

stephprince:
> _units makes sense (to a systems neuroscientist). _single_units or _sua (single unit activity) is even more explicit, but some rows in the Units table may represent multi-unit activity. _sortedunits may be even better. @bendichter @stephprince - as you have done experimental systems neuroscience, what do you think is most intuitive?

I would lean towards `_sorted_units`. 

I think `_units` on its own is also ok, but is potentially more confusing to a non systems neuroscientist. And then I agree that a Units table in an NWB file could contain neurons tagged as `mua` by spike sorting software, so calling it `_single_units` may lead to confusion.

dandi/metadata/util.py
```diff
@@ -823,23 +823,48 @@ class Neurodatum(TypedDict):
         "approach": "electrophysiological approach",
     },
     "Spectrum": {
-        "module": "ndx-spectrum",
+        "module": "spectrum",
```
yarikoptic, line 826: note: I wonder if this is more to be described in `_desc-`

yarikoptic (changes requested):
dandi/metadata/neurodata_typemap.py seems ot be note used, then what is it for?

type checking seems to fail.

I will move it to draft for now, but I would love for us to finalize and merge this notable improvement!!

bendichter:
I would like to help bring this over the finish line, and I think the way to do that is to land the asset metadata mapping first and split the filename suffix work into a follow-up. Some observations from looking at the current state of the PR and at what is deployed on the archive.

On why the metadata half matters on its own: I scanned the `assetsSummary` of all 895 dandisets, and the only `approach` and `measurementTechnique` values in the entire archive are the ones this map emits. The photometry and pose dandisets carry no marker at all, so there is no facet or query that can find them (see #1754, which this PR would address).

On scope: nothing in the codebase currently reads the `module` field of this map. `process_ndtypes` uses only `approach` and `technique`, and the `_ecephys`-style filename suffix in `organize.py` comes from `_populate_modalities`, which derives modalities by introspecting pynwb submodules and is untouched here. So the `misc` to `sortedunits` rename and the `_units` suffix discussion above will have no effect until `organize.py` is changed to consult this map. That is a small, separable change, and I suggest we do it in a follow-up PR rather than hold this one on it.

On the semicolon-joined terms: `"fiber photometry technique; optical technique"` becomes a single `MeasurementTechniqueType` name, not two techniques, so nothing faceting on "optical technique" would ever match it. The existing `"microscopy approach; cell population imaging"` set that precedent, but I do not think we should extend it. The schema already models both `approach` and `measurementTechnique` as lists at the asset and summary level, so no dandischema change is needed: we can make the typemap values lists (`technique: list[str]`) and have `process_ndtypes` add each term individually.

On the photometry approach term: `"calcium imaging; cell population imaging"` is wrong for a large fraction of photometry dandisets, since dLight, GRAB-DA, and GRAB-ACh sensors measure neurotransmitters rather than calcium (000251, 000351, and 001434 are examples). I would use "fiber photometry approach", which is accurate regardless of indicator and follows the pattern of "optogenetic approach".

On the new `dandi/metadata/neurodata_typemap.py`: as @yarikoptic noted, it is not imported anywhere, and it duplicates the map in `util.py`, so the two would immediately drift. I would either make it the single source of truth that `util.py` imports from, or drop it from this PR and handle the expanded map separately. If we keep it, note that matching is by bare type name, so two extensions defining the same type name will collide silently in the dict literal.

The one failing test looks like the map and the test expectation simply have `approach` and `technique` swapped for `FiberPhotometryResponseSeries`.

@rly happy to push these changes to this branch if that would help.

bendichter:
I implemented the above in a PR against this branch: https://github.com/rly/dandi-cli/pull/1. It switches the photometry entries to "fiber photometry approach" / "fiber photometry technique" as plain single terms (with single terms per type, no change to the map structure or to dandischema is needed), adds the table and device types from both photometry extensions, and drops the unused neurodata_typemap.py for now. @rly if you merge it into add_ndtypes it will show up here.

dandi/metadata/util.py
yarikoptic: so -- what was the verdict on this one -- remove this entirely or just adjust `technique` or ... ?

dandi/metadata/util.py
rly: I see. That's not great. Let's tackle this in a separate issue. I'll comment on the main PR about this.

dandi/metadata/util.py
rly: > @rly here we have 2 extensions listed to assist with importing [`HEAD`/dandi/metadata/nwb.py#L99](https://github.com/dandi/dandi-cli/blob/HEAD/dandi/metadata/nwb.py?rgh-link-date=2025-09-22T17%3A18%3A23Z#L99) -- should we also add extension information into this 'registry' and use it there too? or may be it is no longer even needed?

That very limited registry is no longer needed, because not long after NWB 2 was released, the NWB APIs started caching the specs used, including extensions, by default in the file, which makes those files readable without external packages. I sampled from the 20 lowest-numbered dandisets and found only two that do not have cached schema: 000004 and 000019. Neither uses extensions. Other extensions in those dandisets (ndx-dandi-icephys, ndx-aibs-ecephys, ndx-mies, ndx-grayscalevolume, ndx-icephys-meta, ndx-spectrum, ndx-events) have the extension cached, so an extra import is not necessary.

I think we should remove the registry and loop there.

Also, ndx-hed and ndx-events can be removed from pyproject.toml. They should not be needed unless dandi-cli is using convenience functions from them.

rly: I adjusted the `PoseTraining` mapping "technique" to "pose estimation training technique". That's good to me.

rly:
Yes, let's get this over the finish line! 🏁

Thanks @bendichter for suggesting we drop the much bigger refactor of the mapping of neurodata types, which requires larger discussion and agreement. Let's focus on getting these extension data types in and correcting longstanding issues.

I reverted the changes to the `module` mapped to `Units` and `Spectrum` since apparently `module` is unused and suffix is determined by [get_neurodata_types_to_modalities map](https://github.com/dandi/dandi-cli/blob/d91a041ec9945613889a55555148b9f0b529f256/dandi/pynwb_utils.py#L156), which uses the pynwb module hierarchy of the associated pynwb class. (Aside: this code is incomplete and misses all `DynamicTable` subtypes. I'll create a separate issue for this.)

We'll fix up `Units` module "misc" -> "sortedunits" and "Spectrum" module "ndx-spectrum" -> "spectrum" in a separate PR where `module` might end up actually being used.

### Updated PR summary

1. Add `"technique": "optogenetic stimulation technique"` to the `OptogeneticSeries` mapping, a term not previously used in the map.

3. Fix typo `"OptogeneticStimulusSIte"` -> `"OptogeneticStimulusSite"` and add `"technique": "surgical technique"`. The typo never matched real data, so newly extracted files containing an `OptogeneticStimulusSite` gain `approach: "optogenetic approach"`, `measurementTechnique: "surgical technique"`, and `variableMeasured: "OptogeneticStimulusSite"`. The first two may already be present from other types (`OptogeneticSeries`, `ElectrodeGroup`); the `variableMeasured` entry is always new.

3. Add types from ndx-pose, the deprecated ndx-photometry, and ndx-fiber-photometry. All photometry types get `"approach": "fiber photometry approach"`, a new term. `FibersTable` and `FiberPhotometryTable` get `"technique": "surgical technique"` since they carry stereotactic implant coordinates, matching `ElectrodeGroup`. The response series and the commanded voltage series get a new `"technique": "fiber photometry technique"`.

PR #1666, opened 2025-07-23
Move retry status to uploads only

### LINKED ISSUE #1662: Allow Retrying when Uploading Big Assets

CodyCBakerPhD:
Attempting to fix #1662 

applied to post & put since `iter_uploads` relies on both

### DIFF
--- a/dandi/consts.py
+++ b/dandi/consts.py
@@ -170,14 +170,7 @@ DRAFT = "draft"
 
 #: HTTP response status codes that should always be retried (until we run out
 #: of retries)
-RETRY_STATUSES = (
-    403,  # Forbidden, but can occur during normal operation on carefully controlled networks
-    429,
-    500,
-    502,
-    503,
-    504,
-)
+RETRY_STATUSES = (429, 500, 502, 503, 504)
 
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
@yarikoptic Check the PR base; this one ought not to need labels if you approve of basic approach

The base PR awaits confirmation that it actually does anything for the user before applying labels and moving forward to a merge

PR #1667, opened 2025-07-23
RF: in our example skip going through all dandisets/assets!

yarikoptic:
As archive grew, it became silly (if not stupid) to always go through all the assets of all dandisets. And we run this example for dandi-cli across all pythons/OSes.  It started to time out for some (OSX) since runs now nearly 6 hours.

With this change, which do not reduce the value of the example, but it runs in just a few seconds from home network:

    python docs/source/examples/dandiapi-example.py  0.67s user 0.05s system 10% cpu 6.973 total

Now we will still list all dandisets (but that is quick), but explore assets only of a few (5) dandisets, where some are known to have desired assets.  Comments are added so people could adjust accordingly.

This might also positively affect our server performance since we would reduce its bombardment with requests from CI!

### DIFF
--- a/docs/source/examples/dandiapi-example.py
+++ b/docs/source/examples/dandiapi-example.py
@@ -4,6 +4,11 @@ from dandi.dandiapi import DandiAPIClient
 
 with DandiAPIClient.for_dandi_instance("dandi") as client:
     for dandiset in client.get_dandisets():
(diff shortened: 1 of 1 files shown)


yarikoptic:
ok, seems to be still green, let's proceed!

PR #1668, opened 2025-07-25
Rename dandiarchive `staging` => `sandbox`

mvandenburgh:
Related issue: https://github.com/dandi/dandi-archive/issues/1030

This PR updates all hardcoded URLs within DANDI CLI to point at the new staging deployment URLs (as described in this design doc https://github.com/dandi/dandi-archive/pull/2229).

This PR must be merged/deployed before these:
- https://github.com/dandi/dandi-infrastructure/pull/233
- https://github.com/dandi/dandi-archive/pull/2449

@yarikoptic @jwodder as discussed in https://github.com/dandi/dandi-archive/pull/2229, we will need to deprecate all past versions of DANDI CLI when this PR releases - please advise on how to do that.

cc @waxlamp @jjnesbitt


mvandenburgh:
I set this as a `minor` release, but feel free to change if that isn't the right tag for this.

mvandenburgh:
> Well, I am remaining [of an opinion](https://github.com/dandi/dandi-archive/pull/2229#discussion_r2023846880) that it is ok to not bother/fail and expect people to upgrade. Requesting newer version for everyone while catering to the cases of people using sandbox IMHO is not needed. But if you insist - just boost `cli-minimal-version` returned by dandi-archive `/info` API:

We have the 308 redirect in place for api-staging.dandiarchive.org, so I don't have any strong feelings either way. I'm fine with leaving it alone and not doing any deprecation.

> that happens because the API server (and frontend proxying it) even for sandbox still returns staging URLs:

Thanks! The sandbox /info endpoint has since been updated with the new URLs, so I've just rebased this PR and the tests seem to be passing now.

yarikoptic (approved):
THANK YOU! Let's merge... I removed `release` for now as we are troubleshooting ATM and might bundle more into it

PR #1669, opened 2025-07-25
Use name/email corresponding to our dandibot so we could allow it to push releases

yarikoptic:


yarikoptic:
I need your approval now @CodyCBakerPhD ;-)

CodyCBakerPhD (approved):

CodyCBakerPhD:
Not sure - never used `.autorc` myself

If it's an external bot though one usually needs to add permissions through some mechanism

Can check If https://github.com/dandi/dandi-archive/settings/installations  to see if it says anything about such a bot

For example, CodedCov and pre-commit bots make a screen that looks like this

image

PR #1673, opened 2025-08-06
organize: operate on unique file-paths in case of multiple paths leading to the same file(s)

yarikoptic:
Originally reported in the slack #debug channel with an odd error message being displayed which looked like the one in the test in case we remove use of set():

    dandi.exceptions.OrganizeImpossibleError: Two files ('/home/yoh/.tmp/pytest-of-yoh/pytest-130/simple20/simple2.nwb' and '/home/yoh/.tmp/pytest-of-yoh/pytest-130/simple20/simple2.nwb') have the same object_id 2c6c16ec-43cd-4be5-ac3f-2c4d8a117138. Must not happen. Either files are duplicates (remove one) or were not saved correctly using NWB>=2.1.0 standard (supported by pynwb>=1.1.0).

With use of set() we would overall avoid having multiple instances of the same path being mentioned even if provided args point to the same file (directly or via folder) multiple times.

### DIFF
--- a/dandi/organize.py
+++ b/dandi/organize.py
@@ -33,6 +33,7 @@ from .utils import (
     is_url,
     load_jsonl,
     move_file,
+    pluralize,
     yaml_load,
 )
 from .validate_types import (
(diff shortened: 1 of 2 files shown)


yarikoptic:
on nfs fail was

```
ERROR dandi/tests/test_dandiapi.py::test_get_assets_with_path_prefix - dandi.exceptions.UploadError: failed to compute digest: Persisting input arguments took 0.52s to run.If this happens often in your code, it can cause performance problems (results will be correct in all cases). The reason for this is probably some large input arguments for a wrapped function.
= 834 passed, 1 skipped, 6 deselected, 2 xfailed, 9 xpassed, 1 error in 803.46s (0:13:23) =
```

CodyCBakerPhD:
Now MyPy complains :( otherwise warning seems fine

CodyCBakerPhD (approved):
LGTM

PR #1684, opened 2025-08-19
Fix auto release

CodyCBakerPhD:
@yarikoptic 

From https://github.com/intuit/auto/issues/2377

### DIFF
--- a/.autorc
+++ b/.autorc
@@ -4,6 +4,7 @@
     "email": "dandibot@mit.edu",
     "noVersionPrefix": true,
     "plugins": [
+        "protected-branch",
         "git-tag",
         [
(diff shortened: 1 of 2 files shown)


yarikoptic:
- test are failing due to sandbox renaming...
- sure thing we could try this but
  - https://www.npmjs.com/package/@auto-it/protected-branch#configuration talks about  some  to be also set `PROTECTED_BRANCH_REVIEWER_TOKEN`?
  - how does it work for dandi-archive which does **not** use that plugin?  potential answer -- they use "admin" level token?
  - but dandibot is also `admin` here!

image

so why should it help? ;-)

PR #1688, opened 2025-08-22
Revert back to use of GITHUB_TOKEN for actual push of PR to be merged with changelog

yarikoptic:
See details of analysis in
 https://github.com/dandi/dandi-cli/pull/1686#issuecomment-3214652629

I think we need two tokens now -- one to submit PR (default token should be enough) and then another to approve PR and get it merged...

This whole dance somehow for protected branch handling... last attempt!


yarikoptic:
@CodyCBakerPhD please approve and merge so we could see if this helps.

CodyCBakerPhD (approved):

PR #1708, opened 2025-09-18
Document options for the `dandi download --existing` flag

kabilar:
For `dandi download --existing` I am a bit confused as to when to use `overwrite-different` and `refresh` so perhaps we can clarify that here.

### DIFF
--- a/dandi/cli/cmd_download.py
+++ b/dandi/cli/cmd_download.py
@@ -64,9 +64,14 @@ Download files or entire folders from DANDI.
     "--existing",
     type=click.Choice(list(DownloadExisting)),
     # TODO: verify-reupload (to become default)
-    help="What to do if a file found existing locally. 'refresh': verify "
-    "that according to the size and mtime, it is the same file, if not - "
-    "download and overwrite.",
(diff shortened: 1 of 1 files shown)


docs/source/cmdline/download.rst
```diff
@@ -20,6 +20,15 @@ Options
...
 
+    For ``error``, if the local file exists, display an error and skip downloading that asset.
+
+    For ``skip``, if the local file exists, skip downloading that asset.
+
+    For ``overwrite``, if the local file exists, overwrite that asset.
+
+    For ``overwrite-different``, if the local file's hash is the same as on the
+    server, the asset is skipped; otherwise, it is redownloaded.
+
```
yarikoptic, line 31: pushed corresponding changes to the code -- docs maintained manually in parallel for such ATM

yarikoptic:
failing test is unrelated:

```
FAILED dandi/cli/tests/test_service_scripts.py::test_reextract_metadata - Failed: Timeout (>300.0s) from pytest-timeout.
```

yarikoptic (approved):

PR #1714, opened 2025-09-29
feat: add session end time extraction from NWB files

### LINKED ISSUE #1713: duration of sessions
I am involved in a project where it is useful to estimate the number of hours of publicly available data across the archive for different modalities (ephys, ca imaging, etc.). This currently requires examination of the contents of (all) NWB files individually, but I think this is a good candidate for pulling more metadata into the asset metadata. This would help a lot of anyone interested in filtering neurophysiology recordings by the duration of the session. We could also provide a total duration attribute in the assets summary, which might be of interest to data users. At the very least it will help us communicate the volume of data available in the archive.

I think it would be possible to represent this without modification to the schema using the wasGeneratedBy [Activity](https://github.com/dandi/dandi-schema/blob/12b071a890b187cd55d3d01d352ecd02da36f34f/dandischema/models.py#L1196), which has fields for startDate and endDate (which are in fact datetimes).

This would just involve creating a new "acquisition" activity where the start date is taken directly from the NWB session_start_time and the end date is calculated by the getting the max duration of all the time series and time intervals tables.

If there is interest in this, I can put together a few functions that could plug into the DANDI CLI to accomplish this.

bendichter:
fix #1713 

- Calculate session duration from NWB TimeSeries and DynamicTable timestamps
- Add new `_get_session_duration()` function to extract min/max timestamps
- Compute session end time by adding duration to session start time
- Include endDate field in Session metadata extraction
- Add comprehensive tests for session duration calculation

This enhancement improves metadata completeness by automatically determining when recording sessions ended based on the actual data timestamps in NWB files, rather than relying solely on manually-provided metadata.

### DIFF
--- a/dandi/pynwb_utils.py
+++ b/dandi/pynwb_utils.py
@@ -261,8 +261,85 @@ def _get_pynwb_metadata(path: str | Path | Readable) -> dict[str, Any]:
 
         # get external_file data:
         out["external_file_objects"] = _get_image_series(nwb)
+        
+        # Calculate session duration for metadata
+        session_duration = _get_session_duration(nwb)
+        if session_duration is not None and out.get("session_start_time") is not None:
+            # Convert to absolute datetime by adding duration to session_start_time
+            from datetime import timedelta
+            start_time = out["session_start_time"]
+            out["session_end_time"] = start_time + timedelta(seconds=session_duration)
 
     return out
+
+
+def _get_session_duration(nwb: pynwb.NWBFile) -> float | None:
+    """Calculate the duration of a recording session from NWB file contents.
+    
+    This function finds the minimum and maximum timestamps across all TimeSeries and 
+    DynamicTable objects with time information, then returns the duration as max - min.
+    
+    Parameters
+    ----------
+    nwb: pynwb.NWBFile
+        An open NWB file object
+        
+    Returns
+    -------
+    float or None
+        The session duration in seconds (max_time - min_time),
+        or None if no time information could be extracted
+    """
+    min_time = None
+    max_time = None
+    
+    # Iterate through all objects in the NWB file
+    for obj in nwb.objects.values():
+        try:
+            # Handle TimeSeries objects
+            if isinstance(obj, pynwb.base.TimeSeries):
+                if obj.timestamps is not None and len(obj.timestamps) > 0:
+                    # Use first and last timestamps
+                    start_val = float(obj.timestamps[0])
+                    end_val = float(obj.timestamps[-1])
+                    if min_time is None or start_val < min_time:
+                        min_time = start_val
+                    if max_time is None or end_val > max_time:
+                        max_time = end_val
+                elif obj.starting_time is not None and obj.rate is not None and obj.data is not None:
+                    # Calculate start and end time
+                    start_val = float(obj.starting_time)
+                    num_samples = len(obj.data)
+                    end_val = float(obj.starting_time + (num_samples / obj.rate))
+                    if min_time is None or start_val < min_time:
+                        min_time = start_val
+                    if max_time is None or end_val > max_time:
+                        max_time = end_val
+            
+            # Handle DynamicTable objects with start_time and stop_time
+            elif isinstance(obj, hdmf.common.DynamicTable):
+                if 'start_time' in obj.colnames:
+                    start_times = obj['start_time'].data[:]
+                    if len(start_times) > 0:
+                        start_val = float(min(start_times))
+                        if min_time is None or start_val < min_time:
+                            min_time = start_val
+                if 'stop_time' in obj.colnames:
+                    stop_times = obj['stop_time'].data[:]
+                    if len(stop_times) > 0:
+                        end_val = float(max(stop_times))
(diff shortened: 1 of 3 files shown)


rly:
`Units` tables also have timestamps in a `spike_times` column that are not in `start_time` or `stop_time` columns. The new `EventsTable` also has a column named `timestamp` and an optional column `duration`.

rly:
The `start_time`, `stop_time`, `spike_times`, and `timestamp` columns will use a new `TimestampVectorData` type, which will improve discovery of these data in the future.

dandi/pynwb_utils.py
```diff
@@ -262,9 +262,97 @@ def _get_pynwb_metadata(path: str | Path | Readable) -> dict[str, Any]:
...
+            # Handle TimeSeries objects
+            if isinstance(obj, pynwb.base.TimeSeries):
+                if obj.timestamps is not None and len(obj.timestamps) > 0:
+                    # Use first and last timestamps
+                    start_val = float(obj.timestamps[0])
+                    end_val = float(obj.timestamps[-1])
+                    if min_time is None or start_val < min_time:
+                        min_time = start_val
+                    if max_time is None or end_val > max_time:
+                        max_time = end_val
```
yarikoptic, line 309: too much of repeated `if`s -- I think it would be much cleaner if you just collect all `start_times, end_times: list[ float]` and then do min/max invocation on those

dandi/pynwb_utils.py
```diff
@@ -262,9 +262,97 @@ def _get_pynwb_metadata(path: str | Path | Readable) -> dict[str, Any]:
         # get external_file data:
         out["external_file_objects"] = _get_image_series(nwb)
 
+        # Calculate session duration for metadata
+        session_duration = _get_session_duration(nwb)
+        if session_duration is not None and out.get("session_start_time") is not None:
+            # Convert to absolute datetime by adding duration to session_start_time
+            from datetime import timedelta
```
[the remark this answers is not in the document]
bendichter: ok, they have been moved up

dandi/tests/test_metadata.py
```diff
@@ -470,6 +470,140 @@ def test_time_extract_gest() -> None:
     )
 
 
+def test_session_duration_extraction(tmp_path: Path) -> None:
+    """Test that session duration is extracted and included in Session activity"""
+    from datetime import datetime
+    import numpy as np
+    from pynwb import NWBFile, TimeSeries, NWBHDF5IO
```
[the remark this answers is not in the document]
bendichter: I like these tests as stand-along functions. They are testing specific functionality, both for TimeSeries and for TimeIntervals objects. I added the decorator

bendichter:
I've added code for the types @rly suggested

dandi/pynwb_utils.py
```diff
@@ -262,9 +263,84 @@ def _get_pynwb_metadata(path: str | Path | Readable) -> dict[str, Any]:
...
+            # Handle DynamicTable objects with start_time and stop_time
+            elif isinstance(obj, hdmf.common.DynamicTable):
+                if "start_time" in obj.colnames:
+                    start_data = obj["start_time"].data[:]
+                    if len(start_data) > 0:
+                        start_times.append(float(min(start_data)))
+                if "stop_time" in obj.colnames:
+                    stop_data = obj["stop_time"].data[:]
+                    if len(stop_data) > 0:
+                        end_times.append(float(max(stop_data)))
```
yarikoptic, line 327: could be shortened even further but at least let's do

```suggestion
                if "start_time" in obj.colnames:
                    if len(start_data := obj["start_time"].data):
                        start_times.append(float(min(start_data)))
                if "stop_time" in obj.colnames:
                    if len(stop_data := obj["stop_time"].data):
                        end_times.append(float(max(stop_data)))
```

unless there was some meaning for explicit slicing with `[:]`, then those should get back I guess

dandi/pynwb_utils.py
```diff
@@ -262,9 +263,82 @@ def _get_pynwb_metadata(path: str | Path | Readable) -> dict[str, Any]:
         # get external_file data:
         out["external_file_objects"] = _get_image_series(nwb)
 
+        # Calculate session duration for metadata
+        session_duration = _get_session_duration(nwb)
+        if session_duration is not None and out.get("session_start_time") is not None:
```
yarikoptic, line 268:
```suggestion
        # Calculate session duration for metadata
        if out.get("session_start_time") is not None \
            and ((session_duration := _get_session_duration(nwb) is not None):
```
would need to be checked but the point -- there is no point to call `_get_session_duration` at all if there is no "session_start_time".  

But it also made me wonder if it is legit at all to get duration as `max(ends) - min(starts)` and add to some start and claim that it is a start of session... Why then not just to take `max(ends)` as the end of the session?  E.g. what if session started at 0, first recording at 1, and ended at 2?

yarikoptic: but it is just an assumption that whatever minimal timestamp you find (besides `session_start_time`) corresponds to the same time point. Is that correct?
Did you try running this through nwb's in the archive to see what you get and whether it makes sense?

PR #1716, opened 2025-10-06
Add strain field to subject metadata and corresponding test case

### LINKED ISSUE #1715: strain information not being parse from NWB files
Strain is not one of the attributes listed here:

https://github.com/dandi/dandi-cli/blob/2827711bdf2265673599cb4c6bbd91d1e4eeeb24/dandi/consts.py#L24-L31

We should be able to parse it basically the same way we do genotype.

bendichter:
fix #1715

### DIFF
--- a/dandi/consts.py
+++ b/dandi/consts.py
@@ -27,6 +27,7 @@ metadata_nwb_subject_fields = (
     "genotype",
     "sex",
     "species",
+    "strain",
     "subject_id",
 )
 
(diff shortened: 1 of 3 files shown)


bendichter:
@yarikoptic can you take it from here?

yarikoptic:
Thank you @bendichter ! Trying to figure it out but goes slow -- what I can say only that it seems that it does the job but we do need some better way to depict what pytest found different in that bloody comparison...

bendichter:
I don't think it's a fluke. I think it has happened multiple times

yarikoptic (approved):

PR #1718, opened 2025-10-06
OSX-compatible installation command

bendichter:


yarikoptic (approved):

CodyCBakerPhD:
> I do not think it is OSX-specific one... what shell are you using and what error do you see if not quoted?

The error it gives is `zsh: no matches found: somepackage[extra]`

It only occurs on Mac but I have not attempted other unix variants together with their Zsh

Either way always wrapping in quotes will always solve the issue on all platforms

PR #1728, opened 2025-10-15
Set maintainers to whole team and fixup label CI error message

asmacdo:
Not sure if this is what we want, but IMO our maintainers list should be more than just yarik.

Second commit fixes up the error message when one of the required labels is missing. Happy to move that to another PR if preferred. Tested by pushing the change (no label)  getting the failure, adding the label, and now the Check PR label test passes.

### DIFF
--- a/.github/workflows/labels.yml
+++ b/.github/workflows/labels.yml
@@ -12,13 +12,13 @@ jobs:
         uses: actions/github-script@v8
         with:
           script: |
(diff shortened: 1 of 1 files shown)


yarikoptic:
thanks for reducing the burden from my soul.  But for the 2nd one to be melded
- need to adjust title so properly describes bundle of two changes
- and also I am still to grasp how wrong it was before -- could you cut/paste screenshots on how it looked before and after the fix? (so might be better just for a separate PR frankly)

Issue #1597, opened 2025-03-24
Replace `bidsschematools` with bids-validator-deno invocation/processing


yarikoptic:
Now that there is https://pypi.org/project/bids-validator-deno/

The problems to expect:
- there will be lots more of new errors -- would be worth checking on where our bids datasets already present in the archive stand
- need to re-approach `dandi validate` output -- should also ensure it does proper validation and we provide reasonable capabilities to filter and summarize (per type error , per folder, path) etc

yarikoptic:
well, we do validation against bida-validator-deno ATM but we still use `bidsschematools` to extract basic metadata based on file paths.

candleindark:
@yarikoptic We stopped at the conclusion of the first checkpoint. No official progress on the last two except that in doing checkpoint 1, we reorganized https://github.com/dandi/dandi-cli/blob/master/dandi/validate_types.py quite a bit which implementation of checkpoint 2 and 3 can build on.

PR #1738, opened 2025-11-09
Remove `ValidationResult.path_regex` as possible return of `purview()`

candleindark:
A purview is expected to be the string expression of a path or `None` in its only usages in `_process_issues()` as indicated in the comments shown below, so it is inappropriate to return `ValidationResult.path_regex` which is a regex.

https://github.com/dandi/dandi-cli/blob/e734c68994f1a0e8c4826c6ae041bf6fc0d57e25/dandi/cli/cmd_validate.py#L157-L158

Note: This change will pave the way to implement a solution for https://github.com/dandi/dandi-cli/issues/1597.

### DIFF
--- a/dandi/validate_types.py
+++ b/dandi/validate_types.py
@@ -219,8 +219,6 @@ class ValidationResult(BaseModel):
     def purview(self) -> str | None:
         if self.path is not None:
             return str(self.path)
-        elif self.path_regex is not None:
-            return self.path_regex
         elif self.dataset_path is not None:
             return str(self.dataset_path)
         else:


candleindark:
> well, it could be the actual string representation of the regex, and that is what I guess it was expected to be. That would have allowed to point to group of files right away. 

You are right. The `purview` method is really just a part of an organization scheme for displaying the errors. With that said, the following is some considerations.

In the current definition of `purview`, the definition before this proposed change, the other two possible non-none values are `ValidaitonResult.path` and `ValidationResult.dataset_path`. Both values are the absolute path of a file/folder that exists in the filesystem according to usage instances of `ValidationResult` I have analyzed. On the other hand, `ValidationResult.path_regex` is a regex used in validating the last component of a path as in https://github.com/dandi/dandi-cli/blob/e734c68994f1a0e8c4826c6ae041bf6fc0d57e25/dandi/organize.py#L1141-L1169

Particularly, it is a regex used in validating the last component of a path. It is not necessarily the case that the component matches the regex. It can be the case that the component doesn't match the regex as in 

https://github.com/dandi/dandi-cli/blob/e734c68994f1a0e8c4826c6ae041bf6fc0d57e25/dandi/organize.py#L1141-L1169


> I guess the real underlying issue is that with bids-validator-deno we do not have regexes and thus it would not be applicable. But is it really required for #1597 ?

No, you are right about this. The `purview` method is really just a part of an organization scheme for displaying the errors.

Let's close this PR if you are OK with what `purview` currently is after reading the analysis I provided.

PR #1743, opened 2025-11-16
Add filtering of issues by type/ID or by file location

candleindark:
This addresses checkpoint 2 of https://github.com/dandi/dandi-cli/issues/1597. It adds two options to `dandi validate` specified in the "Release Notes" section below.

TODO:

- [x] Decide whether to add or `minor` or `patch` label.
- [ ] ATM awaits for https://github.com/dandi/dandi-cli/pull/1822 to be merged first so it could operate on reloaded entrries

## Release Notes
The `dandi validate` now has two additional options

1. `--match` which takes in a list of comma-separated regex patterns to filter issues in validation results by their ID.
2. `--include-path` which takes in multiple paths to filter issues in the validation results to those associated with the paths.

### DIFF
--- a/dandi/cli/cmd_validate.py
+++ b/dandi/cli/cmd_validate.py
@@ -108,7 +108,7 @@ def validate_bids(
 def validate(
     paths: tuple[str, ...],
     ignore: str | None,
-    match: Optional[list[re.Pattern]],
+    match: Optional[set[re.Pattern]],
     grouping: str,
     min_severity: str,
     schema: str | None = None,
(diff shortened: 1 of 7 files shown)


dandi/cli/cmd_validate.py
```diff
@@ -80,6 +86,26 @@ def validate_bids(
...
+    help=(
+        "Comma-separated regex patterns used to filter issues in validation results "
+        "by their ID. Only issues with an ID matching at least one of the given "
+        "patterns are included in the eventual result. "
+        "(No pattern should contain a comma.)"
+    ),
+    callback=parse_regexes,
+)
+@click.option(
+    "--include-path",
```
yarikoptic: - name was a little confusing as I thought "include name in the result"
- we might want to simply (ab)use existing `paths` posarg as it is to define what paths we care about. So even if within bids dataset, bids-validator-deno would run on entire dataset ATM and then we need to filter for those paths
  - we should also warn about that -- that we do run bids-validator on the full dandiset although interested only in some

yarikoptic (changes requested):

yarikoptic (changes requested):
I left comments in line, without compiling into this report.  I will move into `draft` to avoid paying attention until the next round (take out of draft then)

Issue #1762, opened 2025-12-02
Archive-CLI integration tests hanging forever


jjnesbitt:
The CLI integration tests (`master`) are hanging after the following line:

```
tests/test_keyring.py::test_dandi_authenticate_no_env_var PASSED
```

I've observed this multiple times, on multiple different branches/PRs. [Here](https://github.com/dandi/dandi-archive/actions/runs/19839644234/job/56845501680?pr=2644) is just one example.

It seems to hang indefinitely, finally being killed by github after 6 hours.

yarikoptic:
<details>
<summary>FWIW -- tests pass for me locally using identical run</summary> 

```shell
    Updated https://github.com/dandi/dandi-cli (a18619e1b05b8f6f58c055c98cd5d98f55dadd5c)
      Built dandi @ git+https://github.com/dandi/dandi-cli@a18619e1b05b8f6f58c055c98cd5d98f55dadd5c
Installed 110 packages in 99ms
==================================================================================================== test session starts ====================================================================================================
platform linux -- Python 3.11.12, pytest-9.0.1, pluggy-1.6.0 -- /home/yoh/.cache/uv/archive-v0/35jqbSVyH5j-Tkxn1BtZ0/bin/python
cachedir: .pytest_cache
rootdir: /tmp/test-dandi-cli
plugins: rerunfailures-16.1, mock-3.15.1, timeout-2.4.0, cov-7.0.0, dandi-0.73.2+22.ga18619e1
collected 874 items / 591 deselected / 283 selected                                                                                                                                                                         

cli/tests/test_service_scripts.py::test_reextract_metadata PASSED                                                                                                                                                     [  0%]
cli/tests/test_service_scripts.py::test_update_dandiset_from_doi[10.1101/2020.01.17.909838-biorxiv] PASSED                                                                                                            [  0%]
cli/tests/test_service_scripts.py::test_update_dandiset_from_doi[10.1523/JNEUROSCI.6157-08.2009-jneurosci] PASSED                                                                                                     [  1%]
cli/tests/test_service_scripts.py::test_update_dandiset_from_doi[10.1016/j.neuron.2019.10.012-neuron] PASSED                                                                                                          [  1%]
cli/tests/test_service_scripts.py::test_update_dandiset_from_doi[10.7554/eLife.48198-elife] PASSED                                                                                                                    [  1%]
[...]

PR #1763, opened 2025-12-02
bf(tests): pass DBUS_SESSION_BUS_ADDRESS env var for docker

yarikoptic:
Otherwise we observe

  error getting credentials - err: exit status 1, out: `Cannot autolaunch D-Bus without X11 $DISPLAY`

when running undex tox.

FWIW, shouldn't help 

- #1762 

since that one is not done via tox.


candleindark (approved):
Would you give a bit more details why we need this? Running tests with tox in a Linux environment when accessing the credential operations? How is this related to docker, as specified in the title.

Issue #1764, opened 2025-12-05
`keyring.py` is shadows and create confusion with the `keyring` package


candleindark:
The [`keyring.py`](https://github.com/dandi/dandi-cli/blob/f5b8bc3eec06c375d73bdc6a05994585dc220d86/dandi/keyring.py) file shadows the [keyring dependency](https://github.com/dandi/dandi-cli/blob/master/pyproject.toml) in some environments. For example, the shadowing happens in the debug environment of PyCharm which starts Python with something like

```
.../pydevd.py --file .../_jb_pytest_runner.py --path <path to local dandi package source>
```

which adds the package directory `.../dandi-cli/dandi` directly to sys.path. Additionally, the setup of [importing from the `keyring` dependency](https://github.com/dandi/dandi-cli/blob/f5b8bc3eec06c375d73bdc6a05994585dc220d86/dandi/keyring.py#L9-L12) in side a file named `keyring.py` is quite confusing.

I propose renaming the `keyring.py` file to `keyring_utils.py` or some other similar name.

PR #1765, opened 2025-12-05
Rename `keyring.py` to `keyring_utils.py`

candleindark:
This PR closes https://github.com/dandi/dandi-cli/issues/1764.


candleindark:
> Although I don't expect but before we merge please search GitHub for dandi and keyring to ensure that no third party imports it.

I think we are good to go. The only query that located some references of module is [`dandi.keyring NOT repo:dandi/dandi-cli`](https://github.com/search?q=dandi.keyring%20NOT%20repo%3Adandi%2Fdandi-cli&type=code). None of the matches is of concern. They are either in forks of dandi-cli or an auto-generated repo that doesn't actually import `dandi.keyring`.

See https://github.com/dandi/dandi-cli/pull/1766 for detials.

PR #1768, opened 2025-12-12
Improve help message for the `--validation` option in `dandi upload`

candleindark:
This PR improves help message for the `--validation` option in `dandi upload`.

### Reviewer notes:

The meaning of the choice to [ignore](https://github.com/dandi/dandi-cli/blob/35011905a73d86aa13a54eb5fc1d1a038cc792ae/dandi/upload.py#L88) validation is unclear. In fact, the [choice](https://github.com/dandi/dandi-cli/blob/35011905a73d86aa13a54eb5fc1d1a038cc792ae/dandi/upload.py#L88) is not used in the production code. I made an educated guess on the meaning of the choice in constructing the help message.

### DIFF
--- a/dandi/cli/cmd_upload.py
+++ b/dandi/cli/cmd_upload.py
@@ -39,7 +39,11 @@ from ..upload import UploadExisting, UploadValidation
 )
 @click.option(
     "--validation",
-    help="Data must pass validation before the upload.  Use of this option is highly discouraged.",
+    help="Controls validation requirements before upload. (Setting this option to a "
+    "value other than 'require' is highly discouraged.) "
+    "'require' - data must pass validation before upload; "
+    "'skip' - no validation is performed on data before upload; "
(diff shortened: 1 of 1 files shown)


dandi/cli/cmd_upload.py
```diff
@@ -39,7 +39,10 @@
 )
 @click.option(
     "--validation",
-    help="Data must pass validation before the upload.  Use of this option is highly discouraged.",
+    help="Controls validation requirements before upload. "
```
yarikoptic: Let's keep the message!

```suggestion
    help="Controls validation requirements before upload. Use of this option is highly discouraged. "
```

candleindark:
```suggestion
    help="Controls validation requirements before upload. (Setting this option to a value other than 'require' is highly discouraged.)"
```

candleindark: @yarikoptic I applied my own suggested change mentioned above. Please re-review.

PR #1779, opened 2026-01-29
Allow for newer versions of `zarr` package

kabilar:
cc @satra @calvinchai @ayendiki


yarikoptic:
actually -- here we have only macos and only intel failing on unrelated tests -- let's just proceed with this

yarikoptic (approved):

PR #1809, opened 2026-02-24
Bump nwbinspector to ~= 0.7.0

bendichter:
## Summary
- Bumps `nwbinspector` dependency from `>= 0.4.28,!=0.4.32` to `~= 0.7.0`

## Test plan
- [ ] CI passes with the new nwbinspector version


yarikoptic:
on the other hand -- why we need to depend/restrict to 0.7 series of `nwbinspector`, are you planing to break in 0.8 series?

note that we tend to maintain minimal compatible versions, and not to restrict top versions unless necessary.

@bendichter what was actual motivation now, as why not before?

CodyCBakerPhD (approved):

PR #1830, opened 2026-04-02
Refactoring of codebase into dandi/validate/ subpackage for larger #1822

yarikoptic:
by *un*popular demand of @CodyCBakerPhD  in

- #1822 

to simplify review of that beast which I now will force push - it was rebased on top of this, so we could concentrate on functional changes.

Marking as 'minor' since chance structure and thus API.  Besides that, there should be no functional changes and should be mergeable as long as CI is green


CodyCBakerPhD (approved):

CodyCBakerPhD:
Ahh so easy to review this one!

Thank you :)

PR #1832, opened 2026-04-09
Fix broken external_file paths when organizing on Windows

h-mayorquin:
When `dandi organize --update-external-file-paths` runs on Windows, the `external_file` paths written into NWB files use backslashes instead of forward slashes.

Expected:
```
ses-2026-02-12-1_image/21b52ce3_external_file_0.mp4
```

On Windows before this fix:
```
ses-2026-02-12-1_image\21b52ce3_external_file_0.mp4
```

Downstream tools expecting forward slashes (as DANDI/S3 asset keys always use) fail silently when resolving these paths against the archive. This was observed in Dandiset 001771 and reported in [catalystneuro/nwb-video-widgets#33](https://github.com/catalystneuro/nwb-video-widgets/issues/33), where `get_dandi_video_info()` returned empty results. A defensive workaround was merged in [catalystneuro/nwb-video-widgets#38](https://github.com/catalystneuro/nwb-video-widgets/pull/38), but the root cause is here in `dandi organize`.

The fix has two parts:

- **Write side** (`organize.py`): `_create_external_file_names()` used `os.path.join` to build external file paths. On Windows that produces backslashes. Replaced with `posixpath.join` since these are DANDI/S3 keys, not local filesystem paths.
- **Read side** (`pynwb_utils.py`): `_get_image_series()` wrapped external file strings in `Path()`, which on Windows converts forward slashes back to backslashes. Replaced with `PurePosixPath`.

I have also added a regression assertion in `test_video_organize` that `external_file` values never contain backslashes. The initial push included only the write-side fix; Windows CI caught that `Path()` on the read side was re-introducing backslashes, confirming the assertion works as intended.

### DIFF
--- a/dandi/pynwb_utils.py
+++ b/dandi/pynwb_utils.py
@@ -408,8 +408,8 @@ def _get_image_series(nwb: pynwb.NWBFile) -> list[dict]:
             if isinstance(ob, pynwb.image.ImageSeries) and ob.external_file is not None:
                 out_dict = dict(id=ob.object_id, name=ob.name, external_files=[])
                 for ext_file in ob.external_file:
-                    if PurePosixPath(ext_file).suffix in VIDEO_FILE_EXTENSIONS:
-                        out_dict["external_files"].append(PurePosixPath(ext_file))
+                    if (path := PurePosixPath(ext_file)).suffix in VIDEO_FILE_EXTENSIONS:
+                        out_dict["external_files"].append(path)
                     else:
                         lgr.warning(
                             "external file %s should be one of: %s",
--- a/dandi/pynwb_utils.py
+++ b/dandi/pynwb_utils.py
@@ -17,7 +17,7 @@ from datetime import timedelta
 import inspect
 import os
 import os.path as op
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 import re
 from typing import IO, Any, TypeVar, cast
 import warnings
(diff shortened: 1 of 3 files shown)


h-mayorquin:
I can't add labels (no write access), a maintainer will need to add patch. The codecov checks are failing but the actual changes shouldn't reduce coverage.

dandi/pynwb_utils.py
```diff
@@ -408,8 +408,8 @@ def _get_image_series(nwb: pynwb.NWBFile) -> list[dict]:
             if isinstance(ob, pynwb.image.ImageSeries) and ob.external_file is not None:
                 out_dict = dict(id=ob.object_id, name=ob.name, external_files=[])
                 for ext_file in ob.external_file:
-                    if Path(ext_file).suffix in VIDEO_FILE_EXTENSIONS:
-                        out_dict["external_files"].append(Path(ext_file))
+                    if PurePosixPath(ext_file).suffix in VIDEO_FILE_EXTENSIONS:
+                        out_dict["external_files"].append(PurePosixPath(ext_file))
```
CodyCBakerPhD:
```suggestion
                    if (path := PurePosixPath(ext_file)).suffix in VIDEO_FILE_EXTENSIONS:
                        out_dict["external_files"].append(path)
```

WDYT

CodyCBakerPhD:
@h-mayorquin Perhaps worth an NWB Inspector check as well?

dandi/pynwb_utils.py
h-mayorquin: The issue is that `_create_external_file_names` calls [`ext_file.suffix`](https://github.com/dandi/dandi-cli/blob/4cc311b2d7a1916a9e4642e73c83218815ecc4bc/dandi/organize.py#L280) on these values, so storing plain strings would break that. `PurePosixPath` keeps `.suffix` available while ensuring forward slashes on all platforms.

CodyCBakerPhD: @h-mayorquin OK, refactored the walrus

h-mayorquin:
> @h-mayorquin Perhaps worth an NWB Inspector check as well?

Yes:
https://github.com/NeurodataWithoutBorders/nwbinspector/issues/697

dandi/pynwb_utils.py
h-mayorquin: You mean I should add the walrus? if so, I did.

CodyCBakerPhD (approved):

CodyCBakerPhD:
LGTM

Thanks for the catch

h-mayorquin:
what's up with the code coverage?

CodyCBakerPhD:
> what's up with the code coverage?

Something to do with how it counts import lines

All source code lines look to be hit

PR #1839, opened 2026-04-22
Use multi-part upload on zarr chunks larger than 5GB

jjnesbitt:
Depends on and adds support for https://github.com/dandi/dandi-archive/pull/2784

### DIFF
--- a/dandi/files/zarr.py
+++ b/dandi/files/zarr.py
@@ -1037,15 +1037,9 @@ class EntryUploadTracker:
     @staticmethod
     def _mkitem(e: LocalZarrEntry) -> UploadItem:
         # Avoid heavy import by importing within function:
-        from dandi.support.digests import md5file_nocache, multipart_md5file_nocache
+        from dandi.support.digests import md5file_nocache
 
-        file_size = e.filepath.stat().st_size
-        digest = (
-            md5file_nocache(e.filepath)
-            if file_size <= ZARR_LARGE_CHUNK_THRESHOLD
-            else multipart_md5file_nocache(e.filepath)
-        )
-
+        digest = md5file_nocache(e.filepath)
         return UploadItem.from_entry(e, digest)
 
     def get_items(self, jobs: int = 5) -> Generator[UploadItem, None, None]:
--- a/dandi/files/zarr.py
+++ b/dandi/files/zarr.py
@@ -743,10 +743,10 @@ class ZarrAsset(LocalDirectoryAsset[LocalZarrEntry]):
                     # Items to upload in this batch (may be retried e.g. due to
                     # 403 errors because of timed-out upload URLs)
                     all_items = list(items)
-                    large_items = [
+                    multipart_items = [
                         it for it in all_items if it.size > ZARR_LARGE_CHUNK_THRESHOLD
                     ]
-                    items_to_upload = [
+                    singlepart_items = [
                         it for it in all_items if it.size <= ZARR_LARGE_CHUNK_THRESHOLD
                     ]
                     max_retries = 5
@@ -756,7 +756,7 @@ class ZarrAsset(LocalDirectoryAsset[LocalZarrEntry]):
                         zcc.add_leaf(Path(it.entry_path), it.size, it.digest)
 
                     # Upload chunks above 5GB individually via multipart upload
-                    for it in large_items:
+                    for it in multipart_items:
                         # Yield uploading status
                         yield from _multipart_upload(
                             client=client,
@@ -780,9 +780,9 @@ class ZarrAsset(LocalDirectoryAsset[LocalZarrEntry]):
                         }
 
                     # Upload the remaining files using single part upload
-                    while items_to_upload and retry_count <= max_retries:
+                    while singlepart_items and retry_count <= max_retries:
                         # Prepare upload requests for current items
-                        uploading = [it.upload_request() for it in items_to_upload]
+                        uploading = [it.upload_request() for it in singlepart_items]
 
                         if retry_count == 0:
                             lgr.debug(
@@ -814,7 +814,7 @@ class ZarrAsset(LocalDirectoryAsset[LocalZarrEntry]):
                                     upload_url=signed_url,
                                     item=it,
                                 )
-                                for (signed_url, it) in zip(r, items_to_upload)
+                                for (signed_url, it) in zip(r, singlepart_items)
                             ]
 
                             changed = True
@@ -846,20 +846,20 @@ class ZarrAsset(LocalDirectoryAsset[LocalZarrEntry]):
                                 )
 
                             # Prepare for next iteration with retry items
-                            if items_to_upload := retry_items:
+                            if singlepart_items := retry_items:
                                 retry_count += 1
                                 if retry_count <= max_retries:
                                     lgr.info(
                                         "%s: %s got 403 errors, requesting new URLs",
                                         asset_path,
-                                        pluralize(len(items_to_upload), "file"),
+                                        pluralize(len(singlepart_items), "file"),
                                     )
                                     # Small delay before retry
                                     sleep(1 * retry_count)
 
                     # Check if we exhausted retries
-                    if items_to_upload:
-                        nfiles_str = pluralize(len(items_to_upload), "file")
+                    if singlepart_items:
+                        nfiles_str = pluralize(len(singlepart_items), "file")
                         raise UploadError(
                             f"{asset_path}: failed to upload {nfiles_str} "
                             f"after {max_retries} retries due to repeated 403 errors"
--- a/dandi/files/zarr.py
+++ b/dandi/files/zarr.py
@@ -1037,9 +1037,15 @@ class EntryUploadTracker:
     @staticmethod
     def _mkitem(e: LocalZarrEntry) -> UploadItem:
         # Avoid heavy import by importing within function:
-        from dandi.support.digests import md5file_nocache
+        from dandi.support.digests import md5file_nocache, multipart_md5file_nocache
 
-        digest = md5file_nocache(e.filepath)
+        file_size = e.filepath.stat().st_size
+        digest = (
+            md5file_nocache(e.filepath)
+            if file_size <= ZARR_LARGE_CHUNK_THRESHOLD
+            else multipart_md5file_nocache(e.filepath)
+        )
+
         return UploadItem.from_entry(e, digest)
 
     def get_items(self, jobs: int = 5) -> Generator[UploadItem, None, None]:
--- a/dandi/files/zarr.py
+++ b/dandi/files/zarr.py
@@ -749,6 +749,24 @@ class ZarrAsset(LocalDirectoryAsset[LocalZarrEntry]):
                     singlepart_items = [
                         it for it in all_items if it.size <= ZARR_LARGE_CHUNK_THRESHOLD
                     ]
+                    # TODO: remove once all servers are > 0.23.0 (i.e. ship
+                    # dandi-archive#2784) and the multipart zarr upload
+                    # endpoints are universally available; the capability
+                    # check would then be unnecessary.
+                    if multipart_items and not client.supports_zarr_multipart_upload:
+                        largest = max(multipart_items, key=lambda it: it.size)
+                        total_large_size = sum(it.size for it in multipart_items)
+                        raise UploadError(
+                            f"{asset_path}:"
+                            f" {pluralize(len(multipart_items), 'Zarr chunk')}"
+                            f" totaling {total_large_size / 1024**3:.2f} GiB"
+                            f" exceed the S3 single-part upload limit of"
+                            f" {ZARR_LARGE_CHUNK_THRESHOLD / 1024**3:.0f} GiB"
+                            f" (largest: {largest.entry_path},"
+                            f" {largest.size / 1024**3:.2f} GiB);"
+                            f" the server does not support multipart zarr"
+                            f" uploads (dandi-archive#2784)."
+                        )
                     max_retries = 5
                     retry_count = 0
                     # Add all items to checksum tree (only done once)
--- a/dandi/files/zarr.py
+++ b/dandi/files/zarr.py
@@ -10,10 +10,13 @@ from enum import Enum
 import json
 import os
 import os.path
+import re
 from pathlib import Path
+from threading import Lock
 from time import sleep
 from typing import Any, Optional
 import urllib.parse
+from xml.etree.ElementTree import fromstring
 
 from dandischema.models import BareAsset, DigestType
 from pydantic import BaseModel, ConfigDict, ValidationError
@@ -25,6 +28,7 @@ from dandi import get_logger
 from dandi.consts import (
     MAX_ZARR_DEPTH,
     ZARR_DELETE_BATCH_SIZE,
+    ZARR_LARGE_CHUNK_THRESHOLD,
     ZARR_MIME_TYPE,
     ZARR_UPLOAD_BATCH_SIZE,
 )
@@ -46,7 +50,7 @@ from dandi.utils import (
     pre_upload_size_check,
 )
 
-from .bases import LocalDirectoryAsset
+from .bases import LocalDirectoryAsset, _upload_blob_part
 from ..validate._types import (
     ORIGIN_VALIDATION_DANDI_ZARR,
     Origin,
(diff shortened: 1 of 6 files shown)


dandi/files/zarr.py
```diff
@@ -57,6 +57,7 @@
     ValidationResult,
     Validator,
 )
+from .bases import LocalDirectoryAsset, LocalFileAsset
```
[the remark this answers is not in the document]
jjnesbitt: For any humans reading this, the import cycle both already existed (`.bases` was already being imported), and this change is invalid anyway.

dandi/files/zarr.py
```diff
@@ -903,6 +934,115 @@ def _handle_failed_items_and_raise(
     raise failed_items[0][1]
 
 
+def upload_zarr_file_multipart(
+    item: UploadItem,
+    zarr_id: str,
+    dandiset: RemoteDandiset,
+    jobs: int | None = None,
+):
+    # Avoid heavy import by importing within function:
```
yarikoptic: how heavy is it for you?  filed a dedicated
- https://github.com/con/fscacher/issues/108
since may be from there?

dandi/files/zarr.py
```diff
@@ -903,6 +934,115 @@ def _handle_failed_items_and_raise(
...
+                "value": filetag,
+            },
+            "zarr": {
+                "chunk_key": item.entry_path,
+                "zarr_id": zarr_id,
+            },
+        },
+    )
+
+    try:
```
yarikoptic: I would need to look at it but I do not see why actual upload and etag comparison should anyhow be specific to zarrs, may be we need to extract some common helper to reuse if there is none yet

dandi/files/zarr.py
jjnesbitt: That comment is copy pasted from elsewhere

dandi/files/zarr.py
jjnesbitt: I agree, this is just a draft.

jjnesbitt: This has been refactored.

jjnesbitt:
@yarikoptic I believe this is in a good state. 

Regarding the test failures, it's worth noting that tests against the API are almost guaranteed to fail, as they blindly pull the `dandiarchive/dandiarchive-api` image from dockerhub, and the required changes to make this work are contained in https://github.com/dandi/dandi-archive/pull/2784. There is also a new release of RabbitMQ that is causing celery to not work correctly, which means that any tests that depend on the API updating the status of things behind the scenes via celery (a good amount) are broken. We're looking into it.

yarikoptic:
Thank you @jjnesbitt !  I meanwhile submitted 

- #1841 

to mitigate CI fails. Claude identified the same "cause" and provide some of its "aiplanation" but it could be slop, so beware.

yarikoptic:
@jjnesbitt FYI - I have rebased on now green master

dandi/tests/test_files.py
```diff
@@ -536,6 +536,38 @@ def test_upload_zarr_entry_content_type(new_dandiset, tmp_path):
...
 
+@pytest.mark.ai_generated
+def test_upload_zarr_large_chunks(new_dandiset, tmp_path):
+    """Chunks above ZARR_LARGE_CHUNK_THRESHOLD are uploaded via multipart upload."""
+    filepath = tmp_path / "example.zarr"
+    zarr.save(filepath, np.arange(1000), np.arange(1000, 0, -1))
+    zf = dandi_file(filepath)
+    assert isinstance(zf, ZarrAsset)
+
+    from ..files.bases import _multipart_upload as real_multipart_upload
```
yarikoptic: imports should go on top of the file, if there is a valid reason to stay inside function -- comment should be added to state why

dandi/files/zarr.py
```diff
@@ -742,13 +743,43 @@ def mkzarr() -> str:
                 ):
                     # Items to upload in this batch (may be retried e.g. due to
                     # 403 errors because of timed-out upload URLs)
-                    items_to_upload = list(items)
+                    all_items = list(items)
+                    large_items = [
+                        it for it in all_items if it.size > ZARR_LARGE_CHUNK_THRESHOLD
```
yarikoptic, line 750: the 5GB is really the upper threshold for single part upload, and IIRC AWS recommend much lower threshold on deciding to use multipart... need to review if possible to identify if server supports multipart in general, and thus start requesting multipart for e.g. `ZARR_LARGE_CHUNK_THRESHOLD/10` or so.  Also would require analysis on the effect of validation etc at dandi-archive side where we might then need to know/account which version was used to upload the zarr... (might be not easily feasible for existing ones which were at some point initiated to be uploaded by older client/API but could be feasible for new, if we annotate them somehow)

jjnesbitt:
@yarikoptic It would be good to get your review/approval on this PR. The only remaining test failure is the following:

```
FAILED dandi/tests/test_files.py::test_upload_zarr_large_chunks - dandi.exceptions.HTTP404Error: Error 404 while sending POST request to http://localhost:8000/api/uploads/zarr/initialize/: <!DOCTYPE html>
```

This is due to the fact that the version of the API that the tests are being run against doesn't contain the new URLs required for this change.

yarikoptic:
re tests:

> ```
> FAILED dandi/tests/test_files.py::test_upload_zarr_large_chunks - dandi.exceptions.HTTP404Error: Error 404 while sending POST request to http://localhost:8000/api/uploads/zarr/initialize/: <!DOCTYPE html>
> ```
> 
> This is due to the fact that the version of the API that the tests are being run against doesn't contain the new URLs required for this change.

So let's make such test(s) conditional and skip if ran against released/older version of the archive. Is there anything in the `/server-info` which would be indicative of this feature to be there?

jjnesbitt:
> So let's make such test(s) conditional and skip if ran against released/older version of the archive. Is there anything in the `/server-info` which would be indicative of this feature to be there?

We could do that, but it would achieve very little, since as soon as the feature is merged to master and the new image is built, they will immediately be re-enabled again. There's nothing at the `/server-info` endpoint at the moment to delineate the change.

Since the API change is backwards compatible, I suggest we simply merge the API change once it's ready (hopefully soon), at which point the tests will use the new image, and things are back to normal.

yarikoptic:
re server-capability conditioning -- and then client ideally be aware of that too at run time (not only in the tests) so it could just refuse/error if trying to upload 5TB files as I have implemented awhile back in https://github.com/dandi/dandi-cli/pull/1827 and will be merging shortly, instead of hitting a non-existing end point at EMBER and LINC (if they keep it afloat).  I will leave it to you to decide on how client could discover server capability here:

- could be just version based if you are positive that next minor would have it
- I thought I filed an issue proposing to add some 'features: [ ]' listing to `/api/info` but I fail to find 
- could sense if there is an expected endpoint (once per instance) and if not - assume absent feature.

yarikoptic:
Sorry @jjnesbitt , I will not be merging a PR with totally Red CI, it is out of question.

yarikoptic:
ideally it should actually test that server works with at least some TEMP commit which would somehow provision container with that new mighty version of the server from the PR.  But I guess that is for the future, to do smth like I did for bids-examples where in PR description we can point to perspective PRs for bids-specification and bids-validator to test against: https://github.com/bids-standard/bids-examples/blob/master/.github/workflows/validate_datasets.yml#L35

jjnesbitt:
> Sorry @jjnesbitt , I will not be merging a PR with totally Red CI, it is out of question.

That is not what I was suggesting. Once the API PR is merged, this CI will become green, because it will be using the latest image of `dandiarchive/dandiarchive-api`.

yarikoptic:
ok, let's wait then.

yarikoptic (changes requested):
some minor ones about imports but also a question to clarify on why not to use smaller threshold

dandi/files/zarr.py
jjnesbitt: Indeed it seems AWS recommends 100MB as the threshold https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html

I can update the threshold const.

yarikoptic: as I alluded in my original comment we might then need to analyze on how it would affect zarr checksumming etc, so it isn't trivial.  Even now, I haven't yet checked on how we would compute checksum: before we could just always use md5sum per local file and it would correspond to ETag per key on S3, with multipart - we would now use dandi-etag .  

- what happens with already uploaded zarrs with single part upload, finalized or not

thus, making threashold of 5GB is "easy solution" but even for that we might need to adjust https://pypi.org/project/zarr-checksum/, or am I overcomplicating, and it isn't an issue?

jjnesbitt: Indeed, uploading via multipart will result in different checksums than single part. I'm working on a commit to address this.

jjnesbitt: I've addressed this issue in [`4c36712`](https://github.com/dandi/dandi-cli/pull/1839/commits/4c3671268cbc6534772f3ed82003fff51153b431), and the integration tests (previously failing due to the checksum mismatch) now pass https://github.com/dandi/dandi-archive/actions/runs/25011647320/job/73435958213

jjnesbitt:
To avoid complicated conditioning of this feature based on the archive's capability (for which they may quickly upgrade anyway), we could do the following:

1. Merge and release this change
2. In the prod instance, set [`cli-minimal-version`](https://github.com/dandi/dandi-archive/blob/c696bce61d5f08ec6bc44e5d7d221ea739d7e281/dandiapi/api/views/info.py#L88) to the newly released version
3. In the downstream archives, add the newly released version to [`cli-bad-versions`](https://github.com/dandi/dandi-archive/blob/c696bce61d5f08ec6bc44e5d7d221ea739d7e281/dandiapi/api/views/info.py#L90), so that anyone with a CLI version that is newer than what is supported gets a CLI error.

This removes any conditional logic based on archive capabilities in the CLI (which should be discouraged anyway) and puts the compatibility check solely on the archive side. Once LINC/EMBER merge in the supported change, the CLI version can be removed from `cli-bad-versions`, and things are back to normal.

@yarikoptic Do you have objections to this?

dandi/support/digests.py
```diff
@@ -137,6 +137,30 @@ def md5file_nocache(filepath: str | Path) -> str:
...
+    part_md5s = b""
+    num_parts = 0
+    with open(filepath, "rb") as f:
+        while True:
+            chunk = f.read(part_size)
+            if not chunk:
+                break
+            part_md5s += hashlib.md5(chunk).digest()
+            num_parts += 1
+    return f"{hashlib.md5(part_md5s).hexdigest()}-{num_parts}"
```
yarikoptic: shouldn't this be the same as the `get_dandietag`?  or multipart upload of zarr chunks should follow some different partitioning etc?

yarikoptic:
yes, I have objections,  as 
- I think conditioning should not be complicated, let me know if I should take stab at that,
- again -- it seems would require red CI to be merged /released as current CI is red (didn't check details)?
- we do have other instances running so I do not want to break new client release functioning with them ...

But before we jump any ways, I think we still need to elaborate on overall details of the change in behaviors! Thanks for quickly addressing the question of diverging checksum/etag'ing -- I left a comment there as I think multipart upload etag compute must be consistent with how we compute for regular blobs, shouldn'it?

And I (at least) still do not have a complete picture how it 

- would all be consistent with existing zarrs in the archive (already having >100MB single part uploads),
- zarr-checksum library behavior (does not do any multipart splitting, and we do not want different checksums by the library and uploader/archive ) etc.   

Potential "resolutions" to consider:

- switch back to 5GB limit for deciding going multipart with zarr blobs -- even though [I was the one who asked/reminded](https://github.com/dandi/dandi-cli/pull/1839/changes#r3142276238) about AWS recommendation on when to switch to multipart uploads.
   - pros: This would allow for existing zarrs with single part uploads of large files (which succeeded) to be ok; finalizing upload of huge parts via multipart, if we fixup zarr-checksum library to behave similarly etc - we have consistent state
   - cons: single part uploads are more fragile etc
- switch to this new logic of 100MB limit as now to go multipart for zarr
  - pros: consistent with blobs
  - cons: need to be figured out how we make it consistent with library etc . e.g.
     -  it could be some new form of checksum, (e.g. `zarr-checksum-v2`, or `zarr-checksum-mp100` where `mp100` signals on multipart past 100M or `zarr-dandietag-checksum` to signal that individual blobs follow dandietag) and thus we just improve digester to understand both and then existing ones with old `zarr-checksum` and then for new uploads `zarr-checksum-v2`.  will require zarr-checksum library to get support for both I guess
     - ... ???...

or, once again my question -- am I overcomplicating?

dandi/support/digests.py
jjnesbitt: I've pushed a commit to address this. Now instead the multi-part behavior is folded into the `md5file_nocache` function, as that function is used several places in the zarr upload lifecycle.

jjnesbitt:
> * again -- it seems would require red CI to be merged /released as current CI is red (didn't check details)?

And I'll again reiterate, the archive PR is not yet merged, which is why the CI is red. Once that happens, it will be green.

> * we do have other instances running so I do not want to break new client release functioning with them ...

What other clients besides LINC/EMBER? Why is simply excluding the new CLI version not sufficient to support these archives?

> * I think conditioning should not be complicated, let me know if I should take stab at that,

The CLI code for zarr upload is extremely convoluted as it is, adding this conditional logic further exacerbates that. However, if you're really determined to implement this conditional logic then I'll leave it to you. 

> But before we jump any ways, I think we still need to elaborate on overall details of the change in behaviors! Thanks for quickly addressing the question of diverging checksum/etag'ing -- I left a comment there as I think multipart upload etag compute must be consistent with how we compute for regular blobs, shouldn'it?
> 
> And I (at least) still do not have a complete picture how it
> 
> * would all be consistent with existing zarrs in the archive (already having >100MB single part uploads),
> * zarr-checksum library behavior (does not do any multipart splitting, and we do not want different checksums by the library and uploader/archive ) etc.

The zarr-checksum library doesn't need to know specifics of how the digests are computed. As it stands now, we simply pass it the items containing the path and digest, so we control what digests it uses. There should be no issue here.

> Potential "resolutions" to consider:
> 
> * switch back to 5GB limit for deciding going multipart with zarr blobs -- even though [I was the one who asked/reminded](https://github.com/dandi/dandi-cli/pull/1839/changes#r3142276238) about AWS recommendation on when to switch to multipart uploads.

I think short term this is probably the best solution, as it immediately unblocks people trying to upload zarrs with parts over this limit. The underlying issue should still be re-addressed, but we can hold off until there's not a funding gap looming.

dandi/support/digests.py
```diff
@@ -133,7 +134,16 @@ def md5file_nocache(filepath: str | Path) -> str:
...
     present in Zarrs
+
+    If the file is larger than `ZARR_LARGE_CHUNK_THRESHOLD`, the computed checksum is not a
+    traditional md5 checksum, but is instead an S3 multipart ETag.
     """
+    if isinstance(filepath, str):
+        filepath = Path(filepath)
+
+    if filepath.stat().st_size > ZARR_LARGE_CHUNK_THRESHOLD:
+        return get_dandietag(filepath).as_str()
```
jjnesbitt: I cannot see any reason why the type checker is failing here. `DandiETag.as_str` is clearly annotated as returning a `str`.

yarikoptic:
> > * again -- it seems would require red CI to be merged /released as current CI is red (didn't check details)?
> 
> And I'll again reiterate, the archive PR is not yet merged, which is why the CI is red. Once that happens, it will be green.

great -- when it gets green we can merge!  I feel like in a groundhog day ;) I guess we are missing each other's point on this one.

> > * we do have other instances running so I do not want to break new client release functioning with them ...
> 
> What other clients besides LINC/EMBER? Why is simply excluding the new CLI version not sufficient to support these archives?

because it would require THEM to exclude, whenever WE introduce breakage, so I would prefer US to avoid breakage that rely on clients to mitigate, and then remitigate whenever we update or release yet another version.

Long term -- it is client/server's couple job imho to agree on acceptable interfaces to be used.

> > * I think conditioning should not be complicated, let me know if I should take stab at that,
> 
> The CLI code for zarr upload is extremely convoluted as it is, adding this conditional logic further exacerbates that. However, if you're really determined to implement this conditional logic then I'll leave it to you.

gotcha, I will look into that... tomorrow morning. 

My question would only remain of absent testing against WiP archive I guess.  did you have chance to test it against some instance?  what was its `/info` output -- I wonder what would be the best indicator if instance supports this -- probing new interface?

> > But before we jump any ways, I think we still need to elaborate on overall details of the change in behaviors! Thanks for quickly addressing the question of diverging checksum/etag'ing -- I left a comment there as I think multipart upload etag compute must be consistent with how we compute for regular blobs, shouldn'it?
> > And I (at least) still do not have a complete picture how it
> > 
> > * would all be consistent with existing zarrs in the archive (already having >100MB single part uploads),
> > * zarr-checksum library behavior (does not do any multipart splitting, and we do not want different checksums by the library and uploader/archive ) etc.
> 
> The zarr-checksum library doesn't need to know specifics of how the digests are computed. As it stands now, we simply pass it the items containing the path and digest, so we control what digests it uses. There should be no issue here.

doesn't it provide compute across filetree?:
https://github.com/dandi/zarr_checksum/blob/73bfdf56e822065623b282049e7c70783d87072c/zarr_checksum/generators.py#L108


> > Potential "resolutions" to consider:
> > 
> > * switch back to 5GB limit for deciding going multipart with zarr blobs -- even though [I was the one who asked/reminded](https://github.com/dandi/dandi-cli/pull/1839/changes#r3142276238) about AWS recommendation on when to switch to multipart uploads.
> 
> I think short term this is probably the best solution, as it immediately unblocks people trying to upload zarrs with parts over this limit. The underlying issue should still be re-addressed, but we can hold off until there's not a funding gap looming.

agreeed! let's stick to 5GB limit as decision point for now

jjnesbitt:
> My question would only remain of absent testing against WiP archive I guess. did you have chance to test it against some instance? what was its `/info` output -- I wonder what would be the best indicator if instance supports this -- probing new interface?

I've done this both through running the integration tests with this branch of the CLI code ([successful run](https://github.com/dandi/dandi-archive/actions/runs/25011647320/job/73435958213)), as well as running the corresponding branch of the API locally and pointing the tests at that, instead of utilizing the docker-compose pytest fixture.

As far as what test can be done to check for server compatibility, the existence of the `/uploads/zarr/initialize/` endpoint is the simplest and easiest answer.

> doesn't it provide compute across filetree?: [dandi/zarr_checksum@`73bfdf5`/zarr_checksum/generators.py#L108](https://github.com/dandi/zarr_checksum/blob/73bfdf56e822065623b282049e7c70783d87072c/zarr_checksum/generators.py#L108)

What you've linked is the one of the "default" file generators included in the `zarr-checksum` package. We do not use that generator, and instead [directly](https://github.com/dandi/dandi-cli/blob/96bf2dfb2a88fb6f07a89b9b6842b48c51afb3bf/dandi/support/digests.py#L125-L128) provide the files and digests used to compute the zarr checksum.

> agreeed! let's stick to 5GB limit as decision point for now

I've switched this back.

yarikoptic:
Sent 
- https://github.com/dandi/dandi-cli/pull/1846

conditioning part is quite minimal, as I expected.

jjnesbitt:
Superseded by #1925

PR #1846, opened 2026-04-30
Zarr multipart upload tune ups: condition on server capability, fix progress reporting, harmonize variable names

yarikoptic:
Individual commits would have more of 'claudetail' ;-)

Target:
- #1839

### DIFF
--- a/dandi/files/zarr.py
+++ b/dandi/files/zarr.py
@@ -743,10 +743,10 @@ class ZarrAsset(LocalDirectoryAsset[LocalZarrEntry]):
                     # Items to upload in this batch (may be retried e.g. due to
                     # 403 errors because of timed-out upload URLs)
                     all_items = list(items)
-                    large_items = [
+                    multipart_items = [
                         it for it in all_items if it.size > ZARR_LARGE_CHUNK_THRESHOLD
                     ]
-                    items_to_upload = [
+                    singlepart_items = [
                         it for it in all_items if it.size <= ZARR_LARGE_CHUNK_THRESHOLD
                     ]
                     max_retries = 5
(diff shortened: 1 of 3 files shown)


dandi/files/zarr.py
```diff
@@ -743,22 +743,44 @@ def mkzarr() -> str:
                     # Items to upload in this batch (may be retried e.g. due to
                     # 403 errors because of timed-out upload URLs)
                     all_items = list(items)
-                    large_items = [
+                    multipart_items = [
                         it for it in all_items if it.size > ZARR_LARGE_CHUNK_THRESHOLD
                     ]
-                    items_to_upload = [
+                    singlepart_items = [
```
yarikoptic, line 749: without name changes code would be hard(er) to read going forward. Cleaner names at least make more sense IMHO for future ourselves

dandi/files/zarr.py
```diff
@@ -743,22 +743,44 @@ def mkzarr() -> str:
...
-                        # Yield uploading status
-                        yield from _multipart_upload(
+                    # Upload chunks above 5GB individually via multipart upload.
+                    # ``_multipart_upload`` reports ``current`` as bytes within
+                    # the single chunk being uploaded; translate it to bytes
+                    # uploaded across the whole zarr so progress reporting
+                    # stays monotonic for downstream consumers.
+                    for it in multipart_items:
+                        cumulative_before = bytes_uploaded
+                        for status in _multipart_upload(
```
yarikoptic, line 783: potentially we could change what record we yield in that helper to make it easier here, e.g. could provide "bytes_uploaded" to  be provided inside to be taken into account for progress reporting or alike -- I will leave to you to choose

dandi/tests/test_files.py
```diff
@@ -30,6 +29,7 @@
     dandi_file,
     find_dandi_files,
 )
+from ..files.bases import _multipart_upload as real_multipart_upload
```
yarikoptic, line 32: did you have pre-commit installed for your local clone of dandi-cli @jjnesbitt ?

jjnesbitt (approved):
CI is green by simply not running the new tests, why didn't I think of that 👍

yarikoptic:
> CI is green by simply not running the new tests, why didn't I think of that 👍

;) but it might turn red on your PR in dandi-archive if you test against that instance ... as I noted before -- ideally we should add support for such matrix run to point to PRs in dandi-archive (#1847?) and aim for both stock run and dandi-archive PR run to stay green!

PR #1847, opened 2026-04-30
Allow test server to be provided externally

brianhelba:
This adds support for an optional `DANDI_TESTS_API_URL` environment variable, which if set, will cause the integration test suite to connect to that external server and not boot up a Docker Compose stack. If `DANDI_TESTS_API_URL` is set, then `DANDI_TESTS_DJANGO_API_KEY` must also be set.

This makes it possible to run the CLI integration tests more without having to build a Docker image of the API server, allowing simpler validation of any existing running server against the CLI.

### DIFF
--- a/dandi/tests/fixtures.py
+++ b/dandi/tests/fixtures.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from collections.abc import Callable, Iterator
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace
 from datetime import datetime, timezone
 import logging
 import os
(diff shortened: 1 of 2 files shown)


dandi/tests/fixtures.py
```diff
@@ -373,6 +373,20 @@ def _make_subdirs_dandisets(path: Path) -> None:
 @pytest.fixture(scope="session")
 def docker_compose_setup() -> Iterator[dict[str, str]]:
     skipif.no_network()
+
+    if external_url := os.environ.get("DANDI_TESTS_API_URL"):
+        api_key = os.environ.get("DANDI_TESTS_DJANGO_API_KEY")
+        if not api_key:
```
yarikoptic, line 379: we are in the bright future already ;) could be 

```suggestion
        if not (api_key := os.environ.get("DANDI_TESTS_DJANGO_API_KEY")):
```

but just to expedite -- let's proceed

yarikoptic (changes requested):
actually -- we would need to list them in https://github.com/dandi/dandi-cli/blob/master/DEVELOPMENT.md#environment-variables so human and silicon beasts could discover this feature more easily

yarikoptic:
@claude if you are enabled here -- add minimalistic description of added here env vars into DEVELOPMENT.md

yarikoptic (approved):
Added docs, we should be all set now if no surprises from CI

brianhelba:
Thanks for the quick review and for fixing the deficiencies yourself! I appreciate it.

PR #1848, opened 2026-04-30
Add support for updated error text upon Zarr name conflicts

brianhelba:
Newer versions of the DANDI API will return a different error text when a duplicate Zarr (same name, same dandiset) is created. This is due to more rigorous early checking by Django Rest Framework and cannot easily be changed.

This ensures that any version of the API is supported, as the old error text is still matched too.


brianhelba:
This is necessary to address integration test failures in https://github.com/dandi/dandi-archive/pull/2589.

yarikoptic (approved):

yarikoptic:
will release with some subsequent PR like 

- #1847 

to debounce releases a little ;)

PR #1849, opened 2026-04-30
BF: Add timeout to follow_redirect HEAD requests; retry on Timeout

yarikoptic:
Original motivator from me:  we had windows keep failing here and there with getting stuck on redirects in test_follow_redirect  and potentially others.  So it made sense to add explicit timeout and retry on it. 

claude summary of agreed upon changes:

`follow_redirect()` calls `requests.head(url, allow_redirects=True)` to resolve a redirect chain.  No timeout was passed, so a stalled hop (e.g. a third-party shortener like bit.ly that accepts the TCP/TLS handshake but never sends a response) blocks the call indefinitely.  This has been observed in `test_follow_redirect` on the Windows CI runner — pytest's fault-handler stack ends in `socket.recv_into() -> ssl.read()` inside `requests.sessions.resolve_redirects`.

The retry wrapper around `requests.head()` only caught `requests.ConnectionError`, which does not include `requests.ReadTimeout` — so even adding a timeout would have surfaced as a hard failure rather than the existing transient-error retry path.

Changes:
- Add `REDIRECT_HEAD_TIMEOUT = 30` constant in `dandi/consts.py`.
- Pass it as `timeout=` to the `requests.head()` call.
- Catch `requests.Timeout` in addition to `requests.ConnectionError` inside the retry loop, so `ReadTimeout` and `ConnectTimeout` are now retried with the same exponential backoff.
- Widen the `connection_error` annotation to `requests.RequestException | None` to accommodate both exception types.
- Add `test_follow_redirect_retry_on_timeout` parametrized over `ReadTimeout` and `ConnectTimeout`, mirroring the existing `test_follow_redirect_retry_on_connection_error`.
- Update mock signatures in the existing tests to accept the new `timeout=` kwarg.

### DIFF
--- a/dandi/dandiarchive.py
+++ b/dandi/dandiarchive.py
@@ -44,6 +44,7 @@ from . import get_logger
 from .consts import (
     DANDISET_ID_REGEX,
     PUBLISHED_VERSION_REGEX,
+    REDIRECT_HEAD_TIMEOUT,
     RETRY_STATUSES,
     VERSION_REGEX,
     DandiInstance,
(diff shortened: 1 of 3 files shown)


dandi/dandiarchive.py
```diff
@@ -896,12 +897,18 @@ def follow_redirect(url: str) -> str:
...
+        # CI when a hop in the redirect chain accepts the connection but never
+        # responds.  Without this, requests.head() blocks indefinitely.
+        connection_error: requests.RequestException | None = None
         resp: requests.Response | None = None
         for attempt_idx in range(max_attempts):
             try:
-                resp = requests.head(url, allow_redirects=True)
-            except requests.ConnectionError as e:
+                resp = requests.head(
+                    url, allow_redirects=True, timeout=REDIRECT_HEAD_TIMEOUT
```
CodyCBakerPhD, line 909: Yes, I often use this myself when using the `requests` package - it is quite nice

CodyCBakerPhD (approved):

PR #1865, opened 2026-05-28
enh: add Macaca radiata (bonnet macaque) to species_map

yarikoptic:
Adds NCBI taxon 9548 (Macaca radiata, common name "bonnet macaque") to the species lookup table so the species can be recognized by common name, "radiata", or its NCBITaxon URL.

Requested by a user via email.

### DIFF
--- a/dandi/metadata/util.py
+++ b/dandi/metadata/util.py
@@ -395,6 +395,12 @@ species_map = [
         "Macaca nemestrina",
     ),
     (
+        ["bonnet macaque", "bonnet monkey", "radiata"],
+        None,
+        NCBITAXON_URI_TEMPLATE.format("9548"),
+        "Macaca radiata - Bonnet macaque",
+    ),
+    (
         ["mongolian gerbil", "mongolian jird"],
         None,
         NCBITAXON_URI_TEMPLATE.format("10047"),


dandi/metadata/util.py
```diff
@@ -394,6 +394,12 @@ def extract_cellLine(metadata: dict) -> str | None:
         NCBITAXON_URI_TEMPLATE.format("9545"),
         "Macaca nemestrina",
     ),
+    (
+        ["bonnet macaque", "bonnet monkey", "radiata"],
+        None,
+        NCBITAXON_URI_TEMPLATE.format("9548"),
+        "Macaca radiata - Bonnet macaque",
```
yarikoptic, line 401: I think this version is better than just 

```suggestion
        "Macaca radiata",
```

as we did for "Macaca nemestrina" above . I think we should expand all others into similar 

`{current name} - {Genbank common name}` form which seems they all follow.

@copilot -- prepare a PR where all other entries follow the same format for their "name" and also expand description above on the expected format of the "name" field

yarikoptic:
unfortunately no further reviews but also no objections, so let's proceed with this

yarikoptic-gitmate (approved):

PR #1871, opened 2026-06-02
Update spike time calculations for scattered non-spiking units

CodyCBakerPhD:
Update session duration spike-time boundary handling to correctly support multiple non-spiking units scattered throughout a Units table, not just when the first unit is non-spiking.

The implementation now keeps only unit boundaries where cumulative spike count increases, which avoids invalid indexing from repeated boundaries contributed by non-spiking rows while preserving correct min/max spike-time extraction.

Also adds a regression test covering interleaved spiking and non-spiking unit rows to validate the expected session duration behavior.

### DIFF
--- a/dandi/pynwb_utils.py
+++ b/dandi/pynwb_utils.py
@@ -339,19 +339,21 @@ def _get_session_duration(nwb: pynwb.NWBFile) -> float | None:
             # Read only the first and last spike time from each unit
             if "spike_times" in obj.colnames and len(obj["spike_times"]):
                 idxs = obj["spike_times"].data[:]
-                spiking_idxs = np.unique(idxs[idxs > 0])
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
@copilot Please add one new test for this case (of having multiple non-spiking units scattered throughout the rows of a untis table)

CodyCBakerPhD:
@yarikoptic I recommend merging #1872 first

CodyCBakerPhD:
@yarikoptic OK ready - a user needed this to upload some data

CodyCBakerPhD:
@yarikoptic ping again

yarikoptic (approved):
coolio, thanks!   filed

- https://github.com/dandi/dandi-cli/issues/1877

for longer term.  IMHO we seems to be getting too much "analytical" here ;)

Issue #1873, opened 2026-06-02
Tests optional install should be a dependency group instead


CodyCBakerPhD:
Noticed this in https://github.com/dandi/dandi-cli/blob/master/pyproject.toml#L102

Read more about this convention: https://peps.python.org/pep-0735/

@candleindark knows how to do this

candleindark:
Agree. I would say that `style` should be a dependency-group instead as well.

candleindark:
@CodyCBakerPhD Pushing back on migrating the **`test`** extra specifically (I still agree `style` should become a dependency group).

Unlike most packages, `dandi` ships its tests inside the distribution, and the `test` extra provides the dependencies to run those shipped tests, so they're truly an extra, in the sense that they're needed by the distributed package. Dependency groups are deliberately *not* published in built distributions, so migrating `test` would remove the ability to install the test dependencies from the distributed package, e.g., `pip install --group test dandi` doesn't work without a local `pyproject.toml`.

Concretely, this would break four downstream uses with no simple flag-level fix; making them work again would require cloning `dandi-cli` into the test environment, which defeats the purpose of the cases that test against the released `dandi` from PyPI.

<details>
<summary>The four downstream uses of <code>dandi[test]</code></summary>

dandi-schema (`test-dandi-cli.yml`):
- [`pip install "dandi[test,extras] @ git+…/dandi-cli.git"`](https://github.com/dandi/dandi-schema/blob/d75273828bb7742be74f23d9f8cf8a941f4af8e2/.github/workflows/test-dandi-cli.yml#L73) (master)
- [`pip install "dandi[test,extras]"`](https://github.com/dandi/dandi-schema/blob/d75273828bb7742be74f23d9f8cf8a941f4af8e2/.github/workflows/test-dandi-cli.yml#L77) (released)

dandi-archive (`cli-integration.yml`):
- [`uvx --with dandi[test] …`](https://github.com/dandi/dandi-archive/blob/97cb450a53068da56a0c440f785b8f464226a4bc/.github/workflows/cli-integration.yml#L70) (released)
- [`uvx --with "dandi[test] @ git+…/dandi-cli" …`](https://github.com/dandi/dandi-archive/blob/97cb450a53068da56a0c440f785b8f464226a4bc/.github/workflows/cli-integration.yml#L78) (master)

The two `released` ones fetch `dandi` from PyPI with no local checkout by design.

</details>

[...]

PR #1875, opened 2026-06-04
Migrate style extra to a PEP 735 dependency group

candleindark:
Migrate the `style` extra (`flake8`, `pre-commit`) to a PEP 735 `[dependency-groups]` table, since those are pure development tooling never installed by consumers of the published package. Also trims the hatch environments down to the ones actually used.

Closes the `style` part of #1873. The `test` extra is intentionally **not** migrated (it backs the shipped test suite that downstream repos run against released `dandi`, see [this comment](https://github.com/dandi/dandi-cli/issues/1873#issuecomment-4609935064)).

### Changes

Commit 1, migrate `style`:
- Move `style` out of `[project.optional-dependencies]` into `[dependency-groups]`.
- Drop `dandi[style]` from the `all` extra (extras cannot reference dependency groups).
- Remove `[tool.hatch.envs.style]` (the style deps were never used in an isolated env; lint runs via `tox -e lint`).
- Update the env-list example in `DEVELOPMENT.md`.

Commit 2, consolidate hatch envs:
- Remove the thin per-extra envs `extensions`, `extras`, `tools`, `all`.
- Keep `default`, `test`, and add `dev-all` (`features = ["all"]` + `dependency-groups = ["style"]`): one environment with every dependency for local development.

<details>
<summary>Verification that the <code>style</code> and <code>all</code> extras have no consumers</summary>

Searched this repo, every `dandi-*` sibling repo on `master`, and a workspace-wide literal sweep for `dandi[style]` / `dandi[all]`: the only reference anywhere was dandi-cli's own `pyproject.toml`. So dropping them breaks nothing downstream.

An isolated environment for an individual extra is still one command away when needed: `hatch shell`, then `uv pip install -e ".[<extra>]"`.

</details>

### Test plan

- [x] `hatch env show --json` lists user envs `default`, `test`, `dev-all`; `dev-all` resolves to `features: ['all']` + `dependency-groups: ['style']`.
- [x] `pyproject.toml` parses; `style` is gone from extras and present in `[dependency-groups]`; `all` extra no longer lists `dandi[style]`.
- [x] CI lint/typing/test jobs pass (they use tox, unaffected by these changes).

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -7,7 +7,7 @@
 This project's hatch environments use [uv](https://github.com/astral-sh/uv)
 as the installer, which allows for significant improvements in environment
 setup speed. Each environment corresponds to a project
-extra (e.g., `test`, `style`, `tools`). To enter a shell in an environment:
+extra (e.g., `test`, `extras`, `tools`). To enter a shell in an environment:
 
 ```
 hatch shell <env-name>
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -240,12 +240,6 @@ init_forbid_extra = true
 installer = "uv"
 python = "3.10"
 
-[tool.hatch.envs.extensions]
-features = ["extensions"]
-
-[tool.hatch.envs.extras]
-features = ["extras"]
-
 [tool.hatch.envs.test]
 features = ["test"]
 [tool.hatch.envs.test.env-vars]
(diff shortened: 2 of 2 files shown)


yarikoptic (approved):
I kinda see why it might need to be different but imho it is a bit of overengineering as only introduces yet another way to install components necessary for my work on it

candleindark:
> I kinda see why it might need to be different but imho it is a bit of overengineering as only introduces yet another way to install components necessary for my work on it

It seems to be over-engineering at this point since there is only one dependency group. It will not be so once there are more groups of dependencies define for development usage. I proposed https://github.com/dandi/dandi-cli/issues/1876 to consolidate more group of dependencies used for development.

PR #1881, opened 2026-06-23
Only emit "deleting extra files" status when cleanup is needed

CodyCBakerPhD:
## Summary
Optimize the Zarr download cleanup logic to avoid emitting a misleading "deleting extra files" status message on fresh downloads where no orphaned files exist.

## Changes
- **Two-pass cleanup scan**: Split the Zarr cleanup into a preliminary scan phase that checks for orphaned files without modifying anything, followed by actual deletion only if needed
- **Conditional status emission**: Only emit the "deleting extra files" status when the preliminary scan detects actual orphaned files or empty directories to remove
- **Added debug logging**: Include debug-level logging when deleting extra files and removing empty directories for better troubleshooting
- **New tests**: Added two test cases marked with `@pytest.mark.ai_generated`:
  - `test_download_zarr_clean_no_deleting_status`: Verifies that fresh downloads don't emit the deletion status
  - `test_download_zarr_deletes_orphan_files`: Verifies that orphaned files are properly cleaned up on re-download with the status correctly emitted

## Implementation Details
The cleanup logic now:
1. First scans the Zarr directory tree to detect if any files/directories need removal
2. Only yields the "deleting extra files" status if the scan found orphaned content
3. Performs the actual deletion in a second pass if needed
4. Maintains the same cleanup behavior while avoiding false status messages on clean downloads

https://claude.ai/code/session_01LFHXuWAevoNHvEgw3QhGtU

### DIFF
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -1058,39 +1058,39 @@ def _download_zarr(
     empty_dirs: deque[Path] = deque()
     dirs: deque[Path] = deque([zarr_basepath])
     while dirs:
-        dir = dirs.popleft()
+        d = dirs.popleft()
         is_empty = True
-        for path in list(dir.iterdir()):
-            if exclude_from_zarr(path):
+        for p in list(d.iterdir()):
+            if exclude_from_zarr(p):
                 is_empty = False
             elif (
-                path.is_file()
-                and path.relative_to(zarr_basepath).as_posix() not in remote_paths
+                p.is_file()
+                and p.relative_to(zarr_basepath).as_posix() not in remote_paths
             ):
                 if not announced:
                     announced = True
                     yield {"status": "deleting extra files"}
                 try:
-                    lgr.debug("Deleting extra Zarr file %s", path)
-                    path.unlink()
+                    lgr.debug("Deleting extra Zarr file %s", p)
+                    p.unlink()
                 except OSError:
                     is_empty = False
-            elif path.is_dir():
-                dirs.append(path)
+            elif p.is_dir():
+                dirs.append(p)
                 is_empty = False
             else:
                 is_empty = False
-        if is_empty and dir != zarr_basepath:
-            empty_dirs.append(dir)
+        if is_empty and d != zarr_basepath:
+            empty_dirs.append(d)
     while empty_dirs:
-        dir = empty_dirs.popleft()
+        d = empty_dirs.popleft()
         if not announced:
             announced = True
             yield {"status": "deleting extra files"}
-        lgr.debug("Removing now-empty Zarr directory %s", dir)
-        dir.rmdir()
-        if dir.parent != zarr_basepath and not any(dir.parent.iterdir()):
-            empty_dirs.append(dir.parent)
+        lgr.debug("Removing now-empty Zarr directory %s", d)
+        d.rmdir()
+        if d.parent != zarr_basepath and not any(d.parent.iterdir()):
+            empty_dirs.append(d.parent)
 
     if "skipped" not in final_out["message"]:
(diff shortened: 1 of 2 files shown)


dandi/download.py
```diff
@@ -1066,22 +1071,48 @@ def downloads_gen():
...
+                        p.unlink()
+                    except OSError:
+                        is_empty = False
+                elif p.is_dir():
+                    dirs.append(p)
+                    is_empty = False
+                else:
+                    is_empty = False
+            if is_empty and d != zarr_basepath:
+                empty_dirs.append(d)
```
CodyCBakerPhD:
```suggestion
                empty_dirs.append(d)

```

dandi/download.py
```diff
@@ -1051,37 +1051,64 @@ def downloads_gen():
...
+                        path.unlink()
+                    except OSError:
+                        is_empty = False
+                elif path.is_dir():
+                    dirs.append(path)
+                    is_empty = False
+                else:
+                    is_empty = False
+            if is_empty and dir != zarr_basepath:
+                empty_dirs.append(dir)
```
CodyCBakerPhD:
```suggestion
                empty_dirs.append(dir)

```

dandi/download.py
```diff
@@ -1051,37 +1051,65 @@ def downloads_gen():
         else:
             return
 
-    yield {"status": "deleting extra files"}
     remote_paths = set(map(str, entries))
     zarr_basepath = Path(download_path)
+
+    # Scan the tree (without modifying it) to see if any stale files or now-empty directories need removing, so a clean download doesn't misleadingly announce "deleting extra files".
```
CodyCBakerPhD:
```suggestion
    # Scan the tree to see if any stale files or now-empty directories need removing
    # So a clean download doesn't misleadingly announce "deleting extra files"
```

yarikoptic:
> * First scans the Zarr directory tree to detect if any files/directories need removal
> * Only yields the "deleting extra files" status if the scan found orphaned content

I didn't check original code in detail but would 2pass be really needed, since it sounds like a potential additional burden... e.g. could it be just a matter of a flag like `yield_delete_files_msg: bool` and adding a conditional to yield on first delete thus while keeping original 1 pass implementation?

dandi/download.py
```diff
@@ -1051,37 +1051,49 @@ def downloads_gen():
...
             return
 
-    yield {"status": "deleting extra files"}
     remote_paths = set(map(str, entries))
     zarr_basepath = Path(download_path)
-    dirs = deque([zarr_basepath])
+
+    # Walk the tree, removing stale files (and the directories they leave empty).
+    # The "deleting extra files" status is yielded lazily on the first removal so a
+    # clean download doesn't misleadingly announce it when there's nothing to delete.
```
CodyCBakerPhD:
```suggestion
```

dandi/download.py
```diff
@@ -1051,37 +1051,49 @@ def downloads_gen():
...
+    announced = False
```
CodyCBakerPhD:
```suggestion
    announced: bool = False
```

CodyCBakerPhD:
> I didn't check original code in detail but would 2pass be really needed, since it sounds like a potential additional burden... e.g. could it be just a matter of a flag like yield_delete_files_msg: bool and adding a conditional to yield on first delete thus while keeping original 1 pass implementation?

Restored 1-pass implementation

CodyCBakerPhD:
The two dev-deps failures (also occuring on daily tests) are fixed in https://github.com/dandi/dandi-cli/pull/1882

yarikoptic (approved):

PR #1883, opened 2026-06-26
Support click >= 8.2.0

### LINKED ISSUE #1631: Support click >= 8.2.0
refs:
 
- https://github.com/pallets/click/issues/2911
- #1630 limits to before 8.2.0 for now

rly:
Closes #1631.

## Summary

Removes the `click < 8.2.0` upper bound so dandi-cli can be installed alongside packages that require newer click (for example `linkml-runtime >= 1.11`, which declares `click >= 8.2`). The bound guarded a behavior change in click 8.2's `Choice` rewrite; this PR adapts dandi's enum based options to that change while preserving the existing command line surface.

## Background

click 8.2 made `Choice` generic. For an `Enum`, `click.Choice(list(SomeEnum))` now matches on the enum member **names** instead of their values. dandi builds twelve options from `(str, Enum)` types whose values are the documented lowercase tokens (`error`, `skip`, `refresh`, ...), so under click 8.2+ those options accept `--existing SKIP` and reject `--existing skip`, breaking every documented invocation. The 8.2.2 release (referenced in click #2911) only changed how the default is rendered in `--help`; the matching behavior is unchanged and still reproduces on click 8.4.

## Changes

- Add `EnumChoice` to `dandi/cli/base.py`: a `click.Choice` subclass built from the enum member values that converts the matched value back to the enum member. This keeps the lowercase value tokens on the command line and passes the enum member to the command callback, matching pre 8.2 behavior on every supported click version.
- Migrate the twelve `click.Choice(list(SomeEnum))` options in `cmd_upload`, `cmd_download`, `cmd_organize`, and `cmd_move` to `EnumChoice(SomeEnum)`. `click.Choice(list(dandi_layout_fields))` is left as is because it is a dict of strings, unaffected by the enum change.
- Drop the `< 8.2.0` upper bound in `pyproject.toml`.
- Add `dandi/cli/tests/test_base.py`.

`EnumChoice` follows the approach already used for `--min-severity` in `cmd_validate.py`, which builds its `Choice` from a list of strings derived from `Severity` (`[i.name for i in Severity]`) rather than from enum members, and is therefore unaffected by the click 8.2 change. `EnumChoice` generalizes that into a reusable type that is value-based and returns the enum member, which the upload, download, organize, and move callbacks expect. (NOTE: there are now two enum-to-Choice idioms in the CLI (value-based `EnumChoice` for str-enums, inline `[i.name for i in …]` for Severity). I kept the change minimal; happy to unify them in a follow-up if preferred.)

## Compatibility

- Behavior is identical on click 8.1.8 and 8.4.2: lowercase value accepted and converted to the enum member, string default converted to the member, member name rejected, and `--help` lists the lowercase values.
- No use of other APIs removed in click 8.2 (`BaseCommand`, `MultiCommand`, the parser module, `split_arg_string`, `OptionParser`).

## Testing

- `dandi/cli/tests/test_base.py`, `test_command.py`, `test_move.py`, and the local `organize` choice mode tests pass on click 8.4.2; the new tests also pass on click 8.1.8.
- `black`, `isort`, `flake8`, and `mypy` pass on all changed files.
- Manual check of `upload`, `download`, `organize`, and `move` `--help` shows lowercase value choices; `--existing skip` resolves to the enum member and `--existing SKIP` is rejected as before.

### DIFF
--- a/dandi/cli/base.py
+++ b/dandi/cli/base.py
@@ -1,3 +1,4 @@
+from enum import Enum
 from functools import wraps
 import os
 
@@ -27,6 +28,24 @@ class IntColonInt(click.ParamType):
 
     def get_metavar(self, param, ctx=None):
         return "N[:M]"
+
+
+class EnumChoice(click.Choice):
+    """A ``click.Choice`` over a ``str``-valued ``Enum``, matched on member values.
+
+    The available choices presented on the command line are the enum member
+    values (e.g. ``error``, ``skip``), and ``convert`` returns the corresponding
+    enum member. A string default is converted to its member as well.
+    """
+
+    def __init__(self, enum_cls: type[Enum], case_sensitive: bool = True) -> None:
+        self.enum_cls = enum_cls
+        super().__init__([e.value for e in enum_cls], case_sensitive=case_sensitive)
+
+    def convert(self, value, param, ctx):
+        if value is None or isinstance(value, self.enum_cls):
+            return value
+        return self.enum_cls(super().convert(value, param, ctx))
 
 
 class ChoiceList(click.ParamType):
--- a/dandi/cli/cmd_download.py
+++ b/dandi/cli/cmd_download.py
@@ -5,7 +5,13 @@ import os
 
 import click
 
-from .base import ChoiceList, IntColonInt, instance_option, map_to_click_exceptions
+from .base import (
+    ChoiceList,
+    EnumChoice,
+    IntColonInt,
+    instance_option,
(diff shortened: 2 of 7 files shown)


yarikoptic (approved):
Thank you @rly . Let the perfect not be enemy of the good (I feel that we are digging deeper into ad-hoc solution over something which should be a simple common construct... but ATM can't wrap my head around to see that proper alternative).

rly:
Thanks @yarikoptic !

PR #1886, opened 2026-07-07
Set uploaded asset metadata schemaKey to "Asset"

candleindark:
## Summary

Pin `schemaKey` to `"Asset"` on every metadata dict written to an asset endpoint. This is the lockstep client change for dandi/dandi-schema#419, which changes `BareAsset.schemaKey` from `"Asset"` to `"BareAsset"`.

Whatever the metadata was gathered as (e.g. a `BareAsset`), what is created on the server is an `Asset`. The archive stores `schemaKey` verbatim and does not normalize it, so once `BareAsset.schemaKey` becomes `"BareAsset"`, `BareAsset`-derived metadata would be stored with the wrong `schemaKey`. Setting it at the point of upload keeps this correct regardless of how the metadata was gathered.

It is a **no-op against the current dandi-schema release** (where `BareAsset.schemaKey` is still `"Asset"`), so it can merge and release ahead of the dandi-schema change.

<details>
<summary>What changed</summary>

`schemaKey` is set to `"Asset"` at every sink that writes to an asset create/update endpoint:

- `LocalFileAsset.iter_upload` and `ZarrAsset.iter_upload` (POST create / PUT replace), next to the existing `metadata.setdefault("path", ...)` step. Covers the normal CLI upload, the deprecated `upload_raw_asset` / `iter_upload_raw_asset`, and direct `LocalAsset.upload()` / `iter_upload()` API use.
- `RemoteBlobAsset.set_raw_metadata` and `RemoteZarrAsset.set_raw_metadata` (PUT update). The `reextract` service command reaches these with `BareAsset`-derived metadata (`nwb2asset` returns a `BareAsset`) without going through `iter_upload`.

`RemoteDandisetVersion.set_raw_metadata` is left untouched: it writes Dandiset metadata, not an asset.
</details>

<details>
<summary>Test adaptations for the <code>BareAsset.schemaKey</code> change</summary>

Two metadata tests hard-coded the top-level `schemaKey` of a `BareAsset` as `"Asset"`, which becomes `"BareAsset"` under dandi/dandi-schema#419. Both now track the default of the `schemaKey` field in `BareAsset` rather than a literal, so they pass against the current dandischema and #419 alike:

- `test_nwb2asset` / `test_nwb2asset_remote_asset`: drop the redundant explicit `schemaKey="Asset"`, which merely restated the field default.
- `test_prepare_metadata`: set the expected `schemaKey` to the default of the `schemaKey` field in `BareAsset` before the comparison (the four `metadata2asset*.json` files now store a placeholder for it), and promote the object to an `Asset` before the `validate()` call that checks it against the `Asset` class.
</details>

## Test plan

- [x] `black --check` clean on the changed files (verified out-of-band; the repo's pinned black currently crashes under Python 3.14, unrelated to this change)
- [x] `test_upload_download` extended to assert the uploaded asset's stored `schemaKey == "Asset"` (needs the Docker-backed archive fixtures in CI)
- [x] Confirm against the consolidated schema (`BareAsset.schemaKey == "BareAsset"`) via #1888, which repoints the `dandischema` dependency at the dandi/dandi-schema#419 branch

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -70,6 +70,20 @@ if TYPE_CHECKING:
 lgr = get_logger()
 
 
+def set_asset_schema_key(metadata: dict[str, Any]) -> None:
+    """
+    Set ``schemaKey`` to ``"Asset"`` on metadata bound for the server as asset metadata.
+
+    However the metadata was gathered (e.g. as a ``BareAsset``), if it is to be uploaded
+    to the server to populate, in part, an ``Asset`` instance, set the ``schemaKey`` of
+    the metadata to ``"Asset"`` here. The server expects the uploaded metadata to carry
+    ``"Asset"`` as its ``schemaKey``.
+
+    The metadata is modified in place.
+    """
+    metadata["schemaKey"] = "Asset"
+
+
 class AssetType(Enum):
     """
     .. versionadded:: 0.36.0
@@ -1993,11 +2007,7 @@ class RemoteBlobAsset(RemoteAsset, BaseRemoteBlobAsset):
         Set the metadata for the asset on the server to the given value and
         update the `RemoteBlobAsset` in place.
         """
-        # Whatever the metadata was gathered as (e.g. a ``BareAsset``),
-        # what is being created on the server is an ``Asset`` as result of the upload.
-        # The server stores ``schemaKey`` verbatim, so set it
-        # here, at the point of upload, rather than relying on the caller.
-        metadata["schemaKey"] = "Asset"
+        set_asset_schema_key(metadata)
         data = self.client.put(
             self.api_path, json={"metadata": metadata, "blob_id": self.blob}
         )
@@ -2021,11 +2031,7 @@ class RemoteZarrAsset(RemoteAsset, BaseRemoteZarrAsset):
         Set the metadata for the asset on the server to the given value and
         update the `RemoteZarrAsset` in place.
         """
-        # Whatever the metadata was gathered as (e.g. a ``BareAsset``),
-        # what is being created on the server is an ``Asset`` as result of the upload.
-        # The server stores ``schemaKey`` verbatim, so set it
-        # here, at the point of upload, rather than relying on the caller.
-        metadata["schemaKey"] = "Asset"
+        set_asset_schema_key(metadata)
         data = self.client.put(
             self.api_path, json={"metadata": metadata, "zarr_id": self.zarr}
         )
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -1993,6 +1993,11 @@ class RemoteBlobAsset(RemoteAsset, BaseRemoteBlobAsset):
         Set the metadata for the asset on the server to the given value and
         update the `RemoteBlobAsset` in place.
         """
+        # Whatever the metadata was gathered as (e.g. a ``BareAsset``),
+        # what is being created on the server is an ``Asset`` as result of the upload.
+        # The server stores ``schemaKey`` verbatim, so set it
+        # here, at the point of upload, rather than relying on the caller.
+        metadata["schemaKey"] = "Asset"
         data = self.client.put(
             self.api_path, json={"metadata": metadata, "blob_id": self.blob}
         )
@@ -2016,6 +2021,11 @@ class RemoteZarrAsset(RemoteAsset, BaseRemoteZarrAsset):
         Set the metadata for the asset on the server to the given value and
         update the `RemoteZarrAsset` in place.
         """
+        # Whatever the metadata was gathered as (e.g. a ``BareAsset``),
+        # what is being created on the server is an ``Asset`` as result of the upload.
+        # The server stores ``schemaKey`` verbatim, so set it
+        # here, at the point of upload, rather than relying on the caller.
+        metadata["schemaKey"] = "Asset"
         data = self.client.put(
             self.api_path, json={"metadata": metadata, "zarr_id": self.zarr}
         )
--- a/dandi/tests/data/metadata/metadata2asset.json
+++ b/dandi/tests/data/metadata/metadata2asset.json
@@ -1,6 +1,6 @@
 {
   "id": "dandiasset:0b0a1a0b-e3ea-4cf6-be94-e02c830d54be",
-  "schemaKey": "TO BE SPECIFIED DYNAMICALLY",
+  "schemaKey": "ANYTHING GOOD NOW -- `schemaKey` of a `BareAsset` instance will be adjusted to the default value of the `schemaKey` field in the Pydantic `BareAsset` model",
   "schemaVersion": "0.4.1",
   "keywords": [
     "test",
--- a/dandi/tests/data/metadata/metadata2asset.json
+++ b/dandi/tests/data/metadata/metadata2asset.json
@@ -1,6 +1,6 @@
 {
   "id": "dandiasset:0b0a1a0b-e3ea-4cf6-be94-e02c830d54be",
-  "schemaKey": "Asset",
+  "schemaKey": "TO BE SPECIFIED DYNAMICALLY",
   "schemaVersion": "0.4.1",
   "keywords": [
     "test",
--- a/dandi/files/bases.py
+++ b/dandi/files/bases.py
@@ -25,7 +25,12 @@ from pydantic_core import ErrorDetails
 import requests
 
 import dandi
-from dandi.dandiapi import RemoteAsset, RemoteDandiset, RESTFullAPIClient
+from dandi.dandiapi import (
+    RemoteAsset,
+    RemoteDandiset,
+    RESTFullAPIClient,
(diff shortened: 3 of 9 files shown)


candleindark:
@yarikoptic Please take a look at this PR. We will need this merged and released in order to have the integration tests in dandi/dandi-schema#419 to become green.

dandi/dandiapi.py
```diff
@@ -1993,6 +1993,11 @@ def set_raw_metadata(self, metadata: dict[str, Any]) -> None:
         Set the metadata for the asset on the server to the given value and
         update the `RemoteBlobAsset` in place.
         """
+        # Whatever the metadata was gathered as (e.g. a ``BareAsset``),
+        # what is being created on the server is an ``Asset`` as result of the upload.
+        # The server stores ``schemaKey`` verbatim, so set it
```
yarikoptic: I twitch constantly  due to my duplication allergy and seeing this length comment + the same code change all around.  We have 4 places with that.   Even adding a `set_asset_schemaKey(metadata)` to concentrate that comment + change would be a better solution IMHO. 

But I also wonder if we could just add to base class and then extend in derived smth like `def prepare_metadata_for_server(metadata: dict) -> metadata` where then `Asset` class would do its assignment there. And then we would call the method when providing metadata.

This would also allow for downgrades of the version as we are yet to do for properly allowing newer client + schema operation with older deployment where feasible.

dandi/tests/data/metadata/metadata2asset.json
```diff
@@ -1,6 +1,6 @@
 {
   "id": "dandiasset:0b0a1a0b-e3ea-4cf6-be94-e02c830d54be",
-  "schemaKey": "Asset",
+  "schemaKey": "TO BE SPECIFIED DYNAMICALLY",
```
yarikoptic: is that to ensure that we would fail unless we do that and otherwise could as well still be prior `"Asset"`?

then may be smth along the lines ?


```suggestion
  "schemaKey": "ANYTHING GOOD NOW -- Asset VALUE MUST BE PROVIDED DURING UPLOAD TO SERVER",
```

yarikoptic (changes requested):
duplication needs to be addressed.

In the longer run I feel that it might need to be the server which would do the upgrade from BaseAsset to Asset!?

dandi/dandiapi.py
candleindark: The actual code duplication here was a single line (`metadata["schemaKey"] = "Asset"`); what really repeated was the explanatory comment. I've bound that comment and the operation together in one `set_asset_schema_key()` helper in `dandi.dandiapi`, called from all four upload sites, so the documentation and the assignment now live in one place.

On the `prepare_metadata_for_server` method form specifically: the four sites don't share a base class to hang it on. Two are `LocalAsset.iter_upload` overrides (`dandi/files/`) and two are `RemoteAsset.set_raw_metadata` overrides (`dandi/dandiapi.py`), so an overridable method would have to be duplicated onto two unrelated bases (or added via a mixin), whereas a plain module-level helper is callable from all four. The version-downgrade motivation is real, but I'd argue that belongs in dandischema (schema migration) rather than the CLI, and it isn't needed for this PR.

The right long-term home for this is the server: a client should upload an unmodified `BareAsset`, and dandi-archive should validate it and transform it into an `Asset` (the dandi/dandi-schema#205 spirit). I kept the change client-side deliberately, since the proper server version needs endpoint hardening, likely a database migration, and cross-team discussion, so it is out of scope for this PR. The `set_asset_schema_key()` helper already gives a single point of truth on the client, and there is a design reason not to push a quick fix into dandi-archive instead (details below). Shall I file a follow-up issue?

<details>
<summary>Why not patch dandi-archive now: the smell, the migration, and endpoint hardening</summary>

**Why not a quick server-side patch.** There is a shortcut that avoids any migration: setting `schemaKey="Asset"` inside [`strip_metadata`](https://github.com/dandi/dandi-archive/blob/5e94e06db0b1f9b47e2f08ae4e4201c1f027be61/dandiapi/api/models/asset.py#L330) stamps every stored asset without touching `ASSET_COMPUTED_FIELDS` or the constraint (`strip_metadata` runs on both asset write paths, and `schemaKey` stays a non-computed field, so a draft still has no computed keys). But it makes the server silently rewrite client-supplied metadata. dandi-archive already discards client-supplied computed fields in `strip_metadata`, yet that is at least explicit; hiding a `schemaKey` rewrite in the same place amplifies the "the client sends data the server quietly treats as wrong" smell, which is itself the argument for the formal validate-and-transform endpoint. So I'd rather not build further on the current design.

**Done cleanly, it entails a migration.** The natural way for the server to own `schemaKey` is to treat it as a computed field, i.e. add it to [`ASSET_COMPUTED_FIELDS`](https://github.com/dandi/dandi-archive/blob/5e94e06db0b1f9b47e2f08ae4e4201c1f027be61/dandiapi/api/models/asset.py#L32-L42) so [`strip_metadata`](https://github.com/dandi/dandi-archive/blob/5e94e06db0b1f9b47e2f08ae4e4201c1f027be61/dandiapi/api/models/asset.py#L330) drops it from the stored (draft) [`metadata`](https://github.com/dandi/dandi-archive/blob/5e94e06db0b1f9b47e2f08ae4e4201c1f027be61/dandiapi/api/models/asset.py#L122) and [`full_metadata`](https://github.com/dandi/dandi-archive/blob/5e94e06db0b1f9b47e2f08ae4e4201c1f027be61/dandiapi/api/models/asset.py#L294) sets it to `"Asset"` on read. But `ASSET_COMPUTED_FIELDS` is baked into the [`asset_metadata_no_computed_keys_or_published`](https://github.com/dandi/dandi-archive/blob/5e94e06db0b1f9b47e2f08ae4e4201c1f027be61/dandiapi/api/models/asset.py#L166-L172) CheckConstraint (a draft's metadata must contain none of those keys), so changing the list needs a `RemoveConstraint`/`AddConstraint` migration; and since every existing asset stores `schemaKey` today (it isn't stripped), every existing draft row would then violate the amended constraint, so a data migration must first strip `schemaKey` from all draft rows. Migration [`0011_asset_access_metadata.py`](https://github.com/dandi/dandi-archive/blob/5e94e06db0b1f9b47e2f08ae4e4201c1f027be61/dandiapi/api/migrations/0011_asset_access_metadata.py) is the precedent: it [rewrote every asset's `metadata`](https://github.com/dandi/dandi-archive/blob/5e94e06db0b1f9b47e2f08ae4e4201c1f027be61/dandiapi/api/migrations/0011_asset_access_metadata.py#L8-L14) and [re-created this same constraint](https://github.com/dandi/dandi-archive/blob/5e94e06db0b1f9b47e2f08ae4e4201c1f027be61/dandiapi/api/migrations/0011_asset_access_metadata.py#L31-L35) for the `access` field.

**The received `BareAsset` should be validated at the endpoint, and that is a separate check from validating the outcome.** A `BareAsset` arriving at the endpoint should be validated server-side, synchronously at the point of receipt, against the Pydantic `BareAsset` model (or its JSON schema). That is distinct from validating the transformed outcome, the `Asset` (and, at publish, `PublishedAsset`). Today neither of those is true: ingestion is permissive and does no `BareAsset` validation, [`AssetRequestSerializer.metadata`](https://github.com/dandi/dandi-archive/blob/5e94e06db0b1f9b47e2f08ae4e4201c1f027be61/dandiapi/api/views/asset.py#L191) is a bare `serializers.JSONField()` whose [`validate()`](https://github.com/dandi/dandi-archive/blob/5e94e06db0b1f9b47e2f08ae4e4201c1f027be61/dandiapi/api/views/asset.py#L210) only checks blob/zarr exclusivity, that `path` is present and well-formed, and defaults `schemaVersion`; [`_create_asset`](https://github.com/dandi/dandi-archive/blob/5e94e06db0b1f9b47e2f08ae4e4201c1f027be61/dandiapi/api/services/asset/__init__.py#L27) stores the metadata after `full_clean(validate_constraints=False)`; and the only dandischema check runs *after* storage, in an [async task](https://github.com/dandi/dandi-archive/blob/5e94e06db0b1f9b47e2f08ae4e4201c1f027be61/dandiapi/api/tasks/__init__.py#L66) → [`validate_asset_metadata`](https://github.com/dandi/dandi-archive/blob/5e94e06db0b1f9b47e2f08ae4e4201c1f027be61/dandiapi/api/services/metadata/__init__.py#L48), which validates the *outcome* against `PublishedAsset` and merely records a VALID/INVALID status without rejecting the upload. So a proper design would add synchronous `BareAsset` validation at receipt, kept separate from that outcome validation.
</details>

dandi/tests/data/metadata/metadata2asset.json
candleindark: > is that to ensure that we would fail unless we do that and otherwise could as well still be prior `"Asset"`?

yes

> then may be smth along the lines ?

yes

yarikoptic (approved):
ok, let's do it and see how it goes

failed CI is unrelated flake

PR #1890, opened 2026-07-09
Drop python 3.10 and simplify for  click 8.2 enum choice 

yarikoptic:
## Summary

Follow-up to #1883 that collapses the click-8.2 compatibility shim
and rebases the CLI enums on the standard-library `enum.StrEnum`
(Python 3.11+). No behavior change — same command-line surface,
same `--help` output, same accept/reject semantics.

Two independent commits:

1. **click 8.2 exposes `Choice.normalize_choice`** as the explicit hook
   for choice-string normalization. Overriding it to return `.value`
   for `Enum` members (instead of the default `.name`) is *all* that
   is needed to preserve the pre-8.2 CLI surface, because
   `Choice.convert` then matches by normalized string and returns the
   matched enum member automatically. The custom `__init__` +
   `convert` + `self.enum_cls` bookkeeping in #1883's `EnumChoice`
   become unnecessary. Bumps the `click` floor from `>= 7.1` to
   `>= 8.2` — the earlier bound predates `normalize_choice`.

2. **Migrate the 11 CLI-facing `(str, Enum)` classes to
   `enum.StrEnum`** and drop their hand-written
   `__str__ -> self.value` overrides. `StrEnum` was introduced in
   Python 3.11 and provides exactly these semantics natively. This
   requires bumping the minimum Python from 3.10 (EOL 2026-10-04)
   to 3.11: `pyproject.toml`, `tox.ini`, and 5 GitHub Actions
   workflows are updated accordingly (matrix drops the 3.10 leg;
   pinned single-Python `include:` entries move to 3.11).

Net change vs merged #1883: **+39 / −83** across 14 files.

## Commits

- b2d3c648 `Simplify EnumChoice via click 8.2 normalize_choice hook`
- 5d2cec5e `Adopt enum.StrEnum for CLI choice enums; drop Python 3.10`

<details><summary>Per-commit details</summary>

### b2d3c648 `Simplify EnumChoice via click 8.2 normalize_choice hook`

`EnumChoice` shrinks from ~15 lines (custom `__init__`, `convert`,
`self.enum_cls`, defensive `None`/`isinstance` guards) to a single
3-line `normalize_choice` override. Callsites are unchanged —
iterating an Enum class yields members, and `click.Choice.__init__`
already does `tuple(choices)` on any iterable, so
`EnumChoice(DownloadExisting)` still works verbatim.

`pyproject.toml`: `click >= 7.1` → `click >= 8.2`. The
`normalize_choice` hook is documented "added in 8.2.0" upstream;
retaining a lower bound would just be dead compatibility.

### 5d2cec5e `Adopt enum.StrEnum for CLI choice enums; drop Python 3.10`

Migrated: `SyncMode` (`consts.py`), `DownloadExisting` /
`DownloadFormat` / `PathType` (`download.py`), `UploadExisting` /
`UploadValidation` (`upload.py`), `MoveExisting` / `MoveWorkOn`
(`move.py`), `FileOperationMode` / `CopyMode` / `OrganizeInvalid`
(`organize.py`). Each gets its `(str, Enum)` base changed to
`StrEnum` and its `__str__ -> self.value` override deleted.

`_Existing` in `dandi/cli/tests/test_base.py` also migrated for
consistency.

`EnumChoice` itself is unchanged — `StrEnum` still exposes distinct
`.name` and `.value`, and click 8.2 still defaults to matching on
`.name`, so the `normalize_choice` override remains necessary.

Python floor moves from 3.10 to 3.11 to unlock `enum.StrEnum`:

- `pyproject.toml`: `requires-python = ">=3.11"`, drop the 3.10
  classifier, `[tool.hatch.envs.default] python = "3.11"`.
- `tox.ini`: drop the `tomli; python_version < '3.11'` conditional
  from the lint env (unreachable under the new floor).
- `.github/workflows/run-tests.yml`: matrix drops `3.10`; the five
  `include:` entries pinned to `'3.10'` (`dandi-api`, `dandi-api` +
  `ember-dandi`, `dev-deps`, `nfs`, and the `ember-dandi` sandbox
  vendored config) move to `'3.11'`.
- `.github/workflows/{lint,docs,release,typing}.yml`: single
  `python-version` bumps to `'3.11'`.

**Intentionally not migrated**: `dandi/utils.py:StrEnum` (a
`(str, Enum)` shim used by `dandi/validate/_types.py` and
`dandi/bids_validator_deno/_models.py`). Its
`_generate_next_value_` returns `name` (case-preserving), while
`enum.StrEnum._generate_next_value_` returns `name.lower()`.
Migrating it would change every `auto()`-derived value that appears
in serialized `ValidationResult` output (`Standard.BIDS` → `"BIDS"`
vs `"bids"`, etc.) — a separate audit worth its own PR.

</details>

## Test plan

- [x] `pytest dandi/cli/tests/{test_base,test_command,test_move,test_download}.py` — all pass on click 8.4.2 + Python 3.13
- [x] Broader `dandi/cli/tests/ + dandi/tests/test_organize.py` — 154 pass (only pre-existing `bids-validator-deno`-binary-dependent tests skipped in local env)
- [x] `dandi organize --files-mode SIMULATE` still rejected; lowercase `simulate` still accepted
- [x] `dandi {download,upload,organize,move} --help` render lowercase choice lists unchanged
- [x] Pre-commit hooks (black, isort, codespell, flake8, reuse) pass on both commits
- [x] Test locally interactively
- [x] CI green on the PR (Ubuntu / macOS / Windows × Python 3.11–3.14)

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -13,8 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         python:
-          - '3.10'
-          #- 3.11
+          - '3.11'
     steps:
     - name: Check out repository
       uses: actions/checkout@v7
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,7 +15,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v6
       with:
-        python-version: '3.10'
+        python-version: '3.11'
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
-          python-version: '3.10'
+          python-version: '3.11'
 
       - name: Install build & twine
         run: python -m pip install build twine
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -34,8 +34,7 @@ jobs:
           - macos-latest  # M1 architecture
           - macos-15-intel  # non-M1 architecture
         python:
-          - '3.10'  # Needs quotes so YAML doesn't think it's 3.1
-          - '3.11'
+          - '3.11'  # Needs quotes so YAML doesn't think it's 3.1
           - '3.12'
           - '3.13'
           - '3.14'
@@ -48,10 +47,10 @@ jobs:
           - default
         include:
           - os: ubuntu-latest
-            python: '3.10'
+            python: '3.11'
             mode: dandi-api
           - os: ubuntu-latest
-            python: '3.10'
+            python: '3.11'
             mode: dandi-api
             vendored_dandiapi: ember-dandi
             instance_name: EMBER-DANDI
@@ -61,13 +60,13 @@ jobs:
             python: 3.14
             mode: obolibrary-only
           - os: ubuntu-latest
-            python: '3.10'
+            python: '3.11'
             mode: dev-deps
           - os: ubuntu-latest
             python: 3.14
             mode: dev-deps
           - os: ubuntu-latest
-            python: '3.10'
+            python: '3.11'
             mode: nfs
 
     steps:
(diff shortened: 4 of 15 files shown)


CodyCBakerPhD:
hint: if you tell your agent to 'watch the pr' it will automatically fix any CI failures, such as the docs build error

yarikoptic:
thanks! I typically do not permit the beasts to push without my supervision, so would need another manual step in the loop

CodyCBakerPhD (approved):

PR #1892, opened 2026-07-15
Document validation caveats of the `get_metadata` methods

candleindark:
Docstring-only change to `RemoteDandiset.get_metadata()` and `BaseRemoteAsset.get_metadata()`. No behavior change.

Both methods validate fetched metadata against the `Dandiset`/`Asset` model of the **installed** `dandischema`, not against the version of that model the metadata was published under. The existing note only warns that unpublished versions may have invalid metadata, which leaves two things unstated:

- **Even a published version's metadata can fail to validate**, because of drift between the model in the installed `dandischema` and the version of the model the metadata was published under.
- **The methods can raise `pydantic.ValidationError`** when the fetched metadata does not validate against `dandischema.models.Dandiset` / `dandischema.models.Asset`.

Both are now documented, the latter as a `:raises:` field matching the convention already used elsewhere in `dandiapi.py`.

## Test plan

- [x] pre-commit hooks pass (black, isort, flake8, codespell, reuse lint)
- [x] `flake8` and `black --check` clean on `dandi/dandiapi.py`

### DIFF
--- a/dandi/dandiapi.py
+++ b/dandi/dandiapi.py
@@ -1167,6 +1167,15 @@ class RemoteDandiset:
             Only published Dandiset versions can be expected to have valid
             metadata.  Consider using `get_raw_metadata()` instead in order to
             fetch unstructured, possibly-invalid metadata.
+
+            Even a published version's metadata can fail to validate.  The
+            fetched metadata is validated against the `Dandiset` model of the installed
+            `dandischema`, not against the version of `Dandiset` model that the metadata
+            was published under, so drift between the two can render
+            previously-valid metadata invalid.
+
+        :raises pydantic.ValidationError: if the fetched metadata does not
+            validate against `dandischema.models.Dandiset`
         """
         return models.Dandiset.model_validate(self.get_raw_metadata())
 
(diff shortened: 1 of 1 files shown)


candleindark:
The failing CI checks here are unrelated to this docstring-only change.

- `codecov/project` is a coverage-target miss (expected: a docstring-only diff adds no covered lines), and `codecov/patch` passed.
- The only failing test jobs are the two `dev-deps` variants; every `normal`, `dandi-api`, and `nfs` job passed across all OS/Python combinations. The same `dev-deps` failures reproduce on `master` in the scheduled run [29477710764](https://github.com/dandi/dandi-cli/actions/runs/29477710764), which does not include this PR, so a docstring change cannot be the cause.

<details>
<summary>Root cause of the <code>dev-deps</code> failures</summary>

The `dev-deps` jobs install `hdmf` from git (`6.1.1.dev12+g50769ca55`), which now emits a new `DeprecationWarning` while writing an NWB file:

```
DeprecationWarning: Override function for attribute 'experimenter' returned None, so the container's attribute value is used. In HDMF 8.0, a None return will set the attribute to None; return ObjectMapper.NO_OVERRIDE to keep using the container's attribute value.
```

Under `filterwarnings = error` this is promoted to an exception, which aborts the shared NWB-file fixture (`dandi/tests/fixtures.py` → `dandi/pynwb_utils.py::make_nwb_file` → `io.write`). Every test that depends on a written NWB file (`test_validate_*`, `test_organize_*`, `test_video_organize*`, `test_organized_nwb_dir*`, `test_get_asset_metadata`, `test_smoke*`) then errors at setup, and the failures cascade to the archive-backed tests. Released `hdmf 6.1.0` (used by the pinned-dependency jobs) does not emit this warning, so only `dev-deps` is affected. This will keep failing until `hdmf`/`pynwb` upstream stops emitting the warning or dandi-cli filters it.
</details>

candleindark:
Follow-up: the `dev-deps` failure is already tracked upstream by pynwb PR [NeurodataWithoutBorders/pynwb#2224](https://github.com/NeurodataWithoutBorders/pynwb/pull/2224), which migrates the offending override functions (including `NWBFileMap.experimenter_obj_attr`, the exact one in the traceback) to return `ObjectMapper.NO_OVERRIDE` instead of `None`. It depends on the hdmf change that introduced the deprecation ([hdmf-dev/hdmf#1167](https://github.com/hdmf-dev/hdmf/pull/1167)). Nothing to fix in dandi-cli; the `dev-deps` jobs should go green once those land and dev-deps picks up the new git tips.

yarikoptic (approved):

PR #1894, opened 2026-07-27
Build expected metadata with the server's DANDI schema version

candleindark:
## Problem

`test_update_dandiset_from_doi` fills the `schemaVersion` and `@context` of its expected metadata from the `DANDI_SCHEMA_VERSION` of the client's `dandischema`, but in the served metadata under test both carry the server's. While dandi-cli and the dandi-archive instance are built against the same `dandischema` the two coincide and the test passes; once they differ, every parametrization fails on those two fields even when the DOI-derived metadata the test exists to check is correct. The test thus asserts something it was never meant to: that both sides run the same `DANDI_SCHEMA_VERSION`.

## Fix

Build those two fields from the schema version reported by the server's `/info/`, as the test already does for the other instance-dependent fields (`url`, `repository`, `manifestLocation`).

<details>
<summary>Where this surfaced</summary>

dandi/dandi-schema#419 bumps `DANDI_SCHEMA_VERSION` to `0.8.0`. In that repo's `test-dandi-cli` matrix, the `schema_install: server` job installs the branch into the API image only, leaving a released client against a `0.8.0` server. All five parametrizations then fail, while the other 1001 tests in the job pass:

```
{'schemaVersion': '0.8.0'} != {'schemaVersion': '0.7.0'}
{'@context': '.../releases/0.8.0/context.json'} != {'.../releases/0.7.0/context.json'}
```
</details>

## Test plan

- [x] pre-commit hooks pass
- [x] CI, as a no-regression check. It cannot show the benefit, since both sides there run the same released `dandischema`; that is verified by re-running the `server` job of dandi/dandi-schema#419 once this is merged.

### DIFF
--- a/dandi/cli/tests/test_service_scripts.py
+++ b/dandi/cli/tests/test_service_scripts.py
@@ -9,7 +9,6 @@ import sys
 
 import anys
 from click.testing import CliRunner
-from dandischema.consts import DANDI_SCHEMA_VERSION
 from dandischema.models import ID_PATTERN
 import pytest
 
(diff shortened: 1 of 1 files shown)


yarikoptic (approved):
hm, I think it does make more sense this way! Nice find!  Most likely we had this test created BEFORE we exposed version of the schema used by the server.  So let's proceed this way indeed.

PR #1900, opened 2026-08-05
Give a descriptive error when a DOI cannot be resolved to citation metadata

### LINKED ISSUE #1855: update-dandiset-from-doi errors non-descriptively
For @candleindark effort on testing linkml produced jsonschema driven meditor, I was trying to find the script we had to sweep populate staging (now sandbox)  with dandisets based on the ones in the main instance metadata for testing etc... failed to find that script yet (might be on drogon) but found helper command I thought I could use -- the `update-dandiset-from-doi`  but it errored out:

```bash
❯ DANDI_DEVEL=1 dandi service-scripts update-dandiset-from-doi -d 216023 -e overwrite -F all -i dandi-sandbox https://doi.org/10.48324/dandi.001827/0.260505.1322
  File "/home/yoh/proj/dandi/dandi-cli/.venv/lib/python3.13/site-packages/requests/models.py", line 978, in json
    return complexjson.loads(self.text, **kwargs)
           ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/json/__init__.py", line 352, in loads
    return _default_decoder.decode(s)
           ~~~~~~~~~~~~~~~~~~~~~~~^^^
  File "/usr/lib/python3.13/json/decoder.py", line 345, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/json/decoder.py", line 363, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

During handling of the above exception, another exception occurred:

  File "/home/yoh/proj/dandi/dandi-cli/.venv/bin/dandi", line 10, in <module>
    sys.exit(main())
             ~~~~^^
  File "/home/yoh/proj/dandi/dandi-cli/.venv/lib/python3.13/site-packages/click/core.py", line 1161, in __call__
           ~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/yoh/proj/dandi/dandi-cli/.venv/lib/python3.13/site-packages/click/core.py", line 1082, in main
    rv = self.invoke(ctx)
  File "/home/yoh/proj/dandi/dandi-cli/.venv/lib/python3.13/site-packages/click/core.py", line 1697, in invoke
                           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/home/yoh/proj/dandi/dandi-cli/.venv/lib/python3.13/site-packages/click/core.py", line 1697, in invoke
                           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/home/yoh/proj/dandi/dandi-cli/.venv/lib/python3.13/site-packages/click/core.py", line 1443, in invoke
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/yoh/proj/dandi/dandi-cli/.venv/lib/python3.13/site-packages/click/core.py", line 788, in invoke
    return __callback(*args, **kwargs)
  File "/home/yoh/proj/dandi/dandi-cli/dandi/cli/cmd_service_scripts.py", line 268, in update_dandiset_from_doi
    doidata = doiclient.get(doi)
  File "/home/yoh/proj/dandi/dandi-cli/dandi/dandiapi.py", line 332, in get
    return self.request("GET", path, **kwargs)
           ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/yoh/proj/dandi/dandi-cli/dandi/dandiapi.py", line 312, in request
    return result.json()
           ~~~~~~~~~~~^^
  File "/home/yoh/proj/dandi/dandi-cli/.venv/lib/python3.13/site-packages/requests/models.py", line 982, in json
    raise RequestsJSONDecodeError(e.msg, e.doc, e.pos)
requests.exceptions.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

adityasingh2400:
Fixes #1855

`update-dandiset-from-doi` crashed with a bare `json.JSONDecodeError: Expecting value: line 1 column 1 (char 0)` whenever doi.org answered with anything other than JSON.

There are two separate problems behind that traceback.

The first is a header bug. The CSL Accept header was set on the `RESTFullAPIClient` session, but `RESTFullAPIClient.request()` sets `accept: application/json` on the request whenever `json_resp` is true, and in requests a per-request header wins over a session header. The resolver therefore never saw the citation format we meant to ask for. The checked-in VCR cassettes record this, every captured request carries `accept: application/json` and never the CSL type. Crossref happens to serve JSON for that anyway, because it redirects to `api.crossref.org/.../transform`, which is why the existing tests pass. A registration agency that does not will redirect to the landing page and return HTML with a 200, which is the reported failure on a DataCite `10.48324` DOI.

The second is that there was no error handling at all on that path, so the user got a `JSONDecodeError` raised from inside requests with nothing naming the DOI.

The fetch moves into `fetch_doi_citation_metadata()`, which requests the raw response so the intended CSL Accept header survives, and turns a 404, any other HTTP error, a non-JSON body, and a non-object body each into a `click.ClickException` naming the DOI, the URL, and the content type actually received. `normalize_doi()` now accepts a bare DOI, a `doi:` URI, or a resolver URL, and rejects anything else with a `click.UsageError` instead of a traceback. That also fixes the `relatedResource` record, whose url was built as `https://doi.org/{doi}` and came out doubled when the user passed a resolver URL. The lookup now happens before connecting to the archive, so a bad DOI fails fast without needing credentials.

Only `title`, `abstract`, and `author[*].given/family/ORCID/affiliation` are read, and all of those are present in both CSL JSON and the Crossref record the cassettes captured, so replaying the existing cassettes is unaffected. vcrpy matches on method and URI, not headers.

Verified against the base ref with doi.org mocked to serve HTML at 200. Before, the run ends in `requests.exceptions.JSONDecodeError` from `dandiapi.py` line 326. After, it reports that the DOI answered with `text/html` instead of CSL JSON and explains that the registration agency likely does not serve citation metadata. The Accept header actually sent went from `application/json` to `application/vnd.citationstyles.csl+json; charset=utf-8`.

New tests are marked `@pytest.mark.ai_generated`. They give 18 passed. The 6 deselected are the VCR `test_update_dandiset_from_doi` cases, which need the docker archive fixture and could not run locally.

AI assistance disclosure: this change was written with the help of Claude Code, and the added tests are marked `ai_generated` as CLAUDE.md asks. I reviewed and tested everything before submitting.

### DIFF
--- a/dandi/cli/cmd_service_scripts.py
+++ b/dandi/cli/cmd_service_scripts.py
@@ -7,6 +7,7 @@ from difflib import unified_diff
 import json
 import os
 from pathlib import PurePosixPath
+import re
 from textwrap import indent
 from typing import Any, TypeVar
 import urllib.parse
@@ -16,7 +17,7 @@ import click
 from dandischema.consts import DANDI_SCHEMA_VERSION
 from packaging.version import Version
 from requests.auth import HTTPBasicAuth
-from requests.exceptions import HTTPError
+from requests.exceptions import HTTPError, RequestException
 
 from dandi.consts import known_instances
 
@@ -24,7 +25,7 @@ from .base import ChoiceList, instance_option, map_to_click_exceptions
 from .. import __version__, lgr
 from ..dandiapi import DandiAPIClient, RemoteBlobAsset, RESTFullAPIClient
 from ..dandiarchive import parse_dandi_url
-from ..exceptions import NotFoundError
+from ..exceptions import HTTP404Error, NotFoundError
 from ..utils import yaml_dump
 
 T = TypeVar("T")
@@ -34,6 +35,112 @@ DOI_HUMAN_URLS = {
     "https://api.test.datacite.org/dois": "https://doi.test.datacite.org/dois",
     "https://api.datacite.org/dois": "https://doi.datacite.org/dois",
 }
+
+#: Base URL of the DOI resolver used to look up citation metadata
+DOI_RESOLVER_URL = "https://doi.org/"
+
+#: Content type requested from the DOI resolver for citation metadata
+DOI_CSL_ACCEPT = "application/vnd.citationstyles.csl+json; charset=utf-8"
+
+#: Matches a bare DOI, e.g. ``10.48324/dandi.001827/0.260505.1322``
+DOI_REGEX = re.compile(r"10\.\d{4,9}/\S+")
+
+#: Prefixes a DOI may be spelled with, in the order they are stripped
+DOI_PREFIX_REGEXES = (r"doi:", r"(?:https?://)?(?:dx\.)?doi\.org/")
+
+
+def normalize_doi(doi: str) -> str:
+    """Reduce a DOI given in any of its usual spellings to the bare DOI.
+
+    A bare DOI (``10.48324/dandi.001827/0.260505.1322``), a ``doi:`` URI, and a
+    resolver URL (``https://doi.org/...``, ``http://dx.doi.org/...``) are all
+    accepted and reduced to the bare form.
+
+    Parameters
+    ----------
+    doi : str
+        The DOI as given by the user
+
+    Returns
+    -------
+    str
+        The bare DOI
+
+    Raises
+    ------
+    ValueError
+        If `doi` is not a syntactically valid DOI in any accepted spelling
+    """
+    value = doi.strip()
+    for prefix_regex in DOI_PREFIX_REGEXES:
+        if m := re.match(prefix_regex, value, flags=re.I):
+            value = value[m.end() :].strip()
+            break
+    if not DOI_REGEX.fullmatch(value):
+        raise ValueError(
+            f"{doi!r} does not look like a DOI.  Expected something like "
+            "'10.48324/dandi.001827/0.260505.1322', optionally prefixed with "
+            "'doi:' or 'https://doi.org/'."
+        )
+    return value
+
+
+def fetch_doi_citation_metadata(doi: str) -> dict[str, Any]:
+    """Fetch the CSL JSON citation metadata for a bare `doi` from doi.org.
+
+    Parameters
+    ----------
+    doi : str
+        A bare DOI, as returned by `normalize_doi()`
+
+    Returns
+    -------
+    dict
+        The parsed CSL JSON record
+
+    Raises
+    ------
+    click.ClickException
+        If the DOI cannot be resolved, or if the resolver answers with
+        something other than a CSL JSON object.  The exception message
+        describes what went wrong, so that the user is not left with a bare
+        `json.JSONDecodeError` traceback.
+    """
+    url = f"{DOI_RESOLVER_URL}{doi}"
+    with RESTFullAPIClient(
+        DOI_RESOLVER_URL, headers={"Accept": DOI_CSL_ACCEPT}
+    ) as doiclient:
+        try:
+            r = doiclient.get(doi, json_resp=False)
+        except HTTP404Error:
+            raise click.ClickException(
+                f"DOI {doi} is not registered: {url} returned 404.  Check the "
+                "DOI for typos and make sure it has already been published."
+            )
+        except HTTPError as e:
+            status = e.response.status_code if e.response is not None else "?"
+            raise click.ClickException(
+                f"Failed to look up DOI {doi}: {url} returned HTTP {status}."
+            )
+        except RequestException as e:
+            raise click.ClickException(f"Failed to look up DOI {doi} at {url}: {e}")
+        content_type = r.headers.get("Content-Type", "<unset>")
+        try:
+            doidata = r.json()
+        except ValueError:
+            raise click.ClickException(
+                f"DOI {doi} did not resolve to citation metadata: {url} answered "
+                f"with {content_type!r} instead of CSL JSON (final URL: {r.url}).  "
+                "This usually means the DOI's registration agency does not serve "
+                "citation metadata for it, and doi.org fell back to redirecting "
+                "to the landing page."
+            )
+    if not isinstance(doidata, dict):
+        raise click.ClickException(
+            f"DOI {doi} resolved to a JSON {type(doidata).__name__} rather than "
+            f"the expected CSL JSON object (final URL: {r.url})."
+        )
+    return doidata
 
 
 @click.group()
@@ -247,7 +354,15 @@ def update_dandiset_from_doi(
     """
     Update the metadata for the draft version of a Dandiset with information
     from a given DOI record.
+
+    DOI may be given bare (``10.48324/dandi.001827/0.260505.1322``), as a
+    ``doi:`` URI, or as a resolver URL (``https://doi.org/...``).
     """
+    try:
+        doi = normalize_doi(doi)
+    except ValueError as e:
+        raise click.UsageError(str(e))
+
     known_instance_names = [k.upper() for k in known_instances.keys()]
 
     # Strip instance name prefix from dandiset ID, if present
@@ -258,15 +373,10 @@ def update_dandiset_from_doi(
             break
 
     start_time = datetime.now().astimezone()
+    # Resolve the DOI before talking to the archive, so that a bad DOI fails
+    # fast and without requiring credentials
+    doidata = fetch_doi_citation_metadata(doi)
     with DandiAPIClient.for_dandi_instance(dandi_instance, authenticate=True) as client:
-        with RESTFullAPIClient(
-            "https://doi.org/",
-            headers={
-                "Accept": "application/vnd.citationstyles.csl+json; charset=utf-8"
-            },
-        ) as doiclient:
-            doidata = doiclient.get(doi)
-
         d = client.get_dandiset(dandiset, "draft", lazy=False)
         original_metadata = d.get_raw_metadata()
         new_metadata = deepcopy(original_metadata)
--- a/dandi/cli/cmd_service_scripts.py
+++ b/dandi/cli/cmd_service_scripts.py
@@ -42,11 +42,17 @@ DOI_RESOLVER_URL = "https://doi.org/"
 #: Content type requested from the DOI resolver for citation metadata
 DOI_CSL_ACCEPT = "application/vnd.citationstyles.csl+json; charset=utf-8"
 
-#: Matches a bare DOI, e.g. ``10.48324/dandi.001827/0.260505.1322``
-DOI_REGEX = re.compile(r"10\.\d{4,9}/\S+")
+#: Matches a bare DOI, e.g. ``10.48324/dandi.001827/0.260505.1322``.
+#: The prefix may be subdivided by a registrant (``10.1000.10/123``), which the
+#: DOI Handbook allows, so the leading number is followed by zero or more
+#: ``.``-separated groups.
+DOI_REGEX = re.compile(r"10\.\d{4,9}(?:\.\d+)*/\S+")
 
-#: Prefixes a DOI may be spelled with, in the order they are stripped
+#: Prefixes a DOI may be spelled with, in the order they are stripped.  The
+#: second is a resolver URL, and only that spelling may carry a query string or
+#: fragment that is part of the URL rather than of the DOI.
 DOI_PREFIX_REGEXES = (r"doi:", r"(?:https?://)?(?:dx\.)?doi\.org/")
+DOI_URL_PREFIX_REGEX = DOI_PREFIX_REGEXES[1]
 
 
 def normalize_doi(doi: str) -> str:
@@ -75,6 +81,13 @@ def normalize_doi(doi: str) -> str:
     for prefix_regex in DOI_PREFIX_REGEXES:
         if m := re.match(prefix_regex, value, flags=re.I):
             value = value[m.end() :].strip()
+            if prefix_regex is DOI_URL_PREFIX_REGEX:
+                # A resolver URL copied from a browser can carry a query string
+                # or fragment (``...?locatt=mode:legacy``).  Those belong to the
+                # URL, not to the DOI, and keeping them would store a corrupted
+                # identifier in the Dandiset metadata.  Bare DOIs are left alone,
+                # since ``?`` and ``#`` are legal (if rare) DOI characters.
+                value = re.split(r"[?#]", value, maxsplit=1)[0]
             break
     if not DOI_REGEX.fullmatch(value):
         raise ValueError(
@@ -112,10 +125,24 @@ def fetch_doi_citation_metadata(doi: str) -> dict[str, Any]:
     ) as doiclient:
         try:
             r = doiclient.get(doi, json_resp=False)
-        except HTTP404Error:
+        except HTTP404Error as e:
+            # doi.org 302-redirects a registered DOI to its registration
+            # agency's content-negotiation endpoint, which can itself 404 when
+            # the record is not served as CSL.  Only a 404 that came back from
+            # doi.org itself means the DOI is unregistered.
+            final_url = e.response.url if e.response is not None else url
+            final_netloc = urllib.parse.urlparse(str(final_url)).netloc
+            if final_netloc == urllib.parse.urlparse(DOI_RESOLVER_URL).netloc:
+                raise click.ClickException(
+                    f"DOI {doi} is not registered: {url} returned 404.  Check "
+                    "the DOI for typos and make sure it has already been "
+                    "published."
+                )
             raise click.ClickException(
-                f"DOI {doi} is not registered: {url} returned 404.  Check the "
-                "DOI for typos and make sure it has already been published."
+                f"DOI {doi} is registered but no citation metadata is available "
+                f"for it: {url} redirected to {final_url}, which returned 404.  "
+                "The registration agency may not serve CSL JSON for this record, "
+                "or the metadata may not have propagated yet."
             )
         except HTTPError as e:
             status = e.response.status_code if e.response is not None else "?"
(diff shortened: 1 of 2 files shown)


dandi/cli/cmd_service_scripts.py
```diff
@@ -35,6 +36,112 @@
...
 }
 
+#: Base URL of the DOI resolver used to look up citation metadata
+DOI_RESOLVER_URL = "https://doi.org/"
+
+#: Content type requested from the DOI resolver for citation metadata
+DOI_CSL_ACCEPT = "application/vnd.citationstyles.csl+json; charset=utf-8"
+
+#: Matches a bare DOI, e.g. ``10.48324/dandi.001827/0.260505.1322``
+DOI_REGEX = re.compile(r"10\.\d{4,9}/\S+")
```
yarikoptic-gitmate, line 46: This regex rejects valid legacy DOIs with sub-registrant codes, e.g. `10.1000.10/123` — the DOI Handbook allows registrant subdivision like `10.1000.10`. The old code accepted any string, so such a (rare but valid) DOI that previously worked now fails with a hard `UsageError` before any lookup is attempted.

```suggestion
DOI_REGEX = re.compile(r"10\.\d{4,9}(?:\.\d+)*/\S+")
```

---

dandi/cli/cmd_service_scripts.py
```diff
@@ -35,6 +36,112 @@
...
+    Raises
+    ------
+    ValueError
+        If `doi` is not a syntactically valid DOI in any accepted spelling
+    """
+    value = doi.strip()
+    for prefix_regex in DOI_PREFIX_REGEXES:
+        if m := re.match(prefix_regex, value, flags=re.I):
+            value = value[m.end() :].strip()
+            break
```
yarikoptic-gitmate, line 78: Since `\S+` matches `?` and `#`, a resolver URL pasted straight from a browser — e.g. `https://doi.org/10.1234/foo?locatt=mode:legacy` — normalizes to `10.1234/foo?locatt=mode:legacy`. On a successful lookup that corrupted identifier is then persisted into the Dandiset's `relatedResource` metadata (and a `#fragment` variant silently looks up a *different* DOI than the one stored, since `requests` drops the fragment before sending).

Stripping the query string/fragment only in the URL spellings keeps bare DOIs containing `?` (technically legal, if rare) working:

```suggestion
    for prefix_regex in DOI_PREFIX_REGEXES:
        if m := re.match(prefix_regex, value, flags=re.I):
            # a resolver URL pasted from a browser may carry a query string
            # or fragment that is not part of the DOI itself
            value = re.split(r"[?#]", value[m.end() :].strip(), maxsplit=1)[0]
            break
```

---

dandi/cli/cmd_service_scripts.py
```diff
@@ -35,6 +36,112 @@
...
+    with RESTFullAPIClient(
+        DOI_RESOLVER_URL, headers={"Accept": DOI_CSL_ACCEPT}
+    ) as doiclient:
+        try:
+            r = doiclient.get(doi, json_resp=False)
+        except HTTP404Error:
+            raise click.ClickException(
+                f"DOI {doi} is not registered: {url} returned 404.  Check the "
+                "DOI for typos and make sure it has already been published."
+            )
```
yarikoptic-gitmate, line 119: A 404 here doesn't necessarily mean the DOI is unregistered: doi.org 302-redirects registered DOIs to the registration agency's content-negotiation endpoint (e.g. `data.crossref.org`), which can itself return 404 when the metadata hasn't propagated yet or the record type isn't served as CSL. In that case the user is told to "check the DOI for typos" for a DOI that *is* registered. `HTTP404Error` is raised with `response=result`, so the final URL is available to distinguish the two cases, like the non-JSON branch below does with `r.url`:

```suggestion
        except HTTP404Error as e:
            resolved_url = e.response.url if e.response is not None else url
            if resolved_url.rstrip("/") == url.rstrip("/"):
                raise click.ClickException(
                    f"DOI {doi} is not registered: {url} returned 404.  Check "
                    "the DOI for typos and make sure it has already been "
                    "published."
                )
            raise click.ClickException(
                f"DOI {doi} is registered, but its registration agency did not "
                f"provide citation metadata: {resolved_url} returned 404."
            )
```

---

dandi/cli/cmd_service_scripts.py
```diff
@@ -258,15 +373,10 @@ def update_dandiset_from_doi(
             break
 
     start_time = datetime.now().astimezone()
+    # Resolve the DOI before talking to the archive, so that a bad DOI fails
+    # fast and without requiring credentials
+    doidata = fetch_doi_citation_metadata(doi)
```
yarikoptic-gitmate, line 378: `fetch_doi_citation_metadata()` only validates that the record is a dict, but downstream this command still hard-indexes `doidata["author"]` (line 387) and `doidata["title"]` (line 442). A DOI that resolves to valid CSL JSON *without* one of those keys — real for some Crossref record types such as editorials/corrections, or records with only organizational creators — still crashes with a raw `KeyError` traceback, the exact failure class this PR sets out to eliminate. (`name`/`description` are safe: `copy_str_from_doi_to_metadata()` uses `.get()`.)

Validating up front, only for the fields actually being updated, keeps the descriptive-error guarantee:

```suggestion
    doidata = fetch_doi_citation_metadata(doi)
    for field, key in (("contributor", "author"), ("relatedResource", "title")):
        if field in fields and key not in doidata:
            raise click.ClickException(
                f"Citation metadata for DOI {doi} has no {key!r} field, which "
                f"is needed to update {field!r}."
            )
```

---

yarikoptic-gitmate (commented):
Reviewed the diff — the core fix is sound (`json_resp=False` correctly stops the client from overriding the CSL `Accept` header, except-clause ordering is right, the 18 new tests pass locally, and flake8 is clean). Left 4 inline comments with suggestions:

1. **`doidata["author"]` / `doidata["title"]` can still raise raw `KeyError` tracebacks** for registered DOIs whose CSL record omits those keys — the failure class this PR targets (most severe; anchored at the `fetch_doi_citation_metadata()` call since the indexing sites are outside the diff).
2. **`normalize_doi()` keeps `?query`/`#fragment` from pasted resolver URLs** as part of the DOI, which can persist a corrupted identifier into Dandiset metadata.
3. **The 404 message misdiagnoses RA-endpoint 404s** ("not registered — check for typos") for DOIs that are registered but whose agency doesn't serve CSL metadata.
4. **`DOI_REGEX` rejects valid legacy sub-registrant DOIs** (`10.1000.10/123`) that the old code accepted.

---

yarikoptic:
@adityasingh2400 thanks for the PR. did you use that service script in battlefield for some dandisets?

yarikoptic:
note that we have an interactive AI-assisted (for better or for worse) helper to provide extended metadata entry/curation assistance

image

so it could fetch even more than what DOI provides or can provide in principle.  Did you try it out?

adityasingh2400:
Thanks both. All four review points are addressed in `fee1ed1`.

1. `DOI_REGEX` now allows registrant subdivision, `10\.\d{4,9}(?:\.\d+)*/\S+`, so `10.1000.10/123` is accepted again.
2. Query strings and fragments are stripped only from the resolver-URL spelling, so `https://doi.org/10.1234/foo?locatt=mode:legacy` normalizes to `10.1234/foo` while a bare `10.1234/foo?bar` keeps its `?`.
3. A 404 is only reported as unregistered when it came from doi.org itself. If doi.org redirected first and the agency 404s, the message now says the DOI is registered but has no citation metadata, and names the final URL.
4. New `check_doi_fields()` validates only the CSL keys the requested `--fields` will index, `author` for contributor and `title` for relatedResource, so a record without them fails with a message instead of a `KeyError`.

Five tests cover these. Against the pre-review source 7 of the new cases fail, including the 404 one asserting the old wording, and the 2 that pass are the bare-DOI cases that were already correct. Full file is 35 passed, 6 skipped.

@yarikoptic on your two questions, honestly: no to both. I have not run this against a real Dandiset, only against mocked resolver responses and the live DOI resolver for the happy path, and I have not tried the interactive metadata helper. So I cannot claim this is better than what that helper already gets you. If the helper covers this ground, the useful part of this PR is probably just the error handling rather than the feature, and I am happy for it to be scoped down or closed.

Still a draft. Say the word and I will mark it ready.

dandi/cli/cmd_service_scripts.py
adityasingh2400: Done, with one deliberate deviation. Your prose says to strip only in the URL spellings, but the suggestion loop strips for every entry in `DOI_PREFIX_REGEXES`, which includes `doi:`. That would take `doi:10.1234/foo?bar` down to `10.1234/foo`, which is the case your prose sets out to protect.

So I gated it on the URL prefix specifically:

```python
for prefix_regex in DOI_PREFIX_REGEXES:
    if m := re.match(prefix_regex, value, flags=re.I):
        value = value[m.end() :].strip()
        if prefix_regex is DOI_URL_PREFIX_REGEX:
            value = re.split(r"[?#]", value, maxsplit=1)[0]
        break
```

Covered by `test_normalize_doi_prefix_and_url_suffix`, which asserts both directions: `https://doi.org/10.1234/foo?locatt=mode:legacy` becomes `10.1234/foo`, while `10.1234/foo?bar` and `doi:10.1234/foo#bar` keep their suffix.

dandi/cli/cmd_service_scripts.py
adityasingh2400: Done. Same logic, lifted into a named helper so it can be tested directly rather than only through the command:

```python
DOI_REQUIRED_KEYS = {"contributor": "author", "relatedResource": "title"}

def check_doi_fields(doi, doidata, fields):
    missing = {
        key: field
        for field, key in DOI_REQUIRED_KEYS.items()
        if field in fields and key not in doidata
    }
    ...
```

called right after `fetch_doi_citation_metadata(doi)`. Only the requested `--fields` are checked, so `--fields name,description` still works on a record with no `author` or `title`, which `copy_str_from_doi_to_metadata()` handles with `.get()` anyway.

`test_check_doi_fields_missing` and `test_check_doi_fields_ok` cover both directions.

dandi/cli/cmd_service_scripts.py
```diff
@@ -35,6 +36,180 @@
...
+DOI_RESOLVER_URL = "https://doi.org/"
+
+#: Content type requested from the DOI resolver for citation metadata
+DOI_CSL_ACCEPT = "application/vnd.citationstyles.csl+json; charset=utf-8"
+
+#: Matches a bare DOI, e.g. ``10.48324/dandi.001827/0.260505.1322``.
+#: The prefix may be subdivided by a registrant (``10.1000.10/123``), which the
+#: DOI Handbook allows, so the leading number is followed by zero or more
+#: ``.``-separated groups.
+DOI_REGEX = re.compile(r"10\.\d{4,9}(?:\.\d+)*/\S+")
```
yarikoptic, line 49: here, should use one from `DOI_PREFIX_PATTERN` etc defined in dandi-schema
https://github.com/dandi/dandi-schema/blob/master/dandischema/models.py#L72
as the instance could be "vendored" and we might change / adjust how DOIs look  so knowledge should be 'centralized'

dandi/cli/cmd_service_scripts.py
```diff
@@ -35,6 +36,180 @@
...
+                # or fragment (``...?locatt=mode:legacy``).  Those belong to the
+                # URL, not to the DOI, and keeping them would store a corrupted
+                # identifier in the Dandiset metadata.  Bare DOIs are left alone,
+                # since ``?`` and ``#`` are legal (if rare) DOI characters.
+                value = re.split(r"[?#]", value, maxsplit=1)[0]
+            break
+    if not DOI_REGEX.fullmatch(value):
+        raise ValueError(
+            f"{doi!r} does not look like a DOI.  Expected something like "
+            "'10.48324/dandi.001827/0.260505.1322', optionally prefixed with "
```
yarikoptic, line 95: here specific current instance `.doi_prefix` could be used in principle...

yarikoptic (approved):
> @yarikoptic on your two questions, honestly: no to both. I have not run this against a real Dandiset, only against mocked resolver responses and the live DOI resolver for the happy path, and I have not tried the interactive metadata helper.

<details>
<summary>ok, I have tried on original to https://github.com/dandi/dandi-cli/issues/1855 case and it was still failing on master version and passed fine here</summary> 

```shell
❯ DANDI_DEVEL=1 dandi service-scripts update-dandiset-from-doi -d 216023 -e overwrite -F all -i dandi-sandbox https://doi.org/10.48324/dandi.001827/0.260505.1322
2026-09-14 17:02:31,797 [ WARNING] A newer version (0.78.0) of dandi/dandi-cli is available. You are using 0.76.8+8.gd15b9c76
2026-09-14 17:02:33,521 [    INFO] Adding new value to Dandiset contributor field:

    affiliation: []
    includeInCitation: true
    name: Gutekundt, Claire-Anne
    roleName:
    - dcite:Author
    schemaKey: Person


2026-09-14 17:02:33,521 [    INFO] Setting Dandiset name to ': Direct Delivery of Modulated Kilohertz Frequency Waveforms Enable Simultaneous Electrical Stimulation and Recording with Minimal-Artifact'
2026-09-14 17:02:33,521 [    INFO] Setting Dandiset description to 'This dataset contains the raw in vivo data for the manuscript in the title. Matlab code to reproduce the manuscript figures is available, contact authors to access the scripts.'
2026-09-14 17:02:33,523 [    INFO] Adding new value to Dandiset relatedResource field:

    identifier: 10.48324/dandi.001827/0.260505.1322
    name: ': Direct Delivery of Modulated Kilohertz Frequency Waveforms Enable Simultaneous
      Electrical Stimulation and Recording with Minimal-Artifact'
    relation: dcite:IsDescribedBy
    schemaKey: Resource
    url: https://doi.org/10.48324/dandi.001827/0.260505.1322


Show diff from old metadata to new? [Y/n] Y
--- dandiset.yaml:old
+++ dandiset.yaml:new
@@ -17,17 +17,47 @@
   roleName:
   - dcite:ContactPerson
   schemaKey: Person
+- affiliation: []
+  includeInCitation: true
+  name: Gutekundt, Claire-Anne
+  roleName:
+  - dcite:Author
+  schemaKey: Person
 dateCreated: '2025-01-20T23:27:26.240363+00:00'
-description: 'test '
+description: This dataset contains the raw in vivo data for the manuscript in 
+  the title. Matlab code to reproduce the manuscript figures is available, 
+  contact authors to access the scripts.
 id: DANDI-SANDBOX:216023/draft
 identifier: DANDI-SANDBOX:216023
 license:
 - spdx:CC0-1.0
 manifestLocation:
 - https://api.sandbox.dandiarchive.org/api/dandisets/216023/versions/draft/assets/
-name: test
+name: ': Direct Delivery of Modulated Kilohertz Frequency Waveforms Enable Simultaneous
+  Electrical Stimulation and Recording with Minimal-Artifact'
+relatedResource:
+- identifier: 10.48324/dandi.001827/0.260505.1322
+  name: ': Direct Delivery of Modulated Kilohertz Frequency Waveforms Enable Simultaneous
+    Electrical Stimulation and Recording with Minimal-Artifact'
+  relation: dcite:IsDescribedBy
+  schemaKey: Resource
+  url: https://doi.org/10.48324/dandi.001827/0.260505.1322
 repository: https://sandbox.dandiarchive.org
 schemaKey: Dandiset
 schemaVersion: 0.7.0
 url: https://sandbox.dandiarchive.org/dandiset/216023/draft
 version: draft
+wasGeneratedBy:
+- description: Metadata (contributor, name, description, relatedResource) was 
+    enhanced with data from DOI 10.48324/dandi.001827/0.260505.1322 by DANDI cli
+  endDate: '2026-09-14 17:02:33.524033-04:00'
+  id: urn:uuid:1ae5c80c-f697-416c-8fa6-216c00a5f1df
+  name: Metadata extraction from DOI
+  schemaKey: Activity
+  startDate: '2026-09-14 17:02:31.800287-04:00'
+  wasAssociatedWith:
+  - identifier: RRID:SCR_019009
+    name: DANDI Command Line Interface
+    schemaKey: Software
+    url: https://github.com/dandi/dandi-cli
+    version: 0.76.8+8.gd15b9c76

Save modified Dandiset metadata? [Y/n] Y
2026-09-14 17:03:12,752 [    INFO] Saving ...
❯ DANDI_DEVEL=1 dandi service-scripts update-dandiset-from-doi -d 216023 -e overwrite -F all -i dandi-sandbox https://doi.org/10.48324/dandi.001827/0.260505.1322
2026-09-14 17:03:18,388 [ WARNING] A newer version (0.78.0) of dandi/dandi-cli is available. You are using 0.76.8+8.gd15b9c76
2026-09-14 17:03:20,280 [    INFO] Dandiset contributor field already up to date
2026-09-14 17:03:20,280 [    INFO] Dandiset name already up to date
2026-09-14 17:03:20,280 [    INFO] Dandiset description already up to date
2026-09-14 17:03:20,284 [    INFO] Dandiset relatedResource field already up to date
2026-09-14 17:03:20,284 [    INFO] No changes to Dandiset metadata

```
</details>

so let's not make the perfect to be the enemy of the good and proceed!  If you open it up for review, we could merge AFAIK.  It could be improved but since it is just a service script, I think the chances that it even gets used again are slim anyways, so likely not worth much more investsment and it does address prior BUG. Thank you @adityasingh2400 !

dandi/cli/cmd_service_scripts.py
adityasingh2400: Done in a69127e. `DOI_REGEX` is now built on `UNVENDORED_DOI_PREFIX_PATTERN` from `dandischema.conf`, so what a DOI prefix looks like is no longer defined here. That constant exists in every dandischema version the pin allows (0.12.0, 0.13.0 and 0.14.0 all define it).

I used the general pattern rather than the instance-specific `DOI_PREFIX_PATTERN` from `dandischema.models` on purpose. On a vendored instance that one is the escaped instance prefix, and this command is normally given a publication's DOI, which `relatedResource` records as `dcite:IsDescribedBy`, so matching only the instance's own prefix would reject the main input. The registrant subdivision case from the earlier review (`10.1000.10/123`) is still accepted, since the suffix `(?:\.\d+)*` is composed on top of the dandischema prefix. One small effect of following dandischema: the prefix no longer has an upper bound of nine digits, which matches how dandischema itself validates a DOI prefix.

`pytest dandi/cli/tests/test_service_scripts.py`: 38 passed, 6 skipped. flake8, codespell, black, isort and mypy are clean on both files.

dandi/cli/cmd_service_scripts.py
adityasingh2400: Done in a69127e. The example in the "does not look like a DOI" message now comes from a small `example_doi()` helper that reads `get_instance_config().doi_prefix`. On a vendored instance with its own prefix the example is a DOI that instance could have minted, `{doi_prefix}/{instance_name.lower()}.000001/0.240101.1234`, and I checked that this shape matches `DANDI_DOI_PATTERN` when `DANDI_DOI_PREFIX` and friends are set in the environment. Without a configured prefix it falls back to the real DANDI DOI that was hardcoded before.

I left the static example in the `--help` docstring as is, since a docstring cannot follow the runtime configuration. Two tests cover the helper, one for each case, by patching `get_instance_config` in the module so the global dandischema configuration is not touched.

adityasingh2400:
@yarikoptic thank you for running this against the original #1855 case, that is the confirmation I could not give myself. Both inline points are addressed in a69127e: the DOI prefix pattern now comes from dandischema, and the example DOI in the error message follows the instance's `doi_prefix` when one is configured. Details are in the two thread replies. Marked ready for review as you suggested.

PR #1905, opened 2026-08-26
Organize and upload images referenced by NWB files with `ExternalImage`

h-mayorquin:
`ExternalImage` went into the schema in NeurodataWithoutBorders/nwb-schema#604 and into pynwb in NeurodataWithoutBorders/pynwb#2079, and I am writing the [neuroconv interface](https://github.com/catalystneuro/neuroconv/pull/1987) that produces one right now, so I went to look at what dandi does with a file carrying them: 

Currently, `organize --update-external-file-paths` turns out to do the whole job for `ImageSeries.external_file` and nothing for images, and `upload` drops a `.png` as an unrecognized file.

This PR gives images what videos already have: an `ImageAsset` type so they upload at all, and `ExternalImage.data` collected, relocated and rewritten next to `external_file`. The images keep the existing video naming, `<nwbfile-stem>/<object_id>_external_file_<n>`. BEP044 and BEP047 are not merged, and moving media naming to something BIDS-shaped should move video and images together.

I checked it end to end against the sandbox, the plain upload and the organized one both:
https://sandbox.dandiarchive.org/dandiset/218663/draft


CodyCBakerPhD:
Thanks!

One small item: I notice the commit co-authorships with Claude, but none of the tests are marked as `ai_generated`

Please confirm origin, then we should be good to go

h-mayorquin:
Thanks for the quick chck

Test are indeed generated with AI, should I mark them as such? how?

I was not aware of this convention.

CodyCBakerPhD:
@h-mayorquin I'm actually quite surprised it wasn't done automatically given the instructions in https://github.com/dandi/dandi-cli/blob/master/CLAUDE.md#mandatory-read-before-making-any-code-changes

Also mentioned in human docs: https://github.com/dandi/dandi-cli/blob/master/DEVELOPMENT.md?plain=1#L250

h-mayorquin:
> @h-mayorquin I'm actually quite surprised it wasn't done automatically given the instructions in https://github.com/dandi/dandi-cli/blob/master/CLAUDE.md#mandatory-read-before-making-any-code-changes
> 
> Also mentioned in human docs: https://github.com/dandi/dandi-cli/blob/master/DEVELOPMENT.md?plain=1#L250

Oh, yes, my bad, **I** did not read the human documentation. 

Regarding the agent. I have been using an agent orchestrator workflow (`claude agents`) and those are not start from the repo root so the local CLAUDE.md is not consumed automatically. Concretely, the agent did not pick the pick the repo instructions because it was not launched from the repo directory. This is a new failure model that I was not aware of.

h-mayorquin:
Marked in bbd85309.

I left `test_find_dandi_files` alone, since I only added the `.png` case to an existing test and marking the whole function would relabel work that is not generated. Happy to mark it too if you would rather.

CodyCBakerPhD:
> Regarding the agent. I have been using an agent orchestrator workflow (claude agents) and those are not start from the repo root so the local CLAUDE.md is not consumed automatically. Concretely, the agent did not pick the pick the repo instructions because it was not launched from the repo directory. This is a new failure model that I was not aware of.

Such things can often be resolve by having some base instructions tailored to your usage for sub-agents to 'always check and follow the local agent instructions per repository'

> I left test_find_dandi_files alone, since I only added the .png case to an existing test and marking the whole function would relabel work that is not generated. Happy to mark it too if you would rather.

Nope that's as intended then

Thank you

CodyCBakerPhD (approved):

CodyCBakerPhD:
Bad luck on the CI, seems very recent on the failures

Fixing in https://github.com/dandi/dandi-cli/pull/1906, will merge soon as that is through

PR #1908, opened 2026-08-27
Tolerate the filesystem's mtime granularity in `download -e refresh`

### LINKED ISSUE #1907: download -e refresh never skips anything on filesystems without sub-second mtimes

## Problem

## Mechanism

```python
```

```python
```

CodyCBakerPhD:
Closes #1907.

## Problem

`dandi download -e refresh` never skips anything on a filesystem that does not store
sub-second mtimes, so it re-transfers every asset of every dandiset on every invocation.

The refresh branch of `_download_file()` compares the asset's recorded mtime against the
mtime read back from the local file — but that mtime is one dandi set itself with
`os.utime()` at the end of the previous download, so the comparison is really a
filesystem round trip.  It was performed with `is_same_time()`'s default `tolerance` of
one microsecond, i.e. it assumed the value round-trips exactly.

- ext4/XFS/tmpfs store nanosecond mtimes, so the round trip *is* exact and refresh
  works.  That is why the bug is invisible to most developers and to CI.
- A filesystem that truncates or rounds mtimes (a mounted Windows volume, FAT/exFAT,
  some network mounts) reads back a value up to a full second off — six orders of
  magnitude past the tolerance.  `same` ends up `["size"]`, the
  `same == ["mtime", "size"]` check fails, and the file is redownloaded.  Every file,
  every time, permanently.

Reported against a nightly cron job pulling four dandisets (~32 GB) into `/mnt/g/...`
under WSL2: ~16 minutes of re-transfer per run with nothing changed upstream.

## Approach

Three options were weighed (see the issue):

1. **Measure the destination filesystem's granularity and widen the tolerance to it** —
   chosen.
2. Compare with `>=` as `_populate_dandiset_yaml()` does — **rejected**: truncation
   makes the local mtime *smaller* than the record, so `local >= record` is false there
   too.  `dandiset.yaml` skips because of the content-equality check that precedes the
   `>=` branch (hence its `no change` message), not because of the comparison, so
   copying that comparison would not have fixed the asset path.  Rounding filesystems
   (FAT) can also err in either direction.
3. Size + digest fallback when mtime disagrees — correct, but pays a rehash of the whole
   dandiset on every run, and would mask rather than fix the granularity assumption.
   The existing `TODO` in that branch is left in place.

## Changes

**`dandi/utils.py`** — new `get_mtime_granularity(path)`.  It measures how precisely a
filesystem stores mtimes by writing a series of timestamps to a temporary file with
`os.utime()` and reading them back with `os.stat()`, then rounds the largest observed
round-trip deviation up to one of the granularities filesystems are known to use
(1 µs / 1 ms / 1 s / 2 s).  Results are cached per filesystem (keyed by `st_dev`), so
the probing happens at most once per filesystem per process, and `DEFAULT_MTIME_GRANULARITY`
(1 s) is returned with a warning if the filesystem cannot be probed.

The probe timestamps are anchored at a present-era epoch value so that the float
resolution of the probe matches that of the mtimes actually compared, and include
offsets crossing a second boundary so that filesystems rounding to a multiple of two
seconds are detected too.

**`dandi/download.py`** — the refresh comparison now passes that granularity as
`is_same_time()`'s `tolerance`.  This is self-correcting across filesystems: it resolves
to the current 1 µs tolerance on ext4/XFS/tmpfs and widens only where the filesystem
genuinely cannot do better.

The rejected-skip debug message now reports both timestamps, their delta, the tolerance
applied and both sizes, instead of only naming which attributes matched:

```
'/mnt/g/000027/sub-RAT123/sub-RAT123.nwb' - same attributes: ['size'].  Redownloading.
Local mtime: 2026-08-22 15:21:20+00:00 (1787412080.000000),
record mtime: 2026-08-22 15:21:20.651000+00:00 (1787412080.651000),
delta: 0.651000 s, tolerated mtime granularity: 1 s, local size: 19, record size: 19
```

## Tests

A real ext4 temp dir cannot exercise this class of bug, so a new `coarse_mtime_fs`
fixture patches `os.utime()` to quantize the times it is asked to set — which is what a
coarse-granularity filesystem effectively does — and clears the granularity cache around
the test.

- `test_download_file_refresh_coarse_mtime_fs[0.0/1.0/2.0]` downloads a file through
  `_download_file()` and then refreshes it, asserting the second pass is
  `skipped / same time and size` and that the downloader is not called again.  This
  exercises the full write-then-compare round trip rather than a synthetic mtime.
- `test_download_file_refresh_reports_mtime_mismatch` asserts a genuinely out-of-date
  file is still redownloaded and that the debug message carries the timestamps, delta
  and tolerance.
- `test_get_mtime_granularity_{fine,coarse,unprobeable}` cover the probe directly.

All new tests carry `@pytest.mark.ai_generated`.  Verified that the three
coarse-granularity tests fail on the pre-fix code and pass after; the `[0.0]`
(ext4 baseline) case passes both ways, as expected.

`black`/`isort` (at the pinned pre-commit versions), `flake8` and `mypy` are clean, and
the full suite still collects.  Test failures in this environment were limited to the
pre-existing network/proxy-dependent ones, which fail identically on `master`.

https://claude.ai/code/session_01VGH4mHn6chhLQfGrAYgrBY

---
_Generated by [Claude Code](https://claude.ai/code/session_01VGH4mHn6chhLQfGrAYgrBY)_

### DIFF
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -59,6 +59,7 @@ from .utils import (
     ensure_datetime,
     exclude_from_zarr,
     flattened,
+    get_mtime_granularity,
     get_retry_after,
     is_same_time,
     path_is_subpath,
@@ -700,9 +701,20 @@ def _download_file(
                     f"{path!r} - no mtime or ctime in the record, redownloading"
                 )
             else:
-                stat = os.stat(op.realpath(path))
+                realpath = op.realpath(path)
+                stat = os.stat(realpath)
+                # The mtime we compare against is the one we set ourselves via
+                # os.utime() after the previous download, so the comparison is
+                # really a filesystem round trip.  Not every filesystem stores
+                # mtimes at the resolution os.stat() reports them at (mounted
+                # Windows volumes, FAT/exFAT and some network filesystems
+                # truncate or round), so tolerate whatever granularity the
+                # destination filesystem actually has instead of assuming that
+                # the value round-trips exactly.  See
+                # https://github.com/dandi/dandi-cli/issues/1907
+                tolerance = get_mtime_granularity(op.dirname(realpath))
                 same = []
-                if is_same_time(stat.st_mtime, mtime):
+                if is_same_time(stat.st_mtime, mtime, tolerance=tolerance):
                     same.append("mtime")
                 if size is not None and stat.st_size == size:
                     same.append("size")
@@ -712,7 +724,22 @@ def _download_file(
                     # TODO: add recording and handling of .nwb object_id
                     yield _skip_file("same time and size", size=size)
                     return
-                lgr.debug(f"{path!r} - same attributes: {same}.  Redownloading")
+                lgr.debug(
+                    "%r - same attributes: %s.  Redownloading.  "
+                    "Local mtime: %s (%.6f), record mtime: %s (%.6f), "
+                    "delta: %.6f s, tolerated mtime granularity: %g s, "
+                    "local size: %s, record size: %s",
+                    str(path),
+                    same,
+                    ensure_datetime(stat.st_mtime),
+                    stat.st_mtime,
+                    mtime,
+                    mtime.timestamp(),
+                    abs(stat.st_mtime - mtime.timestamp()),
+                    tolerance,
+                    stat.st_size,
+                    size,
+                )
 
     if size is not None:
         yield {"size": size}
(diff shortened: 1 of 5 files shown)


CodyCBakerPhD:
Closing since considerably more complex than #1910 (this one was proposed by the source investigator downstream - both Opus)

PR #1909, opened 2026-08-27
Tolerate coarse filesystem mtimes in `download -e refresh` with a fixed 2 s window

### LINKED ISSUE #1907: download -e refresh never skips anything on filesystems without sub-second mtimes

CodyCBakerPhD:
Closes #1907.  Simpler alternative to #1908.

## Problem

`dandi download -e refresh` never skips anything on a filesystem that does not store sub-second mtimes, so it re-transfers every asset of every dandiset on every invocation.

The refresh branch of `_download_file()` compares the asset's recorded mtime against the mtime read back from the local file — but that mtime is one dandi set itself with `os.utime()` at the end of the previous download, so the comparison is really a filesystem round trip.  It was performed with `is_same_time()`'s default `tolerance` of one microsecond, i.e. it assumed the value round-trips exactly.  ext4/XFS/tmpfs store nanosecond mtimes, so the bug is invisible to most developers and to CI; a filesystem that truncates or rounds mtimes (a mounted Windows volume, FAT/exFAT, some network mounts) reads back a value up to 1–2 s off, so the skip check fails for every file, every time, permanently.  See #1907 for the full mechanism and reproducer.

## Approach: a fixed modify-window instead of per-filesystem probing

#1908 resolves this by measuring each destination filesystem's actual mtime granularity at runtime (probe with a temp file, cache per `st_dev`) and using the measured value as the tolerance.  This PR instead compares with a fixed tolerance of **2 s** — the coarsest mtime granularity a filesystem is known to use (FAT) — the same approach as rsync's `--modify-window` and rclone's `--modify-window`.

The precision that probing buys is protection against a scenario the rest of the check already covers:

- The local mtime being compared is not an organic "was this file touched" signal — it is a marker dandi itself stamped from the very record value it later compares against.  The comparison's only job is "did our own stamp survive the filesystem round trip?".
- A skip additionally requires an **exact size match**, and the recorded mtime (`blobDateModified`) only moves when the asset is actually replaced on the server.  A false skip under a 2 s window therefore requires a same-size re-upload whose recorded mtime moved by less than 2 s — not a case that occurs in practice.
- Widening the tolerance can only convert redownloads into skips, never the reverse.
- On the coarse filesystems that actually exhibit the bug, the probe converges on 1–2 s anyway; its hard-coded granularity table already concedes 2 s as the ceiling.  A fixed window is the same fix minus the measurement apparatus.

Compared to probing, this drops ~120 lines of machinery and its operational failure modes: temp-file writes inside the user's dandiset (with stray probe files left behind if the process dies mid-probe), warnings on read-only destinations where an all-skipped refresh would otherwise work silently, a per-device cache and lock, and the assumption that quantization is one of a few known schemes.  There is also codebase precedent for treating mtimes as coarse: `_populate_dandiset_yaml()` compares at whole-second `>=` semantics and leans on content equality.

## Changes

**`dandi/download.py`** — the refresh comparison now passes `REFRESH_MTIME_TOLERANCE = 2.0` as `is_same_time()`'s `tolerance`, with the rationale documented on the constant.

The rejected-skip debug message now reports both timestamps, their delta, the tolerance applied and both sizes, instead of only naming which attributes matched (the diagnosability half of #1907):

```
'/mnt/g/000027/sub-RAT123/sub-RAT123.nwb' - same attributes: ['size'].  Redownloading.
Local mtime: 2026-08-22 15:21:20+00:00 (1787412080.000000),
record mtime: 2026-08-22 15:21:20.651000+00:00 (1787412080.651000),
delta: 0.651000 s, tolerance: 2 s, local size: 19, record size: 19
```

## Tests

A real ext4 temp dir cannot exercise this class of bug, so a new `coarse_mtime_fs` fixture patches `os.utime()` to quantize the times it is asked to set — which is what a coarse-granularity filesystem effectively does.

- `test_download_file_refresh_coarse_mtime_fs[0.0/1.0/2.0]` downloads a file through `_download_file()` and then refreshes it, asserting the second pass is `skipped / same time and size` and that the downloader is not called again.  This exercises the full write-then-compare round trip rather than a synthetic mtime.
- `test_download_file_refresh_reports_mtime_mismatch` asserts a genuinely out-of-date file is still redownloaded and that the debug message carries the timestamps, delta and tolerance.

All new tests carry `@pytest.mark.ai_generated`.  Verified that the `[1.0]` and `[2.0]` cases and the mismatch test fail on the pre-fix code and pass after; the `[0.0]` (ext4 baseline) case passes both ways, as expected.

`black`/`isort` (at the pinned pre-commit versions), `flake8`, `codespell` and `mypy` are clean, the full suite still collects, and the offline download unit tests pass.  Test failures in this environment were limited to pre-existing network/proxy-dependent ones in `test_utils.py`, which fail identically on an unmodified tree.

https://claude.ai/code/session_01TarBMUbTumRRVoSSgRoDCs

---
_Generated by [Claude Code](https://claude.ai/code/session_01TarBMUbTumRRVoSSgRoDCs)_

### DIFF
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -68,6 +68,21 @@ from .utils import (
 
 lgr = get_logger()
 
+#: Tolerance (in seconds) for comparing a local file's mtime against the mtime
+#: recorded for the asset when deciding whether ``DownloadExisting.REFRESH``
+#: may skip a redownload.  The local mtime is one we set ourselves via
+#: `os.utime()` at the end of the previous download, so the comparison is
+#: really a filesystem round trip — and not every filesystem stores mtimes at
+#: the resolution `os.stat()` reports them at: mounted Windows volumes,
+#: FAT/exFAT and some network filesystems truncate or round the value, by up
+#: to FAT's 2 s granularity.  Tolerate that coarsest known granularity
+#: everywhere (cf. rsync's ``--modify-window``) instead of assuming the value
+#: round-trips exactly.  A skip additionally requires an exact size match, and
+#: the recorded mtime only moves when the asset is actually replaced, so the
+#: widened window cannot in practice mistake an updated asset for an unchanged
+#: one.  https://github.com/dandi/dandi-cli/issues/1907
+REFRESH_MTIME_TOLERANCE = 2.0
+
 
 class DownloadExisting(StrEnum):
     ERROR = "error"
@@ -702,7 +717,9 @@ def _download_file(
             else:
                 stat = os.stat(op.realpath(path))
                 same = []
-                if is_same_time(stat.st_mtime, mtime):
+                if is_same_time(
+                    stat.st_mtime, mtime, tolerance=REFRESH_MTIME_TOLERANCE
+                ):
                     same.append("mtime")
                 if size is not None and stat.st_size == size:
                     same.append("size")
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
Essentially the same as #1910 (Opus- this one was Fable) but I'd call that one a little better organized

PR #1910, opened 2026-08-27
Tolerate the filesystem's mtime granularity in `download -e refresh`

### LINKED ISSUE #1907: download -e refresh never skips anything on filesystems without sub-second mtimes

CodyCBakerPhD:
Closes #1907.  Supersedes #1908.

## Fix

The refresh branch of `_download_file()` compared the recorded mtime against the local
one with `is_same_time()`'s default 1 µs tolerance, i.e. it assumed the value round-trips
through the filesystem exactly.  Pass a `MTIME_TOLERANCE` of 2 s instead.

#1908 fixed this by *measuring* each destination filesystem's granularity — 121 lines: a
temp file probed inside the user's download directory, a table of known granularities, an
`st_dev`-keyed cache behind a `threading.Lock`, and a shared fixture reaching into that
private cache so tests could clear it.  Two properties of the comparison site make a
constant sufficient:

1. The local mtime is one this package wrote itself, a few lines below in the same function
   (`os.utime(path, (time.time(), mtime.timestamp()))`).  The comparison is therefore
   only a filesystem round trip, and the sole error it must absorb is that filesystem's
   own quantization — worst case FAT's 2 s.
2. The skip also requires `stat.st_size == size`, so a wider tolerance can only mis-skip
   an asset that changed with an identical size *and* an mtime moving under 2 s.

https://claude.ai/code/session_015TCMNL9Jkrz4ooAf2bmSfE

### DIFF
--- a/dandi/consts.py
+++ b/dandi/consts.py
@@ -257,3 +257,12 @@ REDIRECT_HEAD_TIMEOUT = 30
 
 #: Suffix used for temporary download directories
 DOWNLOAD_SUFFIX = ".dandidownload"
+
+#: Tolerance (in seconds) when comparing an asset's recorded mtime against the
+#: mtime read back from the downloaded file under ``-e refresh``.  That local
+#: mtime is one we set ourselves with ``os.utime()``, so the comparison is
+#: really a filesystem round trip, and not every filesystem stores mtimes at
+#: the resolution ``os.stat()`` reports them at: mounted Windows volumes,
+#: exFAT and some network filesystems truncate, and FAT rounds to a multiple
+#: of two seconds.  See https://github.com/dandi/dandi-cli/issues/1907
+MTIME_TOLERANCE = 2.0
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -13,7 +13,7 @@ from __future__ import annotations
 from collections import Counter, deque
 from collections.abc import Callable, Iterable, Iterator, Sequence
 from dataclasses import InitVar, dataclass, field
-from datetime import datetime
+from datetime import datetime, timezone
 from enum import Enum, StrEnum
 from functools import partial
 import hashlib
@@ -724,14 +724,16 @@ def _download_file(
                     # TODO: add recording and handling of .nwb object_id
                     yield _skip_file("same time and size", size=size)
                     return
+                # Both timestamps are reported in UTC, which is what
+                # is_same_time() normalizes to before comparing them
                 lgr.debug(
                     "%r - same attributes: %s.  Redownloading.  "
-                    "local mtime: %f, record mtime: %f, delta: %f s, "
+                    "local mtime: %s, record mtime: %s, delta: %f s, "
                     "tolerance: %g s, local size: %s, record size: %s",
                     str(path),
                     same,
-                    stat.st_mtime,
-                    mtime.timestamp(),
+                    ensure_datetime(stat.st_mtime, tz=timezone.utc),
+                    ensure_datetime(mtime, tz=timezone.utc),
                     abs(stat.st_mtime - mtime.timestamp()),
                     MTIME_TOLERANCE,
                     stat.st_size,
--- a/dandi/download.py
+++ b/dandi/download.py
@@ -38,7 +38,13 @@ from interleave import FINISH_CURRENT, lazy_interleave
 import requests
 
 from . import get_logger
-from .consts import DOWNLOAD_SUFFIX, RETRY_STATUSES, SyncMode, dandiset_metadata_file
+from .consts import (
+    DOWNLOAD_SUFFIX,
+    MTIME_TOLERANCE,
+    RETRY_STATUSES,
+    SyncMode,
+    dandiset_metadata_file,
+)
 from .dandiapi import AssetType, BaseRemoteZarrAsset, RemoteDandiset
 from .dandiarchive import (
     AssetItemURL,
@@ -702,7 +708,13 @@ def _download_file(
             else:
                 stat = os.stat(op.realpath(path))
                 same = []
-                if is_same_time(stat.st_mtime, mtime):
+                # The mtime compared against here is the one we set ourselves
+                # with os.utime() after the previous download, so this is
+                # really a filesystem round trip; tolerate the coarsest
+                # granularity filesystems are known to store mtimes with
+                # instead of assuming the value round-trips exactly.  See
+                # https://github.com/dandi/dandi-cli/issues/1907
+                if is_same_time(stat.st_mtime, mtime, tolerance=MTIME_TOLERANCE):
                     same.append("mtime")
                 if size is not None and stat.st_size == size:
                     same.append("size")
@@ -712,7 +724,19 @@ def _download_file(
                     # TODO: add recording and handling of .nwb object_id
                     yield _skip_file("same time and size", size=size)
                     return
-                lgr.debug(f"{path!r} - same attributes: {same}.  Redownloading")
+                lgr.debug(
+                    "%r - same attributes: %s.  Redownloading.  "
+                    "local mtime: %f, record mtime: %f, delta: %f s, "
+                    "tolerance: %g s, local size: %s, record size: %s",
+                    str(path),
+                    same,
+                    stat.st_mtime,
+                    mtime.timestamp(),
+                    abs(stat.st_mtime - mtime.timestamp()),
+                    MTIME_TOLERANCE,
+                    stat.st_size,
+                    size,
+                )
 
     if size is not None:
         yield {"size": size}
--- a/dandi/tests/test_download.py
+++ b/dandi/tests/test_download.py
@@ -190,10 +190,22 @@ def coarse_mtime_fs(
     real_utime = os.utime
     granularity = 0.0
 
-    def quantizing_utime(path: Any, times: Any = None, **kwargs: Any) -> None:
-        if granularity and times is not None:
-            times = tuple(t // granularity * granularity for t in times)
-        real_utime(path, times, **kwargs)
+    def quantizing_utime(
+        path: Any, times: Any = None, *, ns: Any = None, **kwargs: Any
+    ) -> None:
+        # `os.utime()` accepts the times either as seconds (`times`) or as
+        # nanoseconds (`ns`, used by e.g. `shutil.copystat()`), never both, so
+        # quantize whichever was given and forward it the same way
+        if granularity:
+            if times is not None:
+                times = tuple(t // granularity * granularity for t in times)
+            if ns is not None:
+                ns_granularity = int(granularity * 1_000_000_000)
+                ns = tuple(n // ns_granularity * ns_granularity for n in ns)
+        if ns is not None:
+            real_utime(path, ns=ns, **kwargs)
+        else:
+            real_utime(path, times, **kwargs)
 
     def set_granularity(value: float) -> None:
         nonlocal granularity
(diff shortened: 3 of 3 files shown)


CodyCBakerPhD:
@yarikoptic Confirmed this solved my issue on 'my system'

dandi/tests/test_download.py
```diff
@@ -170,6 +174,138 @@ def test_download_000027_resume(
...
+        nonlocal granularity
+        granularity = value
+
+    monkeypatch.setattr(os, "utime", quantizing_utime)
+    yield set_granularity
+
+
+#: An arbitrary asset mtime with a non-zero sub-second component, i.e. one that
+#: does not survive a round trip through a filesystem storing whole seconds only
+COARSE_MTIME_RECORD = datetime(2026, 8, 22, 15, 21, 20, 651000, tzinfo=timezone.utc)
```
yarikoptic: This timestamp's epoch second is **even** (`1787412080.651`), so flooring it to a multiple of `2.0` lands on exactly the same value as flooring to `1.0`. The `granularity=2.0` parametrization is therefore a silent duplicate of `granularity=1.0` — the suite never observes a delta larger than 0.651 s:

```
g=1.0  stored 1787412080.0  delta 0.651
g=2.0  stored 1787412080.0  delta 0.651
```

Consequence: nothing in the suite pins the 2 s constant. I set `MTIME_TOLERANCE = 0.7` locally and all four tests still pass — so a regression to under a third of the documented FAT worst case would go unnoticed, which is the one value this PR exists to establish.

Shifting the record by one second onto an odd epoch second makes `g=2.0` truncate by 1.651 s and pins it:

```suggestion
COARSE_MTIME_RECORD = datetime(2026, 8, 22, 15, 21, 21, 651000, tzinfo=timezone.utc)
```

Verified with that change: `MTIME_TOLERANCE` of `0.7` and `1.0` both fail `coarse_mtime_fs[2.0]`, and `2.0` passes.

Worth widening the comment above it too. "does not survive a round trip through a filesystem storing whole seconds" doesn't hint that the parity of the epoch second is load-bearing; something like "chosen so that both the 1 s and 2 s cases truncate by a distinct, near-worst-case amount" would keep the next editor from quietly neutering the 2.0 case again.

---

yarikoptic (commented):
The fix is right, and choosing a constant over #1908's probe-and-cache machinery is well argued. Both properties in the description hold up against the code: the compared mtime is one `_download_file()` set itself at `download.py:862`, and the `stat.st_size == size` conjunct is a real backstop. Happy to see this go in once the test constant is fixed.

Verified locally: all four new tests pass; reverting just the `tolerance=` argument fails `coarse_mtime_fs[1.0]` and `[2.0]`, so the regression coverage is genuine. `flake8`, `mypy`, `black` and `isort` are clean on the three touched files (the `black` diff on `test_download.py` is pre-existing `with mock.patch(...)` code this PR doesn't touch). CI is green across all 30 matrix jobs.

One finding worth fixing, inline on `COARSE_MTIME_RECORD`: the `granularity=2.0` case is a silent no-op, so nothing in the suite actually pins the 2 s constant.

## `_populate_dandiset_yaml()` makes the same assumption

`download.py:588` is the same bug class in the same command, and it's the more interesting half of the story:

```python
elif existing is DownloadExisting.SKIP or (
    existing is DownloadExisting.REFRESH
    and os.lstat(dandiset_yaml).st_mtime >= mtime.timestamp()
):
```

Same premise as the site being fixed — that an mtime we wrote with `os.utime(dandiset_yaml, (time.time(), mtime.timestamp()))` twelve lines below reads back exactly. On a coarse filesystem it reads back *below* the value we set, so the `>=` fails.

The symptom is quite different from #1907, though, and much milder — worth spelling out so the issue doesn't get closed on a wrong model of the blast radius:

1. **No re-transfer churn.** The `yaml_load(fp, typ="safe") == metadata` check above short-circuits with `_skip_file("no change")` whenever the content matches, which is the steady state. The mtime comparison is only reached when the metadata genuinely differs, and there redownloading is the *correct* outcome.
2. **The one real loss is protection of local edits.** That `>=` means "local copy is ahead of the record, leave it alone". Truncation makes a locally-modified `dandiset.yaml` look up to 2 s behind the record when it is in fact level with or slightly ahead of it, and `ds.update_metadata(metadata)` overwrites it. Needs the edit to land within the quantization window of `dandiset.modified`, so it's a corner — but it's silent data loss when it hits, whereas #1907 was merely slow.

Note the direction differs from the `_download_file()` site: this one is a one-sided `>=`, not a symmetric equality, so it wants `st_mtime >= mtime.timestamp() - MTIME_TOLERANCE` rather than an `is_same_time()` tolerance. One line, same constant, and it keeps the two round-trip assumptions in `download.py` from drifting apart. Either fold it in here, or leave a note on #1907 so it isn't marked fully addressed.

## Minor, non-blocking

- **Log readability.** The new debug line prints raw epoch floats (`local mtime: %f, record mtime: %f`). Since sharper diagnostics is part of the point, ISO timestamps would serve someone reading a log far better than `1787412080.651000`. Separately, `%r` applied to `str(path)` renders `'/x/y'` where the old `f"{path!r}"` rendered `PosixPath('/x/y')` — an improvement, just noting the format changed.
- **Fixture hazard.** `quantizing_utime` forwards `times` positionally, so a caller using `os.utime(path, ns=...)` would hit `ValueError: you may specify either 'times' or 'ns' but not both`. Nothing in `dandi` does that today; only a trap if the fixture gets reused.
- `test_download_file_refresh_reports_mtime_mismatch` regex-parses the debug message to recover the delta, coupling it to an incidental log format. Defensible since the diagnostics are part of this change, but asserting `same == ["size"]` more directly would be sturdier.

---

dandi/tests/test_download.py
CodyCBakerPhD: wow, good catch

CodyCBakerPhD:
> The one real loss is protection of local edits. That >= means "local copy is ahead of the record, leave it alone". Truncation makes a locally-modified dandiset.yaml look up to 2 s behind the record when it is in fact level with or slightly ahead of it, and ds.update_metadata(metadata) overwrites it. Needs the edit to land within the quantization window of dandiset.modified, so it's a corner — but it's silent data loss when it hits, whereas https://github.com/dandi/dandi-cli/issues/1907 was merely slow.

I'm not really concerned about this case TBH - we're talking about the CLI usage in particular here and that kind of time lag just isn't realistic (might be if we were to make some kind of interactive app that needed to constantly talk to the database though)

Addressed the other minor suggestions

dandi/download.py
yarikoptic, line 1: **`_populate_dandiset_yaml()`, at line 594, makes the same round-trip assumption this PR is removing from `_download_file()`** — and it's left untouched. (File-level comment because the line is outside this PR's diff hunks, so GitHub won't take an applicable `suggestion` there.)

```python
        elif existing is DownloadExisting.SKIP or (
            existing is DownloadExisting.REFRESH
            and os.lstat(dandiset_yaml).st_mtime >= mtime.timestamp()
        ):
```

That local mtime is also one we set ourselves, twelve lines below:

```python
os.utime(dandiset_yaml, (time.time(), mtime.timestamp()))
```

so on a quantizing destination it reads back *below* what we wrote, and the `>=` fails.

Suggested (verified `black` and `flake8` clean; `MTIME_TOLERANCE` is already imported in this module):

```python
        elif existing is DownloadExisting.SKIP or (
            existing is DownloadExisting.REFRESH
            # This mtime is also one we set ourselves with os.utime() below, so
            # the same filesystem round trip applies here as in
            # _download_file(); without the tolerance a destination that
            # quantizes mtimes makes the local copy look older than the record
            # it was set from.  See https://github.com/dandi/dandi-cli/issues/1907
            and os.lstat(dandiset_yaml).st_mtime >= mtime.timestamp() - MTIME_TOLERANCE
        ):
```

Note the direction differs from the `_download_file()` site — this is a one-sided `>=`, not a symmetric equality, so it wants the constant subtracted rather than an `is_same_time()` tolerance.

The symptom is also quite different from #1907, and worth being precise about so the issue doesn't get closed on the wrong model of the blast radius:

1. **No re-transfer churn here.** The `yaml_load(fp, typ="safe") == metadata` check above short-circuits with `_skip_file("no change")` whenever the content matches, which is the steady state. This comparison is only reached when the metadata genuinely differs — and there, redownloading is the *correct* outcome.
2. **What's actually lost is protection of local edits.** That `>=` means "the local copy is ahead of the record, leave it alone". Truncation makes a locally-modified `dandiset.yaml` look up to 2 s behind the record when it is in fact level with or ahead of it, and `ds.update_metadata(metadata)` then overwrites it. It needs the edit to land within the quantization window of `dandiset.modified`, so it is a corner — but it is silent data loss when it hits, where #1907 was merely slow.

Fine by me to defer this to a follow-up rather than widen the PR — but then #1907 shouldn't be closed as fully addressed, since the same faulty premise survives in the same command.

---

yarikoptic (commented):
Re-reviewed the current head (`32b7bab`). The test constant, the ISO timestamps, the `ns=` fixture hardening and dropping the regex parse all look good — and thanks for catching that `os.utime()` only rejects `times`/`ns` when both are non-None, my `ValueError` claim was wrong.

Verified locally on this head: `flake8`/`mypy`/`black`/`isort` clean on the touched files, all four tests pass, and the constant is now genuinely pinned — `MTIME_TOLERANCE = 1.0` fails `coarse_mtime_fs[2.0]`, which it did not before. `ensure_datetime(stat.st_mtime, tz=timezone.utc)` also normalizes the same way `is_same_time()` does internally, so the two numbers in the log really are in the frame the comparison used.

One item from my first pass is still open — file comment on `download.py` for `_populate_dandiset_yaml()`.

---

dandi/consts.py
```diff
@@ -257,3 +257,12 @@ def urls(self) -> Iterator[str]:
...
 DOWNLOAD_SUFFIX = ".dandidownload"
+
+#: Tolerance (in seconds) when comparing an asset's recorded mtime against the
+#: mtime read back from the downloaded file under ``-e refresh``.  That local
+#: mtime is one we set ourselves with ``os.utime()``, so the comparison is
+#: really a filesystem round trip, and not every filesystem stores mtimes at
+#: the resolution ``os.stat()`` reports them at: mounted Windows volumes,
+#: exFAT and some network filesystems truncate, and FAT rounds to a multiple
+#: of two seconds.  See https://github.com/dandi/dandi-cli/issues/1907
+MTIME_TOLERANCE = 2.0
```
yarikoptic, line 268: initially I thought that we must go instead the way "you" approached in 
- #1908 

instead since this property is really file system dependent and not per se a global const... But there are indeed tolerances to observe but also I do not want to loosen it up needlessly a lot on systems where there is sufficient mtime records precision...  thinking also about the fact that if precision is not there -- it is rounded to nearest number not just "noise" in the time value thus giving us an alternative and somewhat more appropriate decision process to consider it "the same".

I am preparing a potential PR to this PR to see if we could refine that way

yarikoptic: also this is a great example for where to start using https://github.com/con/eval-under, sweeping over filesystems, at least for some portion of the likely related tests (download/uploads)... here fires up one more claude code session to formalize the action to (re)use ;-)

yarikoptic: also now I wonder how it would affect fscacher used intensively here in dandi-cli , also unrelated to this changes, just identifying other affected blocks which might tune up to work "as expected" on such wonderful file systems

CodyCBakerPhD: Why would we need nanosecond resolution when supporting this feature on any filesystem? I'm thinking about the actual human use of the affected behavior (downloading a dataset)

My point to favoring this approach was simplicity, which yes means a 'global solution' but also does not mean overfitting to any one system

So I do feel like a lot of this is just 'overthinking it'

CodyCBakerPhD:
> One item from my first pass is still open — file comment on download.py for _populate_dandiset_yaml().

I did address that, actually, in my [human response](https://github.com/dandi/dandi-cli/pull/1910#issuecomment-5447740272)

Similar to my other comment https://github.com/dandi/dandi-cli/pull/1910#discussion_r3881792640, if you really want it to include the incredibly small (and more likely to be updated at a regular basis on remote) `dantiset.yaml` file too I think you can do that in a follow-up. All I need to address MY human concerns here is what is on this PR

The fscacher question is an interesting one too but I haven't run any operations that felt 'slow' where it could make it feel 'fast'

yarikoptic:
ok, please consider/review
- https://github.com/dandi/dandi-cli/pull/1912

CodyCBakerPhD:
> ok, please consider/review
> 
> https://github.com/dandi/dandi-cli/pull/1912

I already did, 30 minutes earlier https://github.com/dandi/dandi-cli/pull/1912#issuecomment-5454419802

yarikoptic (approved):

PR #1916, opened 2026-09-04
Preserve legacy ndx-pose video references during organize

### LINKED ISSUE #1817: `dandi organize` breaks `PoseEstimation.original_videos` paths
If a user carefully sets the [`original_videos`](https://github.com/rly/ndx-pose/blob/a5e4c2e1a2ef0f86df06963258bd5dabe19c1483/spec/ndx-pose.extensions.yaml#L80-L90) field in an ndx-pose `PoseEstimation` container to match the relative path of their source video files, `dandi organize` will break them. 

`dandi organize --update-external-file-paths` renames the `ImageSeries.external_file` paths but leaves `PoseEstimation.original_videos` untouched, silently breaking the connection between pose estimation data and its source video. Currently, [`_get_image_series()`](https://github.com/dandi/dandi-cli/blob/a3dd9f96/dandi/pynwb_utils.py#L376-L409) only scans for `pynwb.image.ImageSeries` objects in the `VIDEO_FILE_MODULES` and [`rename_nwb_external_files()`](https://github.com/dandi/dandi-cli/blob/a3dd9f96/dandi/pynwb_utils.py#L412-L454) only rewrites `container.external_file`, with no awareness of ndx-pose's `PoseEstimation.original_videos` field.

A possible fix: `rename_nwb_external_files()` already builds a mapping of old paths to new paths for each video file it renames. It could also scan for `PoseEstimation` containers in the NWB file and apply that same mapping to any `original_videos` entries that match an old path.

Would this make sense?

AtomicGlance:
Fixes #1817

Older ndx-pose files can store source-video paths in the deprecated `PoseEstimation.original_videos` field. `dandi organize --update-external-file-paths` previously renamed the linked `ImageSeries.external_file` values but left those legacy strings pointing at the old location.

This change reuses the exact source-video old-to-new mapping while an NWB file is open and updates only matching, non-URL `original_videos` entries. It identifies the container by its `neurodata_type`, so ndx-pose is not a required runtime dependency. Newer `source_video` links and unrelated containers are left untouched.

Tests cover path-separator normalization, byte/string values, remote URLs, and unrelated containers.

Validation: focused regression test and flake8 pass.

### DIFF
--- a/dandi/pynwb_utils.py
+++ b/dandi/pynwb_utils.py
@@ -527,11 +527,10 @@ def rename_nwb_external_files(metadata: list[dict], dandiset_path: str) -> None:
                             ext_file_dict["external_files_renamed"],
                         )
                     ):
-                        if not is_url(str(name_old)):
+                        reference = _external_reference_key(name_old)
+                        if not is_url(reference):
                             container.external_file[no] = str(name_new)
-                            pose_video_renames[_external_reference_key(name_old)] = str(
-                                name_new
-                            )
+                            pose_video_renames[reference] = str(name_new)
                 _rename_pose_estimation_original_videos(nwb, pose_video_renames)
         if external_images:
             _rename_external_images(dandiset_nwbfile_path, external_images)
@@ -565,9 +564,10 @@ def _rename_pose_estimation_original_videos(
         if original_videos is None or isinstance(original_videos, (str, bytes)):
             continue
         for no, name_old in enumerate(original_videos):
-            if is_url(str(name_old)):
+            reference = _external_reference_key(name_old)
+            if is_url(reference):
                 continue
-            if (name_new := renames.get(_external_reference_key(name_old))) is not None:
+            if (name_new := renames.get(reference)) is not None:
                 original_videos[no] = name_new
 
 
--- a/dandi/pynwb_utils.py
+++ b/dandi/pynwb_utils.py
@@ -508,6 +508,7 @@ def rename_nwb_external_files(metadata: list[dict], dandiset_path: str) -> None:
                 dandiset_nwbfile_path, mode="r+", load_namespaces=True
             ) as io:
                 nwb = io.read()
+                pose_video_renames: dict[str, str] = {}
                 for ext_file_dict in image_series:
                     # retrieve nwb neurodata object of the given object id:
                     container_list = [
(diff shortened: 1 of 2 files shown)


AtomicGlance:
I added the remaining regression coverage before asking for review: a close/reopen HDF5 persistence test, coverage through `rename_nwb_external_files()`, and cases for empty mappings, missing/scalar values, and mixed matched/unmatched paths. Byte-valued references are also decoded before URL detection now.

The full `test_pynwb_utils.py` module, lint, typing, every CI test environment (including NFS), and both Codecov checks are green. I kept the change limited to the deprecated `original_videos` field. @h-mayorquin, would you prefer `labeled_videos` to be handled as part of the same compatibility fix, or left for a separate follow-up? If this scope looks right, could a maintainer add the `patch` label and review it?

yarikoptic-gitmate (commented):
Nice, targeted fix. I checked it beyond just reading the diff: built a real NWB file with an actual `ndx-pose` `PoseEstimation` + `ImageSeries`, ran `rename_nwb_external_files` end-to-end, and confirmed `original_videos` and `external_file` both get renamed and persisted correctly — including with `ndx-pose` *uninstalled* (pynwb reconstructs `PoseEstimation` dynamically from the file's cached namespace spec), which validates the "not a required runtime dependency" claim. The unit tests are thorough on the edge cases that matter (byte/str values, path-separator normalization, URLs left alone, unrelated containers, longer replacement strings for variable-length HDF5 strings).

Two small things before this is good to merge:

1. Were the new tests in `test_pynwb_utils.py` written with AI assistance? If so, per `DEVELOPMENT.md` they should carry the `@pytest.mark.ai_generated` marker, as is done elsewhere in the suite (e.g. `test_cmd_validate.py`, `test_base.py`).
2. Could you extend the `rename_nwb_external_files` docstring to mention that it now also updates legacy `PoseEstimation.original_videos` references (not just `ImageSeries.external_file`)? Would help future readers who don't dig into `_rename_pose_estimation_original_videos`.

Otherwise this looks solid to me.

---

AtomicGlance:
Addressed both points from the review:

- Marked the four new tests in `dandi/tests/test_pynwb_utils.py` with `@pytest.mark.ai_generated`, as required by `DEVELOPMENT.md`.
- Expanded `rename_nwb_external_files()`’s docstring to describe updates to legacy `PoseEstimation.original_videos` references alongside `ImageSeries.external_file`.

The focused utility tests pass locally (7 passed), and the touched files pass the targeted flake8 check. The update is in commit `4cc085d5`; the branch is ready for another look.

AtomicGlance:
The code and required checks for #1916 and #1917 are green; the remaining `check_labels` failures are the repository label gate. When convenient, could a maintainer apply `patch` to #1916 and `minor` to #1917 and review them together? No additional code commit is needed for the label check. @yarikoptic

yarikoptic-gitmate (approved):
Re-reviewed at `f1a3e99`. Both prior points are addressed — the docstring now documents the `original_videos` side effect, and all four new tests carry `@pytest.mark.ai_generated`. The core logic in `_rename_pose_estimation_original_videos`/`_external_reference_key` is unchanged from the version I already exercised end-to-end (real `ndx-pose` file, both with the extension installed and with it uninstalled via pynwb's dynamic-class path), so that verification still stands.

On the performance question specifically: `_rename_pose_estimation_original_videos` walks `nwb.objects.values()`, which is the container graph pynwb already fully materializes during the pre-existing `io.read()` call in this same function — so the scan adds no extra I/O or re-parsing, just an in-memory dict iteration with a cheap `getattr` per container. I benchmarked this directly: a synthetic file with 5000 containers took ~18s to `read()` (dominated by pynwb/hdmf construction, unrelated to this PR) versus ~0.5ms to run the new scan over all 5003 resulting objects — over 4 orders of magnitude smaller than the cost the feature already pays to open the file. It's also gated by `if not renames: return`, so files with no non-URL external-file renames (the common case) skip the scan entirely, and it runs once per file rather than once per renamed path. I don't see an avoidable hit here.

Logic-wise nothing new to flag: matching is still by normalized path string (backslash→forward-slash, bytes decoded) against the exact old→new map built from the `ImageSeries` rename pass, URLs are skipped on both sides, and non-list `original_videos` values (missing/str/bytes) are safely skipped.

LGTM.

---

